PR #17, opened 2018-09-05
Use tox, the "test automation tool", on Travis

amarburg:
tox is suggested by pytest as a tool for testing your code for packaging errors.  Basically, run `pytest` most of the time to test code, then run `tox` to test how it's packaged and to test against multiple python versions.   Pull request also changes .travis.yml to use tox.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ __pycache__/
 *.png
 *.tex
 *~
+
+*.egg-info/
+.tox/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
   - "3.6"
+install: pip install tox-travis
 script:
   - tox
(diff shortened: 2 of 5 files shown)


leewujung:
I feel the code is not really at a package stage yet -- there're quite a bit of change to be done for reading/saving different file formats into the standard format. Bare bone pytest seems more transparent for getting the test functions in shape before getting into the whole packaging business. Let me do some investigation on this before accepting the PR.

amarburg:
Fair enough.   The one use case where I've found it useful is that it will catch cases where you've added a dependency ... and the pytest work because you've installed in manually on your computer,   but you've forgotten to add it to requirements.txt.   I also think it's a cleaner interface when testing against multiple versions of python (3.5, 3.6), but I recognize the package isn't there yet.

leewujung:
Thanks @amarburg! I'll include tox with the new folder structure in the next PR now that there really stuff to test for... :)

amarburg:
Let me know if I can help!

PR #20, opened 2019-01-16
Folder restructuring and using tox

leewujung:
- Restructured folders based on functionalities
- Created placeholders with pointers to reference scripts
- Removed random old files
- Merged files from PR #17 for using `tox`


leewujung:
@amarburg please don't merge this yet, the way I am doing it now your contribution doesn't show up. Let me try again.

leewujung:
Yes you're right, `setup.py` pulls everything from `version.py` so things should just be in one place. I just reverted it back to as in the Shablona example (i.e., everything in `version.py`).

PR #21, opened 2019-01-28
Sphinx documentation initial set-up

valentina-s:


leewujung:
I see 1 folder named `doc` and the other named `docs`, are they both supposed to be on? or just the `docs`?

PR #32, opened 2019-02-18
OOI Notebook

valentina-s:
This notebook shows how to get data from OOI and process it with `echopype`. Multiple files are simply processed by a for-loop and in the end simultaneously open with `xr.open_mfdataset`. We can wait to merge it until it is improved.


leewujung:
Let's merge this in as a temporary example and update the notebook after sorting out how to best parallelize the under-the-hood stuff in `echopype`.

PR #35, opened 2019-02-20
Fix dependencies install requires

### LINKED ISSUE #34: Install requires doesn't actually do anything.
## Overview

There's an issue with install requires, which is not really listing the dependencies for this package. So when performing pip install, it assumes you have all the dependencies already on your python environment. Otherwise, I can't use the package right away!

https://github.com/OSOceanAcoustics/echopype/blob/e8b735c8ef2855e434f397ccba3b1b3f5babaa4b/setup.py#L29

lsetiawan:
## Overview

This PR should resolve #34, it gets a list of dependencies from `requirements.txt`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,11 @@ with open(ver_file) as f:
 # Long description will go up on the pypi page
 with open('README.rst') as file:
     LONG_DESCRIPTION = file.read()
+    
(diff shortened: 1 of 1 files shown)


leewujung:
Cool, thanks for discovering all the problems! :D

PR #56, opened 2019-08-03
Additions and improvements to AZFP

ngkavin:
- Added notebook that demonstrates downloading from multiple files from OOI and plotting Sv.
- Added notebook that demonstrates getting MVBS and plotting.
- Added notebook that demonstrates plotting Sv 'matlab style' where pings are evenly spaced.
- Moved range calculation to model so that it is not stored in converted .nc file.
- Handle non rectangular arrays for `backscatter_r` by padding with nan
- Added wrapper class around convert and model.


leewujung:
@ngkavin I changed the base branch to azfp-improve so that we can test everything before merging to master.

PR #60, opened 2019-08-23
Testing azfp-improve

cyrf0006:
Thorough testing of azfp-improve branch: improvement and bug fixing: 

- The biggest challenge was to make the toolbox work when an AZFP has only 3 frequencies (it looks like the firmware samples 4 anyways in the binary data). This now works although maybe not aesthetically perfect (I have to flush the 4th channel to avoid dimension problems). It also now only work with 3 or 4 frequencies.

- Added the capability to use Sv_clean to comptue MVBS (the flag was there but not working)

- Improved noise removal algorithm (among other, added SNR and Sv_threshold) and now returned the signal with noise removed. This is hard-coded, and should be improved, however. 

Some other small bug correction that you might want to have a look:
- renamed "depth" by "calc_range" for consistency;
- added "if statements" here and there (to avoid crash);
- remove +1 in "num_tile_range_bin+1" somewhere that was causing an error when binning...
- modified range_bin_size in get_MVBS for dimension problems.

Also modified default T-S, Depth and orientation default values for my own purpose, but you can ignore. I am still not sure how to pass them during the processing.


leewujung:
Fantastic! Thanks @cyrf0006 !! @ngkavin and I will take a closer look at this next week and ping you on any questions!

A quick response to the last point:
> Also modified default T-S, Depth and orientation default values for my own purpose, but you can ignore. I am still not sure how to pass them during the processing.

We've added setter/getter for some of these params for exactly this reason. You have a good point that orientation default that should also be added.

leewujung:
@cyrf0006 it seems that if i simply resolved conflicts and push into this branch, your commits don't show up properly in the history. So what I am going to do is to push to the `Fred_s_azfp` branch in *your* fork and then merge. Not there yet, but you'll get a notification when I push (not sure if I have access to the PR branch) or do a PR.

cyrf0006:
Thx! @leewujung

leewujung:
Merging this PR to `azfp-improve` now so it's easier for @ngkavin and I to work together from that branch point, even though tests for the model module won't pass just yet.

PR #69, opened 2019-10-29
Suggested simplification of convert.py

emiliom:
@leewujung I was thinking back at the flow in convert.py, and realized there was a good chunk of rendundancy and some tests happening before more logical ones (ie, testing for the file extension type before testing for whether the file path exists). So, I went ahead and refactored the code to implement this simplification. 

I didn't want to bother forking, cloning, editing locally (for just one module), then submitting a PR. So I've edited it inline here in Github. I have done no testing whatsoever! My intent is not necessarily for you to try to merge this PR, but just to communicate my suggestions

FYI, in this version, `echosounder_type` is no longer used. But I can see that maybe the redundant structure in your code was intended to support, in the future, an expansion of functionality and echotypes handled. In that case, my single-if-else-block approach may not be a good way to go. Use as you see fit!


leewujung:
Great, thanks @emiliom! I'll take a look and update this along with other bugs. Looks like I can just push to your branch so it'll be straightforward to merge once I test it out.

leewujung:
@emiliom I changed the "empty" wrapper classes to simple functions based on our discussion. Goodness that was 13 days ago!

emiliom:
Great.

PR #70, opened 2019-11-08
Incorporate Convert and Model changes

ngkavin:
These changes factor out commonalities between the two current supported echosounders so that future implementations can be accomplished quickly and in a non-brittle way.
**Convert:**

- convertbase.py added to store platform information
- Updated AZFP to include Matlab 1.5 code

**Model:**
- Add 'self.calc_seawater_absorption()' to EK60 model to match AZFP code.
- TVG and ABS values are calculated when they are needed.
- Changing pressure, temperature, and salinity will recalculate sound_speed, sea_absorption, range, and sample_thickness for AZFP
- AZFP Sv and TS tolerances fixed
- AZFP temperature is averaged by default
- self.seawater absorption moved to modelbase

**Utils:**
Added new underwater acoustics module. Designed to provide static, utility methods currently for calculating underwater absorption and sound speed. This removes dependency on ARLPY.

-------------
The second group of commits include the following changes:
- Add angle data into the parsed output, note using `alongship`/`athwartship` terminology instead of the `major`/`minor` terminology specified in the SONAR-netCDF4 convention to avoid confusion.
- Change `major`/`minor` terminology in receiving and transmitting beamwidth to `alongship`/`athwartship` terminology
- Add Sv data set from Matlab EK60 data unpacking code for testing.
- Add an explicit method `modelbase:recalculate_environment` to recalculate sound speed and absorption, which triggers updates of range and sample thickness.
- Revise TVG calculation for EK60 to conform with Matlab code implementation.


ngkavin:
I merged with the latest version and fixed conflicts, also added the matlab update but I could not push because "Git LFS on github.com does not currently support pushing LFS objects to public forks" so I removed the LFS file. Not sure how to deal with it.

leewujung:
Thanks @ngkavin. I'll merge your PR and #69. Will also take a look at the LFS setup.

leewujung:
@ngkavin: the `platform_name`, `platform_type` and `platform_code_ICES` in the `ConvertBase` class right now are only meaningful/functional in the context of `ConvertAZFP`. Can you revise and merge these properties in `ConvertEK60` into `ConvertBase` and make sure the values gets pushed into the right place when saving to .nc file?

ngkavin:
The pull request is updated with new commits

PR #81, opened 2019-12-26
Convert changes

ngkavin:
EK60 backscatter, range bin, and ping times are no longer saved as pieces in a single .nc or .zarr file. They are split into multiple files with _part_2, _part_3, ... appended to the filename. These files are standalone so they contain other groups.

This pull also supports combining multiple EK60 or AZFP raw files into a single .nc or .zarr file.


leewujung:
@ngkavin: I added a few commits:
- change how the splitting was done based on your code (by indexing in numpy array directly)
- move the data splitting code into a separate method `split_by_range_group`
- fix a bug in the indexing for splitting `ping_time`
- fix a bug in splitting `angle_dict` -- before there was a typo and was actually saving from `power_dict`
- also split the transmit signal parameters and sample interval (`tx_sig` in the code before -- now the operations are under `split_by_range_group`)
- add `beam['angle_dict']` to be encoded as data variable `split_beam_angle` in `set_groups_ek60/set_beam`. I made up this data variable name since it's not in the convention.
- move the initialization of self.save_path, self.nc_path, self.zarr_path to the convertbase parent class

There's still an exception handling case for angle_dict that I need to hunt down. It's related to some of the channels may not be split-beam. I'll push the commit and merge this PR once I got that figured out.

leewujung:
@ngkavin I am done for now for the angle_dict business. There are 2 TODOs you can see in convert/ek60.py that is related to changing methods in set_groups_ek60 to accommodate some parameters if they vary across pings within the same file, but we can leave them there for now, since it happens rarely.

The current behavior that intends to split data into two .nc files (if there are 2 different range_bin lengths) actually spit out 3 files: filename.nc (which contains all groups except for `Beam`), and filename_part01.nc and filename_part02.nc (which contain only the `Beam` group for the 2 range_bin lengths). I know that I added "_part01", but this happens because your use of beam_dict['path'] overwrites line 135 in `set_groups_ek60` that checks file existence first, otherwise my change of adding "_part01" should error out.

Please change the behavior to be splitting data into two .nc files, with each file having the same values for all other groups but different `Beam` group data. This is because each of the files needs to be self-contained. This means that you should also slice pitch, heave, and roll in the `Platform` group according to `ping_time` (check what I fixed with power_dict and angle_dict to make sure you have the right indices). For latitude and longitude, slice `location_time` based on the time points included in the `range_bin`

echopype/convert/convert.py
```diff
@@ -24,15 +24,23 @@ def Convert(path='', xml_path=''):
...
-
     if path:
-        if not os.path.isfile(path):
-            raise FileNotFoundError(f"There is no file named {os.path.basename(path)}")
+        if isinstance(path, list):
+            file_name = os.path.basename(path[0])
+            ext = os.path.splitext(file_name)[1]
+            for p in path:
+                if not os.path.isfile(p):
+                    raise FileNotFoundError(f"There is no file named {os.path.basename(path)}")
```
leewujung: do you mean `p` here where you have `path`?

echopype/convert/convert.py
```diff
@@ -24,15 +24,23 @@ def Convert(path='', xml_path=''):
...
+                if os.path.splitext(p)[1] != ext:
+                    raise ValueError("All files are not in the same format")
```
leewujung: This seems a weird sentence, better phrased as "Not all files are in the same format."

echopype/convert/ek60.py
```diff
@@ -169,75 +160,102 @@ def _read_datagrams(self, fid):
...
+        # Split data based on range_group (when there is a switch of range_bin in the middle of a file)
+        self.split_by_range_group()
 
         # Trim excess data from NMEA object
         self.nmea_data.trim()
 
-    def save(self, file_format):
-        """Save data from EK60 `.raw` to netCDF format.
+    def save(self, file_format, compress=True):
+        """Save data from raw 01A format to a netCDF4 or Zarr file
```
leewujung: here should be "Save data from .RAW format to a netCDF4 or Zarr file"

echopype/convert/ek60.py
```diff
@@ -304,131 +329,108 @@ def _set_nmea_dict():
...
 
-        # Get exported filename
-        filename = os.path.splitext(os.path.basename(self.filename))[0]
-        self.save_path = os.path.join(os.path.split(self.filename)[0], filename + file_format)
-        self.nc_path = os.path.join(os.path.split(self.filename)[0], filename + '.nc')
-        self.zarr_path = os.path.join(os.path.split(self.filename)[0], filename + '.zarr')
+        def copyfiles():
+            self.all_files = []
+            split = os.path.splitext(self.save_path)
+            for piece in range(len(self.range_lengths)):
```
leewujung: this `piece` shadows the same name from outer scope, better to change to something else to avoid confusion

PR #82, opened 2020-01-03
Update modelbase and docs

leewujung:
- update get_MVBS, noise_estimate, and noise removal in modelbase to use xarray apply-map.
- fixed bugs in .calibrate for ek60 after environmental parameters are updated
- change default sound speed calculation to using Mackenzie and absorption calculation to Ainslie and McColm
- remove rzinfo to get rid of warnings from np.datetime64 conversion that is time zone naive


leewujung:
@ngkavin please merge these with your model class changes. Thanks!

PR #94, opened 2020-01-22
add ek60 TS calculation

### LINKED ISSUE #92: Add TS calculation in `ModelEK60`
`ModelAZFP` already has the method to calculate TS. We need this for `ModelEK60` as well just as we already have for calculating Sv via `.calibrate()` for data from both echosounders.

ngkavin:
calibrate_TS can now be called on EK60 like for AZFP

closes #92


leewujung:
@ngkavin could you also add it as a parent class method? Thanks.

PR #95, opened 2020-01-31
Add __init__ to fix utils module import error

lsetiawan:
## Overview

I was getting  `ModuleNotFoundError: No module named 'echopype.utils'` when trying to import echopype. This PR should fix it. `utils` didn't have `__init__.py` which caused python to not know it's a module. I added one, and now things seem fine.


leewujung:
Great, thanks @lsetiawan!

Issue #97, opened 2020-02-04
Allow combining multiple raw files when converting to nc/zarr


leewujung:
Add option for both netcdf and zarr to combine multiple input raw files into 1 big file.
   - under this option, default to combining all files passed in as a list into 1 big .nc file
   - give user the option to save all data within each month to individual big .nc files
   - something like below:
      ```python
      fileconvert = Convert(FILE_LIST)  # pass in a list of files to be converted

      # default to convert each raw file to 1 .nc file
      fileconvert.raw2nc(combine_opt=None)  

      # convert all data in input raw files to 1 big .nc file
      fileconvert.raw2nc(combine_opt='all')
      ```

Currently this is partially implemented: we can convert multiple files into one output file at once,
BUT the converted file use the filename of the first file. Need to change it such that user can specify the combined output filename. The filename part is related to #88 .

leewujung:
I just realized that the way the file unpacking works right now it may be hard to do `combine_opt=None`. @ngkavin is that the case?

In terms of user experience, there doesn't seem to be a difference between passing in a list of files and specify an output filename, vs passing in one file at a time if you just want one output file from each raw file. The difference though is that the Convert object has to be destroyed and created for every file, in the second case. Thoughts?

leewujung:
Ok, let's do
```python
fileconvert = Convert(FILE_LIST)  # pass in a list of files to be converted

# default: convert each raw file to 1 .nc file
fileconvert.raw2nc(combine_files=False)  

# optional: convert all data in input raw files to 1 big .nc file
fileconvert.raw2nc(combine_files=True)
```

Issue #89, opened 2020-01-20
Allow over-writing existing nc file


leewujung:
Users sometimes need to over-write an existing .nc file that has been converted before. Currently the convert module will issue a warning and abort if there is already an existing file with the same filename as the raw binary file. This is a good default behavior but should give user this option.

Something like:
```python
tmp = Convert(DATA_FILE)
tmp.raw2nc(overwrite=True)   # default overwrite=False
```

PR #99, opened 2020-02-13
Handle multiple files and allow overwriting files

### LINKED ISSUE #88: Save generated files at user-specified path
Currently all files generated from raw binary file to netCDF conversion as well as calibration and MVBS calculation are saved in the same folder as the raw binary files. This is not very convenient to use. We should allow users to specify the path to save generated files.

For conversion, this can simply be:
```python
tmp = Convert(RAW_FILE)
tmp.raw2nc(save_path='PATH_TO_SAVE_NC_FILE')
```
For Sv and MVBS calculation, this can probably be:
```python
tmp = EchoData(NC_FILE)
tmp.calibrate(save=True, save_path='PATH_TO_SAVE_SV_FILE')
tmp.get_MVBS(save=True, save_path='PATH_TO_SAVE_MVBS_FILE')
```

leewujung:
This merges #98 and closes #88, #89, #97 .

@ngkavin : I added handling of the case when `save_path=FILENAMEONLY.nc` and use the input filename directory as the save-to directory. Also consolidate everything to `.validate_path()` -- testing `save_path now` is within that function.

### DIFF
--- a/echopype/convert/azfp.py
+++ b/echopype/convert/azfp.py
@@ -263,8 +263,14 @@ class ConvertAZFP(ConvertBase):
                     raise ValueError(f"Header value {field} is not constant for each ping")
         self._checked_unique = True
 
-    def parse_raw(self):
-        """Parses a raw AZFP file of the 01A file format"""
+    def parse_raw(self, raw):
+        """Parses a raw AZFP file of the 01A file format
+
+        Parameters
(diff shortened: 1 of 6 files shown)


leewujung:
Try again: this PR closes #89. This PR also closes #97 .

PR #112, opened 2020-03-18
Change to how multiple files are combined into one .nc or .zarr

ngkavin:
When combining files, all raw files are converted to .nc files in a temporary folder. Then the files are combined by using `xr.open_mfdataset()` and saving each group into a single .nc file. The temporary folder is then deleted.


leewujung:
@ngkavin : Can you write the 2 unresolved issues on the repo about 1) the serialization warning and 2) appending mode for zarr, and reference those here?

Right now we are taking the route of creating multiple temp files and combine them to for the final single file, but it is better in terms resources to do the appending once it the append_dim bug is fixed for zarr.

Thanks!

leewujung:
This PR contains unresolved issues #113 and #114, the latter will allow us to just append the new files to the first created file without creating whole bunch of temp files. Let's loop back to this afterwards.

echopype/model/modelbase.py
```diff
@@ -387,7 +387,8 @@ def remove_n(x, rr):
         if np.unique([np.array(x).size for x in range_bin_tile_bin_edge]).size == 1:
             Sv_clean = pp.groupby_bins('ping_idx', ping_tile_bin_edge).\
                             map(remove_n, rr=range_bin_tile_bin_edge[0])
-            Sv_clean = Sv_clean.drop_vars(['ping_idx', 'ping_idx_bins'])
+            # Sv_clean = Sv_clean.drop_vars(['ping_idx', 'ping_idx_bins'])
```
leewujung: You can just remove this commented line so it doesn't conflict with master.

leewujung:
It seems that the splitting of subset of files into part01, part02, ...  (due to dimension change in `range_bin`)would error in the current implementation. There's also redundancy that may be able cut down by factoring out the operations in `export()` as well as the set_group subfunctions. @ngkavin let's discuss these in our next call.

leewujung:
Merge this now so we can do LFS migration. We'll refactor the code structure afterwards.

PR #115, opened 2020-03-25
Implement EK80 Calibration

ngkavin:
`calibrate()`  has been implemented allowing calibration of Sv and TS using backscatter data from either complex broadband backscatter data or real continuous wave power data. The Sv and TS datasets can be saved to a .nc file similarly to the `calibrate()` methods for EK60 and AZFP.


leewujung:
Merge to EK80 branch for now to allow LFS migration.

PR #120, opened 2020-03-31
Implements angle testing for ek60

ngkavin:
Adds test .nc file containing power and angle data for ek60.


leewujung:
Merge for now to allow LFS migration.

PR #124, opened 2020-04-20
Convert refactor ek80

ngkavin:
- Implement combining raw files into a single NetCDF or Zarr.
- Reorganized convert functions so that it mirrors the refactored EK60.
- Passes EK80 and EK60 test, but not AZFP tests.


leewujung:
Merging this to convert-refactor-ek60 so all refactoring-related updates are in one branch in the master repo regardless of where they come from.

PR #125, opened 2020-04-22
Ek80 model update

leewujung:
This PR 
1. cleans up object initialization for ModelEK80 
2. factors out the generation of transmit signal and pulse compression as separate methods outside of `.calibrate()` to `.calc_transmit_signal()` and `.pulse_compression()`


leewujung:
@ngkavin : could you take charge of resolving the conflicts and merging this? I don't want to break what you're doing with the refactoring. Thanks!

Issue #132, opened 2020-04-28
Fix units of coordinates in EK80 beam group


leewujung:
Currently if read EK80 Beam group data the coordinate `frequency` and `range_bin` will be attached with wrong units. 

Need to:
- explicitly set the units for `frequency`
- figure out what units to use for `range_bin`

Issue #121, opened 2020-04-01
Redundant data in NMEA group


leewujung:
At the moment the entire NMEA group is copied for all files that are split due to changes in `range_bin` in the middle of the raw file. The NMEA datagrams should be split into each of the `part01`, `part02`, ... files. Let's take care of this once we refactor the code that combines/appends data from multiple files.

leewujung:
@ngkavin : as we discussed yesterday!

Issue #13, opened 2018-08-30
RAW data parser (for ek80 code)


emma-ozanich:

leewujung:
Almost there! 😎

@ngkavin I am noting the below items here for us to keep track of things:
- Currently ping time is saved as float64 but should be datetime64.
- Split-beam angle data for CW mode EK80 files are not currently saved. Note can reference EK60 data handling of this.

leewujung:
Update: the above is being addressed in the combine-refactor branch.

PR #136, opened 2020-05-02
Ek60 split nmea

ngkavin:
Reduces redundancy with regards to storing NMEA data when files are split up due to changing range_bin lengths.


ngkavin:
Also fixes 'units' being listed as a coordinate.

echopype/convert/utils/set_groups_ek80.py
```diff
@@ -294,7 +295,7 @@ def set_beam(self, beam_dict):
                 bb = xr.Dataset(
                     {'backscatter_r': (['frequency', 'quadrant', 'ping_time', 'range_bin'], beam_dict['backscatter_r'],
                                        {'long_name': 'Real part of backscatter power',
-                                        'units': 'dB'}),
+                                        'units': 'Hz'}),
```
leewujung, line 298: This is backscatter so obviously not in units Hz, but also not in dB. Since these are raw files, perhaps the best is just set it to volts as the WBT measures voltage over a 1 K
 load.

leewujung:
This PR:
- fixes `ping_time` units problem, which closes #132 and also closes closes #13
- splits NMEA groups correctly, which closes #121

PR #141, opened 2020-05-04
Fix deps

ocefpaf:
@leewujung this is the first step to help with `echotype` installation. I ran `depfinder` and updated the `requirement.txt`, added a `requirement-dev.txt` that will help us with more tests later, and fixed a mino issue with one of the files that prevented AST parsing.

PS: here is `depfinder` report on the dependencies:

```
'builtin': ['__future__',
             'collections',
             'configparser',
             'datetime',
             'distutils',
             'errno',
             'io',
             'json',
             'logging',
             'math',
             'os',
             're',
             'shutil',
             'struct',
             'subprocess',
             'sys',
             'warnings',
             'xml'],
 'questionable': ['ConfigParser', 'cx_Freeze', 'py2exe', 'setuptools'],
 'relative': ['_version',
              'azfp',
              'convert',
              'convertbase',
              'echodata',
              'echogram',
              'ek60',
              'ek60_date_conversion',
              'modelbase',
              'set_groups_azfp',
              'set_groups_base',
              'set_groups_ek60'],
 'required': ['echopype',
              'matplotlib',
              'netcdf4',
              'numpy',
              'pynmea2',
              'pytz',
              'setuptools',
              'versioneer',
              'xarray',
              'zarr']}
```


echopype/convert/utils/ek60_date_conversion.py
```diff
@@ -1,4 +1,4 @@
-﻿"""
```
ocefpaf, line 1: There was a hidden BOM character here preventing us from parsing this file.

environment.yml
```diff
@@ -1,23 +1,21 @@
...
 dependencies:
+  - python=3.8
+  - matplotlib
+  - netCDF4
+  - numpy
+  - pynmea2
+  - pytz
+  - xarray
+  - zarr
+  # extras
```
ocefpaf, line 13: Everything above is required and I kept the rest as "extras" b/c I'm not sure how this file is used. I usually don't avise people to add env files to the repo to avoid confusion. It is hard to keep everything in sync that way. Also, for source code repos I recommend to use the same req files with something like:

```shell
 conda create --name ECHOTYPE python=3.8 --file requirements.txt --file requirements-dev.txt
```

for example.

requirements-dev.txt
```diff
@@ -0,0 +1,21 @@
...
+pylint
+pytest
+pytest-cov
+pytest-flake8
+pytest-xdist
+setuptools_scm
+sphinx
+sphinx-automodapi
+twine
+wheel
```
ocefpaf, line 21: After this is merged I'll add a few tests that will use these req deps. Most of them is to check the tarball and inconsistencies that will make the releases more robust or, at least, easier to debug.

environment.yml
leewujung: I see, makes sense. Most below are for running the example notebooks. Some of them are not actually needed and can be deleted I think (e.g. click, pandas). I guess this is consistent with the suggestion that the notebooks should be separated out to another repo.

ocefpaf: Or this file belongs in the notebook folder. In the root folder it can be confused with a work/test/build env.

requirements-dev.txt
leewujung: Cool! Thanks! 

Question before I merge this: I had to specify sphinx-automodapi and numpydoc because I couldn't get readthedocs to build correctly without them in the requirements.txt file (from what I found it seems that allowing specifying requirements under docs started in v2.0, but RTD didn't support that version then). I could look into if that has changed. But should we try to keep these two in the requirements.txt for now?

environment.yml
leewujung: Sounds good!

requirements-dev.txt
ocefpaf: We can added them as extras-docs in the `setup.py`. I'm not sure how RTD works but that should do it. Let me check...

.readthedocs.yml
```diff
@@ -20,7 +20,7 @@ formats: []
 python:
    version: 3.7
    install:
-      - requirements: requirements.txt
+      - requirements: docs/rtd-requirements.txt
```
ocefpaf, line 23: @leewujung looks like this is the suggested way to have a single req file for RTD. You still need to keep the docs/requirement.txt updated, but only for doc deps.

leewujung: Awesome, thanks! let me give it a try in a test RTD doc. Be back after a meeting...

leewujung: I updated the RTD min required version to 2.0, but haven't gotten the automodapi part to work : [build report](https://readthedocs.org/api/v2/build/10966609.txt). Was testing in [this branch](https://github.com/leewujung/echopype/tree/ocefpaf-fix_deps).

ocefpaf: The build report link opens a "not found" page for me :-/

leewujung: Sorry I guess that's probably somehow a private link in RTD: [new link](https://www.dropbox.com/s/9ajj1ag429qw6ov/10966609.txt?dl=0). 

After requesting to use sphinx>=2.0 (although RTD bulids using v1.8.3), it seems that automodapi is installed correctly. However there's an error for not finding content.rst, which is a little mysterious, because in `conf.py` I did set `master_doc = 'index'` and that should ask sphinx to use `index.rst`. 

P.S. The docs build locally.

ocefpaf: `docs/contents.rst` does nto exist on my fork :-/
I'm not sure how RTD works but this seems unrelated to this PR, no?

leewujung: What I meant was that there shouldn't be a need for a `docs/contents.rst`...
It is related to this PR, because the docs will fail to build if I merge it. 
Could you put sphinx-automodapi and numpydoc back to requirements.txt in the root level for now, and raise an issue on how they shouldn't be in there? I can make that issue too...

ocefpaf: Ah. You run `sphinx-automodapi` to create that content, right?

leewujung: ? I think it autogenerates the context with the directive automodapi.

leewujung:
@ocefpaf just wanted to let you know that I haven't merged this PR because there's something else needed to be fixed for the automodapi-sphinx problem, but I haven't gotten around to do it.

PR #143, opened 2020-05-13
Adds zarr capability for processing.

lsetiawan:
## Overview

This PR adds netcdf/zarr switching based on file extension, this should enable zarr capability on process.


leewujung:
@ngkavin could you test and merge this? The changes should still pass the testing functions you're adding. Thanks! :)

Issue #145, opened 2020-05-20
ek80_raw_parsers.SimradXMLParser._unpack_contents fails on two-in-one transducers


FletcherFT:
Two-in-one transducers (such as the [Simrad ES38-18/200-18C](https://www.kongsberg.com/globalassets/maritime/simrad/transducers/38-and-es38/es38-17200-18c/430754aa_es38_18_200_18c_data_sheet_english.pdf)) appear in the transceiver XML tree with two transducers and two channels, but both transducers have the same SerialNumber attribute (used for matching with a channel's TransducerSerialNumber attribute).

The current implementation overwrites the first channel with the second transducer's configuration instead of assigning the second channel with the second transducer. Additionally there's no current alternative way to match a channel with a transducer, so one has to assume that the first reported channel matches with the first reported transducer, which might not be true.

Any ideas on what information can be used to match them? Would it be easier to adjust the Serial Number attributes in the .raw file so that they are separately identifiable?

I'm trying to address this bug, but it might fail on the original ek80 tests. If I get my two-in-one test working with the unit test, then I'll make a pull request.

PR #146, opened 2020-05-20
Bugfix/two in one transducer compatibility

FletcherFT:
This is a proposed fix for #145 . There is no known alternative for matching a transducer to a channel apart from serial number (which is the same for 2-in-1 transducers). To determine if there is a 2-in-1 transducer attached to a transceiver, I compare the number of detected transducers to the number of transducers attached to the transceiver. If they are unequal than there is likely to be a combination transducer present.

My work-around hack is to look at the name of the channel_short_id attribute and extract the frequency from it with regex, then compare that frequency with the frequency attribute of the transducer element. If they are matching then update the configuration datagram with the transducer information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ OOI-D20170821-T081522.raw
 2015843-D20151023-T190636.raw
 D20170912-T234910.raw
 echopype/test_data/ek80/*.raw
+venv/
--- a/echopype/convert/ek80.py
+++ b/echopype/convert/ek80.py
@@ -318,7 +318,8 @@ class ConvertEK80(ConvertBase):
 
         # Find largest dimensions of array in order to pad and stack smaller arrays
         max_samples = 0
-        max_splits = max([n_c for n_c in self.n_complex_dict.values()])
+        # TODO How to determine if a CW data set is split beam or single beam, and how many splits?
+        max_splits = max([n_c for n_c in self.n_complex_dict.values()]) if bb else 4
         for tx in ch_ids:
             if bb:
                 reshaped = np.array(self.complex_dict[tx]).reshape((ping_num, -1, self.n_complex_dict[tx]))
--- a/echopype/convert/ek80.py
+++ b/echopype/convert/ek80.py
@@ -26,6 +26,7 @@ class ConvertEK80(ConvertBase):
         self.power_dict = {}    # dictionary to store power data
         self.angle_dict = {}    # dictionary to store angle data
         self.complex_dict = {}  # dictionary to store complex data
+        self.n_complex_dict = {}  # dictionary to store the number of beams in split-beam complex data
         self.ping_time = []     # list to store ping time
         self.environment = {}   # dictionary to store environment data
         self.parameters = defaultdict(dict)   # Dictionary to hold parameter data
@@ -120,6 +121,8 @@ class ConvertEK80(ConvertBase):
                         self.power_dict[ch_id].append(tmp_datagram_dict[i]['power'])  # append power data
                         self.angle_dict[ch_id].append(tmp_datagram_dict[i]['angle'])  # append angle data
                         self.complex_dict[ch_id].append(tmp_datagram_dict[i]['complex'])  # append complex data
+                        if self.n_complex_dict[ch_id] < 0:
+                            self.n_complex_dict[ch_id] = tmp_datagram_dict[i]['n_complex']  # update n_complex data
 
             # NME datagrams store ancillary data as NMEA-0817 style ASCII data.
             elif new_datagram['type'].startswith("NME"):
@@ -167,6 +170,7 @@ class ConvertEK80(ConvertBase):
                 self.power_dict[ch_id] = []
                 self.angle_dict[ch_id] = []
                 self.complex_dict[ch_id] = []
+                self.n_complex_dict[ch_id] = -1
 
                 # Parameters recorded for each frequency for each ping
                 self.parameters[ch_id]['frequency_start'] = []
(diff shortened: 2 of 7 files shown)


leewujung:
@FletcherFT : thanks for the issue and PR! It's awesome that you added a test file and function too. :) 

Could you please look into what caused the complaint on NaN padding [reported by Travis](https://travis-ci.org/github/OSOceanAcoustics/echopype/builds/689220466#L260)?

It'll take @ngkavin and I a few days to get to the frequency matching approach you propose, but it seems a nice adaptive solution that would guarantee/find the channel-transducer matching, as long as this pattern is consistent for all other files.

FletcherFT:
@leewujung all fixed and passing now. I've put a TODO where I think there is an assumption about the number of split beams in the data.

emlynjdavies:
Hi, I have an error that the serial number is not found in some data from a WBT-Mini with two transducers. It seems that one of the transducers did not record correctly, but I get a key error here:

"echopype/convert/utils/ek_raw_parsers.py", line 826, in _unpack_contents
    if (td.attrib['SerialNumber'] == c.attrib['TransducerSerialNumber']):
KeyError: 'SerialNumber'

Is this a similar issue?

leewujung:
@emlynjdavies Yes, this does seem like the same issue. We have a file that seems to give out the same error, but if you could give us the file that errors out on your end that will be very helpful.

@FletcherFT Thanks for fixing the bug! @ngkavin and I discussed this yesterday and we thought that with the issue @emlynjdavies reported, we'll try the approach to check if recording from a certain transducer is first checked before doing the frequency matching you put in. We also have a file that seemed to not have the SerialNumber recorded properly, so that would also need to be checked first, and use the default sequence of frequency and channel if needed.

Regarding [e6a9c18](https://github.com/OSOceanAcoustics/echopype/pull/146/commits/e6a9c18d337812ddf63425934bf3dc74aac16bbd), we would need to revert the variable name `range` in Sv, as it is used in other places and it seems to make more sense without the s as the name of the dimension, like for across the channels, the dimension is `frequency` and not `frequencies`.

emlynjdavies:
> @emlynjdavies Yes, this does seem like the same issue. We have a file that seems to give out the same error, but if you could give us the file that errors out on your end that will be very helpful.
> 
> @FletcherFT Thanks for fixing the bug! @ngkavin and I discussed this yesterday and we thought that with the issue @emlynjdavies reported, we'll try the approach to check if recording from a certain transducer is first checked before doing the frequency matching you put in. We also have a file that seemed to not have the SerialNumber recorded properly, so that would also need to be checked first, and use the default sequence of frequency and channel if needed.
> 
> Regarding [e6a9c18](https://github.com/OSOceanAcoustics/echopype/pull/146/commits/e6a9c18d337812ddf63425934bf3dc74aac16bbd), we would need to revert the variable name `range` in Sv, as it is used in other places and it seems to make more sense without the s as the name of the dimension, like for across the channels, the dimension is `frequency` and not `frequencies`.


Thanks! Here is an example file I am having problems with.
[https://drive.google.com/file/d/1gMHmjtm3vCHHretS3eG9rNL5_d-Ys3Ul/view?usp=sharing](url)

leewujung:
Thanks @emlynjdavies for the example file!

It appears that we're dealing with 2 problems here: 
1. Compatibility with WBT mini files
2. Compatibility with files from 2-in-1 transducer

The example in our original test file was from WBT. WBT mini saves data slightly differently, in terms of how different transducer channels are grouped under the transceiver section of the configuration XML datagram.

@FletcherFT and @emlynjdavies could you confirm that your data are indeed from WBT mini?

Thanks!

emlynjdavies:
> Thanks @emlynjdavies for the example file!
> 
> It appears that we're dealing with 2 problems here:
> 
> 1. Compatibility with WBT mini files
> 2. Compatibility with files from 2-in-1 transducer
> 
> The example in our original test file was from WBT. WBT mini saves data slightly differently, in terms of how different transducer channels are grouped under the transceiver section of the configuration XML datagram.
> 
> @FletcherFT and @emlynjdavies could you confirm that your data are indeed from WBT mini?
> 
> Thanks!

Yes, this is from a WBT mini with a split 200kHz and a 333kHz. (I think the 200 was not configured correctly, but the 333 seemed to work).

FletcherFT:
I’m also working with a WBT-mini transceiver. Transducer is 2-in-1 combi transducer (38 kHz split + 200 kHz single).

leewujung:
@FletcherFT : could you check if the parsed results of [this branch](https://github.com/leewujung/echopype/tree/two-in-one) are correct? I tested the code against the file you provided. I couldn't push to yours or @ngkavin's fork directly, seems to be something about git LFS permission.

I took into account other files that I could get my hands on, and the matching of transducer info under the `Transducers` section of the XML to the correct channels under the `Transceivers` section of the XML is rather messy: the worst file I saw had 2 two-in-one transducers and the only thing that distinguishes them is part of the transceiver channel_id that somehow makes its way in there! That's not always the case: not in @emlynjdavies's file for example.

@emlynjdavies : the config section now works for your data, but errors out in a later part for your file 😕, probably because of the problem you mentioned about the 200kHz not being set up correctly. I'll move that to a separate issue to see if how to give error messages but still saves the data from the 333 kHz channel.

leewujung:
@emlynjdavies : could you open up a new issue and put the error messages you have from using code in [this branch](https://github.com/leewujung/echopype/tree/two-in-one) there? Please also include information on how the 200 kHz channel is not set up correctly there, so that we have better ref for printing out error messages when the raw files are not correctly recorded. Thanks!

emlynjdavies:
> @emlynjdavies : could you open up a new issue and put the error messages you have from using code in [this branch](https://github.com/leewujung/echopype/tree/two-in-one) there? Please also include information on how the 200 kHz channel is not set up correctly there, so that we have better ref for printing out error messages when the raw files are not correctly recorded. Thanks!

Sure. :)

FletcherFT:
> @FletcherFT : could you check if the parsed results of [this branch](https://github.com/leewujung/echopype/tree/two-in-one) are correct? I tested the code against the file you provided. I couldn't push to yours or @ngkavin's fork directly, seems to be something about git LFS permission.
> 
> 
> 
> I took into account other files that I could get my hands on, and the matching of transducer info under the `Transducers` section of the XML to the correct channels under the `Transceivers` section of the XML is rather messy: the worst file I saw had 2 two-in-one transducers and the only thing that distinguishes them is part of the transceiver channel_id that somehow makes its way in there! 🤯 That's not always the case: not in @emlynjdavies's file for example.
> 
> 
> 
> @emlynjdavies : the config section now works for your data, but errors out in a later part for your file 😕, probably because of the problem you mentioned about the 200kHz not being set up correctly. I'll move that to a separate issue to see if how to give error messages but still saves the data from the 333 kHz channel.
> 
> 

I’m on acoustic survey until mid-July, I’ll check the results around then :)

leewujung:
@FletcherFT I am going to close this now since #178 should have fixed the problem. Feel free to reopen if the issue continues.

PR #151, opened 2020-06-04
Fix RTD issues

leewujung:
This PR fixes the previous problem in using 2 requirement files in root and docs, now that RTD supports version >2.0. This PR also merges #141 .


leewujung:
Looks like RTD updated allowed sphinx version, so now specifying two separate requirement files in `.readthedocs.yml works`. :)

Issue #148, opened 2020-05-27
raw2zarr, raw2nc, save: rename, consolidate?


emiliom:
I was originally going to open this issue as a suggestion to rename `raw2zarr` and `raw2nc` to something like `to_zarr` and `to_netcdf`. I find the `raw2...` names somewhat confusing and misleading. In reality, the export (serialization) is not actually *from* a raw format, but rather from an in-memory xarray dataset, right? In other words, the raw data has already been read and mapped to the data convention in xarray. The names I'm suggesting match xarray export functions. 

But looking at [the code](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/convertbase.py#L129), I see that at this time there's no reason to have two separate export functions `raw2zarr` and `raw2nc`. They're effectively identical, except for the `param` passed to `self.save`. 

I also noticed that `raw2zarr` currently seems to be limited to writing to a file system path and doesn't seem to accommodate writing to object storage like S3 (which xarray's `to_zarr` supports). If that's correct, that's a separate issue. Adding that functionality would be a good reason to maintain two separate functions, as xarray does.

leewujung:
I think changing `raw2zarr` and `raw2nc` to `to_zarr` and `to_netcdf` is a great suggestion. This way it is consistent with xarray and pandas methods naming so familiar to users.

In terms of having separate methods, the conversion to zarr and netcdf needs to be separate right now because the way to combine data from multiple raw files into a single zarr or nc file is different for zarr and netcdf under xarray (`to_netcdf` does not allow appending variables directly). @ngkavin : could you finalize the block diagram we went through together last time and upload it to under /docs/img so that we can discuss further based on that? Thanks!

For the `raw2zarr` method to accommodate writing to an object storage, under the hood it is using xarray's `to_zarr` method, so the `validate_path` method is probably where the tweaking needs to go. @emiliom : could you raise a separate issue on this? Thanks!

emiliom:
> I  think changing raw2zarr and raw2nc to to_zarr and to_netcdf is a great suggestion. This way it is consistent with xarray and pandas methods naming so familiar to users.

Thinking through things a bit further, I wonder if reusing the names from xarray could cause confusion, if the parameters are actually different ... I'll look into it further, but not right away.

[...]

PR #154, opened 2020-06-12
Save EK80 power data for all channels that record it

ngkavin:
EK80 no longer checks for all frequency channels before saving power. Backscatter data is saved as it comes in.


leewujung:
Thanks @ngkavin! The same checking structure is in convert/ek60.py also, but let's hold on to change that, and have a discussion on how much overlap/divergence between the two there is to see if we can factor out a cleaner structure, in conjunction with addressing #148 . I'll make an issue about this.

PR #177, opened 2020-06-30
save toplevel when combining multiple files

### LINKED ISSUE #175: Cannot process file created using combine_opt=True option in raw2nc
I've been exploring the use of echopype for the OOI AZFP data and have created a [notebook](https://nbviewer.jupyter.org/github/cwingard/ooi-data-explorations/blob/echopype/python/examples/notebooks/zplsc/CE01ISSM-00004%20Echogram%20for%202015-10-09.ipynb) showing the steps I followed based off of your example notebooks and [documentation](https://echopype.readthedocs.io/en/latest/).

When using:

```python
dc = Convert(list_of_files, 'calibration.xml')
dc.raw2nc(combine_opt=True, 'combined_data.nc')
```
to create a single, combined data file for a group of files, I get an error when trying to then load the file via the `Process` function.

```python
AttributeError                            Traceback (most recent call last)
~\Anaconda3\envs\ooi\lib\site-packages\echopype\process\process.py in Process(nc_path)
     41         try:
---> 42             echo_type = nc_file.keywords
     43         except AttributeError:

~\Anaconda3\envs\ooi\lib\site-packages\xarray\core\common.py in __getattr__(self, name)
    232         raise AttributeError(
--> 233             "{!r} object has no attribute {!r}".format(type(self).__name__, name)
    234         )

AttributeError: 'Dataset' object has no attribute 'keywords'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-9-f17411c69779> in <module>
      1 # Using the combined data file for processing
----> 2 data = Process(os.path.join(save_dir, '20151009.nc'))

~\Anaconda3\envs\ooi\lib\site-packages\echopype\process\process.py in Process(nc_path)
     42             echo_type = nc_file.keywords
     43         except AttributeError:
---> 44             raise ValueError("This file is incompatible with echopype functions.")
     45 
     46     # Returns specific Process object

ValueError: This file is incompatible with echopype functions.
```

ngkavin:
When combining raw files into a single netcdf file, the toplevel was not being saved.

This PR resolves #175


echopype/convert/ek80.py
```diff
@@ -545,6 +546,31 @@ def _export_zarr(self, save_settings, file_idx=0):
...
+        # Create the variables in the file
+        for k, v in ds_vend.variables.items():
+            data = np.empty(len(v), complex64)
+            var = vdr.createVariable(k, complex64_t, v.dimensions)
+            var[:] = data
+
+        src.close()
+        trg.close()
+
     def _combine_files(self):
```
leewujung, line 574: I think the name is different from the parent class, so the method is not inherited correctly?

echopype/convert/ek80.py
```diff
@@ -545,6 +546,31 @@ def _export_zarr(self, save_settings, file_idx=0):
         else:
             self._set_groups(raw_file, out_file, save_settings=save_settings)
 
+    def _copy_group(self, src_file, trg_file):
```
leewujung, line 549: Could you give this method a better name so that it is more descriptive of what the function is? It is hard coded to only copy the Vendor group so the name does not make sense.

echopype/convert/ek80.py
```diff
@@ -545,6 +546,31 @@ def _export_zarr(self, save_settings, file_idx=0):
...
+        # Utility function for copying the filter coefficients from one file into another
+        src = netCDF4.Dataset(src_file)
+        trg = netCDF4.Dataset(trg_file, mode='a')
```
leewujung, line 552: Could you check if the saved data variables can be read correctly using xarray? I am also wondering why you choose to use netCDF4 here and not xarray?

ngkavin: xarray does not handle complex numbers

PR #179, opened 2020-07-10
Save EK80 configuration xml

### LINKED ISSUE #171: Save or export EK80 XML configuration datagram
Currently only the fields of the EK80 XML configuration datagram that correspond to requirements in the SONAR-netCDF convention is saved into netCDF/zarr. However, there may be other parameters that will be useful for users later. Therefore it seems a good idea to give users access to the entire datagram once the data is parsed.

@ngkavin and I discussed the options we'll move forward with:
- [ ] implement a separate method to allow user to export a companion XML file with netCDF/zarr
- [ ] test saving the entire config datagram into the Vendor group

@emiliom @lsetiawan @valentina-s : any suggestions welcome.

ngkavin:
Saves the configuration xml to the vendor group.
Also provides the option for exporting the configuration to an xml file by calling `export_xml()` on a `Convert `object

Closes #171

### DIFF
--- a/echopype/convert/ek80.py
+++ b/echopype/convert/ek80.py
@@ -144,8 +144,6 @@ class ConvertEK80(ConvertBase):
         ----------
         raw : str
             raw filename
-        export_xml : bool
-            whether or not to export the configuration as an xml file
         """
         print('%s  converting file: %s' % (dt.now().strftime('%H:%M:%S'), os.path.basename(raw)))
         with RawSimradFile(raw, 'r') as fid:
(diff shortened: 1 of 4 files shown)


leewujung:
@ngkavin : in the current implementation `ConvertEK80.load_ek80_raw()` is called again to parse the entire set of files the second time to export the XML file. This is not very efficient. I can see that this is because you try to do it as part of `ek_raw_parsers.py` and `ek_raw_io.py`. 

What about instead of changing those low-level classes, just make `ConvertEK80.export_xml()` to save the XML string since you already have it in `self.config_datagram['xml']` after the file is parsed the first time?

Issue #185, opened 2020-07-23
Handle missing GPS and/or MRU data


ngkavin:
For some EK80 data files, pitch, roll, heave, and NMEA data are not recorded, so they are saved as empty lists. When creating an xarray dataset, pitch, roll, heave, lat, and lon are assumed to have coordinates of time, so an error is raised when these values are empty.

The best solution is likely to fill empty pitch, roll, heave, lat, and lon values with NaNs with length of ping_time so that it can be saved to the xarray dataset.

leewujung:
Ugh, upon digging into the code some more, I found a bug 😎

Under the MRU datagram ([here](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/ek80.py#L133))
`self.mru_data['timestamp']` records the actual MRU timestamp
but when saving to platform group, we currently have 
`out_dict['ping_time'] = self.ping_time` 
which is the ping time from the RAW datagram (echo data) ([here](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/ek80.py#L266)).

Lat/lon have another time base from the NME datagram and that timestamp is now correctly saved into the Platform group.

PR #187, opened 2020-07-23
Fixed Saildrone filename parsing and water_level_draft

Chuck-A:
* Added regex pattern search for grabbing file timestamp data.
* added check to see if water_level_draft in environment dictionary before adding to out_dict.

This still throws errors when calling raw2nc method but they are related to format and array dimensions differences on mru data which are empty when reading Saildrone files. I think this is the place we'll have to pull in the data from the external files.

### DIFF
--- a/echopype/convert/ek80.py
+++ b/echopype/convert/ek80.py
@@ -1,4 +1,5 @@
 import os
+import re
 import shutil
 from collections import defaultdict
 import numpy as np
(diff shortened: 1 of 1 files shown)


leewujung:
> This still throws errors when calling raw2nc method but they are related to format and array dimensions differences on mru data which are empty when reading Saildrone files. I think this is the place we'll have to pull in the data from the external files.

This is documented in #185 . Let's resolve it there.

Issue #183, opened 2020-07-23
Produce depth-aligned Sv data in process


leewujung:
Currently `process.calibrate()` would produce a data cube with dimension frequency x ping_time x range_bin. The `range_bin` dimension is in sample number but not physical units (meter). It'll be more useful if we can produce data cube with dimension frequency x ping_time x **depth**, so let's make a method that does that. 

I can see think of three use cases:

1. Default case:  interpolate to the one with the longest physical depth range
    ```python
    process.resample_to_depth(self, depth_grid=None)  # default
    ```

2. User selects which frequency depth grid they want to interpolate to
    ```python
    process.resample_to_depth(self, depth_grid='70e3')  # specific which frequency channel in [Hz]
    ```

3. User specify a depth grid they want the output to be aligned to
    ```python
    process.resample_to_depth(self, depth_grid=np.arange(0, 100, 0.5))  # specify a depth grid in [m]
    ```

@ngkavin let me know if anything needs clarification. Thanks.

leewujung:
Closing this since we're going to add helper functions for converting `range_bin` to range and depth in physical units in #259.

PR #191, opened 2020-08-01
Added function for producing depth-aligned Sv data

ngkavin:
Addresses #183 by allowing users to go from range bins to range in physical units by resampling the data to fit a defined depth grid.

The function `resample_to_depth()` is part of the process module and works for EK60, EK80, and AZFP.

### DIFF
--- a/echopype/process/azfp.py
+++ b/echopype/process/azfp.py
@@ -89,7 +89,7 @@ class ProcessAZFP(ProcessBase):
             sth = self.sound_speed * ds_beam.sample_interval / 2
             return sth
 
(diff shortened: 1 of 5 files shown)


leewujung:
@ngkavin : Let's add this while refactoring the process module. One thing I did notice was that since Sv itself is a log domain quantity, the interpolation would need to involve transforming the values back to the linear domain, interpolating, and then transforming the results back to the log domain.

leewujung:
Closing this as the class structure has been revamped and this no longer applies.

Issue #213, opened 2020-12-18
Error when running `to_netcdf` or `to_zarr`


lsetiawan:
## Overview

Within the `class-redesign` branch, when running `to_netcdf` I get the error below.

![Screenshot from 2020-12-18 14-12-14](https://user-images.githubusercontent.com/17802172/102666235-461b4400-413b-11eb-8b99-655b8a948c09.png)

## Troubleshooting
1) Seems like there's no catch for when `self.convert_obj.ping_data_dict['angle']` has none channels ...

```
defaultdict(<class 'list'>, {1: array([[[  -1,   -1],
        [  -2,   -2],
        [  -2,   -2],
        ...,
        [ -78,   33],
        [ -66,   47],
        [  93,  -93]],

       [[  -1,   -1],
        [  -2,   -2],
        [  -2,   -2],
        ...,
        [  99,  -60],
        [-122, -125],
        [-125,  120]],

       [[  -1,   -1],
        [  -2,   -2],
        [  -2,   -2],
        ...,
        [  82,  -71],
        [-119,  -62],
        [  99,  -69]],

       ...,

       [[  -1,   -1],
        [  -2,   -2],
        [  -2,   -2],
        ...,
        [ -66,  -58],
        [  68,    5],
        [  99,  -65]],

       [[  -1,   -1],
        [  -2,   -2],
        [  -2,   -2],
        ...,
        [  74,   87],
        [  53,   77],
        [ -54,   79]],

       [[  -1,   -1],
        [  -2,   -2],
        [  -2,   -2],
        ...,
        [  38, -123],
        [  85, -122],
        [-122,  -17]]], dtype=int8), 2: None, 3: None})
```

2) I tried using the master just to make sure it's not the raw data, and sure enough, it's problem with the code.

Please help! @ngkavin @leewujung

leewujung:
Ah! The check must have been removed at some point...  Thanks for reporting this bug!

This is related to whether the transducer is single beam and split beam. Single beam files would not have the angle data. 
The configuration has a field "beam_type" where 0 = Single, 1 = Split.

@ngkavin : could you add a check for this? 
Some info below (if you're in SetGroupsEK60/EK80):
```python
# EK60
for ch in self.convert_obj.config_datagram['transceivers'].keys():
    self.convert_obj.config_datagram['transceivers'][ch]['beam_type']

# EK80
for ch in self.convert_obj.config_datagram['configuration'].keys():
[...]

PR #215, opened 2020-12-19
Check if angle data exists before indexing

ngkavin:
EK60 SetGroups errored out when trying to save angle data of channels that do not have angle data. 

This fix checks the beam type and only saves the angle data if the beam_type == 1 (split beam mode) because if beam_type == 0 (single beam mode) there is no angle data.

Channels without angle data are saved as NaN (ping_time x range_bin)


ngkavin:
Fixes #213

leewujung:
Great, thanks!

Issue #195, opened 2020-09-25
Enhance Convert to accept file-like objects, not just local file path strings


emiliom:
Currently `Convert` can read only files from the locally accessible file system. It accepts one or more file path strings. That's true of both the current (`master`) version and the refactored version in `class-redesign`.

If we want to read a file not on the file system or a buffered stream in memory, we're out of luck.  The only option right now is to download the file, then read it as a local file.

Having `Convert` also accept [file-like objects](https://docs.python.org/3/glossary.html#term-file-like-object) would open up this capability. It would enable both `io.BytesIO` and `io.StringIO` (see [here](https://medium.com/dev-bits/ultimate-guide-for-working-with-i-o-streams-and-zip-archives-in-python-3-6f3cf96dca50) and [here](https://liutheprogrammer.wordpress.com/2018/08/02/stringio-bytesio-and-file-like-objects/) -- references for myself, if nothing else!) as well as [fsspec](https://filesystem-spec.readthedocs.io/)-based access to remote files (S3 buckets, http-hosted files, etc).

I'm interested in attempting a limited implementation myself. For my own reference when I get back to this, I'm listing here the sequence of method calls for EK60 that currently is traversed to the statement that opens a local file:

1. [convert/convert.py#L11](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/convert.py#L11)
2. [convert/ek60.py#L39](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/ek60.py#L39)
3. [convert/ek60.py#L253](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/ek60.py#L253)
4. [convert/utils/ek_raw_io.py#L64](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/utils/ek_raw_io.py#L64)

emiliom:
An `S3FileSystem` (type:  `s3fs.core.S3FileSystem`) has an `open` method:

```python
Signature: FS.open(path, mode='rb', block_size=None, cache_options=None, **kwargs)
Docstring:
Return a file-like object from the filesystem

The resultant instance must function correctly in a context ``with``
block.

Parameters
----------
path: str
    Target file
mode: str like 'rb', 'w'
    See builtin ``open()``
block_size: int
    Some indication of buffering - this is a value in bytes
cache_options : dict, optional
    Extra arguments to pass through to the cache.
[...]

PR #216, opened 2020-12-19
Modify Convert to be able to read raw files from anywhere using fsspec mapping

lsetiawan:
## Overview

This PR adds the ability for echopype `Convert` object to read and parse raw files located in any filesystem listed [here](https://github.com/intake/filesystem_spec/blob/a5bb95155b450ec2443872a45a4a58ce90748297/fsspec/registry.py#L88-L182). 
Local file system stays the same however for sonal models that require `RawSimradFile` class, meaning it uses python `FileIO` rather than fsspec local file system mapping.

Additionally, `storage_options` attribute is added to the class to allow user to pass in keyword arguments for any of fsspec file system.

## Issue Related
Partial solution to #195

### DIFF
--- a/echopype/convert/parse_base.py
+++ b/echopype/convert/parse_base.py
@@ -57,7 +57,8 @@ class ParseEK(ParseBase):
     def parse_raw(self):
         """Parse raw data file from Simrad EK60, EK80, and EA640 echosounders.
         """
-        with RawSimradFile(self.source_file, 'r') as fid:
+        with RawSimradFile(self.source_file,
+                           'r', storage_options=self.storage_options) as fid:
             self.config_datagram = fid.read(1)
             self.config_datagram['timestamp'] = np.datetime64(
                 self.config_datagram['timestamp'].replace(tzinfo=None), '[ms]')
--- a/echopype/convert/parse_base.py
+++ b/echopype/convert/parse_base.py
@@ -27,7 +27,7 @@ class ParseBase:
 class ParseEK(ParseBase):
     """Class for converting data from Simrad echosounders.
     """
-    def __init__(self, file, storage_options, params):
+    def __init__(self, file, params, storage_options):
         super().__init__(file, storage_options)
 
         # Parent class attributes
--- a/echopype/convert/parse_base.py
+++ b/echopype/convert/parse_base.py
@@ -12,11 +12,12 @@ NMEA_GPS_SENTENCE = 'GGA'
 class ParseBase:
     """Parent class for all convert classes.
     """
-    def __init__(self, file):
+    def __init__(self, file, storage_options):
         self.source_file = file
         self.timestamp_pattern = None  # regex pattern used to grab datetime embedded in filename
         self.nmea_gps_sentence = None  # select GPS datagram in _set_platform_dict()
         self.ping_time = []            # list to store ping time
+        self.storage_options = storage_options
 
     def _print_status(self):
         """Prints message to console giving information about the raw file being parsed.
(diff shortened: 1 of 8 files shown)


lsetiawan:
Just rebased. All merge conflicts have been resolved.

leewujung:
Awesome, thanks @lsetiawan! I noticed the mini tests. Seems that we need to more test sources. :)

Let's use the [NCEI WCSD S3 bucket](https://registry.opendata.aws/ncei-wcsd-archive/). Below goes directly to 2017 data with both EK60 and EK80 files:
https://ncei-wcsd-archive.s3.amazonaws.com/index.html#data/raw/Bell_M._Shimada/SH1707/
The EK80 ones appear to be BB mode files.

lsetiawan:
I'm assuming these files in S3 can be accessed using an anonymous credential?

emiliom:
Hmm, on my email, but not here, I see this comment from @lsetiawan:

> I'm assuming these files in S3 can be accessed using an anonymous credential?

Yes, they're accessible via an anonymous credential.

I'm planning to run some tests of Don's code later today using these files. Adding actual echopype unit tests that use data from this bucket sounds good.

lsetiawan:
@emiliom Yea I did ask that, but realized I only made tests for http for now. So access to S3 is via HTTP atm.

lsetiawan:
Alright, since it's simple, I added the s3 protocol test just now. :smile:

echopype/convert/parse_base.py
```diff
@@ -26,8 +27,8 @@ def _print_status(self):
 class ParseEK(ParseBase):
     """Class for converting data from Simrad echosounders.
     """
-    def __init__(self, file, params):
-        super().__init__(file)
+    def __init__(self, file, storage_options, params):
```
emiliom, line 30: This is just a comment about consistency: how about reversing the order of `params` and `storage_options`, for consistency with what's used elsewhere? Then, of course, calls to `ParseEK` will have to be updated as well

emiliom:
I've gone through the PR. Looks great! I made one minor stylistic suggestion via an inline comment. I also tested one aspect of the read-from-S3 addition, by adapting my code that uses convert-parse code to extract only GPS datagrams from EK60 raw files. Works great.

One thing I noticed relative to the current echopype stable release is that `Convert` no longer accepts `pathlib` `Path` types. I've grown very fond of the convenience and readability of `Path` paths. It looks like the problem here is not with this PR proper, but rather a previous change. The issue is with the instance checks here:
https://github.com/lsetiawan/echopype/blob/convert_input_upgrade/echopype/convert/convert.py#L153
This would be fixed by changing that line to `if isinstance(rawpth, (str, Path)):`, after adding this import statement: `from pathlib import Path`. @lsetiawan would you mind adding that change??

lsetiawan:
> One thing I noticed relative to the current echopype stable release is that Convert no longer accepts pathlib Path types. I've grown very fond of the convenience and readability of Path paths.

I have not used `Path` before. What is the advantage of this and what are the use cases? If we do allow Path types, then this means that we have to check if all the packages downstreams also take this. At the end of the day, if the usecase of people using `Path` is small, maybe we should just put in the docs to encourage people to use string.

lsetiawan:
Following up on my comment above, I've made the minor changes, that @emiliom have suggested. Is there anything else regarding this specific PR? Otherwise, I think the matter of `Path` stuff should be in a separate issue since it might not be a simple change. But yet again, I've never used `Path` so I'm not very familiar with it.

emiliom:
Thanks! I agree that restoring `Path` support is a separate issue. I'll open an issue and describe the reason for it there.
I'll merge this PR now. Thanks again!

PR #218, opened 2020-12-21
changes and bug fixes related to Sv and Sp

leewujung:
Changes:
- change all 'TS' to 'Sp' for EK instruments, still need to do this for AZFP

Bug fixes:
- absorption and sound speed were swapped in set_env
- add sa_correction initialization in init_cal_params
- the computed range is now a data variable within the Sv dataset, since there may be ping-by-ping variation of sound speed changes and/or sample interval


leewujung:
@ngkavin : please note especially the change of `range` now being a data variable instead of a coordinate in `ed.Sv` and `ed.Sp`. There should be corresponding changes for AZFP since the ping-by-ping temperature may be different.

PR #238, opened 2021-01-16
Add ad2cp parsing

imranmaj:
Incomplete


leewujung:
Thanks for the PR @imranmaj !

In testing this I found that only the `Vendor` group exists within each file, which is due to that you issued `mode="w"` for writing the groups. I saw that you have a TODO question on this.

Use `mode='a'` when creating a new file (which is when you are writing the top-level group) and `mode="w"` when adding additional groups will get you all the groups in the same file.

In general when you write a certain group or a file, you will be overwriting the existing group/file, that's why we check for file existence first before proceeding on parsing and saving files.

leewujung:
@imranmaj : thanks for the updates, sorry it took me a while to get to this!

### Overall
I checked all the groups and caught a couple things:

   - **Missing top-level data**: the top-level group (what you get when you do `xr.open_dataset(filename)`) is empty in the example data. Could you take a look at the [base class method](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/set_groups_base.py#L47-L63 ) and its use [here](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/set_groups_azfp.py#L18) and [here](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/set_groups_ek60.py#L24), and use the same method to save this group?

   - **Empty coordinates**: when certain mode is absent in the data set, I think it would be good to not have those coordinates showing up in the saved dataset. For example, in the example files, when only the Average mode data is present, coordinates such as `time_burst`, `time_echosounder`, `range_bin_burst`, `range_bin_echosounder` can be removed. This is my view though, let me know if there are specific reasons why you think we should keep these coordinates around.

### Units and attributes
Once we are good with the above, I think that the next stage we should focus on getting the units right. This goes hand-in-hand with the tests you are writing against the Ocean Contour exported files: 

  - **Units**: in the Integrator Guide Data Format section: there's a column "Resolution/Unit" which shows the conversion from the raw _counts_ you get by parsing the files to physical units, which is crucial. Could you put these conversion into the parser? I think it could be either 1) convert them to the correct units when you parse the raw count, or 2) have a unit conversion stage/function that takes care of that. Let me know which way you decide to go.

  - **Variable attributes**: along with the above conversion to physical units, could you attach the units of the data variables as variable attributes when you organize the datasets? We don't need to go to the full specification like what we did with other sonar model yet (since things like `long_name`, `standard_name`, etc may not have been fully defined), but we should at least have the units in there.

Also tagging @emiliom here for the last point -- AD2CP data vs convention is not in our current scope, but I thought it'll be good to put this on your radar. I'll share the relevant docs @imranmaj and I are working from when it's time, since I am sure you won't want to see them now!  :P

leewujung:
### Beam sequence
One thing I didn't realize but it is obvious looking back to the data format spec is that the `beam` instead of always starting from 0 it should be the actual beam number. For example the 0097_a file has beam 234, and 0097_b file has beam 124, which is related to whether the glider was diving down or up. This would affect velocity, correlation, and amplitude under the Beam group.

@imranmaj : could you also make this change? Thanks!

echopype/tests/test_process_ad2cp.py
```diff
@@ -102,10 +102,7 @@ def test_field(base, test, base_field_name, test_field_name):
         base = xr.open_dataset(parsed_file, group="Data/Avg")
         
         test = xr.open_dataset(test_file, group="Platform")
-        # does not work because ocean contour timestamps have rounding errors
-        # so we can't compare times
-        # for time in base["time"]:
-        #     assert close(base["Heading"].sel(time=time).data[()], test["heading"].sel(time=time).data[()])
+        # ocean contour timestamps have rounding errors so we can't compare times
```
leewujung, line 105: Could you plot a histogram of the timestamp difference and we can discuss in our next meeting? Is it on the order of ns or ms?

leewujung (changes requested):

echopype/convert/set_groups_ad2cp.py
```diff
@@ -11,6 +11,7 @@
 
 class SetGroupsAd2cp(SetGroupsBase):
     def write(self, ds: xr.Dataset, group: str):
+        ds.attrs["Units"] = "velocity: m/s, temperature: deg Celsius, pressure: dBar, direction: degrees, length: m, voltage: V, magnitude: dB, echosounder data: dB/count"
```
leewujung, line 14: Add this as memo from our meeting: Units are attributes for each of the data variables.

leewujung (changes requested):

echopype/tests/test_process_ad2cp.py
```diff
@@ -53,7 +53,12 @@ def test_process():
     "./echopype/test_data/ad2cp/SG198_0099_b.ad2cp.nc"
 ]
 
+THRESHOLD = 0.02
+
 def test_ocean_contour():
+    def close(a, b):
+        return abs(np.float32(a) - np.float32(b)) < THRESHOLD
```
leewujung: How did you decided `THRESHOLD = 0.02`? 

For some of the data variables below I can see the difference may be from rounding errors, but for things such as `base.attrs["Instrument_avg_nCells"]` and `base.attrs["Instrument_avg_nCells"]` why would the values not match exactly?

Also, why don't you use `np.isclose` for the comparison?

leewujung (changes requested):

echopype/tests/test_process_ad2cp.py
```diff
@@ -37,3 +38,111 @@ def test_process():
...
+    "./echopype/test_data/ad2cp/SG198_0099_a.ad2cp",
+    "./echopype/test_data/ad2cp/SG198_0099_b.ad2cp"
+]
+
+OCEAN_CONTOUR_PARSED = [
+    "./echopype/test_data/ad2cp/SG198_0097_a.ad2cp.nc",
+    "./echopype/test_data/ad2cp/SG198_0097_b.ad2cp.nc",
+    "./echopype/test_data/ad2cp/SG198_0099_a.ad2cp.nc",
+    "./echopype/test_data/ad2cp/SG198_0099_b.ad2cp.nc"
+]
```
leewujung, line 54: Could you change:
- `OCEAN_CONTOUR_UNPARSED` to `AD2CP_UNPARSED`, and 
- `OCEAN_CONTOUR_PARSED` to `AD2CP_PARSED_BY_OCEAN_CONTOUR`?
This way we are explicit about what these refer to.

echopype/tests/test_process_ad2cp.py
```diff
@@ -37,3 +38,111 @@ def test_process():
...
+
+def test_ocean_contour():
+    for file, parsed_file in zip(OCEAN_CONTOUR_UNPARSED, OCEAN_CONTOUR_PARSED):
+        test_convert = Convert(file, model="AD2CP")
+        test_file = ".".join(file.split(".")[:-1]) + ".nc"
```
leewujung, line 64: I think it'll be great for us to use `pathlib.Path` consistently for our path operations. Could you change this and other similar path ops to use `Path`?

echopype/convert/parse_ad2cp.py
```diff
@@ -315,13 +315,13 @@ def _postprocess(self, field_name):
...
-                self.data["velocity_data_included"] = self.data["configuration"] & 0b0000_0000_0010_0000
-                self.data["amplitude_data_included"] = self.data["configuration"] & 0b0000_0000_0100_0000
-                self.data["correlation_data_included"] = self.data["configuration"] & 0b0000_0000_1000_0000
+                self.data["pressure_sensor_valid"] = min(1, self.data["configuration"] & 0b0000_0000_0000_0001)
+                self.data["temperature_sensor_valid"] = min(1, self.data["configuration"] & 0b0000_0000_0000_0010)
+                self.data["compass_sensor_valid"] = min(1, self.data["configuration"] & 0b0000_0000_0000_0100)
+                self.data["tilt_sensor_valid"] = min(1, self.data["configuration"] & 0b0000_0000_0000_1000)
+                self.data["velocity_data_included"] = min(1, self.data["configuration"] & 0b0000_0000_0010_0000)
+                self.data["amplitude_data_included"] = min(1, self.data["configuration"] & 0b0000_0000_0100_0000)
+                self.data["correlation_data_included"] = min(1, self.data["configuration"] & 0b0000_0000_1000_0000)
```
leewujung: Could you explain why we need `min(1, parsed_values)` here? If you shift the bits, wouldn't the parsed values always be 1?

imranmaj: I did min here instead of shifting, I should probably shift instead

leewujung:
@imranmaj : A quick note that let me know once you are done changing the behavior of set groups methods and adding the hook with EchoData, then I can review this to see if we are ready to merge. 

Also adding @lsetiawan to the reviewer list for his feedback.

imranmaj:
b635d4a messed up the entire tree, closed in favor of #318

PR #239, opened 2021-01-19
Combine files and SetGroups updates

leewujung:
This PR contains updates related to `Convert.combine_files`. 

### Main changes
- combine-related functions are now in `convert_combine.py` to declutter `convert.py`
- bug fixed related to Platform group dimensions for EK60 and EK80
- remove `combine=nested` from all groups that need to be combined along the `ping_time`/`location_time`/`mru_time` dimensions. This makes file combination only possible when these coordinates are nicely monotonically varying within each individual file prior to combination. This is often not the case for EK60/EK80 files (?? somehow time often goes backward) and we will need to add a handler to clean up these anomalies before combining the files.
- change the default chunk size (ad-hoc but the original ping_time chunk of 100 seems too small):
  ```
  DEFAULT_CHUNK_SIZE = {
      'range_bin': 25000,
      'ping_time': 2500
  }
  ```
### Main changes since first PR
- bug fixed related to Vendor group dimensions for AZFP
- make `set_toplevel` and `set_sonar` consistent across all sonar models

### TODO for future versions
- add a handler to clean up anomalous time entries (e.g. `ping_time` going backward) before file combination
- another possible why to deal with this is to use `combine='nested'` in open_mfdataset but make sure that the list of individual files passed in are listed sequentially, since `combine_nested` does not check this.
- experiment with chunking sizes
- consider separating `location_time` and `mru_time` chunk size from that of `ping_time` since the associated data variables are typically 1D (instead of 2-4D like in the Beam group). However keeping it consistent with `ping_time` may help with data slicing.

### DIFF
--- a/echopype/convert/convert.py
+++ b/echopype/convert/convert.py
@@ -19,10 +19,11 @@ from .set_groups_azfp import SetGroupsAZFP
 from .set_groups_ek60 import SetGroupsEK60
 from .set_groups_ek80 import SetGroupsEK80
 from ..utils import io
-
+from .._version import get_versions
 
 warnings.simplefilter('always', DeprecationWarning)
 
+ECHOPYPE_VERSION = get_versions()['version']
 
 MODELS = {
     "AZFP": {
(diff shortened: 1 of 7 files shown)


lsetiawan (approved):
LGTM! I like that the combine things are now in it's own module :smile:

PR #240, opened 2021-01-22
Convert upgrade to be able to read/write to object store

lsetiawan:
## Overview

This PR includes changes to the `Convert` part of echopype to be able to write to object store.

## Actions
- [X] `_to_file` able to write to object store with `overwrite=True/False`
- [X] Implement tests for conversion of EK60 (no combine)
- [X] Implement tests for conversion of EK80 (no combine)
- [X] Implement tests for conversion of AZFP (no combine)
- [X] Implement CI to run tests
- [X] Implement Python linting CI (currently allow failures)

## TODO

The conversions without combining seems to work and tests passed. However, I think I will make changes for combine in a separate PR since this one is getting large already and have CI additions.

### DIFF
--- a/.ci_helpers/Dockerfile
+++ b/.ci_helpers/Dockerfile
@@ -1,0 +1,2 @@
+FROM minio/minio
+CMD ["minio", "server", "/data"]
--- a/.ci_helpers/py3.7.yaml
+++ b/.ci_helpers/py3.7.yaml
@@ -1,0 +1,18 @@
+name: echopype
+channels:
+  - conda-forge
+dependencies:
+  - python=3.7
+  - dask
+  - matplotlib
+  - netCDF4==1.5.5
+  - numpy
+  - pynmea2
+  - pytz
+  - scipy
+  - xarray==0.16.2
+  - zarr
+  - fsspec>=0.8
+  - requests
+  - aiohttp
+  - s3fs==0.5.2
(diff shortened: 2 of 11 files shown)


leewujung:
Thanks @lsetiawan! I'll look at this in a few hours. Excited that you have added linting and the CI workflow!!

Note that the combine _will_ fail if there if ping_time goes backward in the Beam (contains backscatter data) or Platform (contains GPS and pitch/roll/heave data) group, in case you run into that. This is something we need to add to the docs and add a preprocess function for open_mfdataset in a later release, as it does happen even for the OOI EK data.

lsetiawan:
Ugh, the instability of OOI raw data server broke the 3.7 test... Maybe we should host test data on some other http server that's a lot more stable.

emiliom:
Can we host a couple of OOI raw files on the repo itself, and have the tests use https to read them using a full url path?

Great job @lsetiawan!! I may be able to test the new write functionality in this PR this afternoon.

emiliom:
On another note, I see each Test CI build takes about 20 min. Yikes.

leewujung:
> Can we host a couple of OOI raw files on the repo itself, and have the tests use https to read them using a full url path?

The OOI raw files are too large for the repo itself but we can use the http from GitLFS if we push some OOI files there, or just use the raw files that are already on GitLFS.

emiliom:
Ah. For testing EK60, the NCEI 2017 Hake survey raw files are small (~ 25 MB). We could copy one here, after confirming that it doesn't have a time reversal. It seems like an unnecessary burden to use a large file for testing.

PR #242, opened 2021-01-25
Address calibration issue with NaN ping

leewujung:
This PR attempts to temporarily fix issues related to NaN pings in converted raw files.

@ngkavin : could you take a look to see if this fixes the problems with those files with channel 1 and 3 containing NaN pings?

Also I ran into issues with opening zarr due to self._file_format is None in `echodata.get_beam_from_raw`. It doesn't seem that you ran into this error?

### DIFF
--- a/echopype/process/process_base.py
+++ b/echopype/process/process_base.py
@@ -6,12 +6,17 @@ from ..utils import uwa
 from ..utils import io
 
 
+# TODO: separate out calibration in its own clas since once we get to Sv
+#  the processing is uniform across all sonar models
+
+
 class ProcessBase:
     """Class for processing sonar data.
     """
     def __init__(self, model=None):
         self.sonar_model = model   # type of echosounder
 
+    # TODO: update this using the latest corresponding methods in convert
     def validate_proc_path(self, ed, postfix, save_path=None, save_format='zarr'):
         """Creates a directory if it doesn't exist. Returns a valid save path.
         """
@@ -95,6 +100,7 @@ class ProcessBase:
         ed : EchoDataBase
         """
 
+    # TODO: can remove if we rearrange dimension in calc_range
     def _restructure_range(self, ed, ranges=None):
         """Expands the range to include a ping_time dimension if it does not have one. (Uses first raw ping_time)
         Reorders dimensions so that ranges has dimensions [frequency x ping_time x range_bin]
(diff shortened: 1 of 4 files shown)


echopype/process/process_base.py
```diff
@@ -260,33 +266,39 @@ class ProcessEK(ProcessBase):
...
+            #  /Users/wu-jung/miniconda3/envs/echopype_jan2021/lib/python3.8/site-packages/xarray/core/indexing.py:1369:
+            #  PerformanceWarning: Slicing is producing a large chunk. To accept the large
+            #  chunk and silence this warning, set the option
+            #      >>> with dask.config.set(**{'array.slicing.split_large_chunks': False}):
+            #      ...     array[indexer]
+            #  To avoid creating the large chunks, set the option
+            #      >>> with dask.config.set(**{'array.slicing.split_large_chunks': True}):
+            #      ...     array[indexer]
+            #    return self.array[key]
+            ed.raw = ed.raw.dropna(dim='ping_time', how='any', subset=['transmit_duration_nominal'])
```
ngkavin, line 295: This might fit better in `Echodata `since it is unexpected for the raw to change during `init_cal_params`

echopype/process/process_base.py
```diff
@@ -317,12 +329,12 @@ def _cal_narrowband(self, ed, env_params, cal_params, cal_type,
...
+            Sv = ed.raw['backscatter_r'] + spreading_loss + absorption_loss - CSv - 2 * cal_params['sa_correction']
             Sv.name = 'Sv'
             Sv = Sv.to_dataset()
 
-            # Attach calculated range into data set
-            Sv['range'] = (('frequency', 'ping_time', 'range_bin'), self._restructure_range(ed))
+            # Attach calculated range (with units meter) into data set
+            Sv = Sv.merge(ed.range)
 
             # Save Sv into the calling instance and
```
ngkavin, line 339: The following error occurs during `io.save_file` if the `ping_time `changes.
`ValueError: Zarr requires uniform chunk sizes except for final chunk. Variable named 'Sv' has incompatible dask chunks ...  Consider rechunking using chunk().`

echopype/process/process_base.py
leewujung: Good point. Seem that we can have an echopype.qc to clean up issues like this and the ping_time going backward issue.

PR #246, opened 2021-01-29
Relocate dockerfile and setup http server

lsetiawan:
## Overview

This PR contains dockerfile to create http server to server data for testing to simulate http read.


.ci_helpers/docker/download.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+# ==== ONLY EDIT WITHIN THIS BLOCK =====
+DATA_PATH=/usr/local/apache2/htdocs/data
+
+ek60_source='https://ncei-wcsd-archive.s3-us-west-2.amazonaws.com/data/raw/Bell_M._Shimada/SH1707/EK60/Summer2017-D20170615-T190214.raw'
```
lsetiawan, line 6: @emiliom Could you go over these datasets and let me know if we need more? Thanks!

emiliom: Heh, I don't think I'm the right person for that; @leewujung is. As far as I'm concerned, you've included one file of each instrument type (except ea640, and I don't know anything about that), so what you have looks good.

emiliom:
I'll test this PR later today.

emiliom:
@lsetiawan, some questions and comments:

- Currently the http dockerfile (and the minio dockerfile) are only for "manual" testing locally, right? Initially I assumed this PR involved changes to the CI that would use the docker file, but I see now that there is nothing in the [GH CI `Test CI`](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/.github/workflows/ci.yaml) that uses this docker file, or the minio docker file. The idea is that a new entry would later be added to `ci.yaml` under `jobs.test.services`, pointing to a dockerhub image, analogous to the minio entry, right?
- The GH CI `Test CI` is [failing](https://github.com/OSOceanAcoustics/echopype/runs/1793567036?check_suite_focus=true) with this PR on the miniconda setup. Is this expected?

I haven't run the tests against the http dockerfile yet. I've been wrapping my head around the test suite, the existing CI, and what this PR does. Setting up Git LFS for the first time ever. Etc. But I'm now ready, so I expect to do it tomorrow.

On a somewhat side note, while I expect the local http testing with your docker file will work, I wonder if it wouldn't be "better" (faster and fully within the pytest framework) to use an http mocking library instead of a local http server deployed via docker. I remembered that http mocking was used for ulmo (see [here](https://github.com/ulmo-dev/ulmo/blob/master/test/test_util.py) and [here](https://github.com/ulmo-dev/ulmo/blob/master/test/usgs_nwis_test.py)), and looked into it a bit. It relies on [httpretty](https://httpretty.readthedocs.io). I *think* it would work for echopype http file access, so I might try it myself if I feel inspired.

emiliom:
I tested `http.dockerfile` and a slim, tweaked version of `test_convert.py`, against data from all 3 instrument types after building and running the docker image. All files are downloaded and served on http://localhost/data. Everything worked as expected.

I collected step-by-step instructions, that I can share later. 

I was focused only on testing the meat of this PR: converting files served via a locally/docker deployed http server. So I deactivated s3 reading and writing for my testing (I didn't want to bother with minio, plus there's nothing new vis a vis minio or s3 in this PR).

FYI, I started looking into the `httpretty` avenue, but I think it was a dead end. One thing that could speed things up a bit and be more robust is to host the data files via Git LFS rather than downloading them on the fly.

I'll go ahead and merge. Thanks @lsetiawan !

lsetiawan:
> I tested http.dockerfile and a slim, tweaked version of test_convert.py, against data from all 3 instrument types after building and running the docker image. All files are downloaded and served on http://localhost/data. Everything worked as expected.

Woo hoo! I am so proud of your Docker chops nowadays @emiliom :smile:!

> Currently the http dockerfile (and the minio dockerfile) are only for "manual" testing locally, right? Initially I assumed this PR involved changes to the CI that would use the docker file, but I see now that there is nothing in the GH CI Test CI that uses this docker file, or the minio docker file. The idea is that a new entry would later be added to ci.yaml under jobs.test.services, pointing to a dockerhub image, analogous to the minio entry, right?

Yepp this is correct.

> The GH CI Test CI is failing with this PR on the miniconda setup. Is this expected?

Yea, I cancelled the build since it's not really needed for this.

> One thing that could speed things up a bit and be more robust is to host the data files via Git LFS rather than downloading them on the fly.

Is this for the http docker server stuff or more for running the CI?

PR #247, opened 2021-01-29
Separate out calibrate as a module

leewujung:
This PR contains the following:
- separates out calibration operations in a separate module `calibrate`, with a set of underlying classes handling calibration ops from different sonar models, and a simple function to use with an `EchoDataNew` object
- calibration ops included are for EK60 and AZFP only. EK80 cal needs to be included.
- an `EchoDataNew` class that is a handler fo echo data and filepaths (currently only raw and Sv)
- minimum test functions for EK60 and AZFP

@lsetiawan : it seems easier if we just go back and forth on this PR. Thanks for looking this over!


echopype/calibrate/cal_func.py
```diff
@@ -0,0 +1,24 @@
...
+from .calibrate_azfp import CalibrateAZFP
+
+CALIBRATOR = {
+    'EK60': CalibrateEK60,
+    'EK80': CalibrateEK80,
+    'AZFP': CalibrateAZFP
+}
+
+
+def get_Sv(echodata, env_params=None, cal_params=None):
```
lsetiawan, line 11: I was thinking maybe having a generalized function for the calibration similar to like `to_file`. I think this would make things clearer API wise. I can see it being like:

```python
from echopype import EchoData
from echopype import calibrate

echodata = EchoData(...)

Sv = calibrate(echodata, cal_type='Sv')
```

Something like that.

leewujung: That is actually how it is structured now (see #250). There are other functions under calibrate that are forthcoming, which is why I have it as
From echopype import calibrate
Sv = calibrate.get_Sv(echodata)

leewujung: Some of the other functions would require passing in other arguments, so it'll be too messy to use the same cal_type set up.

lsetiawan: Ah okay that makes sense. Thanks for clearing that up :smile:

leewujung: Thinking about it again, with the current structure, for different functions (let's say there are `calibrate.get_Sv(echodata)` and `calibrate.get_single_target_TS(echodata, target_location)`) under the hood each call creates and destroys a Calibrate object. Can this potentially be a problem?

lsetiawan: I think it would be better if we can do the creation of Calibrate object once. Can the echodata object contain the calibrate object internally? 

You're creating this with echodata attributes already anyways:

```python
cal_obj = CALIBRATOR[echodata.sonar_model](echodata)
```

leewujung: Heh, that was actually how I had it before, and you advised against having it as part of echodata! 😛

Regardless, yes massaging calibrate.get_Sv to use internal calibrator object in echodata is straightforward to do.

lsetiawan: > Heh, that was actually how I had it before, and you advised against having it as part of echodata! 

I advised against having the actual `get_Sv` function as part of echodata :stuck_out_tongue: hehe. Anyways, now thinking about it further I think the `cal_obj` can probably live in echodata.

leewujung:
I think @valentina-s has a point that our modules should probably be named using nouns and the functions be named using verbs. Anyone has thoughts on this?

leewujung:
Apparently I have completely forgotten the existence of this PR, which should be pointed to #266.

Anyway, I feel much better about the status of the calibration code now, and feel once we we settle down the API calling patterns and prune the redundant old scripts, this can be merged into class-redesign. We will need additional changes for backward compatibility like #254.

leewujung:
I am going to merge this and treat the API call change to be a separate task. This PR partially address #266 .

PR #253, opened 2021-02-02
Skip unneeded datagrams

ngkavin:
When selecting certain datatypes to save such as the GPS information, all data in the raw files was parsed and the unneeded datagrams were ignored. This change peeks at the next datagram and parses it only if it is needed, which will save computation time in cases where the whole raw file does not need to be saved.

Also fixes a bug in the peek function from the parser.

### DIFF
--- a/echopype/convert/parse_base.py
+++ b/echopype/convert/parse_base.py
@@ -183,10 +183,20 @@ class ParseEK(ParseBase):
         num_datagrams_parsed = 0
 
         while True:
+            # Skip parsing datagrams that the user did not ask for
+            if 'ALL' not in self.data_type:
+                try:
+                    # Peek file to check the type
+                    dgram_type = fid.peek()['type']
+                except SimradEOF:
+                    break
+                # Skip unused types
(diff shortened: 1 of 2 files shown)


leewujung:
I'll review and comment on this after v0.5.0 is out, because 1) this feature is not already in place for v0.4.1, and 2) the operations, from the call through Convert to assembly of output files, needs to be verified with tests.

emiliom:
ref: This PR is directly related to a subset of [this discussion](https://github.com/OSOceanAcoustics/echopype/discussions/225#discussioncomment-238638)

I, too, can review the PR after 0.5.0 is out.

leewujung:
This PR probably needs to be closed because the code has gone through structural changes in v0.5.0, but we may want to consider reusing some of the changes here. Specifically the part about selecting which type of datagrams to read for EK echosounder data. I'll keep this PR open for now and let's discuss that in the next couple meetings.

lsetiawan:
Stale PR. As Wu-Jung has noted. I put a label on it to easily find it again. Closing for now.

PR #262, opened 2021-02-24
Change source of similarly named AZFP parameters

ngkavin:
The parameters parsed from the XML file will now override those of the same name from the 01A file except for gain which will now use the values from the 01A file.

Also gain is moved to the vendor group.


leewujung:
@ngkavin : some of the changes are not what I intended... I'll explain below:
- it is correct to use the gain values from the 01A file in the Beam group
- can you prefix the gain parameter in the Vendor group with `XML_`?
- for all other parameters that have duplication in the XML and 01A file, the XML parameters should be prefixed with `XML_` and stored in the Vendor group.
Thanks.

ngkavin:
New changes keep everything the same as before the PR except for the gain correction change. Duplicate values from the XML are now saved in the vendor group with the XML_ prefix.

leewujung:
Thanks @ngkavin, this looks great! I'll merge.

PR #270, opened 2021-03-26
Add pypi github action for publishing

lsetiawan:
## Overview

This PR adds pypi publish test github action.

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -1,0 +1,49 @@
+# Publish archives to PyPI and TestPyPI using GitHub Actions
+name: Publish to PyPI
+
+# Only run for tagged releases and pushes to the development branch
+on:
+  release:
+    types:
+      - published
+  push:
+    branches:
+      - class-redesign
+  workflow_dispatch:
+
+jobs:
+  publish-pypi:
+    name: Publish to PyPI
+    runs-on: ubuntu-latest
+    if: github.repository == 'OSOceanAcoustics/echopype'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2.3.4
+
+    - name: Set up Python
+      uses: actions/setup-python@v2.2.1
+      with:
+        python-version: 3.9
+
+    - name: Install dependencies
+      run: python -m pip install setuptools wheel
+      
+    - name: Build source and wheel distributions
+      run: |
+        python setup.py sdist bdist_wheel
+        echo ""
+        echo "Generated files:"
+        ls -lh dist/
+        
+    - name: Publish to Test PyPI
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+
+    - name: Publish to PyPI
+      if: startsWith(github.ref, 'refs/tags')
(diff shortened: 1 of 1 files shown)


lsetiawan:
Note that I manually cancelled the Test CI github actions.

leewujung:
@lsetiawan : great, thanks for setting this up!

Where do you "manually" cancel the CI tests?  
This nice since then we can make sure everything passes before pushing the tag. :D

lsetiawan:
You can go to the actions tab and cancel whichever workflow that you don't want to run :)

lsetiawan:
Is this okay to merge?

leewujung:
> You can go to the actions tab and cancel whichever workflow that you don't want to run :)
Ugh, you mean to do that every time a tag is pushed? 

Could you actually add the skip_ci flag? :P

lsetiawan:
> Ugh, you mean to do that every time a tag is pushed?

When a tag is created, the CI test will not run. So you don't need to do that. The CI will run in these conditions (just pushes to master and class-redesign, and pull requests):

https://github.com/OSOceanAcoustics/echopype/blob/90ebf03730c77fe24229cec9b5857cc806d5cdd1/.github/workflows/ci.yaml#L3-L13

> Could you actually add the skip_ci flag? :P

You can add `[skip ci]` to commit message if you don't want the CI test to run
https://github.com/OSOceanAcoustics/echopype/blob/90ebf03730c77fe24229cec9b5857cc806d5cdd1/.github/workflows/ci.yaml#L22

emiliom:
> You can add [skip ci] to commit message if you don't want the CI test to run

Cool! I hadn't paid attention to that line in `ci.yml` and was not aware of that trick to semi-manually skip a CI.

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,49 @@
...
+name: Publish to PyPI
+
+# Only run for tagged releases and pushes to the development branch
+on:
+  release:
+    types:
+      - published
+  push:
+    branches:
+      - class-redesign
```
emiliom, line 11: Is this here only to test during development? Otherwise, do you expect a PyPI build (test or otherwise) to be initiated from this branch?

emiliom:
Assuming I'm understanding this, this action will result in a test pypi build on **every** push to `class-redesign` (but not a PR, right?). Is that excessive?

Otherwise, it looks good and IMHO it's ok to merge. Thanks!!! I already have other use for this workflow :wink:. I'd been thinking about adding a pypi build action for a different package, and you've made my job easier :smiley:

emiliom (approved):

leewujung:
> Assuming I'm understanding this, this action will result in a test pypi build on every push to class-redesign (but not a PR, right?). Is that excessive?

That's my understanding also. My interpretation is that direct push to class-redesign wouldn't happen other than to push a tag, but let's see what @lsetiawan says. It seems that the trigger should be from master branch?

lsetiawan:
Yepp, this will run only on class-redesign ... but now thinking about it, this should probably also run for master ... tweaking now. The actual publishing to the production pypi only happens when it's a tag creation.

lsetiawan:
Welp ... looks like `pull_request_target` doesn't have the commit message payload so CI still runs ... need to fix that on another PR.

leewujung:
Wow, ok. I'll go ahead to merge this then. Thanks @lsetiawan !

Issue #266, opened 2021-03-02
Overhaul calibration code and API call pattern


leewujung:
This issue is long overdue (both in terms of making its way here and the action).

The major change from v0.4.1 to v0.5.0 (WIP) is to overhaul the underlying module structure for calibration (which is instrument-specific) so that it is easier to maintain and read now that we're expanding to more sonar models. We are also changing the API call pattern (see #250) to be more modularized wrt the functionality, rather than dump everything into the previous `Process` object.

The general module reorganization is in the `class-redesign` branch, whereas the specific overhaul of calibration is in the `echodata-overhaul` branch.

(I kept on needing a place to put interim notes and finally figured out that's what @emiliom has been doing with the docs branch! 🤓 )

emiliom:
> I kept on needing a place to put interim notes and finally figured out that's what @emiliom has been doing with the docs branch! 

It's worth noting that the docs lend themselves nicely to being in a separate branch, b/c the files are completely separate from everything else. Well, except for the API docs build, but I'm just ignoring all the error messages for now!

leewujung:
> except for the API docs build

🤯     
Yes, we need to do a housekeeping soon, once the calling patterns are settled and the tests are in place.

leewujung:
This can now be closed with `class-redesign` merged to `master`. Hurray!

PR #272, opened 2021-03-30
Update convert api

leewujung:
This PR includes the following:
- change module `convert` to `conversion`
- add `open_raw` to be the interface for file conversion, in the form of:
   ```python
   import echopype as ep
   echodata = ep.conversion.open_raw(“raw_file_path”, sonar_model=”EK60”)
   echodata.to_zarr(...)
   ```
   This function simply wraps the original `Convert` object but the change would likely help with the API consistency issues for users as we discussed
- update `test_convert.py` to use `conversion.open_raw` and substitute all the `tmp_` by `echodata_` in the tests 

@emiliom : the link we discussed to access the specific groups in the converted file directly through the echodata object has **not** been implemented 
```python
echodata.environment   # xr.Dataset of the parsed env data
echodata.beam          # xr.Dataset of the parsed backscatter data
```

@lsetiawan : could you make sure I didn't break your file conversion tests?


leewujung:
Now I am having a second thought on whether it should be `convert` or `conversion`. scipy actually has both: `integrate` and `optimize` vs `signal` and `stats`. Maybe it does not matter to have variations and we are spending too much energy on this. For whatever reason I feel having `convert`, `calibrate`, `visualize` is natural, but so is having `metrics` and `qc`.

lsetiawan:
This looks okay. I have to tweak the CI. Seems like it runs double ... hmmm ... If from fork, the `pull_request_target` should run, not `pull_request`. The test that should run are all passing, so that's good :smile:

lsetiawan:
> I feel having convert, calibrate, visualize is natural, but so is having metrics and qc.

These are more verb words, and we are performing an action with these stuff, so I think that makes more sense also.

leewujung:
> The test that should run are all passing, so that's good 😄

great!

>> I feel having convert, calibrate, visualize is natural, but so is having metrics and qc.

> These are more verb words, and we are performing an action with these stuff, so I think that makes more sense also.

Ok, pinging @valentina-s to see if she has some insights.

emiliom:
> Now I am having a second thought on whether it should be convert or conversion. scipy actually has both: integrate and optimize vs signal and stats. Maybe it does not matter to have variations and we are spending too much energy on this. For whatever reason I feel having convert, calibrate, visualize is natural, but so is having metrics and qc.

Thanks for that R&D comparing with what's in `scipy`! I definitely agree with this: "we are spending too much energy on this". I *think* part of the reason why the verbs convert, calibrate, etc feel natural is b/c English is so squishy. We can say "a convert object" with the same ease as "a conversion object" or "a qc object". True, `convert.Convert` seems awkward, but the need for users to class that class explicitly will be minimized now that there'll be a `convert.open_raw`.

So: despite all our sweat and blood that went into this discussion, I think I'd vote for not making the change. But I'm good with proceeding with it, too.

If the decision is to go ahead with your changes to `conversion`, it looks like many of the comments and description strings within the code (as opposed to the code proper) should probably not be changed from convert to conversion, b/c the result is awkward sentences. For example, this comment string: "# Rectangularize all data and conversion to numpy array indexed by channel"

leewujung:
@emiliom : haha, yeah, I thought I caught most of the nonsense convert->conversion replace-all changes, but apparently not! 

I think it'll be better to not make the change, because convert has been there for too long!  I'll revert the last commit.

echopype/convert/convert.py
```diff
@@ -79,8 +99,14 @@ class Convert:
         # get environment XML only (EK80)
         ec.to_netcdf(data_type='ENV_XML')
     """
-
     def __init__(self, file=None, xml_path=None, model=None, storage_options=None):
+        warnings.warn(
+            "Calling `echopype.Convert` directly will be deprecated, "
+            "use `echopyp.conversion.open_raw(file, model, ...)` instead.",
```
emiliom, line 105: Change to `echopype.convert.open_raw(file, model, ...)`

leewujung: Ah great! thanks for catching this!

echopype/convert/convert.py
```diff
@@ -35,6 +35,26 @@
 NMEA_SENTENCE_DEFAULT = ["GGA", "GLL", "RMC"]
 
 
+def open_raw(file=None, model=None, xml_path=None, storage_options=None):
+    """Create a converter object by reading in raw data files.
+
+    The converter object can be used for adding metadata and ancillary data
```
emiliom, line 41: Maybe it'd be a bit more helpful to say "Create a Convert object ...", "The Convert object ..."? But this is a minor point.

leewujung: Agreed, better to be explicit

emiliom:
After your last changes just now, you can merge this PR any time as far as I'm concerned. FYI, my assessment is based just on code inspection, and no local running of tests or examining CI results.

leewujung:
This addresses part of #266 .

PR #275, opened 2021-04-06
Update calibrate api and add echodata class

leewujung:
This PR updates the API call for the `calibrate` module and along with it introduces a class `echodata` that handles the various "groups" in the converted raw file.

I have updated and tested the AZFP-related component and am on the way to do the same for EK-related components. Wanted to do a PR right now because @emiliom @lsetiawan I want to see what you think about `echodata` and allowing using it at the echopype package level, i.e. one can do:
```python
import echopype as ep
echodata = ep.open_converted('some_path_to_converted_files')   # how's this function name?
``` 

Note that under echodata the attributes include
```python
echodata.beam  # data in Beam group
echodata.environment  # data in Environment group
# etc
```
that users can use to access the specific groups.

Previously we decided to use 
```python
echodata = ep.convert.open_raw('some_path_to_raw_files')
```
so this creates a conflict because in this case `echodata` is actually a `Convert` object. Perhaps we should just make the output of `ep.convert` an `EchoData` object? (and create some link under the hood to Convert methods when people do `to_netcdf`/`to_zarr`?)


echopype/echodata/echodata.py
```diff
@@ -0,0 +1,57 @@
...
+
+        self.top = None
+        self.beam = None
+        self.beam_power = None  # only present for EK80 containing both complex and power data, may remain empty
+        self.environment = None
+        self.vendor = None
+
+        self.storage_options = storage_options if storage_options is not None else {}
+
+        # TODO: verify if converted_raw_path is valid on either local or remote filesystem
```
leewujung, line 28: @lsetiawan : this is the place where I need your help with adding checks for remote filesystems. I put in `storage_options` as in Convert.

echopype/echodata/echodata.py
```diff
@@ -0,0 +1,57 @@
...
+        if converted_raw_path:
+            self._load_file(converted_raw_path)
+            self.sonar_model = self.top.keywords
```
leewujung, line 31: @lsetiawan : this is the place where I need your help with adding checks for remote filesystems. I put in storage_options as in Convert.

emiliom:
What's the role of `EchoDataOld` vs `EchoData`? Just from comparing the two class definitions, it's unclear; but my best guess is that there's functionality in `EchoDataOld` that hasn't been transitioned to `EchoData` yet, specially regarding handling of multiple files. Will `EchoDataOld` go away once we reach 0.5?

emiliom:
> I want to see what you think about echodata 

Looks good, I guess. But my answer is contingent on my previous question about `EchoData` vs `EchoDataOld` :wink: 

> and allowing using it at the echopype package level

That seems fine. Speaking of which, I don't think `open_raw` is accessible at the package level yet, but I've had the same thought about it.

`open_converted` is a good name, I think.

emiliom:
> Previously we decided to use 
> ```python
> echodata = ep.convert.open_raw('some_path_to_raw_files')
> ```
> so this creates a conflict because in this case `echodata` is actually a `Convert` object. Perhaps we should just make the output of `ep.convert` an `EchoData` object? (and create some link under the hood to Convert methods when people do `to_netcdf`/`to_zarr`?)

Hmm. This seems trickier. Maybe a topic for our next meeting? Some thoughts that come to mind:
- In principle I do like the consistency of both returning `EchoData` objects
- `open_converted` does "lazy" open, same as xarray `open_*`. It returns rich metadata that fully populates the `EchoData` object.
- `open_raw` does "lazy-lazy" open -- it doesn't open anything, it just sets up and checks paths and some basic metadata. If it were to return an `EchoData` object, it would be a very empty one. Which is not necessarily problem; I'm just pointing it out.
- How large a change would it be? I suppose that `open_raw` could work by doing what it does now via the `Convert` object, then using the results to instantiate an `EchoData` object. But a lot of the `Convert`-specific info would still need to be retained, in order to perform convert functions. Maybe `EchoData` could be expanded to include a `Convert` object that would be empty (`None`) when using `open_converted` and populated when using `open_raw`?

leewujung:
> What's the role of EchoDataOld vs EchoData?
EchoDataOld would allow OO calls and chained command, and the calibration functions (wrapped in "Process" that will no longer exist after 5.0) would operate on EchoDataOld instances instead of xarray Dataset and put the calibrated Sv into an attribute of EchoDataOld. So in short EchoDataOld was the design before we had the lengthy/convoluted discussions on how the API should be like.

The new `EchoData` also needs to be able to handle combining files as we finish that component in `convert` and can just use it as an under the hood mechanism.

emiliom:
> So in short EchoDataOld was the design before we had the lengthy/convoluted discussions on how the API should be like.
>
> The new EchoData also needs to be able to handle combining files as we finish that component in convert and can just use it as an under the hood mechanism.

`EchoDataOld` is still being used in tests and in `process/process.py`. Can you clarify if the intent is to migrate all functionality currently there and nowhere else, to `EchoData` and elsewhere, before or after version 0.5?

leewujung:
@emiliom : `process` will actually go away entirely, and also many test_ scripts, so `EchoDataOld` won't be used. I just keep them there for now as I revise the tests to make sure things are not missed accidentally.

leewujung:
> I don't think open_raw is accessible at the package level yet, but I've had the same thought about it.

Right, and we can discuss whether to do it that way. At the moment the other things under `convert` are not functioning fully (the `convert_combine` and `add_ancillary`) but they will be, so may be good to keep `open_raw` under `convert`.

leewujung:
> How large a change would it be? I suppose that open_raw could work by doing what it does now via the Convert object, then using the results to instantiate an EchoData object. But a lot of the Convert-specific info would still need to be retained, in order to perform convert functions. Maybe EchoData could be expanded to include a Convert object that would be empty (None) when using open_converted and populated when using open_raw?

@emiliom : this last part sounds very nice conceptually that you always get an EchoData object in some way when you `open_` something.

Change wise, if we have something like `echodata.convertor` as an attribute like what you said, the mechanism/code does not need to change much. We can link up `echodata.to_zarr()` and `echodata.convertor` under the hood pretty easily.

emiliom:
> @emiliom : process will actually go away entirely, and also many test_ scripts, so EchoDataOld won't be used. I just keep them there for now as I revise the tests to make sure things are not missed accidentally.

Got it, thanks.

echopype/echodata/echodata.py
```diff
@@ -0,0 +1,57 @@
...
+        self.environment = self._load_groups(raw_path, group='Environment')
+        self.vendor = self._load_groups(raw_path, group='Vendor')
+
+    @staticmethod
+    def _load_groups(filepath, group=None):
+        # TODO: handle multiple files through the same set of checks for combining files
+        suffix = Path(filepath).suffix
+        if suffix == '.nc':
+            return xr.open_dataset(filepath, group=group)
+        elif suffix == '.zarr':
```
lsetiawan, line 54: I think now, you can open_zarr with open_dataset. Just use `engine='zarr'`.

leewujung: Oh cool! Thanks for the note! There are probably a few simplifications we can do in the code because of this.

emiliom: Yes, that's true (starting in xarray [0.16.2](http://xarray.pydata.org/en/stable/whats-new.html#v0-16-2-30-nov-2020)). I don't know whether the plan is to deprecate the zarr-specific function calls, though; so, no harm leaving this as is.

leewujung:
Alright, we made good progress today in the discussion! 

Here's a summary:

One file case: `open_raw` will create 1 EchoData object that corresponds to content of the single input raw file
```python
echodata = ep.convert.open_raw('raw_file')  # parse raw file
echodata.beam  # this will be an xr.Dataset
echodata.to_zarr('zarr_path')  # serialize to zarr
```

Multiple files case: `open_mfraw` will create 1 EchoData object that combines the content of a list of multiple raw files
```python
echodata = ep.convert.open_mfraw(['raw_file1', 'raw_file2'])  # parse and combine content of multiple raw files
echodata.beam  # this will be an xr.Dataset
echodata.to_zarr('zarr_path')  # serialize to zarr
```

leewujung:
I realized that what we discussed today re `open_raw` and `open_mfraw` is actually more related to convert than to calibration, even though it involves changing `EchoData`, so I am going to merge this and start a new branch for that.

PR #276, opened 2021-04-09
Fix CI for testing

lsetiawan:
## Overview

This PR attempts to fix the current failing tests. Initial investigation leads to the need to pin fsspec version to a known working version of `0.8.5` seems like the current fsspec verison of `0.9.0` broke the CI.

### DIFF
--- a/.ci_helpers/docker/setup-services.py
+++ b/.ci_helpers/docker/setup-services.py
@@ -1,0 +1,67 @@
+"""setup-services.py
+
+Script to help bring up docker services for testing.
+
+"""
+import os
+import shutil
+import sys
+from pathlib import Path
+import argparse
+
+HERE = Path('.').absolute()
+BASE = Path(__file__).parent.absolute()
+COMPOSE_FILE = BASE / "docker-compose.yaml"
+TEST_DATA_PATH = HERE / "echopype" / "test_data"
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Setup services for testing')
+    parser.add_argument(
+        '--deploy', action='store_true', help="Flag to setup docker services"
+    )
+    parser.add_argument(
+        '--tear-down',
+        action='store_true',
+        help="Flag to tear down docker services",
+    )
+
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    args = parse_args()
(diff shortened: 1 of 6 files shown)


lsetiawan:
I just bumped fsspec to `0.8.7`. This is the last working version right before `0.9.0`.

emiliom:
Sigh. Looking forward to `fsspec` reaching 1.0 and becoming more stable (fewer breaking changes). Nice detective work

lsetiawan:
> Sigh. Looking forward to fsspec reaching 1.0 and becoming more stable (fewer breaking changes). Nice detective work

Thanks! Yea, it would be good to not have to pin libraries ... I think for now we'll just do this, but eventually would be good to adjust our code to comply with the newest version.

.ci_helpers/docker/setup-services.py
```diff
@@ -0,0 +1,67 @@
+"""setup-services.py
```
lsetiawan, line 1: @emiliom You'd enjoy this script! Hopefully that will make it easier to bring up the necessary services. Try it out!

```bash
python .ci_helpers/docker/setup-services.py --help
```

I'm wondering if this should be an echopype CLI, but this is mostly for testing so idk. What do you think?

lsetiawan:
Merging this for now so that it can be used by new PR's

.ci_helpers/docker/setup-services.py
emiliom: > @emiliom You'd enjoy this script! Hopefully that will make it easier to bring up the necessary services. Try it out!

Nice indeed! I'll try it out, but not today.

emiliom: > I'm wondering if this should be an echopype CLI, but this is mostly for testing so idk. What do you think?

That's an interesting prospect. But my vote would be to banish any discussion of an echopype CLI until 0.5.0 is released! Otherwise, I fear it could be a distraction :disappointed:

leewujung: Hah, there _was_ an echopype CLI, but I removed it at some point. For the workflow stuff we will need it, and I don't think it's hard to put it in place, but like what @emiliom said, let's push that to after v5.0. :)

PR #277, opened 2021-04-09
Use setuptools scm

lsetiawan:
## Overview

Porting auto-versioning to [pypa/setuptools_scm](https://github.com/pypa/setuptools_scm) instead of versioneer. This would slim down the package size and also work toward fixing the pypi publish github actions.

## Size comparisons from (219kB to 104kB)

Screenshot from 2021-04-09 14-12-39


emiliom:
> Size comparisons from (219kB to 104kB)

Wow, that's crazy! 
Thanks!

leewujung:
Wow indeed. Thanks! 
@lsetiawan feel free to merge once you think things are in good order.

lsetiawan:
Had to do a little fix for the pypi github action, but here you go, a built dev version: https://test.pypi.org/project/echopype/0.4.1.dev439/ :smile:

PR #281, opened 2021-04-16
Update echodata to have better repr and dynamic groups attributes

lsetiawan:
## Overview

This PR modifies EchoData object to have better repr when printed on screen. Additionally the creation of groups attributes to the class is now dynamic.

### Closed
Screenshot from 2021-04-16 14-40-55

### Opened
Screenshot from 2021-04-16 14-52-48

### Text repr
Screenshot from 2021-04-16 14-53-00


lsetiawan:
I don't have tests for this yet. I am working on fixing tests, so I'll add test for EchoData with that PR. Thanks.

lsetiawan:
Tests added, but not running successfully since the docker image doesn't contain the test data.

leewujung:
Do you want me to merge this now so that you can add test properly?

leewujung:
This printout looks awesome and dramatically increase the data readability!! 

What do you think if we add the actual attribute names (which are the netCDF group names) to each entry? e.g., `top` for Top-level, `environment` for Environment, `vendor` for Vendor specific. This seems can help avoid confusion and also provide guide for users directly.

lsetiawan:
> What do you think if we add the actual attribute names

Would this be okay: `> top: (Top-level) contains blah blah.`

lsetiawan:
Closing this. Implemented in #282

Issue #228, opened 2020-12-24
separate nc4/xarray dataset group creation from file saving functionality


emiliom:
In the new [class-redesign branch](https://github.com/OSOceanAcoustics/echopype/tree/class-redesign), the `SetGroup*` classes and functionality perform two related but distinct functions: create the xarray dataset for each nc4 group, and save it to file (nc or zarr). Separating those functions would provide more flexibility to power users, without impacting regular users who use `Convert` followed by the new `to_netcdf` and `to_zarr` methods to write to files.

See https://github.com/OSOceanAcoustics/echopype/discussions/225 for more discussions on this.

leewujung:
I was going to implement this yesterday but thought of a possible issue: in the case when a Dataset is large, by returning the Dataset itself as output from all `SetGroup*.set_*` methods and save them all into file in, say, a `save_all` method, it seems that we would be keeping the large Datasets around until saving them. Obviously one way to circumvent this to not have the `save_all` method but create (using `set_*`) --> save to file --> destroy the Dataset object.

The majority of data sits in the `Beam` group that holds the backscatter data. The file size is something the users can choose. The largest file I've seen so far is ~300 MB from EK80.

Thoughts?

emiliom:
I've already forgotten the details of how this stuff works :disappointed: .... (though I did re-read the associated discussion, #225). 

But, my main comment would be to not focus on this issue at this time. There's already plenty to do to get 0.5.0 out the door. This issue can wait.

PR #282, opened 2021-04-19
Change output of open_raw to an EchoData object

leewujung:
This PR:
- changes output of `open_raw` from a `Convert` object to an `EchoData` object
- implements the behavior for `open_raw` to parse the raw data files and organize the parsed data into attributes of the output `EchoData` object. This requires user to use `EchoData.to_zarr/to_netcdf` to save the parsed data to file. This addresses #228 and is related to #225.

Note:
- `open_raw` only works with parsing a single raw data file. 
- `open_raw` and `open_converted` are now under `echodata/open_file.py`. I was debating if `open_raw` should remain in the `convert` module but eventually moved it to under `echodata`. Let me know if you think otherwise.

Todo for later PRs:
- still need to implement `open_mfraw` to parse multiple raw files and combine the parsed data into attributes of a single `EchoData` object.
- we should remove the `Convert` class since its function as an interface for using `ParseXX` and `SetGroupsXX` objects can now be substituted with `Echodata`

## Updates
- `Convert` is now deprecated with all functionalities provided directly by `Parse` and `SetGroup` objects
- Next steps are to:
   -  take care of specific functions under echopype/convert/api.py, such as _validate_path and to_file, to just provide with 1-to-1 (input/output) mapping
   - Other singular changes in https://github.com/leewujung/echopype/pull/3


leewujung:
@lsetiawan : Could you take a look at the use of `storage_options` and `output_storage_options` in `EchoData` this PR? I am pretty sure when trying to transplant `_validate_path` and associated functions I broke your `test_convert`, because the output of `open_raw` is now an `EchoData` object instead of a `Convert` object. Thanks!

lsetiawan:
Looks like you integrated PR #281 into here is that correct?

echopype/echodata/echodata.py
```diff
@@ -1,57 +1,455 @@
...
+        return out_path  # output_path is always a list
+
+    @staticmethod
+    def _normalize_path(out_f, convert_type, output_storage_options):
+        if convert_type == "zarr":
+            return fsspec.get_mapper(out_f, **output_storage_options)
+        elif convert_type == "netcdf4":
+            return out_f
+
+    def _to_file(
```
lsetiawan, line 231: Hmm... looks like you integrated a lot of the stuff from Convert Object ... Not sure if this is the best way to go because I can see EchoData ending up being convoluted ... I think it might be better if you keep all of these functions in the Convert Object and use that within EchoData.

echopype/tests/test_convert.py
```diff
@@ -14,7 +14,7 @@
 import xarray as xr
 import pytest
 from pathlib import Path
-from ..convert import open_raw
+from ..echodata import open_raw
```
lsetiawan, line 17: For tests, it's better if you don't do relative paths since `echodata` should be installed before you run pytest. So you can just do `from echodata import ...`

echopype/echodata/echodata.py
leewujung: Yeah, those transplanted functions are my biggest hesitation, as in whether to keep them in Convert or move them to EchoData. (I was thinking about getting rid of the Convert object!) But looking at this again maybe it'll be better that way.

echopype/tests/test_convert.py
leewujung: I see, cool, thanks for explaining this!

leewujung:
> Looks like you integrated PR #281 into here is that correct?

Yes, wanted to make sure I got your changes in before starting this.

echopype/echodata/echodata.py
lsetiawan: I see. I think it's okay to keep the Convert object. Since we know it already works and this would split up the functionalities, and we can just test individually.

leewujung: Oh now I remember why I did that: the main thing is that now the set groups functions return xr.Dataset and are linked directly to attributes of EchoData in open_raw. That's why I felt the Convert class can actually be removed and substituted by the same functionalities in EchoData. This does make EchoData more convoluted though I agree. Hmm... thoughts?

leewujung:
Noting conversations in https://github.com/leewujung/echopype/pull/3 for more changes on this PR.

leewujung:
@lsetiawan thanks for the changes! I just put updates the PR message to include next steps and will merge this now.

PR #284, opened 2021-04-21
Add summary statistics computation

leewujung:
In this PR I put in a few simple functions to compute local summary statistics from echograms. There are few other functions still to be added, but I am interested in seeing what others' thoughts are on the approach to just put these as simple functions that can be applied on the calibrated Sv xarray Dataset echopype generates, rather than putting things into a class. It seems that for debugging and understanding the operations, this is more straightforward.

@valentina-s : there's this specific thing about `zname` which is related to what we discussed in terms of whether one has `range` or `depth` in the Sv xarray dataset. `zname` is such a bad variable name which should be changed, but what do you think about this way for users to identify that particular variable.

And yes these need docstrings!!


echopype/metrics/summary_statistics.py
```diff
@@ -0,0 +1,56 @@
...
+
+def delta_z(ds, zname='range'):
+    if zname not in ds:
+        raise ValueError(f"{zname} not in the input Dataset!")
+    dz = ds[zname].diff(dim='range_bin')  # here assume range or depth are data variables with the same dims as Sv
+    return dz.where(dz != 0, np.nan)
+
+
+def convert_to_linear(da_Sv):
+    return 10 ** (da_Sv / 10)  # convert Sv to linear domain
```
lsetiawan, line 22: These all look good to me. Maybe instead of comments, you can put docstring of what these functions do? This way users know. Otherwise, should probably make it private functions, if this is planned to be used by other classes/functions.

lsetiawan (dismissed):

echopype/metrics/summary_statistics.py
leewujung: Good point about making it a private function.

And like I said, yes we need docstrings for all of these!

valentina-s:
I guess it is more philosophical question of the scope of `echopype`. Is `echopype` just doing the piping, and other packages focus on analysis, or it will have analysis modules. Although these are simple stats, their interpretation may not be that simple, and one can start thinking about all the other metrics to include. In the context of unmoored echosounders, one might want to also use bottom annotation/detection to be integrated with those. Since there is already an echometrics package one could revamp it to work netcdf format.

leewujung:
The functions here are meant to work with calibration outputs generated by echopype directly. The format or convention for processed data is a wild west. Since these are lightweight functions, I think it is more straightforward for the users to just include them in the package. I added them under `metrics` because other lightweight computations, like frequency-differencing, are also in the same category and can be included here. These functions accepts xr.Dataset as input and returns xr.Dataset, but in an expected format that contains labels. I did look at the echometrics package and ref it on top of the script, but I decided that it is easier to just implement the functions rather than dealing with a specific non-labeled structure, since the computations are themselves very simple.

In terms of philosophy: the "pipe" in my thinking has been a "pipeline," and not simply piping the data from one format to the other.

echopype/metrics/summary_statistics.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+import numpy as np
+
+
+def delta_z(ds, zname='range'):
+    if zname not in ds:
+        raise ValueError(f"{zname} not in the input Dataset!")
+    dz = ds[zname].diff(dim='range_bin')  # here assume range or depth are data variables with the same dims as Sv
+    return dz.where(dz != 0, np.nan)
```
leewujung, line 18: change this line to `dz.where(dz != 0, other=np.nan)` for readability.

stuara:
opened new PR (w/ old commits) that I am able to push to

PR #288, opened 2021-04-22
Setup automated update of docker images that contain test_data.

lsetiawan:
## Overview

This PR setup the docker build and ci workflow to be able to work together and provide the latest test_data.


lsetiawan:
Checks are failing since latest is not currently found. Merging.

Issue #286, opened 2021-04-22
Revamp tests to work with `open_raw` and misc fixes


leewujung:
With #282 now we have a stable and consistent pattern to access/use both raw and converted data, which is great, but that also leaves the original convert tests to revamp due to the API change. 

There are also some more granular changes @lsetiawan and I thought would be good to make changes for with the new `open_raw` code. 

Below is a checklist for tying up these loose ends. 

- [x] change all instances with mixed use of `Path` and `os` to all use `Path` in `echopype/convert/api.py`, such as in [here](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/api.py#L99) and [here](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/api.py#L99)
- [x] use `Path` in all cases where `os` is currently used on `echopype/convert/api.py`
- [x] make [`to_file`](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/api.py#L142) and [`_validate_path`](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/api.py#L50) to work with just one file only. We can do the multiple files part when we get to make `open_mfraw`. For discussions on this see https://github.com/leewujung/echopype/pull/3
- [x] change input arguments `file` --> `raw_file` and `model` --> `sonar_model` under [`open_raw`](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/api.py#L428-L434) -- this has implications with all tests
- [x] overhaul all tests in `test_convert.py` involving using the deprecated `Convert` object to use `open_raw` and the resulting `EchoData`
- [x] overhaul all tests in `test_convert_....py` that currently save an output file before loading the data back in to compare with test datasets, to just compare the attributes of the returned `EchoData` object and the test datasets

@lsetiawan feel free to chime in on things I may have missed or modify the above.

Thanks @ngkavin !

lsetiawan:
@ngkavin There is a helper python script that will setup the services for testing. Run the command below on the root of echopype. Make sure you have `docker` and `docker-compose` installed. Let me know if you have any questions! Make sure to perform tests against these services.

### Deploying 
```
python .ci_helpers/docker/setup-services.py --deploy
```

### Tearing down
```
python .ci_helpers/docker/setup-services.py --tear-down
```

ngkavin:
[...]

PR #289, opened 2021-04-22
Replace os operations with pathlib

ngkavin:
The current Echodata API mixes use with the `os` and `pathlib` libraries in order to perform operations on file paths. This PR replaces all occurances of the `os` library with the corresponding `pathlib` operations.

This PR also makes the `to_file` and `_validate_path` functions work with only 1 file at a time.

These changes are suggested as part of issue #286


lsetiawan:
I am going to close and reopen this PR so that it runs with update CI.

lsetiawan:
The failures in CI is now due to issues in test files now.

PR #293, opened 2021-04-23
Fix convert tests

ngkavin:
Fix convert tests so that they work with the changes to EchoData and the EchoData API as described in issue #286.

This also fixes an issue in the API which caused tests to fail if the source/save path was s3://, http://, etc. This issue was due to Windows machines using backslashes instead of  forward slashes, so a link like `http://azfp/file.raw` would turn into `http:\\azfp\\file.raw` when casted to a string from a Path object. The casted link is invalid. 

The EchoData API now supports passing in a Path object, but it does not cast to a Path object if paths are given as strings.


ngkavin:
@lsetiawan Some of the files in the CI test data seem to be corrupted. In the drive, there are 2 versions of 17031001.01A for AZFP and I think the smaller one is being used in the tests. The tests report 'Error: Unknown file type'  while trying to parse the file. It works when I replace this file with the larger version though.

lsetiawan:
I have updated the test docker image. Rerunning tests now. Assuming that you've updated the tests data in google drive, this should work.

echopype/tests/test_convert.py
```diff
@@ -14,7 +14,7 @@
 import xarray as xr
 import pytest
 from pathlib import Path
-from ..echodata import open_raw
+from .. import open_raw
```
lsetiawan, line 17: Let's try to be explicit here for tests file. You can simply just do `from echopype import open_raw`.

echopype/tests/test_convert_azfp.py
```diff
@@ -11,10 +11,11 @@
 import xarray as xr
 import pandas as pd
 from scipy.io import loadmat
-from ..echodata import open_raw
+from .. import open_raw
```
lsetiawan, line 14:
```suggestion
from echopype import open_raw
```

echopype/tests/test_convert_ek60.py
```diff
@@ -3,9 +3,10 @@
 import xarray as xr
 import pandas as pd
 from scipy.io import loadmat
-from ..echodata import open_raw
+from .. import open_raw
```
lsetiawan, line 6:
```suggestion
from echopype.echodata import open_raw
```

echopype/tests/test_convert_ek80.py
```diff
@@ -4,9 +4,10 @@
 import xarray as xr
 import pandas as pd
 from scipy.io import loadmat
-from ..echodata import open_raw
+from .. import open_raw
```
lsetiawan, line 7:
```suggestion
from echopype import open_raw
```

lsetiawan:
Due to `fsspec` and `s3fs` version incompatibility, tests are currently failing. PR https://github.com/ngkavin/echopype/pull/10 should fix this.

Right now `s3fs=0.6.0` and `fsspec=0.8.7` get installed ... well this combination doesn't work.

echopype/convert/api.py
```diff
@@ -73,14 +74,19 @@ def _validate_path(
         out_path = out_dir / (Path(source_file).stem + file_format)
 
     else:
-        fsmap = fsspec.get_mapper(save_path, **output_storage_options)
+        if isinstance(save_path, str):
+            save_path = Path(save_path)
```
lsetiawan, line 78: I think the issue is here. Using pathlib Path removes the `://` and turns it to `:/`.

Screenshot from 2021-04-26 12-03-18

echopype/convert/api.py
```diff
@@ -90,7 +96,7 @@ def _validate_path(
             out_path = out_dir / (root.stem + file_format)
 
     # Create folder if save_path does not exist already
-    fsmap = fsspec.get_mapper(str(out_dir), **output_storage_options)
+    fsmap = fsspec.get_mapper(out_dir.as_posix().replace(':/', '://'), **output_storage_options)
```
lsetiawan, line 99: This is really hacky, I think there should be a check for `://` instead higher up in the function. If the output string has `://` It's a sign that this is not a local path, so don't use pathlib to parse the save_path string, but rather keep it the way it is and use fsspec.

echopype/tests/test_convert.py
```diff
@@ -143,6 +144,8 @@ def download_files():
 def test_validate_path_single_source(
     model, file_format, input_path, output_save_path, minio_bucket
 ):
+    def normalize_path(path):
+        return path.replace("\\", "/").replace(':/', '://')
```
lsetiawan, line 148: What does this do? Is this to catch cases for windows path?

echopype/convert/api.py
emiliom: Yes, Path `doesn't` handle url's, last time I checked. So you have to test for Path vs (http://, https://) vs s3:// (the last two sets as strings), and make decisions accordingly.

echopype/tests/test_convert.py
```diff
@@ -335,23 +331,30 @@ def test_convert_ek60(
...
+    elif export_engine == "zarr":
+        to_file = getattr(ec, "to_zarr")
+    else:
+        return
+
+    to_file(
         save_path=output_save_path,
         overwrite=True,
-        combine=combine_files,
         storage_options=output_storage_options,
```
lsetiawan, line 354: The attribute should be `output_storage_options`. https://github.com/ngkavin/echopype/blob/86895a8134930f8cd6370c1b441aaa850ee1b873/echopype/convert/api.py#L122

lsetiawan: cc @ngkavin you can either change it on api.py or the test file. This is why it's failing because the correct credentials is not actually being used.

echopype/tests/test_convert.py
```diff
@@ -461,28 +468,32 @@ def test_convert_ek80(
...
+    elif export_engine == "zarr":
+        to_file = getattr(ec, "to_zarr")
+    else:
+        return
+
+    to_file(
         save_path=output_save_path,
         overwrite=True,
         combine=combine_files,
         storage_options=output_storage_options,
```
lsetiawan, line 496:
```suggestion
        output_storage_options=output_storage_options,
```

echopype/tests/test_convert.py
```diff
@@ -417,15 +417,22 @@ def test_convert_azfp(
...
```
lsetiawan, line 432:
```suggestion
        output_storage_options=output_storage_options,
```

echopype/convert/api.py
```diff
@@ -154,6 +154,7 @@ def to_file(
         source_file=echodata.source_file,
         file_format=format_mapping[engine],
         save_path=save_path,
+        output_storage_options=output_storage_options,
```
lsetiawan, line 157: There's the culprit! It's trying to access AWS instead of local minio!

lsetiawan: Works locally because you're actually accessing AWS. :stuck_out_tongue_closed_eyes: 
Screenshot from 2021-04-26 17-22-49

lsetiawan (approved):
All good now!

PR #295, opened 2021-04-24
Add preprocess module

leewujung:
## Overview

This PR adds the `preprocess` module that contains functions that are previously accessed through the `Process` such as `get_MVBS` and `remove_noise`. These functions are renamed and revamped to take xarray datasets produced by `calibrate.compute_Sv` as inputs and produce xarray datasets as outputs to conform with the new API access pattern we plan to roll out in v0.5.

Three functions are included:
- `remove_noise`: equivalent to the previous `remove_noise`
- `estimate_noise`: equivalent to the previous `estimate_noise`
- `compute_MVBS_index_binning`: equivalent to the previous `get_MVBS` function, which uses `xarray.coarsen.mean` to obtain bin-averaged Sv
- `compute_MVBS`: this is new and potentially more useful in practical applications, as it obtain bin-averaged Sv based on physical units in the `range` and `ping_time` dimension (i.e., in meters and time span such as `10S`)

## Notes
The outputs of `compute_MVBS` and `compute_MVBS_index_binning` are different due to the operation involved:
- `compute_MVBS` has a dimension/coordinate `range` which is in units meter. This means that all data variables are aligned to the same 3D grid (`frequency`, `ping_time`, `range`), i.e., the output have the same `range` resolution across all frequency channels.
- `compute_MVBS_index_binning` still has a dimension/coordinate `range_bin` as in the input Sv dataset, and a data variable `range`, which itself is a 3D array along `frequency`, `ping_time`, and `range`. This is because the `xarray.coarsen.mean` operation merely produces binned averages according to the specified number of bins along `range_bin` and `ping_time` without attempting to align the potentially different `range` resolution across all frequency channels.

## TODO (for v0.5)
- [ ] revamp tests for `remove_noise` and `compute_MVBS_index_binning`
- [ ] add tests for `compute_MVBS`

## Next steps (beyond v0.5)

1. I think the `regrid` function should be added in this module here, as all the functions here are related to bringing the calibrated Sv data into a more uniform format.
2. We probably want to add a function that uses `regrid` under the hood to bring the output of `compute_MVBS_index_binning` to the same form as the output of `compute_MVBS`, i.e., with all data variables aligned to the same 3D grid.

@ngkavin : tagging you here for deprecating `Process` in v0.5

@emiliom : tagging you here as we do not follow any convention at the moment for outputs of MVBS, as in #294 

@ngkavin @emiliom @lsetiawan : I am also looking for suggestions on how to make the names of the two compute_MVBS functions more obvious to users.


leewujung:
I'll merge this now so that @ngkavin can continue to deprecate `Process`. Let's do the tests in another PR.

PR #299, opened 2021-04-27
Fix echodata repr test

lsetiawan:
## Overview

This PR fixes the repr test to match with the updated repr for echodata.

### DIFF
--- a/echopype/tests/test_echodata.py
+++ b/echopype/tests/test_echodata.py
@@ -35,13 +35,13 @@ class TestEchoData:
         expected_repr = dedent(
             f"""\
             EchoData: standardized raw data from {zarr_path_string}
-              > Top-level: contains metadata about the SONAR-netCDF4 file format.
-              > Environment: contains information relevant to acoustic propagation through water.
-              > Platform: contains information about the platform on which the sonar is installed.
-              > Provenance: contains metadata about how the SONAR-netCDF4 version of the data were obtained.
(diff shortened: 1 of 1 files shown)


lsetiawan:
I have setup branch protection and master need 2 approvals, so if you two (@emiliom, @leewujung) can approve that would be awesome :smile: Thanks! Let me know what you think of this branch protection! Is it too much to require 2 approvals?

leewujung:
Haha, i was wondering about that, when I forced merge last night. 😛

I suspect we may need to go by 1 reviewer due to time availability issue, but happy to go with 2 if we want to make sure all changes are vetted thoroughly. Let's see what @emiliom thinks.

emiliom:
> master need 2 approvals

So, that's 2 approvals **in addition** to your own? If so, yeah, IMHO that's too much. Anyway, I'll approve right now.

leewujung (approved):

emiliom:
I was able to merge. I guess that's b/c @leewujung had already approved?

lsetiawan:
You both have admin access hence you're able to merge, otherwise outside users would need 2 approval from any of us.

emiliom:
Ah! I thought you meant that even admins needed 2 approvals. Aren't you an admin? So, admins can still self merge into master?

leewujung:
@lsetiawan: I agree, I think for PR from outside users it's a good idea to have 2.

PR #302, opened 2021-04-27
Allow local test run of everything

lsetiawan:
This PR allows `run-test.py` to run locally of all the tests when user put `--local` flag.

### DIFF
--- a/.ci_helpers/run-test.py
+++ b/.ci_helpers/run-test.py
@@ -6,6 +6,7 @@ Script to run tests in Github.
 import argparse
 import glob
 import os
+import shutil
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


emiliom:
I think I'm confused. I remember a few weeks ago you had set up a script to run the tests locally. I don't remember if it was `run-test.py` or something else, but I think I remember that that's what first replaced the explicit use of `docker-compose` for local tests. So, what does the `--local` flag add that wasn't there before? Has that other script already been removed? Or am I conflating this with `python .ci_helpers/docker/setup-services.py --deploy` that sets up the services, and previously you'd then use the usual `pytest` on the tests?

emiliom:
@lsetiawan This PR is closed, but I don't know where else to report this (other than Slack).

I finally was able to get set up to run the tests locally using this mechanism. The conversion tests failed extensively. I see a lot of `s3fs` references in the errors, but there's too much there to get a good global take on the errors. Here's the final, summary line:
```
=== 11 passed, 14 skipped, 1 warning, 154 errors in 67.27s (0:01:07) ===
```

The calibration tests were largely successful:
```
=== 7 passed, 31 warnings in 98.95s (0:01:38) ===
```

That's after successfully running `python .ci_helpers/docker/setup-services.py --deploy`, naturally.

Issue #303, opened 2021-04-27
Revamp/add tests for `preprocess` module


leewujung:
This came from the remaining TODOs of #295 :
- [x] revamp tests for `remove_noise` and `compute_MVBS_index_binning`
- [x] add tests for `compute_MVBS`

PR #304, opened 2021-04-28
Deprecate old Process class

ngkavin:
This PR adds `process_deprecated.py`. This file contains the `Process` class that provides backwards compatibility with echopype version 0.4.1. The new `Process` class saves the new `EchoData` object and passes it to the new Calibrate and Preprocess functions for calibration, calculating MVBS, and removing noise.


echopype/echodata/echodata.py
```diff
@@ -193,7 +193,11 @@ def nc_path(self):
             DeprecationWarning,
             2,
         )
-        return self.converted_raw_path
+        if self.converted_raw_path.endswith('.nc'):
+            return self.converted_raw_path
+        else:
+            path = Path(self.converted_raw_path)
+            return str(path.parent / (path.stem + '.nc'))
```
lsetiawan, line 200: What does this do? Isn't `converted_raw_path` either ends with `.nc` or `.zarr`?

ngkavin: When converting to nc then to  zarr, we lose the .nc filepath so nc_path would have a .zarr extension. Also, having access to both names was supported in the old version.

lsetiawan: Great! Thanks for those explanation :)

leewujung:
@ngkavin : thanks for the PR!
Except for the `.astype(np.int64)` problem I think we can merge this once the other smaller things are changed.
The revamping tests work #303 will be the tests for the linkages you put in here under `process_deprecated.py` .

emiliom:
I'm pasting my comment from Slack, vis a vis the data type for frequency:

> In the convention document, I now see that there are four frequency variables (with units of Hz) listed. All of them are defined to be **float** type. I would've guessed an int type, too. Geez, I guess frequencies defined as tenths of a Hz are not unusual? Anyway, that makes the int64 vs int32 issue less ... um, relevant?

leewujung:
> I'm pasting my comment from Slack, vis a vis the data type for frequency:
> 
> > In the convention document, I now see that there are four frequency variables (with units of Hz) listed. All of them are defined to be **float** type. I would've guessed an int type, too. Geez, I guess frequencies defined as tenths of a Hz are not unusual? Anyway, that makes the int64 vs int32 issue less ... um, relevant?

Ugh, my bad, I just assumed this to be one of those variables that are supposed to be `int32`. Frequency should be `float` to allow fractions. 

@ngkavin could you look into where in the AZFP convertor that encoded `frequency` to be `int`? We will need that to be in a different PR to keep a record of that issue for the release note.

echopype/preprocess/noise_est.py
```diff
@@ -31,7 +31,7 @@ def _compute_transmission_loss(self):
         """
         if 'sound_absorption' not in self.ds_Sv:
             sound_absorption = uwa.calc_absorption(
-                frequency=self.ds_Sv.frequency,
+                frequency=self.ds_Sv.frequency.astype(np.int64),
```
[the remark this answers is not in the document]
leewujung: Alright, let's resolve #307 via a PR based on @emiliom 's comments below and then revert this change here before merging this PR. :)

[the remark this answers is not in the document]
leewujung: #309 resolves this.

leewujung (approved):

PR #297, opened 2021-04-25
Add function to correct ping_time reversal

leewujung:
This PR adds a function to detect and correct reversed `ping_time` sometimes observed in EK60. 

If no `ping_time` reversal is detected the original timestamps are returned.

If reversed `ping_time` exists, the function uses the median `ping_time` difference within a local window to infer and correct the timestamp.

@emiliom : could you try this on the hake survey file you've seen this problem before? Thanks!

PR #305, opened 2021-04-28
Fix `PulseDuration` vs `PulseLength` difference in EK80 config XML

imranmaj:
Fixes #297


leewujung:
@imranmaj : thanks for the PR! could you pull the latest master into your PR branch?

leewujung:
Awesome! We can merge this now that everything passes with the slightly hacky thing to check if `configuration` exists in the parsed data.

A note that this check really should be whether the data is from EK60 (does not have `configuration`) or EK80 (has `configuration`).

PR #306, opened 2021-04-29
Add s3fs with pinned version to requirements

emiliom:
Add s3fs to requirements.txt, pinned to the version currently found to work with the fsspec version that was already pinned.

Having s3fs in requirements.txt (and therefore in the pypi and conda packages) will make it easier for others to use the S3 I/O features.


lsetiawan:
This looks great! Thanks @emiliom!

lsetiawan (approved):

Issue #307, opened 2021-04-29
Compliance of frequency variable encoding in converted files


leewujung:
### Original

It appears that `frequency` has been encoded in `int32` for AZFP converted files, which results into errors when computing for environmental variables #304 . Let's fix this by encoding it as `float` as specified in the convention.

### Updated

Per @emiliom 's comments below, let's expand the scope of this issue beyond AZFP, so that we do not forget about this other side of compliance issue in conjunction with #210 .

leewujung:
Closed by #309 .

emiliom:
The same change will have to be made for the other sensor types (EK60, etc), right? How about expanding the scope of this issue (and reopening it) beyond AZFP, so the others are not forgotten?

leewujung:
Yes I was gonna add a comment in the convention issue. We can reopen this issue and do everything here. :)

leewujung:
I think we can probably make this in a minor release v0.6.X once we have the main (breaking) things down.

emiliom:
The change to float was done for AZFP in PR #309 (originally to the old `frequency` variable).

For EK60 and EK80, `frequency_nominal` is already set to float, too. 

I think we can close this issue.

leewujung:
Sound good. I imagine this will be covered by #1043

emiliom:
Thanks. Yeah, the goal in #1043 is to also check for data types.

I'll close this issue now.

PR #309, opened 2021-04-29
Cast AZFP frequency to float64

ngkavin:
The current AFZP parser automatically assigns the dtype of `int32` or `int64` depending on the user's OS. This may cause issues in the case of frequency being type `int32` where calculations may cause an integer overflow. The SONAR-netCDF4 convention also states that the frequency should be a floating point number to account for fractional frequencies.

This small PR casts the AZFP frequency to be `float64` which fixes the aforementioned issues.

Addresses #307


leewujung:
Thanks @ngkavin !

leewujung:
This closes #307 .

Issue #310, opened 2021-04-29
Error when trying to calibrate an EK80 file with both CW and BB channels


ngkavin:
Calibration fails for the test file `Summer2018--D20180905-T033113.raw` when the calibration parameters are being retrieved. The error message is: `ValueError: operands could not be broadcast together with shapes (5,5) (3,1)`. 

Specifying the waveform_mode and the encode_mode still resulted in the error.

leewujung:
This is due to the fact that `_get_vend_cal_params_power` should not be called in the BB case, and the factors 'sa_correction', 'gain_correction' are not used for BB calibration. This requires some refactoring and let's try to get this in in the upcoming minor releases.

imranmaj:
Fixed in #400

PR #311, opened 2021-04-30
Fix issue with calibrating files containing BB and CW channels

ngkavin:
Fixes the issue described in #310 where trying to calibrate EK80 files containing both CW and BB channels would result in an error.

### DIFF
--- a/echopype/calibrate/api.py
+++ b/echopype/calibrate/api.py
@@ -29,12 +29,13 @@ def _compute_cal(
         env_params=env_params,
         cal_params=cal_params,
         waveform_mode=waveform_mode,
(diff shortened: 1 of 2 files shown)


ngkavin:
This PR is not final and does not actually contain the fix at the moment

leewujung:
@ngkavin : I think the root of the problem is not that you need access to `waveform_mode` and `encode_mode` as class attributes, but that `_get_vend_cal_params_power` should not be called in the BB case. Fixing this will require a bit of refactoring. The factors 'sa_correction', 'gain_correction' are not used for BB calibration. 

Let's delay fixing #311 to after v0.5 is out.

PR #312, opened 2021-04-30
Add tests for preprocess functions

ngkavin:
This PR adds a test for `remove_noise`, `compute_MVBS_index_binning`, and `get_MVBS`. These tests use fake data to test for an expected outcome.

The `remove_noise` test adds 2 points that do not match the background, and tests to see if they are removed. It also tests to see if the correct number of points are removed from a seeded normal distribution.

The `compute_MVBS_index_binning` test constructs data with values that increase every ping_num and range_bin_num so that when binned MVBS is performed, the result is a smaller array that increases by 1 for each row and column.

The `compute_MVBS` test is the same as the `compute_MVBS_index_binning` test only the data once averaged over seconds and meters gives a smaller array that increases by 1 for each `range_meter_bin` and `ping_time_bin`.

### DIFF
--- a/.ci_helpers/run-test.py
+++ b/.ci_helpers/run-test.py
@@ -88,6 +88,15 @@ if __name__ == "__main__":
                 if "echodata" not in test_to_run:
                     test_to_run["echodata"] = []
                 test_to_run["echodata"].append(file_path)
+            elif any(
+                [
+                    (file_path.match("echopype/preprocess/*")),
+                    (file_path.match("echopype/tests/test_preprocess*")),
+                ]
+            ):
+                if "preprocess" not in test_to_run:
+                    test_to_run["preprocess"] = []
+                test_to_run["preprocess"].append(file_path)
 
     total_exit_codes = []
     for k, v in test_to_run.items():
--- a/.ci_helpers/run-test.py
+++ b/.ci_helpers/run-test.py
@@ -97,6 +97,16 @@ if __name__ == "__main__":
                 if "preprocess" not in test_to_run:
                     test_to_run["preprocess"] = []
                 test_to_run["preprocess"].append(file_path)
+            elif any(
+                [
+                    (file_path.match("echopype/convert/convert.py")),
+                    (file_path.match("echopype/process/*"))
+                    (file_path.match("echopype/tests/test_old.py")),
+                ]
+            ):
+                if "old" not in test_to_run:
+                    test_to_run["old"] = []
+                test_to_run["old"].append(file_path)
 
     total_exit_codes = []
     for k, v in test_to_run.items():
--- a/.ci_helpers/run-test.py
+++ b/.ci_helpers/run-test.py
@@ -100,7 +100,7 @@ if __name__ == "__main__":
             elif any(
                 [
                     (file_path.match("echopype/convert/convert.py")),
-                    (file_path.match("echopype/process/*"))
+                    (file_path.match("echopype/process/*")),
                     (file_path.match("echopype/tests/test_old.py")),
                 ]
             ):
--- a/echopype/tests/test_preprocess.py
+++ b/echopype/tests/test_preprocess.py
@@ -169,7 +169,7 @@ def test_compute_MVBS_index_binning():
     test_array = _construct_MVBS_test_data(nfreq, data_binned_shape[1], data_binned_shape[2])
 
     # Test all values in MVBS
-    np.allclose(data_test, test_array, rtol=0, atol=1e-12)
+    assert np.allclose(data_test, test_array, rtol=0, atol=1e-12)
 
 
 def test_compute_MVBS():
@@ -227,12 +227,12 @@ def test_compute_MVBS():
     test_array = _construct_MVBS_test_data(nfreq, data_binned_shape[1], data_binned_shape[2])
 
     # Test all values in MVBS
-    np.allclose(data_test, test_array, rtol=0, atol=1e-12)
+    assert np.allclose(data_test, test_array, rtol=0, atol=1e-12)
 
     # Test to see if ping_time was resampled correctly
     test_ping_time = pd.date_range('1/1/2020', periods=np.ceil(ping_num), freq=f'{ping_time_bin}S')
-    np.array_equal(data_test.ping_time, test_ping_time)
+    assert np.array_equal(data_test.ping_time, test_ping_time)
 
     # Test to see if range was resampled correctly
     test_range = np.arange(0, total_range, range_meter_bin)
-    np.array_equal(data_test.range, test_range)
+    assert np.array_equal(data_test.range, test_range)
--- a/echopype/tests/test_preprocess.py
+++ b/echopype/tests/test_preprocess.py
@@ -97,7 +97,7 @@ def test_compute_MVBS_index_binning():
 
     # Shape test
     shape = np.ceil((nfreq, npings / ping_num, nrange / range_bin_num)).astype(int)
-    assert np.all(ds_MVBS.Sv.shape == shape)
+    assert np.all(data_test.shape == shape)
 
     # Construct test array that increases by 1 for each range_bin and ping_time
     test_array = np.repeat(np.add(*np.indices((shape[1], shape[2])))[None, ...] + 1, nfreq, axis=0)
@@ -105,7 +105,7 @@ def test_compute_MVBS_index_binning():
     for f in range(nfreq):
         test_array[f] = (test_array[0] * (f + 1))
 
-    # Test values along range_bin
+    # Test all values in MVBS
     np.allclose(data_test, test_array, rtol=0, atol=1e-12)
 
 
@@ -126,13 +126,14 @@ def test_compute_MVBS():
 
     # Construct data with values that increase with range and time
     # so that when compute_MVBS is performed, the result is a smaller array
-    # that increases by 1 for each meter_bin and time_bin
+    # that increases by a constant for each meter_bin and time_bin
     data = np.ones((nfreq, npings, nrange))
     for p_i, ping in enumerate(range(0, npings, ping_rate * ping_time_bin)):
         for r_i, rb in enumerate(range(0, nrange, range_bin_per_meter * range_meter_bin)):
             data[0, ping:ping + ping_rate * ping_time_bin, rb:rb + range_bin_per_meter * range_meter_bin] += r_i + p_i
+    # First frequency increases by 1 each row and column, second increses by 2, third by 3, etc.
     for f in range(nfreq):
-        data[f] = data[0]
+        data[f] = data[0] * (f + 1)
 
     data_log = 10 * np.log10(data)      # Convert to log domain
     freq_index = np.arange(nfreq)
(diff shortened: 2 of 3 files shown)


leewujung:
> @leewujung I noticed that `compute_MVBS_index_binning` attaches an `sv` variable to the input Dataset.
> 
> https://github.com/OSOceanAcoustics/echopype/blob/8e62b571160b812d40c7cc3dc59d65fb09904200/echopype/preprocess/api.py#L87
> 
> This isn't present in `compute_MVBS` so it might be a bug.

You are right, it is a bug. Thanks for discovering it!

leewujung:
(Below is from deferring reviewing #304 to here)
@ngkavin : the deprecation in #304 seems mostly work well, but I noticed that in 
https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/tests/test_old.py#L52 and 
https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/tests/test_old.py#L65
the printout was something like: `saving calibrated TS to FILENAME_Sv.nc` which does not make sense. If it is TS then the output should be _TS.nc. This seems either a bug on how the old `Process.calibrate` was called or that something was breaking between calibration for Sv or TS. 

From the AZFP old tests (https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/tests/test_old.py#L87-L88)
it seems that there are some breaking links under the hood.

Could you take a look and see what's going on? Thanks!

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,151 @@
...
+    # so that when compute_MVBS is performed, the result is a smaller array
+    # that increases by 1 for each meter_bin and time_bin
+    data = np.zeros((nfreq, npings, nrange))
+    for p_i, ping in enumerate(range(0, npings, ping_rate * ping_time_bin)):
+        for r_i, rb in enumerate(range(0, nrange, range_bin_per_meter * range_meter_bin)):
+            data[0, ping:ping + ping_rate * ping_time_bin, rb:rb + range_bin_per_meter * range_meter_bin] += r_i + p_i
+    for f in range(nfreq):
+        data[f] = data[0]
+
+    data_log = 10 * np.log10(data)      # Convert to log domain
```
leewujung, line 126: Could you change the values in `data` to avoid warnings in `np.log10(0)` ?

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # Parameters for fake data
+    nfreq, npings, nrange = 1, 60, 600
+    range_meter_bin = 5          # range in meters to average over
+    ping_time_bin = 5            # number of seconds to average over
+    ping_rate = 2                # Number of pings per second
+    range_bin_per_meter = 30     # Number of range_bins per meter
+    ping_num = npings // ping_rate // ping_time_bin                      # number of pings to average over
+    range_bin_num = nrange // range_bin_per_meter // range_meter_bin     # number of range_bins to average over
+    total_range = nrange // range_bin_per_meter
```
leewujung, line 114: To test the `test_compute_MVBS` properly I think it'll be useful to make some changes to the toy data.
- use `nfreq` > 1
- make the total number of pings and the total number of range_bins **not** multiples of the number of pings and range_bins to average over, so that we know if the `groupby_bins(...).resample(...).mean()` operation is done correctly.

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,151 @@
...
+    assert np.count_nonzero(null.isel(frequency=0, range_bin=slice(None, 50))) == 6
+
+
+def test_compute_MVBS_index_binning():
+    """Test compute_MVBS_index_binning on toy data"""
+
+    # Parameters for fake data
+    nfreq, npings, nrange = 4, 40, 400
+    ping_num = 2             # number of pings to average over
+    range_bin_num = 3        # number of range_bins to average over
```
leewujung, line 63: Similar to the comment on `test_compute_MVBS` below, could you change `ping_num` or `nrange` so that `nrange` is not an integer multiple of `ping_num`? 

This will help us test that the `xr.coarsen(..., boundary='pad').mean()` was done correctly. It likely is already since along the `range_bin` dimension the reduction is not an integer multiple.

leewujung:
> @leewujung I noticed that `compute_MVBS_index_binning` attaches an `sv` variable to the input Dataset.
> https://github.com/OSOceanAcoustics/echopype/blob/8e62b571160b812d40c7cc3dc59d65fb09904200/echopype/preprocess/api.py#L87
> 
> This isn't present in `compute_MVBS` so it might be a bug.

Actually, somehow I thought you were referring to the outputs, but you were referring to the operations.

If you look in `compute_MVBS` the private function does include the same operation, it's just in a different place.

ngkavin:
> > @leewujung I noticed that `compute_MVBS_index_binning` attaches an `sv` variable to the input Dataset.
> > https://github.com/OSOceanAcoustics/echopype/blob/8e62b571160b812d40c7cc3dc59d65fb09904200/echopype/preprocess/api.py#L87
> > 
> > This isn't present in `compute_MVBS` so it might be a bug.
> 
> You are right, it is a bug. Thanks for discovering it!
> 
> ### Update
> Actually, somehow I thought you were referring to the outputs, but you were referring to the operations.
> If you look in compute_MVBS the private function does include the same operation, it's just in a different place. So we are fine here.

I was pointing out that the input dataset to this function is modified in-place with an `sv` variable. In `compute_MVBS` the input dataset is not modifed in-place. So after the function is called, the original Sv dataset that was passed in looks like this:
image

I was just wondering if this was an intended addition.

ngkavin:
> the printout was something like: `saving calibrated TS to FILENAME_Sv.nc` which does not make sense. If it is TS then the output should be _TS.nc. This seems either a bug on how the old

The `Process.calibrate` function is correctly calling `echopype.calibrate.compute_Sv` and `Process.calibrate_TS` is correctly calling `echopype.calibrate.compute_Sp`, however, Process.calibrate` incorrectly prints 'TS' when it should use 'Sv'.

.ci_helpers/run-test.py
```diff
@@ -88,6 +88,25 @@
...
+                    (file_path.match("echopype/tests/test_preprocess*")),
+                ]
+            ):
+                if "preprocess" not in test_to_run:
+                    test_to_run["preprocess"] = []
+                test_to_run["preprocess"].append(file_path)
+            elif any(
+                [
+                    (file_path.match("echopype/convert/convert.py")),
+                    (file_path.match("echopype/process/*"))
```
lsetiawan, line 103:
```suggestion
                    (file_path.match("echopype/process/*")),
```

ngkavin:
Addresses #303

lsetiawan (approved):
Tests for deprecated functions and preprocess are now running and passing. This looks good to me. As long as @leewujung also thinks it's good. I think we can merge this :smile: :partying_face:

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # Shape test
+    shape = (nfreq, ping_num, range_bin_num)
+    assert np.all(ds_MVBS.Sv.shape == np.ceil(shape))
+
+    data_test = (10 ** (ds_MVBS.Sv / 10)).round().astype(int)    # Convert to linear domain
+    # Test values along range_bin
+    assert np.all(data_test.isel(frequency=0, ping_time=0) == np.arange(range_bin_num) + 1)
+    # Test values along ping time
+    assert np.all(data_test.isel(frequency=0, range=0) == np.arange(ping_num) + 1)
```
leewujung, line 159: Add tests to check values in both the `range` and `ping_time` coordinates. That is the primary difference between this function and compute_MVBS_index_binning.

leewujung (changes requested):

leewujung:
> I was pointing out that the input dataset to this function is modified in-place with an `sv` variable.

I see, I misunderstood. Yes this should probably be better to be not modified in place. I'll look into the code again and submit a PR to your branch.

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,178 @@
...
+    # Binned MVBS test
+    ds_MVBS = ep.preprocess.compute_MVBS_index_binning(
+        ds_Sv,
+        range_bin_interval=range_bin_num,
+        ping_num_interval=ping_num
+    )
+    data_test = (10 ** (ds_MVBS.Sv / 10))    # Convert to linear domain
+
+    # Shape test
+    shape = np.ceil((nfreq, npings / ping_num, nrange / range_bin_num)).astype(int)
```
leewujung, line 99: I suggest changing `shape` to `data_binned_shape` to be explicit on what this is.

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,178 @@
...
+    range_bin_num = 7        # number of range_bins to average over
+
+    # Construct data with values that increase every ping_num and range_bin_num
+    # so that when compute_MVBS_index_binning is performed, the result is a smaller array
+    # that increases regularly for each row and column
+    data = np.ones((nfreq, npings, nrange))
+    for p_i, ping in enumerate(range(0, npings, ping_num)):
+        for r_i, rb in enumerate(range(0, nrange, range_bin_num)):
+            data[0, ping:ping + ping_num, rb:rb + range_bin_num] += r_i + p_i
+    # First frequency increases by 1 each row and column, second increses by 2, third by 3, etc.
```
leewujung, line 72: noticed a typo on "increses"

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,178 @@
...
+        ping_num_interval=ping_num
+    )
+    data_test = (10 ** (ds_MVBS.Sv / 10))    # Convert to linear domain
+
+    # Shape test
+    shape = np.ceil((nfreq, npings / ping_num, nrange / range_bin_num)).astype(int)
+    assert np.all(data_test.shape == shape)
+
+    # Construct test array that increases by 1 for each range_bin and ping_time
+    test_array = np.repeat(np.add(*np.indices((shape[1], shape[2])))[None, ...] + 1, nfreq, axis=0)
```
leewujung, line 103: This line of code is not very readable. 
Can you explain what the operation is for using `[None, ...]` and `np.repeat`?

I think it is equivalent to the following (separated out each step to be clear):
```
a = np.indices((binned_data_shape[1], binned_data_shape[2]))
b = np.add(*a)
c = np.array([b]*4)
c = c+1
```

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,178 @@
...
+    # Construct data with values that increase with range and time
+    # so that when compute_MVBS is performed, the result is a smaller array
+    # that increases by a constant for each meter_bin and time_bin
+    data = np.ones((nfreq, npings, nrange))
+    for p_i, ping in enumerate(range(0, npings, ping_rate * ping_time_bin)):
+        for r_i, rb in enumerate(range(0, nrange, range_bin_per_meter * range_meter_bin)):
+            data[0, ping:ping + ping_rate * ping_time_bin, rb:rb + range_bin_per_meter * range_meter_bin] += r_i + p_i
+    # First frequency increases by 1 each row and column, second increses by 2, third by 3, etc.
+    for f in range(nfreq):
+        data[f] = data[0] * (f + 1)
```
leewujung, line 136: Since you are reusing this code block for two functions, it'll be good to wrap this in a private function to make the overall logic clear

echopype/tests/test_preprocess.py
```diff
@@ -0,0 +1,178 @@
...
+    # Shape test
+    shape = np.ceil((nfreq, ping_num, range_bin_num)).astype(int)
+    assert np.all(data_test.shape == shape)
+
+    # Construct test array that increases by 1 for each range_meter_bin and ping_time_bin
+    test_array = np.repeat(np.add(*np.indices((shape[1], shape[2])))[None, ...] + 1, nfreq, axis=0)
+    # Increase test data by a multiple each frequency
+    for f in range(nfreq):
+        test_array[f] = (test_array[0] * (f + 1))
+
```
leewujung, line 168: Wrap this in a function too since it is also reused.

leewujung (changes requested):
@ngkavin I added comments on simplification and consolidating repeated code blocks.

leewujung (approved):

echopype/tests/test_preprocess.py
```diff
@@ -54,24 +54,90 @@ def test_remove_noise():
...
+    """
+
+    # Construct test array
+    test_array = np.indices((npings, nrange_bins))
+    test_array = np.add(*test_array)
+    test_array = np.array([test_array] * nfreq) + 1
+
+    # Increase test data by a multiple each frequency
+    for f in range(nfreq):
+        test_array[f] = (test_array[0] * (f + 1))
```
leewujung, line 121: Line 116-120 can be done using a list comprehension?
Results of line 116 are effectively not used and overwritten in 120.

It is fine to combine 114-115. As I said in the comments, I put them in separate lines in my code for clarity.

leewujung (changes requested):
Added another request to reduce redundancy.

leewujung (approved):

PR #319, opened 2021-05-02
Improve EK80 CW angle test

ngkavin:
The EK80 CW angle test now works. The proposed dtype changes to the parser was already implemented, so only the test was touched.

### DIFF
--- a/echopype/tests/test_convert_ek80.py
+++ b/echopype/tests/test_convert_ek80.py
@@ -43,7 +43,7 @@ def test_convert_ek80_complex_matlab():
     output_dir.rmdir()
 
 
-def test_convert_ek80_cw_power_echoview():
+def test_convert_ek80_cw_power_angle_echoview():
     """Compare parsed EK80 CW power/angle data with csv exported by EchoView.
     """
     ek80_raw_path_cw = str(ek80_path.joinpath('D20190822-T161221.raw'))  # Small file (CW)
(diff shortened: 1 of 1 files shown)


leewujung:
@ngkavin : I guess I had fixed that angle parsing back when we started using the pyecholab then, and forgot about it. I know that I did some comparison between the RAW0 and RAW3 angle data.

If I understand correctly the only thing you changed in the code was `atol=1e-5` --> `atol=1e-4` to make the test pass. 
There is a systematic difference from small to large as the range increase (you should plot to check), which I do not understand.

echopype/tests/test_convert_ek80.py
```diff
@@ -52,10 +52,10 @@ def test_convert_ek80_cw_power_echoview():
...
         for freq in freq_list
     ]
-    # ek80_echoview_angle_csv = [
-    #     ek80_path.joinpath('from_echoview/D20190822-T161221/%dkHz.angles.points.csv' % freq)
-    #     for freq in freq_list
-    # ]
+    ek80_echoview_angle_csv = [
+        ek80_path.joinpath('from_echoview/D20190822-T161221/%dkHz.angles.points.csv' % freq)
+        for freq in freq_list
+    ]
```
leewujung, line 58: These files are not on the Google drive.

leewujung (changes requested):

ngkavin:
> If I understand correctly the only thing you changed in the code was `atol=1e-5` --> `atol=1e-4` to make the test pass.

I also changed the `isel` values which were hard-coded for some reason as well as changing the `zip` to `enumerate`, but yes, there were very little changes needed to get the test working.

> There is a systematic difference from small to large as the range increase (you should plot to check), which I do not understand.

Can you elaborate on this? Are you referring to these spikes when plotting the difference?
image

leewujung:
great, thanks for the clarification, I'll look into the changes now. 

> Can you elaborate on this? Are you referring to these spikes when plotting the difference?
yes this is one view of that: I remember seeing more deviation at longer ranges, which seems to show up in your diff plot that the deviations are more numerous at longer ranges.

leewujung:
@ngkavin : I changed a few things by pushing directly to your fork:
- The `ping_idx` selection was actually _not_ supposed to work if the EchoView test data set and echopype outputs have the same number of pings.  The test only passed because the first ping was not exported in EchoView
- Selecting frequency using `isel` in this case is risky because there is no guarantee the frequency coordinate is ordered with increasing values.
- I used a tighter bound for testing. The max diff is actually different for different frequency channels.

I added a TODO to change everything to use echodata attribute directly instead of saving files and read from them. 
Please do this for this file so that we can merge. 
Please do the same for everything in `test_calibrate.py` and `test_convert_...py` in a separate PR.

leewujung (approved):

PR #318, opened 2021-05-02
Add AD2CP parsing, remove redundant MODELS from convert

imranmaj:

PR #326, opened 2021-05-03
Add AD2CP parsing, remove redundant MODELS from convert (with commit history)

imranmaj:
- Adds AD2CP parsing
- Removes a redundant `MODELS` dictionary from `convert.py`
- Preserves the commit history, unlike #318


leewujung:
Just adding an update to keep track of where we are: need tests before merging.

echopype/tests/test_convert_ad2cp.py
```diff
@@ -24,20 +24,28 @@ def test_convert():
...
         echodata.to_netcdf(save_path=output_dir)
+        raw_echodata[filepath] = echodata
 
-
+# TODO: xarray loads np.datetime64 from disk with floating point errors in v0.16.2,
+# so when we get errors when we try to select using timestamps. Instead, we'll just keep
+# the data in memory so we know the representation will be correct. 
+# When xarray version gets bumped, try deleting raw_echodata 
+# and using the first few commented lines in this function
+# (it is known to behave correctly in v0.17.0).
```
leewujung, line 41: Thanks @imranmaj this is a good find! @lsetiawan do you know why we pin xarray at 0.16.2?

leewujung:
Thanks @imranmaj for the PR!  I added the support for converting `.ad2cp` to new features in the release notes and included it in the docs as alpha. 😀

There is an unresolved issue in the test related to xarray version and saving/reading datetime (even with #346), but let's merge this now and push that to v0.5.1 so that @lsetiawan can do the code formatting all at once.

leewujung:
Alright so the selective CI runs again missed the problem with EchoData, because of the additional  `Beam_Complex` group. We'll need to improve the CI so that there is an option to trigger all tests before merging to master. ''

@imranmaj I am going to revert the merge because it is not immediately clear to me why test_echodata.test_repr would fail due to the extra `Beam_Complex` group since the test file is not an AD2CP file, and this group/attribute should not be there.

PR #328, opened 2021-05-04
Change AZFP ping_time date origin to 1900

ngkavin:
The environment group used seconds since 1970. This was changed to 1900

The vendor group used both which resulted in an incorrect datetime. This is fixed.

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -34,7 +34,7 @@ class SetGroupsAZFP(SetGroupsBase):
                                  'calendar': 'gregorian',
                                  'long_name': 'Timestamp of each ping',
                                  'standard_name': 'time',
-                                 'units': 'seconds since 1900-01-01'})},
+                                 'units': 'seconds since 1900-01-01 00:00:00Z'})},
                         attrs={'long_name': "Water temperature",
                                'units': "C"})
         return ds
@@ -142,7 +142,7 @@ class SetGroupsAZFP(SetGroupsBase):
                                                'calendar': 'gregorian',
                                                'long_name': 'Timestamp of each ping',
                                                'standard_name': 'time',
-                                               'units': 'seconds since 1900-01-01'}),
+                                               'units': 'seconds since 1900-01-01 00:00:00Z'}),
                                 'range_bin': (['range_bin'], range_bin)},
                         attrs={'beam_mode': '',
                                'conversion_equation_t': 'type_4',
@@ -209,7 +209,7 @@ class SetGroupsAZFP(SetGroupsBase):
                                'calendar': 'gregorian',
                                'long_name': 'Timestamp of each ping',
                                'standard_name': 'time',
-                               'units': 'seconds since 1900-01-01'}),
+                               'units': 'seconds since 1900-01-01 00:00:00Z'}),
                 'ancillary_len': (['ancillary_len'], list(range(len(unpacked_data['ancillary'][0])))),
                 'ad_len': (['ad_len'], list(range(len(unpacked_data['ad'][0]))))},
             attrs={
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -27,14 +27,14 @@ class SetGroupsAZFP(SetGroupsBase):
         """Set the Environment group.
         """
         # TODO Look at why this cannot be encoded without the modifications -- @ngkavin: what modification?
-        ping_time = (self.parser_obj.ping_time - np.datetime64('1970-01-01T00:00:00')) / np.timedelta64(1, 's')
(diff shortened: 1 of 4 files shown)


echopype/convert/set_groups_azfp.py
```diff
@@ -27,14 +27,14 @@ def set_env(self) -> xr.Dataset:
         """Set the Environment group.
         """
         # TODO Look at why this cannot be encoded without the modifications -- @ngkavin: what modification?
```
leewujung, line 29: What does this line of comment mean?

ngkavin: I think it means why the time can't be saved without subtracting `np.datetime64('1900-01-01T00:00:00')` and dividing by `np.timedelta64(1, 's')`.

leewujung: Have you tried? I had tried this early on, but you should confirm if that is the current behavior for xarray, in both writing to file and read from file.

ngkavin: `(self.parser_obj.ping_time - np.datetime64('1970-01-01T00:00:00')) / np.timedelta64(1, 's')` converts ping_time to a float which removes the `datetime64` dtype from the ping_time. When saving a `datetime64`, ` xarray has its own attributes for 'calendar' and 'units' which we override in setGoups, but is not allowed.

Specifically, the error is:
 ` ValueError
failed to prevent overwriting existing key units in attrs on variable 'ping_time'. This is probably an encoding field used by xarray to describe how a variable is serialized. To proceed, remove this key from the variable's attributes manually.`

So the problem is only with writing and this is still what xarray 0.16.2 does as well as the latest 0.17.0.

echopype/convert/set_groups_azfp.py
```diff
@@ -27,14 +27,14 @@ def set_env(self) -> xr.Dataset:
...
-        ping_time = (self.parser_obj.ping_time - np.datetime64('1970-01-01T00:00:00')) / np.timedelta64(1, 's')
+        ping_time = (self.parser_obj.ping_time - np.datetime64('1900-01-01T00:00:00')) / np.timedelta64(1, 's')
         ds = xr.Dataset({'temperature': (['ping_time'], self.parser_obj.unpacked_data['temperature'])},
                         coords={'ping_time': (['ping_time'], ping_time,
                                 {'axis': 'T',
                                  'calendar': 'gregorian',
                                  'long_name': 'Timestamp of each ping',
                                  'standard_name': 'time',
-                                 'units': 'seconds since 1970-01-01'})},
+                                 'units': 'seconds since 1900-01-01'})},
```
emiliom, line 37: I suggest changing the datetime string to a fully explicit one: 1900-01-01T00:00:00Z

echopype/convert/set_groups_azfp.py
```diff
@@ -209,7 +209,7 @@ def set_vendor(self) -> xr.Dataset:
                                'calendar': 'gregorian',
                                'long_name': 'Timestamp of each ping',
                                'standard_name': 'time',
-                               'units': 'seconds since 1970-01-01'}),
+                               'units': 'seconds since 1900-01-01'}),
```
emiliom, line 212: I suggest changing the datetime string to a fully explicit one: 1900-01-01T00:00:00Z

echopype/convert/set_groups_azfp.py
lsetiawan: Shouldn't it be `seconds since 1900-01-01 00:00:00` rather than isoformat?

emiliom: I'm proud of you @lsetiawan , for going to the source :smile:! The iso8601 spec is a bit squishy in that allows for some variations; for example, you can use a blank space instead of the T separator, and either a UTC offset or a time zone code (eg, Z). So, the CF spec example on that web page is fully iso8601 compliant! But it never says so explicitly, and it seems to suggest that using a Z would not be correct. However, there's this statement: "The reference time string (appearing after the identifier since) may include date alone; date and time; or date, time, and time zone." So, a Z timezone code seems acceptable.

In addition, the SONAR-netCDF4 spec provides this time origin string "1601-01-01 00:00:00Z". 

SO: My suggestion, for now: "seconds since 1900-01-01 00:00:00Z"

ngkavin:
I'm still trying to figure out why the last commit causes tests to fail.

lsetiawan (approved):
This is now passing and the changes looks good to me. :smile: Thanks @ngkavin!

leewujung:
@emiliom @ngkavin : should there be a "T" or no in the middle?

ngkavin:
> SO: My suggestion, for now: "seconds since 1900-01-01 00:00:00Z"

I went with the latest suggestion, and the example in the SONAR-netCDF4 spec is "nanoseconds since 1601-01-01 00:00:00Z", without the T

Issue #308, opened 2021-04-29
EK80 pulse compressed Sv output values


leewujung:
There are currently unresolved issues related to matching echopype pulse compressed Sv with those generated by Matlab code and EchoView. The Matlab and Echoview pulse compression outputs are also different. This was originally in #278 but needs further investigation so better to be an independent issue. 

The mismatch with Matlab pulse compressed Sv are likely due to a hard-coded range offset (which selects for the starting sample index to be one closest to 0.02 m) in the Matlab code which we should investigate for its reason. The mismatch is smaller at longer range, which is consistent with this potential reason. 

We should connect with EchoView to see how they do pulse compression under the hood to make sure that we know and can reproduce their results.

leewujung:
Note from WGFAST 2022: Lars Anderson from Simrad now has python code implemented for these, so we can cross check and resolve this!

PR #330, opened 2021-05-04
Update processing documentation with code from redesign

ngkavin:
New code blocks that work with the current redesigned package replaced the old code.

### DIFF
--- a/docs/source/process.rst
+++ b/docs/source/process.rst
@@ -96,7 +96,7 @@ as a dictionary while calling ``ep.calibrate.compute_Sv()`` like so:
        'salinity': 30,      # salinity in PSU
        'pressure': 50,     # pressure in dbar
    }
-   ep.calibrate.compute_Sv(env_params=environment)
+   Sv = ep.calibrate.compute_Sv(echodata, env_params=environment)
 
 These value will be used in calculating sound speed,
 seawater absorption, thickness of each sonar sample, and range.
@@ -141,7 +141,7 @@ specify ``cal_params`` while calling the calibration functions like so:
    calibration = {
        'equivalent_beam_angle': [-17.47, -20.77, -21.13, -20.4 , -30]
    }
-   ep.calibrate.compute_Sv(cal_params=calibration)
+   Sv = ep.calibrate.compute_Sv(echodata, cal_params=calibration)
 
 To reset the equivalent beam angle for 18 kHz only, one can do:
 
--- a/docs/source/process.rst
+++ b/docs/source/process.rst
@@ -14,17 +14,25 @@ Functionality
 
   - Calibration and echo-integration to obtain
     volume backscattering strength (Sv) from power data.
+    (beta: we found inconsistencies between pulse compression outputs
+    from the current echopype implementation, EchoView, and Matlab EchoLab code, see
+    `#308 <http://https://github.com/OSOceanAcoustics/echopype/issues/308/>`_.
   - Simple noise removal by removing data points (set to ``NaN``) below
     an adaptively estimated noise floor [1]_.
   - Binning and averaging to obtain mean volume backscattering strength (MVBS)
     from the calibrated data.
+  - Compute mean volume backscattering strength (MVBS) based
+    on either the number of pings and sample intervals
+    (the ``range_bin`` dimension in the data set) or a
+    specified ping time interval and range interval in
+    physics units (seconds and meters).
 
 - EK80 and EA640 broadband echosounders:
 
   - Calibration based on pulse compression output in the
     form of average over frequency.
 
-  - The same noise removal and MVBS functionality available
+  - The same noise removal and MVBS computation functionality available
     to the narrowband echosounders.
 
 
@@ -45,13 +53,15 @@ The steps for performing these analyses are summarized below.
 
 The functions in the ``calibrate`` subpackage take in an ``EchoData`` object,
 which is essentially a container for multiple xarray ``Datasets``,
-and return a single xarray ``Dataset`` (either Sv or Sp).
-The Sv ``Dataset`` can then be passed into the functions provided by the
-``preprocess`` subpackage which all return a single processed ``Dataset``.
+and return a single xarray ``Dataset`` containing the calibrated backscatter
+quantities and the samples' corresponding range in meters.
+The inputs and outputs of all functions in the ``preprocess``
+subpackage are xarray ``Datasets``, with the inputs being ``Datasets``
+containing ``Sv`` and ``range`` generated from calibration.
 
 These ``calibrate`` and ``preprocess`` methods do not save the calculation results to disk,
-but the returned xarray Dataset have methods like ``to_netcdf`` and ``to_zarr`` that
-can be used to save.
+but the returned xarray ``Dataset`` can be saved using native xarray methods
+such as ``to_netcdf`` and ``to_zarr``
 
 For example, to save the Sv and MVBS results to disk:
 
@@ -88,19 +98,19 @@ By default, echopype uses the following for calibration:
 - AZFP: Salinity and pressure provided by the user,
   and temperature recorded at the instrument
 
-These parameters should be overwritten when they differ from the actual
+These parameters can be overwritten when they differ from the actual
 environmental condition during data collection.
 To update these parameters, simply pass in the environmental parameters
-as a dictionary while calling ``ep.calibrate.compute_Sv()`` like so:
+as a dictionary while calling ``ep.calibrate.compute_Sv()``:
 
 .. code-block:: python
 
-   environment = {
+   env_params = {
        'temperature': 8,   # temperature in degree Celsius
        'salinity': 30,     # salinity in PSU
        'pressure': 50,     # pressure in dbar
    }
-   Sv = ep.calibrate.compute_Sv(echodata, env_params=environment)
+   Sv = ep.calibrate.compute_Sv(echodata, env_params=env_params)
 
 These value will be used in calculating sound speed,
 sound absorption, and the thickness of each sonar sample,
@@ -151,20 +161,21 @@ specify ``cal_params`` while calling the calibration functions like so:
 .. code-block:: python
 
    # set all channels at once
-   calibration = {
-       'equivalent_beam_angle': [-17.47, -20.77, -21.13, -20.4 , -30]
+   equivalent_beam_angle = xr.DataArray(
+       [-17.47, -20.77, -21.13, -20.4, -30],
+       dims=['frequency'],
+       coords=[echodata.beam.frequency]
+   )
+   cal_params = {
+       'equivalent_beam_angle': equivalent_beam_angle
    }
-   Sv = ep.calibrate.compute_Sv(echodata, cal_params=calibration)
+   Sv = ep.calibrate.compute_Sv(echodata, cal_params=cal_params)
 
 To reset the equivalent beam angle for 18 kHz only, one can do:
 
 .. code-block:: python
 
    echodata.beam.equivalent_beam_angle.loc[dict(frequency=18000)] = 18.02  # set value for 18 kHz only
-
-Make sure you use ``echodata.beam.equivalent_beam_angle.frequency`` to check
-the sequence of the frequency channels first, and always double
-check after setting these parameters!
 
 
 References
@@ -183,7 +194,7 @@ References
 
 .. [4] Demer DA, Berger L, Bernasconi M, Bethke E, Boswell K, Chu D, Domokos R,
    et al. (2015) Calibration of acoustic instruments. `ICES Cooperative Research Report No.
-   1.   133 pp. <https://doi.org/10.17895/ices.pub.5494>`_
+   326.       133 pp. <https://doi.org/10.17895/ices.pub.5494>`_
 
 
 .. TODO: Need to specify the changes we made from AZFP Matlab code to here:
--- a/docs/source/process.rst
+++ b/docs/source/process.rst
@@ -55,8 +55,8 @@ The functions in the ``calibrate`` subpackage take in an ``EchoData`` object,
 which is essentially a container for multiple xarray ``Datasets``,
 and return a single xarray ``Dataset`` containing the calibrated backscatter
 quantities and the samples' corresponding range in meters.
-The inputs and outputs of all functions in the ``preprocess``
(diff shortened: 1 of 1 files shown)


docs/source/process.rst
```diff
@@ -22,75 +21,40 @@ Functionality
...
-   ed.calibrate(save=True, save_postfix='_Cal')
-
-   # Save output to another directory: ./cal_results/convertedfile_Sv.nc
-   ed.calibrate(save=True, save_path='./cal_results')
-
-   # Save output to another directory with an arbitrary name
-   ed.calibrate(save=True, save_path='./cal_results/somethingnew.nc')
+   import echopype as ep
+   nc_path = './converted_files/file.nc'           # path to a converted nc file
+   echodata = ep.open_converted(nc_path)           # create a processing object
```
emiliom, line 37: > create a processing object

That's no longer a `processing` object, per se. Change to "create an EchoData object"

docs/source/process.rst
```diff
@@ -22,75 +21,40 @@ Functionality
...
+   MVBS = ep.preprocess.get_MVBS(Sv_clean)         # obtain MVBS from denoised Sv
 
-By default, for noise removal and MVBS calculation, echopype tries to load Sv
-already stored in memory (``ed.Sv``), or tries to calibrate the raw data to
-obtain Sv. If ``ed.Sv`` is empty (i.e., whe calibration operation has not been
-performed on the object), echopype will try to load Sv from ``*_Sv.nc`` from
-the directory containing the converted ``.nc`` file or from the user-specified
-path. For example:
+The functions in the ``calibrate`` subpackage take in an ``EchoData`` object,
+which is essentially a container for multiple xarray ``Dataset`` groups,
```
emiliom, line 43: Change to "which is essentially a container for multiple xarray ``Datasets``,"

docs/source/process.rst
```diff
@@ -22,75 +21,40 @@ Functionality
...
-   # Save output to another directory: ./cal_results/convertedfile_Sv.nc
-   ed.calibrate(save=True, save_path='./cal_results')
-
-   # Save output to another directory with an arbitrary name
-   ed.calibrate(save=True, save_path='./cal_results/somethingnew.nc')
+   import echopype as ep
+   nc_path = './converted_files/file.nc'           # path to a converted nc file
+   echodata = ep.open_converted(nc_path)           # create a processing object
+   Sv = ep.calibrate.compute_Sv(echodata)          # obtain Sv
+   Sv_clean = ep.preprocess.remove_noise(Sv, ...)  # obtain a denoised Sv Dataset
```
emiliom, line 39: What are those 3 dots?

docs/source/process.rst
```diff
@@ -115,33 +79,29 @@ By default, echopype uses the following for calibration:
...
-
-This will trigger recalculation of all related parameters,
-including sound speed, seawater absorption, thickness of each sonar
-sample, and range. The updated values can be retrieved with:
+   environment = {
+       'temperature': 8,   # temperature in degree Celsius
+       'salinity': 30,      # salinity in PSU
+       'pressure': 50,     # pressure in dbar
+   }
+   ep.calibrate.compute_Sv(env_params=environment)
```
emiliom, line 99: This looks incomplete. Compare it to the example above: `Sv = ep.calibrate.compute_Sv(echodata)`

docs/source/process.rst
```diff
@@ -115,33 +79,29 @@ By default, echopype uses the following for calibration:
...
 
-.. code-block:: python
-
-   ed.seawater_absorption  # absorption in [dB/m]
-   ed.sound_speed          # sound speed in [m/s]
-   ed.sample_thickness     # sample spatial resolution in [m]
-   ed.range                # range for each sonar sample in [m]
+These value will be used in calculating sound speed,
+seawater absorption, thickness of each sonar sample, and range.
```
emiliom, line 102: I think it'd be helpful to show here how those calculated variables can be accessed. Include a brief code block analogous to what was there before.

docs/source/process.rst
```diff
@@ -165,24 +125,31 @@ from the raw data file.
...
+specify ``cal_params`` while calling the calibration functions like so:
 
 .. code-block:: python
 
-   ed.equivalent_beam_angle.loc[dict(frequency=18000)] = -18.02  # set value for 18 kHz only
+   # set all channels at once
+   calibration = {
+       'equivalent_beam_angle': [-17.47, -20.77, -21.13, -20.4 , -30]
+   }
+   ep.calibrate.compute_Sv(cal_params=calibration)
```
emiliom, line 144: This looks incomplete. Compare it to the example above: `Sv = ep.calibrate.compute_Sv(echodata)`

emiliom (commented):
Thanks @ngkavin! I added a bunch of small comments.

docs/source/process.rst
ngkavin: This function requires additional parameters to be passed in. Should I explicitly write down the required arguments?

docs/source/process.rst
ngkavin: Since these are just function that output an xarray `Dataset`, accessing these values would require creating objects that are not exposed to the user via the `__init__.py` of `calibrate`. So, I don't think you can access these values with the current API.

docs/source/process.rst
emiliom: Ah.If they are only a couple or so, my suggestion is to include them with reasonable values. Otherwise, I guess leaving the ellipsis is ok.

docs/source/process.rst
ngkavin: After further investigation, these are provided in the output dataset. I'll update the docs.

emiliom:
Thanks for the changes. Looks good to me now. @leewujung please merge after reviewing -- or just go ahead and merge as is if you prefer.

docs/source/process.rst
```diff
@@ -22,75 +21,44 @@ Functionality
 
 - EK80 and EA640 broadband echosounders:
 
-  - calibration based on pulse compression output in the
+  - Calibration based on pulse compression output in the
```
leewujung, line 24: Please add `(beta: we found inconsistency between pulse compression outputs from the current echopype implement, EchoView, and Matlab EchoLab code, see #308 )` at the end. Make sure the issue is linked correctly.

docs/source/process.rst
```diff
@@ -2,18 +2,17 @@ Data processing
...
 Functionality
 -------------
 
 - EK60 and AZFP narrowband echosounders:
 
-  - calibration and echo-integration to obtain
+  - Calibration and echo-integration to obtain
     volume backscattering strength (Sv) from power data.
   - Simple noise removal by removing data points (set to ``NaN``) below
     an adaptively estimated noise floor [1]_.
```
leewujung, line 18: Add a bullet point:
```
- Compute mean volume backscattering strength (MVBS) based on either the number of pings and sample intervals (the ``range_bin`` dimension in the data set) or a specified ping time interval and range interval in physics units (seconds and meters).
```

docs/source/process.rst
```diff
@@ -22,75 +21,44 @@ Functionality
 
 - EK80 and EA640 broadband echosounders:
 
-  - calibration based on pulse compression output in the
+  - Calibration based on pulse compression output in the
     form of average over frequency.
 
+  - The same noise removal and MVBS functionality available
```
leewujung, line 27: change `MVBS functionality` to `MVBS computation functionality`

docs/source/process.rst
```diff
@@ -22,75 +21,44 @@ Functionality
...
+   Sv_clean = ep.preprocess.remove_noise(          # obtain a denoised Sv Dataset
+      Sv,
+      ping_num=5,
+      range_bin_num=30
+   )
+   MVBS = ep.preprocess.get_MVBS(Sv_clean)         # obtain MVBS from denoised Sv
+
+The functions in the ``calibrate`` subpackage take in an ``EchoData`` object,
+which is essentially a container for multiple xarray ``Datasets``,
+and return a single xarray ``Dataset`` (either Sv or Sp).
```
leewujung, line 48: "(either Sv or Sp)" is not correct. The output dataset contains both Sv and range, or Sp and range.

Change this to:
```
and return a single xarray ``Dataset`` containing the calibrated backscatter quantities and the samples' corresponding range in meters.
```

docs/source/process.rst
```diff
@@ -22,75 +21,44 @@ Functionality
...
+The Sv ``Dataset`` can then be passed into the functions provided by the
+``preprocess`` subpackage which all return a single processed ``Dataset``.
```
leewujung, line 50: Change this to a more top-down style:
```
The inputs and outputs of all functions in the ``preprocess`` subpackage are xarray ``Datasets``, 
with the inputs being ``Datasets`` containing ``Sv`` and ``range`` generated from calibration.
```

docs/source/process.rst
```diff
@@ -22,75 +21,44 @@ Functionality
...
+
+These ``calibrate`` and ``preprocess`` methods do not save the calculation results to disk,
+but the returned xarray Dataset have methods like ``to_netcdf`` and ``to_zarr`` that
+can be used to save.
```
leewujung, line 54: Small change:
```
but the returned xarray ``Dataset`` can be save using native xarray methods such as ``to_netcdf`` and ``to_zarr`` ```

File: docs/source/process.rst [hunk 13]
```diff
@@ -115,33 +83,38 @@ By default, echopype uses the following for calibration:
...
+
+- AZFP: Salinity and pressure provided by the user,
   and temperature recorded at the instrument
 
 These parameters should be overwritten when they differ from the actual
 environmental condition during data collection.
-To update these parameters, simply do the following *before*
-calling ``ed.calibrate()``:
+To update these parameters, simply pass in the environmental parameters
+as a dictionary while calling ``ep.calibrate.compute_Sv()`` like so:
```
leewujung, line 94: These parameters should be overwritten --> These parameters **can be** overwritten

docs/source/process.rst
```diff
@@ -115,33 +83,38 @@ By default, echopype uses the following for calibration:
...
-   ed.temperature = 8   # temperature in degree Celsius
-   ed.salinity = 30     # salinity in PSU
-   ed.pressure = 50     # pressure in dbar
-   ed.recalculate_environment()  # recalculate related parameters
+   environment = {
+       'temperature': 8,   # temperature in degree Celsius
+       'salinity': 30,     # salinity in PSU
+       'pressure': 50,     # pressure in dbar
+   }
+   Sv = ep.calibrate.compute_Sv(echodata, env_params=environment)
```
leewujung, line 103: change the dictionary name `environment` to `env_params` to avoid confusion

docs/source/process.rst
```diff
@@ -165,24 +138,31 @@ from the raw data file.
...
+specify ``cal_params`` while calling the calibration functions like so:
 
 .. code-block:: python
 
-   ed.equivalent_beam_angle.loc[dict(frequency=18000)] = -18.02  # set value for 18 kHz only
+   # set all channels at once
+   calibration = {
+       'equivalent_beam_angle': [-17.47, -20.77, -21.13, -20.4 , -30]
+   }
+   Sv = ep.calibrate.compute_Sv(echodata, cal_params=calibration)
```
leewujung, line 157: Change `calibration` to `cal_params`

Also I think we should demonstrate passing an xr.DataArray with a frequency dimension instead of a list without any frequency labels. 
@ngkavin can you test constructing an xr.DataArray using coordinates from `echodata.beam.frequency` and test passing it in? It should just work, but if it does not let me know.

docs/source/process.rst
```diff
@@ -165,24 +138,31 @@ from the raw data file.
...
 
 .. code-block:: python
 
-   ed.equivalent_beam_angle = [-17.47, -20.77, -21.13, -20.4 , -30]  # set all channels at once
+   echodata.beam.equivalent_beam_angle.loc[dict(frequency=18000)] = 18.02  # set value for 18 kHz only
 
-Make sure you use ``ed.equivalent_beam_angle.frequency`` to check
+Make sure you use ``echodata.beam.equivalent_beam_angle.frequency`` to check
 the sequence of the frequency channels first, and always double
 check after setting these parameters!
```
leewujung, line 167: The construction of the xr.DataArray from the above comment should allow removing this sentence all together. We actually need a check in the calibration function to take care of this, but let's delay that to later release.

docs/source/process.rst
```diff
@@ -203,7 +183,7 @@ References
 
 .. [4] Demer DA, Berger L, Bernasconi M, Bethke E, Boswell K, Chu D, Domokos R,
    et al. (2015) Calibration of acoustic instruments. `ICES Cooperative Research Report No.
-   326. 133 pp. <https://doi.org/10.17895/ices.pub.5494>`_
+   1.   133 pp. <https://doi.org/10.17895/ices.pub.5494>`_
```
leewujung, line 186: Why is this changed to No. 1 ?  It is No. 326!

leewujung (changes requested):
I requested some text changes and a code check. 

Please be careful with your editor if it changes numbers automatically like in the last comment.

docs/source/process.rst
ngkavin: Thanks for pointing this out. I was aware that my editor was doing this to this specific number for some reason (apparently every time I delete a line of text), but I missed it in the last commit.

docs/source/process.rst
```diff
@@ -45,13 +53,15 @@ The steps for performing these analyses are summarized below.
...
+quantities and the samples' corresponding range in meters.
+The inputs and outputs of all functions in the ``preprocess``
+subpackage are xarray ``Datasets``, with the inputs being ``Datasets``
+containing ``Sv`` and ``range`` generated from calibration.
 
 These ``calibrate`` and ``preprocess`` methods do not save the calculation results to disk,
-but the returned xarray Dataset have methods like ``to_netcdf`` and ``to_zarr`` that
-can be used to save.
+but the returned xarray ``Dataset`` can be saved using native xarray methods
+such as ``to_netcdf`` and ``to_zarr``
```
leewujung, line 65: missing a period here.

docs/source/process.rst
leewujung: Ugh, sorry, can you change everything to singular form?

leewujung:
@ngkavin thanks for the quick turnaround! I'll merge this now.

leewujung (approved):

PR #332, opened 2021-05-07
Revert back to not do multi CPU on running tests

lsetiawan:
This PR reverts back the pytest to not use multi CPU for now. Seems like the tests are not setup in a way to be able to handle these, causing some tests to failed even if they are passing when only using single CPU. Maybe in v0.5.1 we can improve to be able to do this! I think this is what's causing #328 to fail. Sorry @ngkavin! Once this is merged, you can close and reopen your PR, and things should pass. Trying to be fancy and speed up the tests, but I guess it wasn't that simple :stuck_out_tongue_closed_eyes:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -94,7 +94,7 @@ jobs:
       - name: Running Tests
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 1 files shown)


emiliom:
Go for it. I won't be at the following meeting either. I have a good feel for most of the CI already, and I can catch up later.

leewujung:
Oh right, i forgot that that's the time frame.

PR #333, opened 2021-05-07
Fix datetime inconsistency and add test

### LINKED ISSUE #329: Time encoding wrt convention and file writing/reading behavior
Currently when using `open_raw` (to parse data from raw instrument files), the time coordinates (e.g., `ping_time`) in the returned `EchoData` object is encoded in `float64`, whereas when using `open_converted` (to read from netcdf or zarr files) the time coordinate in the returned `EchoData` object is encoded in `datetime64`. 

This inconsistency seems can be handled by casting the type to `datetime64` when setting the time coordinates in the set group methods, but we need to experiment with both the writing and reading end of this to make sure things are done correctly. We would need a few tests for the various time coordinates also!

lsetiawan:
This PR fixes the inconsistencies between `open_raw` and `open_converted` on `*_time` variables. It does the following:
1. Set the encodings rather than attributes for `units` and `calendar`. Additionally the encoding also include the specific `dtype` and `fillvalue`, in this case `int64` and `-9999`.
2. Since the unit is seconds since, and xarray rounds the values to the nearest second, we also round during set groups.

I have also added test to ensure that the proper encoding/attribute are store correctly, and data arrays are equal.

Closes #329

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -3,7 +3,7 @@ Class to save unpacked echosounder data to appropriate groups in netcdf or zarr.
 """
 import xarray as xr
 import numpy as np
-from .set_groups_base import SetGroupsBase
+from .set_groups_base import SetGroupsBase, set_encodings
 
 
 class SetGroupsAZFP(SetGroupsBase):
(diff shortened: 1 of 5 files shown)


leewujung:
@lsetiawan : thanks for the PR!

> 2. Since the unit is seconds since, and xarray rounds the values to the nearest second, we also round during set groups.

Isn't the units `nanoseconds since ... ` in the convention?

lsetiawan:
> Isn't the units nanoseconds since ... in the convention?
This is where @emiliom should chime in :smile: I'm merely following what was discussed here: https://github.com/OSOceanAcoustics/echopype/pull/328 and also whatever the code said before. :sweat_smile:

emiliom:
> Isn't the units nanoseconds since ... in the convention?

It is, but the problem @lsetiawan is addressing here is not about about nanoseconds vs seconds, but rather about ensuring that for time coordinates `open_raw` returns datetime64 and values are serialized as integers. If those issues are solved, making a decision to fully adhere to the SONAR-netCDF4 convention will be much easier, later. At a recent meeting (or github/slack exchanges?) we agreed the convention was, umm, excessive in defining nanoseconds since 1601, so let's leave that issue for some other time.

emiliom:
You're awesome @lsetiawan ! :heart: 

I looked through your changes, and they look like they'd do what you say. Though, I'll add that your description of the PR wasn't 100% clear to me wrt to the critical issue of whether `open_raw` will now return datetime64 for `*_time` coordinate variables. Looking at your code, it looks like it does.  But more to the point, I tested your PR (with an EK60 raw file), and sure enough, it does! Woo-hoo!

I haven't looked at your new test, and won't have time to do so. I looked through your changes to the other modules, and they looked good. I vote for merging.

PR #342, opened 2021-05-10
Change echodata.to_netcdf/to_zarr to return nothing

leewujung:
Currently `echodata.to_netcdf/to_zarr` would return an `EchoData` object. This PR changes that to return nothing, which is consistent with the xarray to_netcdf/to_zarr behavior and likely more the anticipated from users.


lsetiawan (approved):
Awesome! :rocket: LGTM :smile_cat:

PR #343, opened 2021-05-10
Update package to ignore unecessary files

lsetiawan:
This PR updates the `MANIFEST.in` and clean up `setup.py` to ignore some unfinished scripts and include some data such as static css and html files.


leewujung:
> ... include some data such as static css and html files.

I saw your took out these, was just going to comment that they don't seem to be needed. So I guess we're good now?

PR #344, opened 2021-05-11
First input argument in `echodata.to_netcdf/zarr`

leewujung:
This PR defaults the first input argument in `echodata.to_netcdf|zarr` the path to save the file to (`save_path`). Before users have to do `echodata.to_netcdf|zarr(save_path="some_path")`, which is a little redundant, and the operation actually would error out if `save_path` was not specified. The xarray/pandas design to have the first argument being the `save_path` is much more convenient and intuitive.

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -141,7 +141,7 @@ class EchoData:
 
         return xr.open_dataset(filepath, group=group, engine=XARRAY_ENGINE_MAP[suffix])
 
-    def to_netcdf(self, save_path, **kwargs):
+    def to_netcdf(self, save_path=None, **kwargs):
         """Save content of EchoData to netCDF.
 
         Parameters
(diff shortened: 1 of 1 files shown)


echopype/echodata/echodata.py
```diff
@@ -141,7 +141,7 @@ def _load_group(filepath, group=None):
 
         return xr.open_dataset(filepath, group=group, engine=XARRAY_ENGINE_MAP[suffix])
 
-    def to_netcdf(self, **kwargs):
+    def to_netcdf(self, save_path, **kwargs):
```
lsetiawan, line 144: This seem to require user now to need `save_path`. Is this intended. We have a warning right now if there is no `save_path`.

lsetiawan: > need is to initialize save_path=None here?

Yes.

PR #345, opened 2021-05-11
Added code to check if the file exists

### LINKED ISSUE #285: Provide clearer error messages for `open_converted`
`open_converted` does not currently check for file existence and gives the following error when the input file does not exist: `UnboundLocalError:local variable 'ds' referenced before assignment`.

This may be indicative of a need to predict a few of the more common errors that users might face and give messages that can point them in the right direction.

Badboy-16:
I had a try to fix #285 

I added some code in the `open_converted` function to check if `converted_raw_path` is a file, and if not, raise a `FileNoteFoundError` with the message that the specified file doesn't exist. 

I also intend to write a test for this additional piece of code but am not sure which test file I should put the test in the `tests` folder. Could anyone please advise? Thanks.

### DIFF
--- a/echopype/echodata/api.py
+++ b/echopype/echodata/api.py
@@ -2,12 +2,24 @@ from os import path
 
 from .echodata import EchoData
 
+import fsspec
+
 
 def open_converted(converted_raw_path, storage_options=None):
     """Create an EchoData object from a single converted zarr/nc files."""
     # TODO: combine multiple files when opening
-    if not path.isfile(converted_raw_path):
-        raise FileNotFoundError(f"File {converted_raw_path} does not exist")
+
+    def _check_if_file_exists(file_path) -> bool:
+        """Helper function to check if file in converted_raw_path exists."""
+        file_exists_locally = path.isfile(file_path)
+        fsmap = fsspec.get_mapper(file_path)
+        file_exists_in_url = fsmap.fs.exists(fsmap.root)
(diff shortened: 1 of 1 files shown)


ngkavin:
Thanks for looking into this! However, I don't think this works if the path is a url like `s3://data-bucket/file.nc` since os.path checks only for the local files.

leewujung:
@Badboy-16 thanks for the PR! 
I think it'll be good to add check for file existence on remote sources also (like on s3 bucket). 
[`_check_file`](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/api.py#L353) under `convert.api` would be a good reference.

@lsetiawan : what do you think in terms of consolidating the need of checking local and remote path as a utility/helper function in the package?

> I also intend to write a test for this additional piece of code but am not sure which test file I should put the test in the tests folder. Could anyone please advise? Thanks.

I think it'll be good have a `test_open_converted` to test for this. Tagging @lsetiawan on this also since we were just talking about test coverage.

Badboy-16:
Thanks all for the suggestions :)

I took the suggestion from @leewujung and borrowed some code from [`_check_file`](https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/convert/api.py#L353) to check if the file also exists in a URL and wrapped it in a helper function. Please see if it looks ok. If yes, I may help to write the test if needed.

lsetiawan:
This is resolved in #349

PR #349, opened 2021-05-14
Update echodata to be able to read cloud data for zarr

### LINKED ISSUE #285: Provide clearer error messages for `open_converted`

lsetiawan:
This PR enables echodata object to read from cloud (s3 bucket/http) for zarr file only. I have introduced 2 utility methods in utils/io.py which are used for sanitizing file path before passing into `xr.open_dataset` and another for file checking. `open_converted` is now able to accept `str`, `pathlib.Path`, and `fsspec.FSMap` paths.

Closes #285

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -3,11 +3,13 @@ from collections import OrderedDict
 from html import escape
 from pathlib import Path
 import warnings
+import fsspec
 
 import xarray as xr
 from zarr.errors import GroupNotFoundError
 
 from ..utils.repr import HtmlTemplate
+from ..utils.io import sanitize_file_path, check_file_existance
 from .convention import _get_convention
 
 XARRAY_ENGINE_MAP = {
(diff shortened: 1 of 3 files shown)


lsetiawan:
Currently it's failing, since the test data is currently being updated.

lsetiawan:
All tests now passes as expected :smile_cat:

leewujung:
Hey @lsetiawan this looks great! Thanks for adding the tests too!

What do you think about using `check_file_existance` and `sanitize_file_path`  for the corresponding section in `open_raw`? 
Obviously that should be another PR, but I think the more we can consolidate these io/path handling functions that better.

lsetiawan:
>What do you think about using check_file_existance and sanitize_file_path for the corresponding section in open_raw? Obviously that should be another PR, but I think the more we can consolidate these io/path handling functions that better.

I was thinking the same thing, but didn't wanna change too many things before the release. I think maybe that clean up can be done in v0.5.1?

leewujung:
> I think maybe that clean up can be done in v0.5.1?

Sounds good!

PR #353, opened 2021-05-17
Add AD2CP fix from (#326)

lsetiawan:
This PR adds a fix to the failure found on echodata test from PR #326


echopype/echodata/echodata.py
```diff
@@ -124,15 +124,16 @@ def _load_file(self, raw_path):
...
             try:
                 ds = self._load_group(
                     raw_path,
                     group=value["ep_group"],
                 )
             except (OSError, GroupNotFoundError):
-                if group == "beam_power":
-                    ds = None
-                    pass
+                # Skips group not found errors for EK80 and ADCP
```
lsetiawan, line 135: @leewujung This was needed in order for echodata to ignore the missing group since the test only uses ek60 data!

PR #317, opened 2021-04-30
Lint the whole codebase and add coverage test.

lsetiawan:
This PR cleans up code formatting and styling to meet pep8 coding conventions. Additionally, code coverage tests are added. Also, the setup configurations are now mostly in `setup.cfg` rather than `setup.py`.

~~Note: This PR is not ready for review or merging as of 4/30/2021.~~
Update: This PR is now ready for review and the lint tests are now passing.

PR #361, opened 2021-05-20
Fix AD2CP Encoding Issues

### LINKED ISSUE #356: AD2CP ping_time values become inaccurate after saving and loading
Saving and loading ping_time values leads to incorrect values reminiscent of floating point imprecision (e.g., saving a ping_time `2021-03-17T19:47:19.188500000` to disk and loading might produce `2021-03-17T19:47:19.188499999`).

Currently, AD2CP tests avoid this by saving data in memory instead of writing to disk:
[https://github.com/OSOceanAcoustics/echopype/blob/66fca2ba519ec00f80647582eff3684bdccf3c9a/echopype/tests/test_convert_ad2cp.py#L36-L38](https://github.com/OSOceanAcoustics/echopype/blob/66fca2ba519ec00f80647582eff3684bdccf3c9a/echopype/tests/test_convert_ad2cp.py#L36-L38)

imranmaj:
* Fixes data encoding issues outlined in #356
* Adds length check to _encode_dataarrays (to fix data encoding issues)
* Renames `time_*` dimensions to `ping_time_*`

Closes #356

### DIFF
--- a/echopype/convert/api.py
+++ b/echopype/convert/api.py
@@ -282,7 +282,7 @@ def _save_groups_to_file(echodata, output_path, engine, compress=True):
             path=output_path,
             mode='a',
             engine=engine,
-            group='Beam',
+            group='Beam_complex',
             compression_settings=COMPRESSION_SETTINGS[engine]
             if compress
             else None,
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -51,6 +51,8 @@ def _encode_dataarray(da, dtype):
         encoded_data, _, _ = coding.times.encode_cf_datetime(
             da, **read_encoding
         )
+    if da.size == 0:
+        return da
     return coding.times.decode_cf_datetime(
         encoded_data, **read_encoding
     )
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -23,6 +23,36 @@ DEFAULT_ENCODINGS = {
         '_FillValue': np.nan,
         'dtype': np.dtype('float64'),
     },
+    'ping_time_burst': {
+        'units': 'seconds since 1900-01-01T00:00:00+00:00',
+        'calendar': 'gregorian',
+        '_FillValue': np.nan,
+        'dtype': np.dtype('float64'),
+    },
+    'ping_time_average': {
+        'units': 'seconds since 1900-01-01T00:00:00+00:00',
+        'calendar': 'gregorian',
+        '_FillValue': np.nan,
+        'dtype': np.dtype('float64'),
+    },
+    'ping_time_echosounder': {
+        'units': 'seconds since 1900-01-01T00:00:00+00:00',
+        'calendar': 'gregorian',
+        '_FillValue': np.nan,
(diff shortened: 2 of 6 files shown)


echopype/convert/set_groups_base.py
```diff
@@ -51,6 +81,8 @@ def _encode_dataarray(da, dtype):
         encoded_data, _, _ = coding.times.encode_cf_datetime(
             da, **read_encoding
         )
+    if da.size == 0:
```
lsetiawan, line 84: @imranmaj This should probably go up top so the computer doesn't spend time trying to `encode_cf_datetime`. Interesting that there's an empty DataArray.

lsetiawan (changes requested):
Overall looks good. Just need to make that small tweak commented above. Thanks!

echopype/convert/set_groups_base.py
imranmaj: Good idea

lsetiawan:
One more thing @imranmaj. Could you add ADCP test case here for this encoding stuff: https://github.com/OSOceanAcoustics/echopype/blob/007e0d9f468f9e24acdc17ba798eacbb3d0c7432/echopype/tests/test_convert.py#L256-L276

lsetiawan (approved):
Great work @imranmaj! Thanks for implementing all the changes. 

I think this is ready to merge. @leewujung could you look through the changes real quick to make sure all is good and merge? Thanks! 😄

echopype/convert/api.py
```diff
@@ -282,7 +282,7 @@ def _save_groups_to_file(echodata, output_path, engine, compress=True):
             path=output_path,
             mode='a',
             engine=engine,
-            group='Beam',
+            group='Beam_complex',
```
leewujung, line 285: for line 279: why do you have to check `hasattr(echodata, "beam_complex")`? From [here](https://github.com/OSOceanAcoustics/echopype/blob/aa9c32f205ae7aeaded217800119b59ef6a175ec/echopype/echodata/echodata.py#L98-L101) it looks like _all_ groups are set to None when EchoData is initialized, regardless of the type of sonar_model. Did I miss anything?

echopype/tests/test_convert_ad2cp.py
```diff
@@ -112,7 +103,9 @@ def test_raw_output():
                 assert np.isclose(
                     echodata.beam_complex["echosounder_raw_transmit_samples_r"]
                     .sel(
-                        ping_time=echodata.beam_complex["time_echosounder_raw_transmit"]
+                        ping_time=echodata.beam_complex[
+                            "ping_time_echosounder_raw_transmit"
+                        ]
```
leewujung, line 108: For this and the next few assert, what's the actual difference between the ping_time numbers? I am wondering about np.isclose as it has a default tolerance.

imranmaj: The ping_time numbers should be the same I believe, but the sample values themselves are slightly off:
image

Thanks for letting me know about allclose!

leewujung:
Merge this now so that we can do linting. #317

PR #362, opened 2021-05-23
Small updates to data processing page

emiliom:
- added text about environmental parameters in EK60, based on exchanges on Slack. @leewujung please take a look at these edits and suggest changes if not quite accurate
- fixed small mistake

### DIFF
--- a/docs/source/process.rst
+++ b/docs/source/process.rst
@@ -126,15 +126,18 @@ critical in biological interpretation of ocean sonar data. They influence:
 By default, echopype uses the following for calibration:
 
 - EK60 and EK80: Environmental parameters saved with the raw data files. 
-  For EK60, temperature, salinity and pressure are entered by the instrument
-  operator but only calculated sound speed is saved into the raw file.
+  For EK60, instrument operators may enter temperature and salinity values into the 
+  `Simrad EK60 software's Environment dialog 
+  <https://www.simrad.online/ek60/ref_english/default.htm?startat=/ek60/ref_english/xxx_para_environment.html>`_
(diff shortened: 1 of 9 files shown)


leewujung:
@emiliom The changes look good. Could you modify the EK60 description a bit to reflect what's in [the manual](https://www.simrad.online/ek60/ref_english/default.htm)?
Screen Shot 2021-05-24 at 2 20 12 PM

emiliom:
Thanks. Will do. I'm pasting her for reference the link that goes straight to that page in the manual:
https://www.simrad.online/ek60/ref_english/default.htm?startat=/ek60/ref_english/xxx_para_environment.html

I'll include that link in the echopype docs change.

emiliom:
@leewujung I've updated the text as you requested; take a look. **BUT, I'm very confused!** There are now other commits in here that are not from me, involving changes to code rather than just the docs. I don't know how that happened.

Also, I guess I submitted this PR for merging into `master` before you created the new `dev` branch. Should it be changed?

leewujung:
Ok, this is weird. My PR for the time reversal stuff was to `dev`, so I am not sure why you have it here if what you pulled was the `master` branch.  I was thinking this PR is the best to be merge to master since it is a doc update that people could use, but since this thing it messed up now, how about just PR to `dev` and let's get the 0.5.1 out soon?

lsetiawan:
@emiliom do you think you could close and then create another similar PR after you pulled the latest master?

lsetiawan:
It is really weird why github is pulling changes from dev ....

PR #366, opened 2021-05-28
Add ABC decorators to SetGroupsBase

imranmaj:
Adds ABC decorators to SetGroupsBase. This will help to catch errors by requiring that subclasses of SetGroupsBase override the required methods.

### DIFF
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -1,3 +1,5 @@
+import abc
+import pynmea2
 from datetime import datetime as dt
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


echopype/convert/set_groups_base.py
```diff
@@ -162,30 +163,35 @@ def set_provenance(self) -> xr.Dataset:
         ds = ds.assign_attrs(prov_dict)
         return ds
 
+    @abc.abstractmethod
     def set_env(self) -> xr.Dataset:
         """Set the Environment group.
         """
-        pass
+        return xr.Dataset()
```
lsetiawan, line 170: These are a lot of instantiation for an abstract interface. Should probably avoid that for a base class, It would probably be better if we have a `raise NotImplementedError`.
```
@abc.abstractmethod
def set_env(self):
    """Set the Environment group.
    """
    raise NotImplementedError
```

imranmaj: Ah right, thank you

PR #375, opened 2021-06-02
Add docker-compose to dev requirement for contributors.

lsetiawan:
This PR adds `docker-compose` as part of the dev requirement so that users would have it rather than manually installing from docker instructions.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 black
 check-manifest
+docker-compose
 flake8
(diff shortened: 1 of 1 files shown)


emiliom:
Wow, I didn't know docker-compose could be installed that way!

Ultimately, the Contributing docs page should be tweaked to reflect this change. Specifically: https://echopype.readthedocs.io/en/stable/contributing.html#running-the-tests

PR #382, opened 2021-06-14
Fix __all__ by using strings

lsetiawan:
This PR fixes `__all__` errors in docs. This variable should be passed in list of strings, not list of objects!


emiliom:
Thank you @lsetiawan !! Gosh, I should have caught that. Seeing the solution now seems so obvious!

My vote is to merge this PR as is. Any reason not to?

Issue #210, opened 2020-12-16
Document and verify compliance with SONAR-netCDF4 convention


emiliom:
Let's go through the current implementation of SONAR-netCDF4 convention. We can maintain a markdown table to manually document what is currently covered, what sections of the convention it corresponds to, confirmation that it's properly implemented, etc

SONAR-netCDF4 convention:
- https://github.com/ices-publications/SONAR-netCDF4/ (see the "latest working draft" linked on the README file)
- http://www.ices.dk/sites/pub/Publication%20Reports/Cooperative%20Research%20Report%20(CRR)/CRR341.pdf

leewujung:
Adding some specifics to this:

- Many variables are not specified in the convention. I've been putting them in the `Beam` and `Vendor` group somewhat arbitrarily depending on their dimension and how central it feels to be bundled together with backscatter measurements. For example, split-beam alongship and athwartship angles are in `Beam` group because they obviously should go with backscatter measurements, but the filter coefficients needed for processing broadband data are in the `Vendor` group. It'd be nice to review and make decisions on these.

- Echopype conversion currently aims at preserving all info from the raw file. This conflicts with SONAR-netCDF4 in places when the raw recording requires some transformation to have physically meaningful units. For example, EK60 raw backscatter and angle data need to be scaled by specific constants to be the real "power" and "angle." This is likely a larger question related to the convention, but I think we should have a clear strategy on this and put it in the documentation.

- GPS data from EK60/EK80 sometimes include different NMEA sentences. The spatial resolution appears to be different. Currently the default is to only save the `GGA` sentences unless others are specified by the user. @emiliom @lsetiawan since you've dealt more with spatial data perhaps you have some insights on this.

emiliom:
Notes from call between @emiliom and @leewujung on 2021-1-20:

We've laid out the sequence of tasks as follows:

1. 1:1 mappings of variables, attributes and groups. Includes things only found in one of them (?).
2. Content of variables, attributes and groups
3. Things that are intentionally different, or diverged unintentionally

emiliom:
Define a reasonable end point / goals where we want to by the time we submit the echopype paper

leewujung:
For AZFP:
- we currently have `tilt_x` and `tilt_y` in the Beam group that are _derived_ from the raw data, and also have the raw sensor record in both the Beam and the Vendor group
[...]

PR #383, opened 2021-06-15
Add method to combine EchoData objects

### LINKED ISSUE #364: Combine multiple EchoData objects
Right now we have `open_raw` that handles parsing raw data from a single file into a nice `EchoData` object. Let's revamp the previous `combine=True` option so that we can combine multiple `EchoData` objects explicitly. This function can then be used under the hood for when we put together an `open_mfraw` function.

Notes from #180 for Environment group:
- For EK60: these parameters are encoded with each ping (each sample datagram)
- For EK80: these parameters are encoded with a single ping_time before all sample datagrams
- For AZFP: temperature and pressure of the instrument is recorded with each ping, but not salinity, which has to be supplied by the user during calibration.

imranmaj:
* Adds an `echopype.echodata.combine` module which contains functions for combining `EchoData` objects
    * Contains a method `combine_echodata` to combine multiple `EchoData` objects together
        * Corrects EK60 ping time, location time, and mru time reversal (#297) and stores the original `ping_time` in the provenance group
        * Stores original attributes from all `EchoData` groups in the provenance group
        * Has a parameter for determining attribute combination strategy
    * Exposes `combine_echodata` in the base `echopype` module
    * Adds tests for the `combine_echodata` function
* Adds an `echopype.core` module which contains common constants and objects
    * Moves the `MODELS` dictionary from `echopype.core.api` to `echopype.core` and renames it to `SONAR_MODELS`
    * Adds a type hint for sonar model names
        * ~~Removed because `typing.Literal` does not exist in Python 3.7~~
        * Re-added using a `typing_extensions` dependency
    * Adds type hints for multiple `EchoData`-related functions
* Changes `EchoData.__group_map` from an instance attribute to a class attribute and removes name mangling

Closes #364


echopype/tests/test_echodata.py
```diff
@@ -59,7 +59,7 @@ def test_constructor(self):
             'vendor',
         ]
 
-        assert ed.sonar_model == 'EK60'
+        # assert ed.sonar_model == 'EK60'
```
imranmaj, line 62: This failed the tests. I don't think sonar_model should be "EK60"; it's not being passed to the constructor.

emiliom: FYI, this must be due to this [bug in EK60 sonar_model parsing](https://github.com/OSOceanAcoustics/echopype/issues/210#issuecomment-821761942), where "ER60" (the software that runs EK60) is being entered by mistake. Ultimately we need to address that issue, but in the meantime I agree that it's better to comment out that assertion.Thanks! cc @leewujung

lsetiawan: Did the file `Summer2017-D20170615-T190214.zarr` sonar model changed? This test should have parsed the sonar model from there, which it used to be "EK60". @imranmaj added "keywords" in ds on https://github.com/OSOceanAcoustics/echopype/blob/5c8d1db8d48f6928532ff44a66b0991271301677/echopype/echodata/echodata.py#L137

Does this check potentially break the intended behavior?

emiliom: >Did the file Summer2017-D20170615-T190214.zarr sonar model changed? This test should have parsed the sonar model from there, which it used to be "EK60"

What I'm referring to seems to be a broader problem and is not specific to that file. See the brief discussion I pointed to. Though, are you saying that `assert ed.sonar_model == 'EK60'` used to pass with that file as a test?

lsetiawan: > Though, are you saying that assert ed.sonar_model == 'EK60' used to pass with that file as a test?

Yepp. See current build:
https://github.com/OSOceanAcoustics/echopype/runs/2822876705#step:12:241

lsetiawan: Looks like this is fixed. Thanks @imranmaj

echopype/echodata/combine.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def combine_echodata(echodatas: List[EchoData]) -> EchoData:
+    """
+    Combines multiple echodata objects into a single EchoData by combining
+        their groups individually.
+
+    If echodatas contain EchoData objects with different or None sonar_models,
+        a ValueError is raised.
+    """
```
lsetiawan, line 48: Since this method will be public and used by users, could you expand the docstring to follow the numpy doc. Here's a documentation about it: https://numpydoc.readthedocs.io/en/latest/format.html

echopype/echodata/echodata.py
```diff
@@ -166,7 +164,7 @@ def _check_suffix(self, filepath):
         return suffix
 
     def _load_group(self, filepath, group=None):
```
lsetiawan, line 166: While you're at this, could you include type hints for the methods within Echodata objects also? Thanks! 😄

lsetiawan (changes requested):
Nice work @imranmaj You did an awesome job on the creation of this functionality! 🥳  I am so excited to expanding and parallelizing this!!!

I have a couple of comments for the current changes. Also, could you add a test for this `combine_echodata`? I think it would be good if we know it works as intended. Plus it should bring up the code coverage I think. Thank you. 🙏🏽

emiliom:
I haven't had a chance to review this PR closely, though I did scan the code. It's exciting!

One comment I have, though not so much about this PR per se (ie, not about the direct functionality), is that we should consider saving the unmodified time coordinate as a provenance record within the dataset; maybe just the timestamps that were changed, saving the old and new values. This functionality modifies the original data, however tiny and inconsequential the change is in most, if not all cases. Saving the original data and the fact that changes were made would preserve the information in the source raw data and allow for downstream checking if it turns out that some of those changes were not so tiny.

leewujung:
@emiliom : I agree with your suggestions. The change is actually _not_ tiny, because _all_ the timestamps starting from the one that was reversed are changed. What is a more compact form of storing the change? How about saving the original timestamp and the "shift" for only the reversed one as dictionary pairs like below?
```python
# original ping time: shift in timedelta64[ns]
{
   "2021-06-23T09:00:02323": 1004867327,
   "2021-06-23T10:05:02383": 1054623427,
   ...
}
```

leewujung:
Adding a note from last week's conversation between me, @lsetiawan and @imranmaj : 
- let's separate out the "smart" encoding of the original `ping_time` to a separate issue (@imranmaj will write this)
- for now just add a data variable in the Provenance group to store the original `ping_time` so that we can merge this

One thing to be resolved from last week is an error in testing when parsed data from different files are combined.
@imranmaj : @emiliom noticed a problem with `location_time` (issue #392, fix in #393), I seem to remember that the error you ran into was with `location_time` also? Not sure, but thought I'd highlight the issue and PR.

emiliom:
One thing I noticed recently is that the `Platform` group includes a sub-group called `NMEA` which has a `location_time` dimension and variable. I assume this is exactly the same as the unmodified `location_time` in `Platform`? If so, that means that an unmodified copy of `location_time` is already being preserved. (BTW, this NMEA sub-group is not exposed by the awesome `EchoData` repr, so it's easy to forget that it's there)

leewujung:
> BTW, this NMEA sub-group is not exposed by the awesome EchoData repr, so it's easy to forget that it's there

Yes, and I have conveniently forgot about it...

But yes, the original `location_time` is preserved in there. Although I wonder if it is still a good idea to have both the original `ping_time` and `location_time` in the Provenance group, so that users can find it easily in one place.

lsetiawan:
@imranmaj I found a bug as I was making the presentation for today. I notice that when combining the echodata, the attributes are not preserved. So if you add new attributes to each then merge, they dissappear.

emiliom:
> But yes, the original location_time is preserved in there. Although I wonder if it is still a good idea to have both the original ping_time and location_time in the Provenance group, so that users can find it easily in one place.

Yeah, I think you're right. The `Provenance` group is the natural place for that data.

imranmaj:
@leewujung @lsetiawan Can you review this please?

echopype/tests/test_combine_echodata.py
```diff
@@ -0,0 +1,114 @@
...
+            ek60_reversed_ping_time_test_data,
+            "EK60",
+            None,
+            True,
+            SONAR_MODELS["EK60"]["concat_dims"],
+            SONAR_MODELS["EK60"]["concat_data_vars"],
+        ),
+    ],
+)
+def test_combine_echodata(
```
lsetiawan, line 62: Could you add a test for the attribute overriding stuff?

echopype/echodata/combine.py
```diff
@@ -0,0 +1,183 @@
...
+            )
+            concat_data_vars = SONAR_MODELS[sonar_model]["concat_data_vars"].get(
+                group, SONAR_MODELS[sonar_model]["concat_data_vars"]["default"]
+            )
+            combined_group = xr.combine_nested(
+                group_datasets,
+                [concat_dim],
+                data_vars=concat_data_vars,
+                coords="minimal",
+                combine_attrs="override",
```
lsetiawan, line 141: This should be optional maybe? Have a `keep_attrs` argument?

echopype/echodata/combine.py
```diff
@@ -0,0 +1,280 @@
...
+
+    # ping time before reversal correction
+    old_ping_time = None
+    # ping time after reversal correction
+    new_ping_time = None
+    # location time before reversal correction
+    old_location_time = None
+    # location time after reversal correction
+    new_location_time = None
+
```
lsetiawan, line 141: So, there are other "time" variables:
https://github.com/OSOceanAcoustics/echopype/blob/e01d57592385324b5747fb5c98d498ff50c42194/echopype/convert/set_groups_base.py#L25-L34

I was wondering if these are also handled? Not sure if like `mru_time` would have reversal or not, or the other `ping_time*` If so, then this would miss it, because it seems like only `ping_time` and `location_time` are handled.

leewujung: yes the `mru_time` needs to be added

lsetiawan:
@imranmaj did you add the `mru_time`? I think we should add that then merge this PR, it's been standing for a long time. Thanks!

PR #388, opened 2021-06-17
AD2CP Parsing Improvements

### LINKED ISSUE #360: Conversion improvements for AD2CP files
Let's use this issue to note down the various things to improve for the AD2CP file conversion. The issue text will continue to be updated.

Note that a lot of these are related to what encoding convention we'd follow since there are _a lot of_ ad-hoc addition to the current convention and that needs to be reviewed also.

### Overall
- many data variables have missing attributes: many of those still need to be defined by the convention, but should at least have the physical units

### Different groups
- `Sonar` group 
   - this group is currently empty but should be encoded
- `Beam_complex` group:
   - `echosounder_raw_transmit_samples_r|i` has the same `ping_time` dimension with NaNs padded to the same dimension as `echosounder_raw_transmit_samples_r|i`, because the transmit samples only contains 1 ping. Since there is a separate `time_echosounder_raw_transmit` coordinate, this NaN padding shouldn't happen?
- `Beam` group:
   - `amplitude_echosounder` is empty for data in Echosounder mode, but there is an additional variable `echosounder` that contains values that look like the Echosounder mode recording. However, there are weird spikes in the data that are not there when compared with OceanContour outputs like below:
   - Screen Shot 2021-05-19 at 10 31 26 PM

imranmaj:
* Uses `np.float64` explicitly for AD2CP numerical data
* Indexes AD2CP data variables by their correct `ping_time_*` dimensions instead of the overall `ping_time` dimension
    * Updates AD2CP tests to reflect this dimension change
* Changes AD2CP tests to use `np.allclose()` instead of `np.isclose().all()`
* Removes the AD2CP `velocity_echosounder` data variable from the `Beam` group
* Removes the AD2CP `amplitude_echosounder` data variable from the `Beam` group and renames the variable previously named `echosounder` to `amplitude_echosounder`
* Moves AD2CP raw echosounder data samples from the `Beam_Complex` group to the `Vendor` group
    * Deletes the `Beam_Complex` group
* Fixes spikes in AD2CP echosounder data
* Renames AD2CP raw echosounder r/i (real/imaginary) samples to i/q samples (i.e., `echosounder_raw_samples_i` -> `echosounder_raw_samples_q`, etc.)
* Encodes the sonar group

Closes #360


echopype/convert/set_groups_ad2cp.py
```diff
@@ -368,4 +368,16 @@ def set_vendor(self) -> xr.Dataset:
...
     def set_sonar(self) -> xr.Dataset:
-        return xr.Dataset()
+        ds = xr.Dataset(
+            attrs={
+                "sonar_manufacturer": "Nortek",
+                "sonar_model": "AD2CP",
+                "sonar_serial_number": "",
+                "sonar_software_name": "",
+                "sonar_software_version": "",
+                "sonar_type": "omnisonar",
```
leewujung, line 378: Could you change the following:
- `sonar_type` --> "acoustic Dopper current profiler"?
- `sonar_model` : @imranmaj do you know if there is a place in the file that encodes string like "Signature 1000" ?  I don't seem to remember one on top of my head.

For `sonar_serial_number`, please use the serial number in the parsed data (p.59 of integrator's guide). It is also in the string following "ID" below:

Also, because of how the instrument is set up, I think we should add an attribute `sonar_firmware_version`.  
@imranmaj : This should be the strings following "GETHW". You can find the spec in the guide.
@emiliom : tagging you here to see if you have any suggestions.

The first lines of GETALL that shows the config:
Screen Shot 2021-06-29 at 11 59 32 AM

imranmaj: > For sonar_serial_number, please use the serial number in the parsed data (p.59 of integrator's guide). It is also in the string following "ID" below:

I believe I did this here: https://github.com/OSOceanAcoustics/echopype/blob/5ea0cc05bdb9f62420b261b6b73fa07652bdf5e9/echopype/convert/set_groups_ad2cp.py#L381-L382

imranmaj: > do you know if there is a place in the file that encodes string like "Signature 1000" ? I don't seem to remember one on top of my head.

I don't recall one, but I will look through the data

imranmaj: @emiliom the `GETHW` portion is in a key-value format; is it ok to have a dictionary in an attribute?
image

emiliom: >@emiliom the GETHW portion is in a key-value format; is it ok to have a dictionary in an attribute?

I don't think attributes can accommodate dictionaries as such. I recommend transforming it into a string form, like this: "FW:2214, FPGA:185, DIGITAL:I-3 ...". That is, a string in a form that could be easily parsed into a dictionary

echopype/convert/set_groups_ad2cp.py
```diff
@@ -346,32 +367,23 @@ def set_vendor(self) -> xr.Dataset:
...
-            },
-            attrs={"pulse_compressed": self.pulse_compressed},
+            attrs={
+                "sonar_manufacturer": "Nortek",
+                "sonar_model": "AD2CP",
+                "sonar_serial_number": "",
+                "sonar_software_name": "",
+                "sonar_software_version": "",
+                "sonar_firmware_version": "",
+                "sonar_type": "acoustic Doppler current profiler",
```
emiliom, line 379: Tiny comment: it seems odd to only use upper case for Doppler, though I see why you'd do that. I couldn't find a specific vocabulary term for ADCP sonar_type in the ICES vocabularies. How about "Acoustic Doppler Current Profiler (ADCP)"?

echopype/convert/set_groups_ad2cp.py
emiliom: > `sonar_model` : @imranmaj do you know if there is a place in the file that encodes string like "Signature 1000" ? I don't seem to remember one on top of my head.

I don't know what the string "Signature 1000" represents, given that you're already describing the `sonar_model` as AD2CP. Looking at https://www.nortekgroup.com/products, I see one model that's described as having "AD2CP capabilities", and one model that's specifically called "Signature1000" (which matches this model descriptor in the NERC Vocabulary Server: http://vocab.nerc.ac.uk/collection/L22/current/TOOL1009/)

leewujung: Signature 1000 is the specific instrument model that is used to collect the data. "Signature" is a family/series of instruments Nortek developed that have the additional "echosounder mode" in addition to the common velocity sampling capability. It seems that it is being used interchangeably with "AD2CP", with `.ad2cp` being the file extension for all instruments in this family.

echopype/convert/set_groups_ad2cp.py
leewujung: Please reverse this change -- it should actually be "acoustic Doppler current profiler (ADCP)". Doppler is capitalized because it is a name. This is what wikipedia uses also.

leewujung:
@imranmaj : Thanks for the changes! I just checked them. A couple of things:

### ping_time_* issues
some of the ping_time_* selections seem still points to `ping_time` instead of `ping_time_echosounder`. Could you do a thorough check up using different types of data? 

For example, the test file 139.00027 under `Beam` group has an empty `amplitude_echosounder` data variable (it should not be empty) and an `echosounder` data variable (this should be in `amplitude_echosounder`. The `echosounder` data variable has coordinates `ping_time` and `range_bin_echosounder`, but the first should be `ping_time_echosounder`.

I have uploaded example files to the test_CI_data/raw/139 with explanations on the data the files contain. Please use these files with data from different modes to make sure the coordinates and variables are encoded correctly -- **in BOTH the `beam` and `vendor`** groups.

-----------------
The two below are from previous comments, in case they are buried:

### sonar_type
Please use "acoustic Doppler current profiler (ADCP)" for the sonar_type. Doppler is capitalized because it is a name.

### sonar_model
Signature 1000 is the specific instrument model that is used to collect the data. "Signature" is a family/series of instruments Nortek developed that have the additional "echosounder mode" in addition to the common velocity sampling capability. It seems that it is being used interchangeably with "AD2CP", with .ad2cp being the file extension for all instruments in this family.

Do you see anywhere in the data that encodes "Signature 1000" ? If not, we can stick with sonar_model="AD2CP" for now.

imranmaj:
> ### sonar_model
> Signature 1000 is the specific instrument model that is used to collect the data. "Signature" is a family/series of instruments Nortek developed that have the additional "echosounder mode" in addition to the common velocity sampling capability. It seems that it is being used interchangeably with "AD2CP", with .ad2cp being the file extension for all instruments in this family.
> 
> Do you see anywhere in the data that encodes "Signature 1000" ? If not, we can stick with sonar_model="AD2CP" for now.

It seems like there is a single byte that specifies the model which can be looked up with a table, but there isn't actually a string with the name "AD2CP" or "Signature 1000":
image

emiliom:
I've found and fixed the problems causing the failures with the `test_convert_time_encodings` test, but I got an error when trying to add my commit to Imran's PR. I followed the [steps](https://tighten.co/blog/adding-commits-to-a-pull-request/) I've followed once or twice to add to someone else's PR, but this time I got an error:

```bash
$ git push imranmaj HEAD:ad2cp_convert_to_merge
ERROR: Authentication error: Authentication required: You must have push access to verify locks
error: failed to push some refs to 'git@github.com:imranmaj/echopype.git'
```

Unless someone can come up with an easy way to solve this, I suggest we merge this PR and then I'll submit another PR with my fixes.

Now for the actual bugs I fixed: There were actually two separate issues, both involving the `Platform/NMEA` group. The first one involved incorrect `location_time` encodings; that's the error we were seeing in the test logs. The second issue came up once I fixed the first one, and it involved an incorrect reference to the group's path (NMEA rather than Platform/NMEA). I'm puzzled why these issues had not come up before in prior test runs, but it doesn't matter.

Note that one visible outcome of the second fix is that the Platform/NMEA group will now show up in the EchoData repr as another group. Previously it was omitted altogether. Here's a sample:

image

leewujung:
@emiliom : Thanks very much for the investigation and the fixes. Let's do what you suggest for merging this first and then you submit another PR, since it is in the dev branch and we are more flexible there. I'll merge this now. 

> The second issue came up once I fixed the first one, and it involved an incorrect reference to the group's path (NMEA rather than Platform/NMEA). I'm puzzled why these issues had not come up before in prior test runs, but it doesn't matter.

I wonder if this is related to the conditional test triggering we have right now, as not all the tests are run when changes are made. I think right now it is configured to only run a thorough tests when merging to master. Since our discussions of workflow have been scattered in various issues/PRs, it seems worthwhile that we put together a brief workflow document and also document the current testing strategy, and from there it should be clear what changes we need to make.

emiliom:
Thanks. I'll submit the PR in a bit.

> I wonder if this is related to the conditional test triggering we have right now, as not all the tests are run when changes are made. I think right now it is configured to only run a thorough tests when merging to master. Since our discussions of workflow have been scattered in various issues/PRs, it seems worthwhile that we put together a brief workflow document and also document the current testing strategy, and from there it should be clear what changes we need to make.

Good point. And I agree, we should create a document that describes the workflow, and possibly also the conditional tests.

imranmaj:
Thank you!

PR #393, opened 2021-06-29
Restore location_time encoding assignment

### LINKED ISSUE #392: Possible regression bug with location_time in Platform group
It looks like with 0.5.1,  `Platform.location_time` is being created without the critical, required `units` attribute, as well as the `calendar` attribute. As a result, it was not a proper time variable and could only be read by xarray as a `float64` variable rather than a `datetime64`.

For `Platform.ping_time`, the `units` attribute is present as expected: `units = "seconds since 1900-01-01T00:00:00+00:00"`. The attribute is also created as expected with echopype 0.5.0! Same goes for the `calendar` attribute too (`calendar = "gregorian"`).

I found this issue when working with the Pacific Hake survey EK60 raw data on S3, eg: `ncei-wcsd-archive/data/raw/Bell_M._Shimada/SH1707/EK60/Summer2017-D20170728-T212916.raw`

*(Updated: Added info about `calendar` attribute)*

emiliom:
- Fixes #392 (in 0.5.1, `Platform.location_time` was no longer assigned encodings)
- Also made small correction to conda env creation instructions in "Contributing to echopype" doc


emiliom:
- Sorry for the mess of several commits that were all about house cleaning in my local clone!
- It looks like a docs commit from a month ago had not been merged into dev, so it's getting merged here. Odd

leewujung:
Thanks @emiliom for merging everything. :)  It's kinda funny that right now master, dev, and stable all point to the same point -- does this mean that we have not been making much (if any!) progress!? 🤯

Anyway, this looks good. Right now the `DEFAULT_` stuff are all in set_groups_base.py. I wonder though if we should put this into a core and from that we could be testing compliance against. Not right now, but soon, as we wade into the convention issues (supposed to be in v0.5.3!)

leewujung (approved):

lsetiawan:
Please wait before merging. If this should be in the same version, then we should have a hotfix, which would then bump the build later on conda.

leewujung:
> this should be in the same version

Do we really want this to be in the same version? (I assume we're talking about the current v0.5.1.)
Wouldn't it be more explicit to have it in something like v0.5.1a (now that I type it out I don't know if that is a good idea) or v0.5.2 ?

lsetiawan:
I guess we can do v0.5.2 just for this. 👍🏽

emiliom:
For conda-forge packages I think there are "builds" that are not necessarily full-blown new releases. As a narrow but important bug fix, I don't know if this is best viewed as a build or as a new release ...

leewujung:
@emiliom : @imranmaj @lsetiawan and I discussed this on Thursday and decided that with this other PR (#388) it makes sense to make minor release. So, I'll be merging this to dev directly.

Issue #378, opened 2021-06-13
workflow for documentation updates


emiliom:
It would be helpful to be able to updates the ReadTheDocs documentation without being tied to code release cycles. Here I'm referring specifically to doc updates that reference the latest *released* API; in other words, **corrections of mistakes, improved explanations, expanded content, etc**.

In the Git development-branch (`dev`) workflow we've adopted for echopype, all changes (code & docs) are merged into `dev` first, and `dev` is merged into `master` only when a new release is ready to be issued. Changes to the API in `dev` potentially affect the API descriptions in the docs, so they may conflict with the documentation for the current release.

I'd like to propose the following workflow for documentation:

1. Use a `docsdev` branch that serves as an analog to the `dev` branch but is used only for doc changes that don't involve changes to the API -- the types of changes listed in the first paragraph. Documentation-only PR's involving such changes would be merged into this branch. 
2. When we feel there are "enough" changes in `docsdev`, we merge into `master` and prepare a new minor, tagged echopype release that only includes documentation updates. We also merge from into `dev`, to keep that branch in sync.
3. To avoid spawning new pypi and conda packages for this minor release, we can tag the release and include a letter at the end of the version; eg, 0.5.1a
4. Use the tag or the letter suffix ensure that the CI that builds packages is not triggered.
5. At https://echopype.readthedocs.io, change the `latest` version so it points to the `dev` branch rather than the `master` branch. The default version is `stable` (as it should be, IMHO), which is updated with new releases.

I realize that adopting this workflow adds complexity, for the benefit of being able to push docs updates to live more quickly than the code driven package release cycle allows. I would manage this process. I foresee that at least for a few more months, there is work to be done on the documentation that would benefit from this faster docs release workflow. Maybe by then the benefits of this faster but more complex cycle will no longer outweigh its costs, and we can go back to just the `dev` > `master` workflow.

I did some searching on the topic of documentation workflows and couldn't find anything helpful. Some of these suggestions stem from my chat with Jessica Scheick, the maintainer of https://icepyx.readthedocs.io/

Thoughts?

lsetiawan:
[...]

PR #395, opened 2021-07-17
Fix Platform/NMEA group location_time encoding and group path issues

emiliom:
Fix the problems causing the failures with the test_convert_time_encodings test (for ek60 and ek80 only); see #388

There were actually two separate issues, both involving the `Platform/NMEA` group. The first one involved incorrect location_time encodings; that's the error we were seeing in the test logs. The second issue came up once I fixed the first one, and it involved an incorrect reference to the group's path (`NMEA` rather than `Platform/NMEA`). I'm puzzled why these issues had not come up before in prior test runs, but it doesn't matter.

Note that one visible outcome of the second fix is that the `Platform/NMEA` group will now show up in the EchoData repr as another group. Previously it was omitted altogether. Here's a screenshot:

image


emiliom:
@leewujung all checks passed, so it looks like it's good to go. I'll self merge, then you can prepare and issue [0.5.2](https://github.com/OSOceanAcoustics/echopype/milestone/20).

In case it helps, here's where the bulk of our github workflow discussions are at: #378

Issue #259, opened 2021-02-10
Support for depth coordinates


valentina-s:
Currently echopype's output contain coordinates in units range from sensor. For further analysis, researchers may want to be working with coordinates in watercolumn depth for scientific visualizations and alignment with other variables.  It would be nice to have an option to have the water column depth coordinates as part of the object (as opposed to calculating them on the fly every time from the water level).  Can the converted data have the option to store depth coordinates? If we have a method  `coordinate.range_to_depth`, we will need to decide if that populates a separate attribute? Is there a danger those are not kept in sync while doing different computations. Thoughts @leewujung, @ngkavin.

leewujung:
Adding a note here while I review AZFP code:

AZFP has a tilt sensor and the tilt is encoded in the `Beam` group. When deployed in a mooring, range*cos(tilt) will be the actual depth. (ASL Env Sci provides both a dual cage that's inline with the mooring, or a mooring with 15 deg tilt.)

These should be included when we add function for this range to depth conversion, and in plotting.

leewujung:
@imranmaj : as we discussed, the `calc_range_meter` method currently used for calibration can be pulled out the repurpose for this. I think this can be in the form of 
```
echodata = ep.open_raw("some_raw_file.raw")
range_in_meter = echodata.get_range(...)
```
Users can just get an xarray DataArray out that they can use in whatever way they want, and we use the same function in the calibration functions as we do right now.

@valentina-s : Do you envision having the depth itself as outputs (as in for EK60/EK80 adding in the water_level parameter)? There is also the intricacy that for AZFP the instruments could have a non negligible tilt that needs to be factor in when converting range to depth (doing a cosine).

@lsetiawan also tagging you here for inputs. :)

leewujung:
@b-reyes : this is very much related to the addition of the water_level/range_offset in the compute_Sv outputs (#516). Could please you take this on? Thanks!

leewujung:
Trying to sort this out a bit more, I think we need to decide if we want to allow compute_Sv to directly output a data variable `depth`,  rather than just giving users access to both `sonar_range` and `water_level` (or `range_offset`).

For keeping this general I vote for the latter one (keeping 2 variables).

Note: `sonar_range` may become `echo_range` or other names (see #570), and we are renaming `water_level` to `range_offset` to comply with the convention (see #516).

Seeking inputs from: @b-reyes @valentina-s @emiliom

leewujung:
[...]

PR #400, opened 2021-07-19
Add compute_range method to EchoData, fix #310

### LINKED ISSUE #310: Error when trying to calibrate an EK80 file with both CW and BB channels

imranmaj:
Part of #259 

* Adds a `compute_range` method to `EchoData`
    * Changes `CalibrateEk.compute_range_meter` and `CalibrateAZFP.compute_range_meter` to call this new method on `EchoData` internally
    * Adds test for `EchoData.compute_range`
* Removes the `tvg_correction_factor` parameter from `CalibrateEK.compute_range_meter`
* Adds ABC to calibrate_base.py
* Fixes #310 
    * Adds test for file with both CW and BB channels


echopype/tests/test_echodata.py
```diff
@@ -133,3 +137,33 @@ def _check_path(zarr_path):
...
+        (ad2cp_path / "raw" / "090" / "rawtest.090.00001.ad2cp", "AD2CP", None)
+    ]
+)
+def test_compute_range(filepath, sonar_model, xml_path):
+    ed = echopype.open_raw(filepath, sonar_model, xml_path)
+    sound_speed = 343
+    if sonar_model == "EK60":
+        range = ed.compute_range(sound_speed, ek_waveform_mode="CW")
+    elif sonar_model == "EK80":
+        range = ed.compute_range(sound_speed, ek_waveform_mode="CW")
```
leewujung, line 157: @imranmaj : Is there a test for `ek_waveform_mode="BB"` or is there a reason why that is not included here?

leewujung (approved):

leewujung:
@imranmaj : thanks, great work! Other than the BB test case above, I think this is ready to go. Let's talk about how to create a function for #259 which is also related to regridding tomorrow.

leewujung:
@imranmaj : Thanks for this great PR with new functionality and under the hood improvements!  😃 

### docstring improvement
I just have one thing for improving the docstring for `compute_range`: While testing `compute_range` for AZFP, I found that it is not obvious to the user that they need to supply `env_params` since sound speed is not recorded in the data and needs to be computed from temperature, salinity, and pressure. And also that salinity and pressure also need to be supplied by the user because they are not part of the data file. 

Could you expand the docstring of `compute_range` in a fashion similar to EK60/80 so that users can have the above info?


### Note for the future
Another thing I wanted to note here is that, ideally we should add a test to run cal on `waveform_mode="BB"` and `encode_mode="complex"` option for the file `Summer2018--D20180905-T033113.raw` (the file with co-existing CW+BB data) -- this test currently is not included. However, the size of the file (305 MB) makes it too long to run it. We'll need to optimize the code for the many convolutions needed in the operation, but I want to note this down here so we do not forget and also get this on @lsetiawan's radar for when we make test data changes for speeding up CI runs.

imranmaj:
The CI is failing because I pulled the dev branch, but the current dev branch has issues with pre-commit so I reverted the pull. It seems like pre-commit won't accept the reversion that I did and permanently gives me an error.

leewujung:
From just the pre-commit.ci messages, it seems that it is a line being too long in combine.py? Is that it or is that a symptom of something else?

lsetiawan:
> From just the pre-commit.ci messages, it seems that it is a line being too long in combine.py? Is that it or is that a symptom of something else?

Seems like that's the only problem I see. Actual python tests are passing.

imranmaj:
Yes, it says that combine.py line 84 is too long, but it's incorrect; in the actual code ([https://github.com/imranmaj/echopype/blob/a83819540c4e1f700428bc46f6a1f6f5dcffb583/echopype/echodata/combine.py#L84](https://github.com/imranmaj/echopype/blob/a83819540c4e1f700428bc46f6a1f6f5dcffb583/echopype/echodata/combine.py#L84)), the line is only 97 characters long and it was not modified by this PR.

lsetiawan:
I see 2 force pushes happened. I feel like pre-commit is confused as to what to check?

lsetiawan:
I think it's safe to just merge it since I just checked locally and it passed.

imranmaj:
> I see 2 force pushes happened.

Yes, I pulled the latest dev branch but the latest dev branch has the pre-commit issue. I fixed the issue in #425 ([https://github.com/OSOceanAcoustics/echopype/pull/425/files#diff-cce6b1b1f93becbc188e01b21cf0a711ad216692eea7e96f98642ebc4579059fR84-R85](https://github.com/OSOceanAcoustics/echopype/pull/425/files#diff-cce6b1b1f93becbc188e01b21cf0a711ad216692eea7e96f98642ebc4579059fR84-R85)) but I didn't want to duplicate the fix, so I tried to revert it.

imranmaj:
Actually, now that #425 has been merged, I can try pulling the dev branch again.

Issue #351, opened 2021-05-15
Consolidate path checking and handling functions


leewujung:
Right now `open_converted` uses helper functions from `io` (`check_file_existance` and `sanitize_file_path`), and `open_raw` uses a separate suite of similar functions in `convert.api`. Let's consolidate and use these helper functions from a single place.

leewujung:
This is addressed in #401.

PR #401, opened 2021-07-23
Improve io, consolidate functions, and unpin dependencies.

lsetiawan:
This PR removes `_validate_path` function in `convert.api` and replace it with `validate_output_path` within the `io` module. Additionally `_normalize_path` function is replaced with `sanitize_file_path`, which is already used within `EchoData`.

The CI Helper script to run tests is now improved to be able to add new modules to test simply by providing another key to `MODULES_TO_TEST`.

Relates to #351


echopype/utils/io.py
```diff
@@ -49,36 +59,145 @@ def get_file_format(file):
...
+                warnings.warn(
+                    "Mismatch between specified engine and save_path found; forcing output format to engine."  # noqa
+                )
+            out_path = str(
+                final_path.parent.joinpath(final_path.stem + file_ext).absolute()
+            )
+    return out_path
+
+
 def check_file_existance(
```
leewujung, line 201: Spotted a typo here: existance --> existence

leewujung:
Thanks @lsetiawan for this work!  I like the test modularization also. :D 

Could you add some explanation in the docstring for `sanitize_file_path` so that it is clear to others why there is `validate_output_path` followed by `sanitize_file_path` for a file to be written?

It seems to me that #351 is actually resolved by this because `open_converted` now uses the same mechanisms under `EchoData` so it is pretty clean.

leewujung:
@lsetiawan : thanks, this is great! We should put out a new release ASAP so that people won't be getting v0.4.2 anymore but our nice new v0.5.x 😀 

This PR addresses #406 .

Issue #269, opened 2021-03-18
Adding support for ES80 data


DLWangSan:
Hello, I used echopype to analyze the EK80 data on board, and the workload is greatly reduced! Now that we have ES80 sonar data, I wonder if subsequent versions of echopype will add support for ES80 devices?

The following is some data of ES80 sonar equipment：
```
https://drive.google.com/drive/folders/1j7GKaE-WpfAhrD3MHQ4PXe1yHo3PaFIb
```

Thank you !

leewujung:
@DLWangSan : glad that you found echopype helping your work! :)

It'll be awesome to include ES80 support in echopype, though at the moment we are short on hands to add support for ES80.

Would you be interested in helping adding the support for ES80?  A quick googling led to [this doc](https://www.simrad.online/es80/ref/es80_ref_en_a4.pdf) that explains the format and it seems reasonable. I can help with getting things into the standardized format.

DLWangSan:
thank you very much! If I can help, I would be happy to contribute a little to echopype

leewujung:
Great! If you are able to get started on a parser to read in the data, I can set up a branch that you could PR to. Just let me know.

gavinmacaulay:
As I understand it, the file format of the ES80 is identical to the EK80 (with a few trivial metadata changes).

leewujung:
@gavinmacaulay : that's great! Do you have some test files that we could try on?

gavinmacaulay:
I have ES80 files that I can offer. Where should I put them? I also have ES70 files (same format as the ES/ER60 files) if desired. 

To go with these, I have a nearly ready PR that makes the fairly trivial changes necessary for echopype to read ES70 files (ES80 will be just as trivial).

leewujung:
Fantastic! I just sent you an invitation to a Google Drive folder we use to store test data files at the moment (as documented [here](https://echopype.readthedocs.io/en/stable/contributing.html#test-data-files)). 

Looking forward to the PR!!  😃  We can add tests to run through your files once they are in the folder, like what we have for all the other formats.

gavinmacaulay:
echopype reads ES80 files without any changes needed if those files only contain data from WBT transceivers (the files are near identical to those produced by the EK80 software). But ES80 (and EK80) files can also contain data from a GPT transceiver. Currently, echopype fails to read those files. See #409 for some more details.

leewujung:
Closing this so that ES80-related issues are consolidated, in #497.

PR #409, opened 2021-08-09
Allow data files of type ES70

gavinmacaulay:
The ES70 file format is identical to the ER60 and ES60 file format, so accept files labelled as ES70 as if they were ER60 format.

### DIFF
--- a/echopype/convert/utils/ek_raw_parsers.py
+++ b/echopype/convert/utils/ek_raw_parsers.py
@@ -1176,7 +1176,7 @@ class SimradConfigParser(_SimradDatagramParser):
         beam_config                     [str] Raw XML string containing beam config. info
 
 
-    Transducer Config Keys (ER60/ES60 sounders):
+    Transducer Config Keys (ER60/ES60/ES70 sounders):
         channel_id                      [str]   channel ident string
         beam_type                       [long]  Type of channel (0 = Single, 1 = Split)
         frequency                       [float] channel frequency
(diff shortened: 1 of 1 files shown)


gavinmacaulay:
This PR also includes some commits to support reading ES80 data files. Some details about this:
- The EK/ES80 software supports EK/ES80 hardware (a WBT transceiver) and EK/ES 60/70 hardware (a GPT transceiver)
- Versions of the ES80 software earlier than 1.4.3.0 writes .raw files that have an ApplicationName="ES80" entry in the XML0 Configuration datagram, while version 1.4.3.0 (the latest available at this time) produces ApplicationName = "EK80". I do not know if this was deliberate or unintended by Simrad. 

- echopype reads ES80 files that contain just WBT transceiver data without any changes needed (tested with example files es80/WBT-D20210620-T012250.raw and es80/WBT-but-internally-marked-as-EK80-D20210710-T204029.raw)
- echopype fails to read ES80 files that contain GPT transceiver data. The above PR resolves one problem when reading GPT data, but the code then fails on the assumption that all EK/ES80 data channels have non-empty FIL1 datagrams (GPT channels don't have FIL1 datagrams). This fail occurs on file es80/WBT-GPT-D20181216-T111107.raw (to complicate things, this file has one channel from a WBT and one channel from a GPT).

Understanding the code enough to resolve the GPT inside a EK/ES80 file will take me some time, so I'll leave that to the more experienced - see issue #269.

gavinmacaulay:
I have uploaded a smaller WBT/GPT file (`WBT-GPT-D20181211-T091119.raw`) and removed the larger file. This is the smallest WBT/GPT file that I have. If you want something smaller I can replay it through the EK80 software and re-record it into smaller pieces, but this would mean it would become an EK80 format file, not ES80. Wouldn't make any practical difference though... So if 5.4 MB is too large, just say so...

leewujung:
Thanks @gavinmacaulay ! 

To make sure I completely understand:
> I have uploaded a smaller WBT/GPT file (WBT-GPT-D20181211-T091119.raw, about 5MB) and removed the larger file. This is the smallest WBT/GPT file that I have. 

This is still an ES80 file, right?  (I haven't tried loading it in with your PR code but will do so soon)

> If you want something smaller I can replay it through the EK80 software and re-record it into smaller pieces, but this would mean it would become an EK80 format file, not ES80. 

Oh wow, I didn't know this trick, that is awesome! so maybe we can make the other EK80 test files smaller too, particularly the ones with complex broadband data.

> Wouldn't make any practical difference though... So if 5.4 MB is too large, just say so...

I think we can do 5.4 MB for now, especially since it's nice to have an actual ES80 file being run through the tests. Let's see how that goes. 😄

gavinmacaulay:
> 
> This is still an ES80 file, right? (I haven't tried loading it in with your PR code but will do so soon)

Correct.
 
> > If you want something smaller I can replay it through the EK80 software and re-record it into smaller pieces, but this would mean it would become an EK80 format file, not ES80.
> 
> Oh wow, I didn't know this trick, that is awesome! so maybe we can make the other EK80 test files smaller too, particularly the ones with complex broadband data.

It's a useful trick - it's easy enough to generate files with a small number of pings (e.g., 1) by using the playback speed settings. The only downside is that you end up with files written by the version of the software that you used to do the re-writing and not the version that wrote the original file.

leewujung:
> It's a useful trick - it's easy enough to generate files with a small number of pings (e.g., 1) by using the playback speed settings. The only downside is that you end up with files written by the version of the software that you used to do the re-writing and not the version that wrote the original file.

Fantastic! More ways to generate test files from the different versions of the software...!

PR #416, opened 2021-08-13
Put tests into designated folders

lsetiawan:
This PR moves tests into their designated folders that reflects the structure of echopype.


lsetiawan:
The coverage reporting is now more accurate with these changes. 😄

lsetiawan:
@leewujung This was supposed to be for the next version release, but I wonder if it's worth it to just merge it now in this release since we haven't released yet. This includes skipping the out of memory file for now. Pinging @imranmaj so that you're aware.

leewujung:
> @leewujung This was supposed to be for the next version release, but I wonder if it's worth it to just merge it now in this release since we haven't released yet. This includes skipping the out of memory file for now. Pinging @imranmaj so that you're aware.

Yes we can merge this now and give it a try with other PRs, both standalone and when merged. I think we can release very soon.

> The coverage reporting is now more accurate with these changes. 😄 At least for this, since it runs almost all the modules... I wonder what happens when it runs just a single module.

Something we will find out soon! :)  I should be able to work in the summary statistics one today with this merged.

PR #423, opened 2021-08-17
doc strings enhancements and corrections

emiliom:
- Enhanced or corrected several docstrings
- Added new functions to api docs page


echopype/echodata/combine.py
```diff
@@ -90,10 +77,12 @@ def combine_echodata(echodatas: List[EchoData], combine_attrs="override") -> Ech
...
         If the `sonar_model` of the input `EchoData` objects is `"EK60"` and any `EchoData` objects
-        have non-monotonically increasing `ping_time`s or `location_time`s,
-        the `ping_time`s or `location_time`sin the output `EchoData` object will be increased
+        have non-monotonically increasing `ping_time` or `location_time` values,
+        the `ping_time` or `location_time` values in the output `EchoData` object will be increased
         starting at the timestamp where the reversal occurs such that all
-        `ping_time`s and `location_time`s in the output are monotonically increasing.
+        `ping_time` and `location_time` values in the output are monotonically increasing.
+        Additionally, the original `ping_time` and `location_time` values will be stored
+        in the provenance group, although this behavior may change in future versions.
```
leewujung, line 85: @emiliom : there's also `mru_time` that was added to this check later on, which didn't make its way to this docstring. Could you add it in?

@imranmaj : wanted to double check with you to make sure that the `ping_time_*` coordinates in AD2CP files are not included in this checking -- from the code it does not seem like they are.

Thanks both!

emiliom: Sure, I can add mru_time in a bit.

leewujung:
Thanks for the updates @emiliom ! The `sphinx-mermaid` diagram is wonderful! 😃

emiliom:
Darn, this pre-commit auto-fix hook messes with the ability to push another commit on top of an existing PR! I don't know what's most efficient here, so I think I'll just wait to see what @lsetiawan recommends tomorrow. All I can think of is to merge this PR, then pull `dev` into my local clone and work on top of that. That's a hassle.

PR #425, opened 2021-08-17
Check for mru_time reversal in EK80, not EK60

imranmaj:
`mru_time` exists only in EK80; it does not exist for EK60.


leewujung:
@imranmaj : `ping_time` and `location_time` actually both exist in EK80 files. The current logic is such that those two won't be checked. Could you make it such that these two are checked for both EK60/EK80, and `mru_time` only for EK80?

Also, the printouts for the `mru_time` should also be EK80 to be consistent with the behavior.

Thanks!

PR #427, opened 2021-08-19
Determine AD2CP raw transmit based on header id

imranmaj:
Fixes a bug where raw transmit data is placed in `echosounder_raw_samples_i/q` instead of `echosounder_raw_transmit_samples_i/q` by interpreting any packet with the header id for echosounder raw transmit data as echosounder raw transmit data.


leewujung:
This looks great, thanks @imranmaj !

PR #428, opened 2021-08-19
Allow AZFP extensions with more numbers and letters

### LINKED ISSUE #418: Using open_raw for AZFP data files with name other than .01A
We can convert AZFP data files using open_raw for the .01A file. However, not all AZFP data files are .01A files. There are 2 parts that the filename can change (the number and letter of the file)
1. The number: 
When we collect data with multiple phases (max 12 phases) in one deployment, the data files will format as .01A (for phase 1), .02A(for phase 2),... So the AZFP file name could be from .01A to .12A
2. The letter:
When we deploy the instrument multiple times in one hour, the data files name will be ascending in alphabetical order each time it's deployed, which the file name could be from .01A to .01Z. 

image

image

@leewujung @imranmaj hope you can fix this problem. Let me know if you need more information about the AZFP files.

imranmaj:
Allows any extension in the form ".XXY" where XX is a number and Y is a letter.

Fixes #418

### DIFF
--- a/echopype/convert/api.py
+++ b/echopype/convert/api.py
@@ -308,12 +308,11 @@ def _check_file(
     # TODO: https://github.com/OSOceanAcoustics/echopype/issues/229
     #  to add compatibility for pathlib.Path objects for local paths
     fsmap = fsspec.get_mapper(raw_file, **storage_options)
-    ext = SONAR_MODELS[sonar_model]["ext"]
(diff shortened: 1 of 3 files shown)


leewujung (approved):

leewujung:
@imranmaj : the `validate_ext` function is awesome! Thanks for fixing this issue!!  😃 

I'll fix that old test function in another PR.

Issue #201, opened 2020-10-15
Method to add missing GPS data to platform group


leewujung:
Some EK80 data sets come without the NMEA datagrams due to hardware config variations. This is a case similar to #198 in which some environmental data are also missing in AZFP files. Let's first deal with the case when the latitude/longitude data are saved in netCDF or zarr files.

### Goal

Add a method to enable adding ancillary lat/lon data into converted files, when the source data file (recorded by the instrument) does not contain these data. 

### Task

- [ ] Add a method `add_platform_data()` to the `Convert` class that allows users to specify one or more nc/zarr files (will call this GPS files below) that can be opened by xarray and contain variables named `latitude` and `longitude`, and save the lat/lon to `Platform` group. The method should:
   - slice out the lat/lon data section that corresponds to the `ping_time` start and end of the acoustic data in the `Beam` group, and
   - be able to combine multiple GPS files before doing the slicing, using xarray's open_mfdataset.

- [ ] Make the `add_platform_data()` function work with either whole bunch of individual files, each individually converted to nc/zarr, or a single combined output file from a list of individual files (i.e., the `combine=True` option).

Note: this functionality should be added to the `class-redesign` branch.

leewujung:
@imranmaj : could you start tackling the first task as a standalone function? check out `set_platform()` in `/echopypt/convert/set_groups_ek80.py` to see how the coordinates and attributes are encoded; the idea here is to substitute (may actually be overwrite) the NaN variables to actual values from the GPS files. I'll point you to where the test files are on slack.

@ngkavin : I think the second task would require collaboration between you and @imranmaj. Could you suggest how the function argument should look like and where it should be called in the conversion sequence, and the 3 of us can discuss as a group?

Thanks :)

ngkavin:
Hi @imranmaj, glad to have you helping to develop echopype .

It seems like this function really only requires 1 input argument although more would accommodate more use-cases. That argument being the list of .nc/.zarr files that contain the platform information. This function could go anywhere between creating the `Convert` object and calling `to_netcdf`.
For example:
```
tmp= Convert(ek80_raw_path, model="EK80")
tmp.add_platform_data(files)
tmp.set_param(params)
tmp.to_netcdf()
```
The base functionality could be splitting the GPS data while saving each .nc file, but more use-cases could involve adding the GPS data after everything has been combine so that `add_platform_dict` would go after `to_netcdf(combine=True)`.

imranmaj:
[...]

PR #434, opened 2021-08-26
Add update_platform method to EchoData

imranmaj:
Part of #201 

Adds an `EchoData.update_platform` method which can be used to update the `EchoData.platform` group with additional external platform data.

Partially adapted from https://github.com/ngkavin/echopype/pull/2, https://github.com/ngkavin/echopype/pull/3

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -328,7 +328,7 @@ class EchoData:
             - `"longitude"`
             - `"water_level"`
 
-        Only `sonar_model`s of `"EK60"`, `"EK80"`, and `"EA640"` are currently supported.
+        Only `sonar_model`s of `"EK60"`, `"EK80"`, `"EA640"`, and `"AZFP"` are currently supported.
 
         Parameters
         ----------
@@ -342,7 +342,7 @@ class EchoData:
         Raises
         ------
         ValueError
-            If the `sonar_model` is not one of `"EK60"`, `"EK80"`, or `"EA640"`.
+            If the `sonar_model` is not one of `"EK60"`, `"EK80"`, `"EA640"`, or `"AZFP"`.
 
         Examples
         --------
@@ -365,6 +365,8 @@ class EchoData:
         if self.sonar_model in ["EK80", "EA640"]:
             time = "mru_time"
         elif self.sonar_model == "EK60":
+            time = "location_time"
+        elif self.sonar_model == "AZFP":
             time = "location_time"
         else:
             raise ValueError("unsupported sonar_model for adding platform data")
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -327,8 +327,7 @@ class EchoData:
             - `"latitude"`
             - `"longitude"`
             - `"water_level"`
-
-        Only `sonar_model`s of `"EK60"`, `"EK80"`, `"EA640"`, and `"AZFP"` are currently supported.
+        The data inserted into the Platform group will be indexed by a dimension named `"time2"`.
 
         Parameters
         ----------
@@ -338,11 +337,6 @@ class EchoData:
         time_dim: str, default="time"
             The name of the time dimension in `extra_platform_data`; used for extracting
             data from `extra_platform_data`.
-
-        Raises
-        ------
-        ValueError
-            If the `sonar_model` is not one of `"EK60"`, `"EK80"`, `"EA640"`, or `"AZFP"`.
 
         Examples
         --------
@@ -376,62 +370,22 @@ class EchoData:
             extra_platform_data = extra_platform_data.drop_vars(trajectory_var)
             extra_platform_data = extra_platform_data.swap_dims({"obs": time_dim})
 
-        if self.sonar_model in ["EK80", "EA640"]:
-            pitch_roll_heave_time = "mru_time"
-            lat_lon_water_level_time = "location_time"
-        elif self.sonar_model == ["EK60", "AZFP"]:
-            pitch_roll_heave_time = "location_time"
-            lat_lon_water_level_time = "location_time"
-        elif self.sonar_model == "AD2CP":
-            pitch_roll_heave_time = "ping_time"
-            lat_lon_water_level_time = "ping_time"
-        else:
-            raise ValueError("unsupported sonar_model for adding platform data")
-
-        # extra_platform_data = extra_platform_data.sel(
-        #     {time_dim: slice(start_time, end_time)}
-        # )
-
-        platform = self.platform
-
-
-        platform = platform.reindex(
-            {
-                pitch_roll_heave_time: extra_platform_data[time_dim].values,
-                lat_lon_water_level_time: extra_platform_data[time_dim].values,
-            }
+        platform = self.platform.assign_coords(
+            time2=extra_platform_data[time_dim].values
         )
-
-
-        # print(platform[pitch_roll_heave_time])
-        # print(xr.DataArray(extra_platform_data[time_dim].values, dims=pitch_roll_heave_time))
-        # new_time_values = xr.concat([xr.DataArray(extra_platform_data[time_dim].values, coords={pitch_roll_heave_time: extra_platform_data[time_dim].values}, dims=pitch_roll_heave_time), platform[pitch_roll_heave_time], platform[lat_lon_water_level_time]], dim=pitch_roll_heave_time)
-        # print(new_time_values)
-        # print(np.unique(new_time_values).shape)
-        # platform = platform.reindex(
-        #     {
-        #         pitch_roll_heave_time: new_time_values,
-        #         lat_lon_water_level_time: new_time_values,
-        #         # pitch_roll_heave_time: xr.concat([xr.DataArray(extra_platform_data[time_dim].values, coords={pitch_roll_heave_time: extra_platform_data[time_dim].values}, dims=pitch_roll_heave_time), platform[pitch_roll_heave_time]], dim=pitch_roll_heave_time),
-        #         # lat_lon_water_level_time: xr.concat([xr.DataArray(extra_platform_data[time_dim].values, coords={lat_lon_water_level_time: extra_platform_data[time_dim].values}, dims=lat_lon_water_level_time), platform[lat_lon_water_level_time]], dim=lat_lon_water_level_time),
-        #     }
-        # )
-        # extra_platform_data = extra_platform_data.reindex(
-        #     {
-        #         time_dim: new_time_values.data
-        #     }
-        # )
-        # print(platform["magnetometer_raw_x"])
-        # print(platform["magnetometer_raw_x"][np.invert(platform["magnetometer_raw_x"].isnull().data)])
-
 
         dropped_vars = []
         for var in ["pitch", "roll", "heave", "latitude", "longitude", "water_level"]:
             if var in platform and platform[var].isnull().all():
                 dropped_vars.append(var)
         if len(dropped_vars) > 0:
-            warnings.warn(f"some variables will be overwritten by platform data: {', '.join(dropped_vars)}")
-        platform = platform.drop_vars(["pitch", "roll", "heave", "latitude", "longitude", "water_level"], errors="ignore")
+            warnings.warn(
+                f"some variables will be overwritten by platform data: {', '.join(dropped_vars)}"
+            )
+        platform = platform.drop_vars(
+            ["pitch", "roll", "heave", "latitude", "longitude", "water_level"],
+            errors="ignore",
+        )
 
         num_obs = len(extra_platform_data[time_dim])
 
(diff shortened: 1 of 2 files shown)


leewujung:
@emiliom : seems like this operation is something we should record in the Provenance group also?

emiliom:
> seems like this operation is something we should record in the Provenance group also?

Great suggestion!! I totally agree.

imranmaj:
@emiliom I was running into issues with trajectory in the saildrone data before so I added that small section to fix it, I can investigate whether it's necessary at all

emiliom:
> @emiliom I was running into issues with trajectory in the saildrone data before so I added that small section to fix it, I can investigate whether it's necessary at all

I think it is necessary. I'm finally getting around to adding the generalized code (done). Now I just need to run the tests.

emiliom:
@imranmaj I'm running into an error with the `~=` operator in `requirements.txt`, again. When doing `conda create -c conda-forge -n echopype_dev --yes python=3.9 --file requirements.txt --file requirements-dev.txt`, I get this error:

```
InvalidVersionSpec: Invalid version '~3.10': invalid operator
```

I think the glitch is different from what I experienced earlier, a couple of weeks ago, but I don't remember the details of that instance. For now, I just need to build a conda env to run the echopype tests, so I changed `~=` to `=`. But this is starting to sound like a persistent problem. If no one else is running into this type of problem, could there be something about my local setup that needs to be updated or changed?

leewujung:
Hmm... I did not run into the ~= problem while creating a new environment with just python=3.8 and echopype. I used conda-forge as usual. @emiliom is the env you are working with a brand new one?

emiliom:
> is the env you are working with a brand new one?

Yeah. I'm creating a new env

echopype/echodata/echodata.py
```diff
@@ -312,6 +312,126 @@ def squeeze_non_scalar(n):
...
+        --------
+        >>> ed = echopype.open_raw(raw_file, "EK60")
+        >>> extra_platform_data = xr.open_dataset(extra_platform_data_file)
+        >>> ed.update_platform(extra_platform_data)
+        """
+
+        # only take data during ping times
+        start_time, end_time = min(self.beam["ping_time"]), max(self.beam["ping_time"])
+
+        # saildrone specific workaround
```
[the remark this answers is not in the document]
emiliom: I've pushed into your PR the code that generalizes the handling of "trajectory" data encoded using the convention used by the Saildrone sample file

echopype/echodata/echodata.py
```diff
@@ -433,7 +440,7 @@ def mapping_get_multiple(mapping, keys, default=None):
                     mapping_get_multiple(
                         extra_platform_data,
                         ["water_level", "WATER_LEVEL"],
-                        np.ones(num_obs),
+                        default=np.ones(num_obs),
```
leewujung, line 448: @emiliom : Why is this `np.ones` instead of `np.zeros` or  use `np.nan` like in other variables?  It seems bizarre to default water level to be 1 m?

echopype/echodata/echodata.py
```diff
@@ -312,6 +312,145 @@ def squeeze_non_scalar(n):
...
+        Data is extracted from `extra_platform_data` by variable name; only the data
+        in `extra_platform_data` with the following variable names will be used:
+            - `"pitch"`
+            - `"roll"`
+            - `"heave"`
+            - `"latitude"`
+            - `"longitude"`
+            - `"water_level"`
+
+        Only `sonar_model`s of `"EK60"`, `"EK80"`, `"EA640"`, and `"AZFP"` are currently supported.
```
leewujung, line 331: @imranmaj @emiliom : Could you please remind me why we don't allow this for AD2CP files? I seem to remember that we discussed this but can't recall what the conversation was.

echopype/echodata/echodata.py
emiliom: I don't know. I didn't pay attention to that at all. The change I added in my PR was only stylistic -- in all the similar calls to `mapping_get_multiple`, the `default` parameter is used explicitly rather than implicitly (positionally).

echopype/echodata/echodata.py
emiliom: I thinks that's for you and @imranmaj to answer. I don't remember being in a conversation about it, and wouldn't know if there's anything specific about AD2CP files that creates a special issue. I assume there isn't, and that its absence was just inertia.

echopype/echodata/echodata.py
leewujung: Ah I see, I somehow read this as adding a line, but you added the `default`.
My question should be for @imranmaj  :)

echopype/echodata/echodata.py
leewujung: Inertia! 😅 

@imranmaj : do you see problems with adding AD2CP support in this? The potential "problem" I can see is that AD2CP files comes with `pitch`, `roll` and `heading` but not `lat` `lon`, so it is a partially filled dataset. So suppose users then want to add `lat` `lon`, the operations would become reading in the existing Platform group and add the new variables. Is this accurate?

Perhaps more generally the approach would be to:
- check if there is duplications in variables in `extra_platform_data` with the existing `ed.platform` dataset, warn user of overwritting if that's what they select
- add an option to allow user to specify if they want to overwrite or use existing variables
  - if allow overwriting then replace everything
- add new variables not already exist in `ed.platform`

What do you think?

emiliom: For what it's worth, the need you've outlined is more or less identical to the situation for the OOI AZFP data. The platform group is created with some variables, but empty lat & lon.

echopype/echodata/echodata.py
imranmaj: I remember talking about this in a meeting, but I don't remember the exact reasoning; I made a PR specifically to make this change https://github.com/ngkavin/echopype/pull/3/files#diff-d35f9ea311a9dd316616bdd6291eb68fdb343808ce6ad80dc6b17402ec56db9aR563

imranmaj: It was np.nan originally

echopype/echodata/echodata.py
leewujung: Agreed!

echopype/echodata/echodata.py
leewujung: I wonder whether it was because in later operations, say for shipboard deployment, the true depth would be `water_level+range`, and if we have `np.nan` the outcomes would be NaN. I think though it should be `np.zeros` instead of `np.ones`

imranmaj:
External data is being indexed by a dimension named `"time2"` now based on 9/16/21 discussion:

* The sonar netCDF convention says that time dimensions in the Platform group should be named time1, time2, etc
* Indexing using a new time dimension will prevent existing time dimensions (such as location_time) from being overwritten, which in turn will prevent other variables indexed by the previous values of those time dimensions from being invalidated

echopype/echodata/echodata.py
```diff
@@ -312,6 +312,145 @@ def squeeze_non_scalar(n):
...
+            errors="ignore",
+        )
+
+        num_obs = len(extra_platform_data[time_dim])
+
+        def mapping_get_multiple(mapping, keys, default=None):
+            for key in keys:
+                if key in mapping:
+                    return mapping[key].data
+            return default
```
leewujung, line 399: @imranmaj: this function would only return the first existing keys in the mapping data set, right?

Seems that this could potentially produce bugs down the line, because the input argument is call `keys`, so users may think that a list with each element being an array of values corresponding to **all** of what's listed in `keys` would be generated, while in fact only the first existing key will be returned.

The current behavior of this function does exactly what you need this function to do in the code below, but in that case, the function name should not be `mapping_get_multiple`, because it would only get data from a single variable.

Which way do you intend this to be? A function that only returns data from 1 variable from all of the possible names, or a function that would return data from multiple variables?

echopype/echodata/echodata.py
```diff
@@ -312,6 +312,145 @@ def squeeze_non_scalar(n):
...
+        >>> extra_platform_data = xr.open_dataset(extra_platform_data_file)
+        >>> ed.update_platform(extra_platform_data)
+        """
+
+        # # only take data during ping times
+        # start_time, end_time = min(self.beam["ping_time"]), max(self.beam["ping_time"])
+
+        # Handle data stored as a CF Trajectory Discrete Sampling Geometry
+        # http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#trajectory-data
+        # The Saildrone sample data file follows this convention
```
leewujung, line 353: @emiliom: reading this section you pointed out makes me wonder whether or not it is worthwhile to point out that the lat/lon data in the SONAR-netCDF4 ver.1 is consistent with the H.4.2. Single trajectory specification. Not that this necessitates any actual changes, but there seems to be values to make this explicit.

leewujung: Also, I think once we get around to put this function into the documentation, we should link this explicitly!

echopype/echodata/echodata.py
```diff
@@ -312,6 +312,145 @@ def squeeze_non_scalar(n):
...
+            "long_name"
+        ] = "time dimension for external platform data"
+
+        dropped_vars = []
+        for var in ["pitch", "roll", "heave", "latitude", "longitude", "water_level"]:
+            if var in platform and platform[var].isnull().all():
+                dropped_vars.append(var)
+        if len(dropped_vars) > 0:
+            warnings.warn(
+                f"some variables will be overwritten by platform data: {', '.join(dropped_vars)}"
```
leewujung, line 386: Suggesting minor wording changes to clarify which data variables will be overwritten:
`f"some variables in the original Platform group will be overwritten: ...`

echopype/echodata/echodata.py
imranmaj: > users may think that a list with each element being an array of values corresponding to all of what's listed in keys would be generated, while in fact only the first existing key will be returned.

It should only be used internally, I don't believe users can use it.

> Which way do you intend this to be? A function that only returns data from 1 variable from all of the possible names, or a function that would return data from multiple variables?

It should return data from 1 variable, it's "searching" for multiple possible names to be flexible.

> the function name should not be `mapping_get_multiple`, because it would only get data from a single variable.

How about `mapping_try_get_multiple`?

echopype/echodata/echodata.py
```diff
@@ -312,6 +312,145 @@ def squeeze_non_scalar(n):
...
+            time2=extra_platform_data[time_dim].values
+        )
+        platform["time2"].attrs[
+            "long_name"
+        ] = "time dimension for external platform data"
+
+        dropped_vars = []
+        for var in ["pitch", "roll", "heave", "latitude", "longitude", "water_level"]:
+            if var in platform and platform[var].isnull().all():
+                dropped_vars.append(var)
```
leewujung, line 383: The second part of the logic seems reversed? 
With `platform[var].isnull().all()`, variables with actual non-NaN values in it will return `False` and won't be included in `dropped_vars`.

In the current test, all 6 variables listed contains a single NaN and were included in the dropped list and then replaced. The replacement part is the intended behavior though.

echopype/echodata/echodata.py
leewujung: > It should only be used internally, I don't believe users can use it.

I should rephrase this to say that it is not obvious to _others_.

> How about mapping_try_get_multiple?

I think the "multiple" part could be confused with multiple variables. How about `mapping_search_variable`?

leewujung:
Thanks for the changes @imranmaj !  We can now merge this 🙂

echopype/echodata/echodata.py
emiliom: > @emiliom: reading this section you pointed out makes me wonder whether or not it is worthwhile to point out that the lat/lon data in the SONAR-netCDF4 ver.1 is consistent with the H.4.2. Single trajectory specification. Not that this necessitates any actual changes, but there seems to be values to make this explicit.

Late reply ... Interesting observation. I agree, the structure is consistent. Maybe the right place to add this observation is the SONAR-netCDF4 version 2 draft text. From what Gavin said, it doesn't look like there are changes in the Platform group in v.2. 

> Also, I think once we get around to put this function into the documentation, we should link this explicitly!

Definitely.

gavinmacaulay: The consistency with the CF single trajectory specification was deliberate :) We tried to use CF conventions whenever it was sensible (and did attempt to meet CF-1.7, although it wasn't feasible for all of sonar-netcdf4 v1).

emiliom: @gavinmacaulay thanks for taking the time to chime in! I've gone ahead and created a new issue in the SONAR-netCDF4 repo. I'll try to handle it myself, hopefully next week.

PR #436, opened 2021-08-30
Add simple plotting functionality

lsetiawan:
## Overview

This PR adds an initial simple plotting functionality by calling `ep.visualize.create_echogram`.

**NOTE: This is still a draft, suggestions are highly welcomed and encouraged.**

## Example

Screenshot from 2021-08-30 15-42-14

## Actions
- [X] If user passes in EchoData, plot the backscatter_r in beam group.
- [X] If user passes MVBS / Sv, then plot those directly.
- [x] Ability to plot EK80
- [x] Ability to plot EK60
- [x] Ability to plot AZFP
- [x] ~~Ability to plot ADCP~~ (Currently not supported at this moment).
- [x] Add docstring to function
- [x] ~~Add actual plot comparison test?~~ (Maybe future improvements, not sure how to do this at this moment).
- [x] Add a custom EK500 colormap along with the cmocean colormaps.

## Notes

There are still some quirks and improvements to add, and some questions to be discussed such as:
1. How to plot the EK80 quadrant dims?
2. `Backscatter_r` is not available in ADCP, at least the one on the test data. In this case, what to do?
3. How do I utilize the water_level to calibrate the range?
4. How to improve the ease of the function call?


lsetiawan:
@leewujung Do you know if there's a better file for me to develop the visualization with that's pretty? 😃

emiliom:
TODO: Investigate what platform or sonar/beam metadata attribute exists or should be added that clearly specifies the "direction" the echosounder is oriented (eg, downward vs upward looking). This need came about in the context of the OOI EK60/80 sensor @lsetiawan is using in his plotting tests, which is upward looking. This functionality would allow the echogram plots of raw data to be automatically oriented correctly with a depth y axis.

Looking through the SONAR-netCDF4 convention, the `Sonar/Beam` `beam_mode` attribute is helpful, but not fully what's needed; its values can be "vertical", "horizontal" and "inspection". `Platform` variables involving orientation and coordinate system may be usable, too, but currently the convention uses a ship-based frame of reference.

gavinmacaulay:
The upcoming version of SONAR-netCDF4 has a more complete orientation and coordinate system that allows for specifying absolute beam pointing directions - details [here](https://gitcdn.link/cdn/ices-publications/SONAR-netCDF4/a75fc215a7b965f833e1d85d5443934e6576f4eb/Formatted_docs/crr341.html#_coordinate_systems). It was designed for multi-beam bathymetric sonars and might be a bit complicated for just upward-looking echosounders, but was intended to be general enough for all purposes and does not assume a ship-based frame of reference (keep in mind that the _platform_ is not necessary a vessel).

The beam_mode in version 1 of SONAR-netCDF4 is a label for the different groups of beams that fisheries omni-directional sonars have (e.g., a set of horizontal beams, a set of vertical beams, and spot beams - the word inspection comes from the usage in such sonars - a narrow beam for inspecting parts of a fish school)

leewujung:
Great! please add EK500 colormap as one of the options. :)

emiliom:
@gavinmacaulay thanks so much for chiming in to elaborate on the new coordinate system framework and the context of the `beam_mode` attribute! I'll look into the coordinate system description more closely.

lsetiawan:
## EK500 Example

Screenshot from 2021-09-20 16-42-05

leewujung:
For quick visualization of EK80 complex data, we can plot the magnitude 
```python
np.abs(ed.beam.backscatter_r + 1j * ed.beam.backscatter_i)
```
of each quadrant separately.

Can have an option doing a coherent sum over all quadrants
```python
(ed.beam.backscatter_r + 1j * ed.beam.backscatter_i).sum(dim="quadrant")
```
and then do the magnitude.

echopype/visualize/api.py
```diff
@@ -0,0 +1,123 @@
...
+        if isinstance(water_level, xr.DataArray):
+            required_dims = ['frequency', 'ping_time']
+            if not all(
+                True if d in water_level.dims else False for d in required_dims
+            ):
+                raise ValueError(
+                    f"Water level must have dimensions: {', '.join(required_dims)}"
+                )
+            # Adds water level to range if it exists
+            ds['range'] = ds.range + water_level
```
leewujung, line 111: I think water level can be the following:
- a single number: this is often the case when all the transducers are mounted on a single plate on a ship or a mooring (in generally good weather so there's no significant change of depth).
- only have a single dimension `frequency`: in this case different transducers are mounted at different depths, but there was no change across `ping_time`
- only have a single dimension `ping_time`: in this case all transducers are mounted on the same plate, but there may be variations of depth for the entire platform (think about the case when you have an AUV with an echosounder on it)
- have both dimensions `frequency` and `ping_time` like what you have here

All of these should add to `ds.range` without problem since the dimension would be broadcast.

echopype/visualize/api.py
```diff
@@ -0,0 +1,123 @@
...
+    water_level : xr.DataArray, optional
+        Water level data array for platform water level correction.
+    **kwargs: optional
+        Additional keyword arguments for xarray plot pcolormesh.
+
+    """
+    range_attrs = {
+        'long_name': 'Range',
+        'units': 'm',
+    }
```
leewujung, line 41: Does CF convention has a long_name for range? It may be worth checking since the same wording/concept is applicable to radar data also.

echopype/visualize/api.py
```diff
@@ -0,0 +1,123 @@
...
+            range_in_meter = data.compute_range(
+                env_params=range_kwargs.get('env_params', {}),
+                azfp_cal_type=range_kwargs.get('azfp_cal_type', None),
+                ek_waveform_mode=range_kwargs.get('waveform_mode', None),
+                ek_encode_mode=range_kwargs.get('encode_mode', 'complex'),
+            )
+            range_in_meter.attrs = range_attrs
+            if 'water_level' in data.platform:
+                # Adds water level to range if it exists
+                range_in_meter = range_in_meter + data.platform.water_level
```
leewujung, line 85: This implementation is to add water_level automatically under-the-hood. 

I think it would be better to allow `water_level` to also be of `bool` type, and when True use the water_level stored in the data file when it exists, and otherwise display range **without** automatically adding water_level.

echopype/visualize/api.py
```diff
@@ -0,0 +1,123 @@
...
+                if 'env_params' not in range_kwargs:
+                    raise ValueError(
+                        "Please provide env_params in range_kwargs!"
+                    )
+            elif data.sonar_model.lower() == 'ek60':
+                if 'waveform_mode' not in range_kwargs:
+                    range_kwargs['waveform_mode'] = 'CW'
+
+                if 'encode_mode' not in range_kwargs:
+                    range_kwargs['encode_mode'] = 'power'
```
leewujung, line 69: EK60 should always be `waveform="CW"` and `encode_mode="power"` — anything else should raise an error

echopype/visualize/api.py
```diff
@@ -0,0 +1,123 @@
...
+            elif data.sonar_model.lower() == 'ek80':
+                if 'waveform_mode' not in range_kwargs:
+                    raise ValueError(
+                        "Please provide waveform_mode in range_kwargs for EK80 sonar model."
+                    )
```
leewujung, line 74: for EK80 we should force users to always provide `waveform_mode ({“BB”, “CW”})` and `encode_mode ({“complex”, ”power”})`. See doctoring in the #477.

leewujung:
@lsetiawan : Thanks for the PR! Excited to finally got around to review this! 😄  I added comments above. Let me know if anything is unclear.

One more thing: 
I think we should default `create_echogram` when data is Sv to have `get_range=True` automatically, so that things are more congruent (all has physically meaningful units), unless user specifically wants to check some potential problems in the data when needs to use `get_range=False` in an optional case.

echopype/visualize/api.py
emiliom: I don't see any "range" CF `standard_name` entry that are relevant for sonars. See http://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html and type "range" in the text box.

echopype/visualize/api.py
emiliom: "doctoring": heh, you probably meant "docstring"

echopype/visualize/api.py
lsetiawan: Yea the only "range" i see is `altimeter_range` which doesn't really makes sense in this case.

echopype/visualize/api.py
```diff
@@ -80,7 +125,7 @@ def create_echogram(
                 ek_encode_mode=range_kwargs.get('encode_mode', 'complex'),
             )
             range_in_meter.attrs = range_attrs
-            if 'water_level' in data.platform:
+            if water_level is True and 'water_level' in data.platform:
                 # Adds water level to range if it exists
                 range_in_meter = range_in_meter + data.platform.water_level
```
leewujung: Right now the `water_level` is int, float, or xr.DataArray cases do not work with when `get_range=True` and data is an `EchoData` object, since this only allows pulling the water_level values from the Platform group. 

It seem that the block you have from below would work for both the data is Echodata and data is Sv cases and can largely be repurposed?

echopype/tests/visualize/test_plot.py
```diff
@@ -172,6 +172,6 @@ def test_plot_mvbs(
         if 'azfp_cal_type' in range_kwargs:
             range_kwargs.pop('azfp_cal_type')
     Sv = echopype.calibrate.compute_Sv(ed, **range_kwargs)
-    mvbs = echopype.preprocess.compute_MVBS(Sv)
-    # plot = echopype.visualize.create_echogram(mvbs)
-    # assert isinstance(plot, FacetGrid) is True
+    mvbs = echopype.preprocess.compute_MVBS(Sv, ping_time_bin='1S')
```
leewujung, line 175: How about if we default to 10s here? ping_time interval at 1S is a common setting so we're not reducing much with ping_time_bin='1S'

leewujung:
@lsetiawan : thanks for the changes, they look great! I put in a couple small comments that should pretty easy to address.

Do you think you could add a test for the water_level options? With all the possibilities it seems a good idea to have a test to catch any potential hiccups.

echopype/visualize/api.py
lsetiawan: Ah I see, so you'd want user to be able to pass in their on water level also on the case with EchoData object?

leewujung: Yes! since there is no guarantee that the water_level saved in the data is correct, and it is often not populated.

echopype/tests/visualize/test_plot.py
lsetiawan: I put it as '1S' because the data I'm currently using is very small, and I end up having only 1 point, and the plotting doesn't work.

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+    )
+
+    range_in_meter = echodata.compute_range(
+        env_params=range_kwargs.get('env_params', {}),
+        azfp_cal_type=range_kwargs.get('azfp_cal_type', None),
+        ek_waveform_mode=range_kwargs.get('waveform_mode', 'CW'),
+        ek_encode_mode=range_kwargs.get('encode_mode', 'power'),
+    )
+    single_array = range_in_meter.isel(frequency=0, ping_time=0).values
+    no_water_level = False
```
leewujung, line 221: suggest changing this to `no_input_water_level` to clarify what this does.

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+    elif isinstance(water_level, bool) and water_level is True:
+        if no_water_level is False:
+            original_array = (
+                single_array
+                + echodata.platform.water_level.isel(frequency=0, ping_time=0).values
+            )
+        else:
+            original_array = single_array
+    elif isinstance(water_level, (int, float)):
+        original_array = single_array + water_level
```
leewujung, line 239: apparently `isinstance(False, int)` returns True! so the (False, True) combination runs through this. The original_array is single_array at the output...

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+                platform_data=echodata.platform,
+            )
+    except Exception as e:
+        assert isinstance(e, ValueError)
+        assert str(e) == 'Water level must have any of these dimensions: frequency, ping_time, range_bin'  # noqa
+
+    if isinstance(results, xr.Dataset):
+        final_array = results.isel(frequency=0, ping_time=0).values
+
+        assert np.array_equal(original_array, final_array)
```
leewujung, line 267: the results here are xr.DataArray so this assert is not being run. Seems a bug?

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+    [
+        (True, True),
+        (False, True),
+        (xr.DataArray(np.array(50.0)).expand_dims({'frequency': 3}), False),
+        (xr.DataArray(np.array(50.0)), False),
+        (10, False),
+        (30.5, False),
+    ],
+)
+def test_water_level_dataset(water_level, expect_warning):
```
leewujung, line 281: suggest changing this to `test_water_level_Sv_dataset` to specify what the dataset is.

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+                data_type=xr.Dataset,
+            )
+    except Exception as e:
+        assert isinstance(e, ValueError)
+        assert str(e) == 'Water level must have any of these dimensions: frequency, ping_time, range_bin'  # noqa
+
+    if isinstance(results, xr.Dataset):
+        final_array = results.isel(frequency=0, ping_time=0).values
+
+        assert np.array_equal(original_array, final_array)
```
leewujung, line 326: similar to the above the `results` here is `xr.DataArray` so the assert here is not being run through.

leewujung (commented):
Hey @lsetiawan : Thanks for the changes. I found a couple bugs in the water_level tests and added a couple suggestions. Once they are fixed this should be good to go!

echopype/tests/visualize/test_plot.py
lsetiawan:
```suggestion
    no_input_water_level = False
```

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+        azfp_cal_type=range_kwargs.get('azfp_cal_type', None),
+        ek_waveform_mode=range_kwargs.get('waveform_mode', 'CW'),
+        ek_encode_mode=range_kwargs.get('encode_mode', 'power'),
+    )
+    single_array = range_in_meter.isel(frequency=0, ping_time=0).values
+    no_water_level = False
+    if isinstance(water_level, list):
+        water_level = water_level[0]
+        echodata.platform = echodata.platform.drop_vars('water_level')
+        no_water_level = True
```
lsetiawan, line 225:
```suggestion
        no_input_water_level = True
```

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+
+    if isinstance(water_level, xr.DataArray):
+        if 'frequency' in water_level.dims:
+            original_array = single_array + water_level.isel(frequency=0).values
+    elif isinstance(water_level, bool) and water_level is True:
+        if no_water_level is False:
```
lsetiawan, line 231:
```suggestion
        if no_input_water_level is False:
```

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+            original_array = (
+                single_array
+                + echodata.platform.water_level.isel(frequency=0, ping_time=0).values
+            )
+        else:
+            original_array = single_array
+    elif isinstance(water_level, (int, float)):
```
lsetiawan, line 238:
```suggestion
    elif not isinstance(water_level, bool) and isinstance(water_level, (int, float)):
```

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+                range_in_meter=range_in_meter,
+                water_level=water_level,
+                data_type=EchoData,
+                platform_data=echodata.platform,
+            )
+    except Exception as e:
+        assert isinstance(e, ValueError)
+        assert str(e) == 'Water level must have any of these dimensions: frequency, ping_time, range_bin'  # noqa
+
+    if isinstance(results, xr.Dataset):
```
lsetiawan, line 264:
```suggestion
    if isinstance(results, xr.DataArray):
```

echopype/tests/visualize/test_plot.py
lsetiawan:
```suggestion
def test_water_level_Sv_dataset(water_level, expect_warning):
```

echopype/tests/visualize/test_plot.py
```diff
@@ -175,3 +177,142 @@ def test_plot_mvbs(
...
+            results = _add_water_level(
+                range_in_meter=range_in_meter,
+                water_level=water_level,
+                data_type=xr.Dataset,
+            )
+    except Exception as e:
+        assert isinstance(e, ValueError)
+        assert str(e) == 'Water level must have any of these dimensions: frequency, ping_time, range_bin'  # noqa
+
+    if isinstance(results, xr.Dataset):
```
lsetiawan, line 323:
```suggestion
    if isinstance(results, xr.DataArray):
```

echopype/tests/visualize/test_plot.py
lsetiawan, line 238:
```suggestion
    elif water_level is not False and isinstance(water_level, (int, float)):
```

echopype/tests/visualize/test_plot.py
```diff
@@ -0,0 +1,326 @@
...
+    )
+    Sv = echopype.calibrate.compute_Sv(echodata, **range_kwargs)
+    ds = Sv.set_coords('range')
+    range_in_meter = ds.range
+    single_array = range_in_meter.isel(frequency=0, ping_time=0).values
+
+    if isinstance(water_level, xr.DataArray):
+        if 'frequency' in water_level.dims:
+            original_array = single_array + water_level.isel(frequency=0).values
+    elif isinstance(water_level, (int, float)):
```
lsetiawan, line 299:
```suggestion
    elif not isinstance(water_level, bool) and isinstance(water_level, (int, float)):
```

leewujung:
@lsetiawan : thanks, the changes look great! I'll merge this after adding in some small notes to the `create_echogram` docstring due to the range vs depth stuff below.

One thing I noticed which we'll need to do something about (also requested in #259 ) is to include the addition of water_level in the Sv dataset (since right now even if users put in `water_level=True` if the input is an Sv dataset no water_level offset is added.  A more accurate way to think about it is probably that we are actually plotting "depth" in the case when we have range_meters + water_level.

The more general description is that people are plotting the range _offset_ from some platform. Let's discuss how that coordinate should be named...  Hence I am wondering if `water_level` should be `range_offset` instead, so that it is more general. 

Pinging @emiliom here since this is related to the whole convention and attribute conversation.

leewujung:
To check how the EK500 colormap looks like (as in this [previous comment](https://github.com/OSOceanAcoustics/echopype/pull/436#issuecomment-923438537)), use `matplotlib.colormaps["ep.ek500"]`.

From a recent deprecation warning: 
> The get_cmap function was deprecated in Matplotlib 3.7 and will be removed two minor releases later. Use `matplotlib.colormaps[name]` or `matplotlib.colormaps.get_cmap(obj)` instead.

PR #438, opened 2021-09-10
AD2CP conversion bug fixes

imranmaj:
* Fixes a bug where strings were not checked for in the config string data record
* Fixes a bug where AHRS variables were not indexes by `ping_time`


leewujung:
Verified that with this fix now we can parse the following without problem:
- raw data from the regular 5-beam Sgi1000
- raw data from the glider-specific 4-beam Sgi1000
- telemetry data from the glider 

Thanks very much @imranmaj !! 😃

leewujung (approved):

PR #446, opened 2021-09-29
Gain interpolate

### LINKED ISSUE #384: Band-averaged broadband calibration needs to use actual gain from file
This is most likely a leftover while we developing the code, that the gain for band-averaged broadband calibration (for pulse compressed outputs) for EK80 now actually hard codes the gain to be 27, instead of pulling it from the data file. There is also a TODO for generating the gain at the center frequency in the actual transmit signal band.

Code section: 
https://github.com/OSOceanAcoustics/echopype/blob/master/echopype/calibrate/calibrate_ek.py#L526-L527

Let's get this fixed very soon.

mbdunn:
Interpolate gain to center frequency from calibrated values in BB, when available
resolves #384


gavinmacaulay:
I have added three ek80 files to the google drive: CI_test_data/ek80_bb_with_calibration.

All files contain broadband data and calibrations. One of the files has two channels (both broadband). The other two have 6 channels, but a variable number of broadband channels (all channels have an included calibration, broadband or narrowband). In those two files some of the channels are not transmitting (so the echograms look odd).

lsetiawan (approved):
This LGTM @leewujung. I agree that we should parametrize the test files in the near future. Thanks @mbdunn for your contribution on this. 🎉

PR #447, opened 2021-09-29
Add PathNotFoundError in _load_convert for open_converted

leewujung:
`open_converted` failed on opening Zarr file without the `beam_power` group. The allowed exception `GroupNotFoundError` appears to only work for netCDF files but not with Zarr files:
```python
PathNotFoundError: nothing found at path 'Beam_power'
```
Adding `PathNotFoundError` resolved the problem.

I also tested the case to only use `PathNotFoundError` _without_ `GroupNotFoundError` and that works for both local netCDF and Zarr files. 

From the zarr code here
https://github.com/zarr-developers/zarr-python/blob/master/zarr/errors.py
`GroupNotFoundError` is for when "group not found at path {0!r}"
and 
`PathNotFoundError` is for when "nothing found at path {0!r}"

@imranmaj @lsetiawan : If I understand correctly right now the `beam_power` group is _always_ created regardless of the sonar_model, but may be empty. So, there will be no case with exception `GroupNotFoundError`. Is this correct?

The current PR replaces `GroupNotFoundError` with `PathNotFoundError`.

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING, Dict, Optional
 import fsspec
 import numpy as np
 import xarray as xr
-from zarr.errors import GroupNotFoundError, PathNotFoundError
+from zarr.errors import PathNotFoundError
 
 if TYPE_CHECKING:
     from ..core import EngineHint, FileFormatHint, PathHint, SonarModelsHint
@@ -472,7 +472,7 @@ class EchoData:
                     raw_path,
                     group=value["ep_group"],
                 )
-            except (OSError, GroupNotFoundError, PathNotFoundError):
+            except (OSError, PathNotFoundError):
                 # Skips group not found errors for EK80 and ADCP
                 ...
             if group == "top" and hasattr(ds, "keywords"):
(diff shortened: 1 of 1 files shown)


lsetiawan:
>  If I understand correctly right now the beam_power group is always created regardless of the sonar_model, but may be empty

I don't remember if it's `always` created. Kavin did a lot of those group creation code, so I do not know. :/

leewujung:
> I don't remember if it's always created. Kavin did a lot of those group creation code, so I do not know. :/

Right now on initialization EchoData this is done:
https://github.com/OSOceanAcoustics/echopype/blob/d9405a4ba7c4c8863f2ebe54abb8d8663e5d2cf0/echopype/echodata/echodata.py#L113-L115
which pulls all keys in group_map created from [_get_convention](https://github.com/OSOceanAcoustics/echopype/blob/d9405a4ba7c4c8863f2ebe54abb8d8663e5d2cf0/echopype/echodata/convention/__init__.py#L8)
https://github.com/OSOceanAcoustics/echopype/blob/d9405a4ba7c4c8863f2ebe54abb8d8663e5d2cf0/echopype/echodata/echodata.py#L37
I think the latter is you 😄 , but it doesn't matter. 

Unless there's anything missing in this sequence, I am pretty sure this is correct.

leewujung:
Although, from the CI test failure, it seems that things may be handled differently on different platforms re what it means to have/no have a group, and whether there is anything in the path or group... I'll try adding GroupNotFoundError back in.

Issue #381, opened 2021-06-14
Matching time or location of env_params used in calibration


leewujung:
Currently the `cal_params` can be specified by user when calling calibration functions, but the values are restricted to be static values. It is a common requirements when analyzing data that we use the actual CTD data for the temperature, salinity, and depth to calibrate the data. The CTD data can be aligned with time or location (or both) with the acoustic data. We should be able to allow `cal_params` to be passed in as an xarray Dataset that can be automatically aligned during calibration.

leewujung:
@imranmaj : I'll get some actual CTD data from the hake survey data for this.

leewujung:
In terms of functionalities, the env parameters could probably come in in a few forms:
1. In the case of a mooring where we may have CTD profiles at specific times, we would want to align the env parameters along the `ping_time` coordinate by a 1D interpolation.
2. In the case of a survey using mobile platforms like ships, we probably want to default to aligning the the env parameters with the acoustic data along with the platform lat/lon. The lat/lon are likely data variables themselves with a time coordinate. The alignment would be interpolation in 2D using scatter data (say each scatter point are a collection CTD data at a specific lat/lon).
3. In the case of an organized, gridded CTD data product with 3 coordinates in lat/lon/time. The alignment would be interpolation in 3D.

@imranmaj : what do you think about putting in support for case 1 first since it is the most straightforward, and then we can work out the path to support case 2 & 3?

imranmaj:
@leewujung case 3 hasn't been implemented yet, should we keep this issue open for that case?

leewujung:
@imranmaj : Good point! How about if we (you 😁) open up another issue to keep track of that?

PR #464, opened 2021-10-28
Add support for interpolating CTD data in calibration

imranmaj:
Part of #381, but does not include support for the `"organized"` case

* Adds a class `EnvParams` that can be used as the `env_params` parameter for calibration
    * Dictionaries of values are still valid `env_params`
    * `EnvParams` organizes dimensional CTD data and interpolates it with the echosounder data used in calibration
* Adds a test for `EnvParams` in the `"stationary"` and `"mobile"` cases

Todo:

- [x] Docstrings
- [x] Tests


echopype/calibrate/calibrate_base.py
```diff
@@ -7,13 +13,158 @@
...
+                    )
+                )
+                interp = scipy.interpolate.griddata(
+                    points, values, xi, method=self.interp_method
+                )
+                result[var] = ("ping_time", interp)
+            env_params = xr.Dataset(
+                data_vars=result, coords={"ping_time": interp_plat["ping_time"]}
+            )
+        else:
```
leewujung, line 84: Since "organized" is not supported for now, this is really just the "stationary" case, right?
If this is the case, the iterations like `for dim in dims` below would be redundant, since there is only 1 dimension `ping_time`.
Am I missing anything?
Could you clean up the code so that this intention is more explicit? or are you planning on reusing some of this for the "organized" case? down the line?

echopype/calibrate/calibrate_base.py
```diff
@@ -7,13 +13,158 @@
...
+            extrap = env_params.interp(
+                {dim: echodata.platform[dim].data for dim in dims},
+                method=self.extrap_method,
+                # scipy interp uses "extrapolate" but scipy interpn uses None
+                kwargs={"fill_value": "extrapolate" if len(dims) == 1 else None},
+            )
+            extrap_unique_idx = {
+                dim: np.unique(extrap[dim], return_index=True)[1] for dim in dims
+            }
+            extrap = extrap.isel(**extrap_unique_idx)
```
leewujung, line 99: Why is this needed? Could you put in more comments to explain this? Thanks!

echopype/calibrate/calibrate_base.py
```diff
@@ -7,13 +13,158 @@
...
+                    }
+                )
+
+                # remove zero length dims (xarray does not support in combine_by_coords)
+                non_zero_dims = [
+                    ds
+                    for ds in (less, middle, greater)
+                    if all(dim_len > 0 for dim_len in ds.dims.values())
+                ]
+                env_params = xr.combine_by_coords(non_zero_dims)
```
leewujung, line 140: Zero length dimension -- is this when one of `less`, `middle`, `greater` is empty?
Could you put in a short comment to clarify?

echopype/calibrate/calibrate_base.py
```diff
@@ -7,13 +13,158 @@
...
     """Class to handle calibration for all sonar models."""
 
-    def __init__(self, echodata):
+    def __init__(self, echodata, env_params=None):
         self.echodata = echodata
-        self.env_params = None  # env_params are set in child class
+        if isinstance(env_params, EnvParams):
+            env_params = env_params._apply(echodata)
+        elif env_params is None:
+            env_params = {}
```
leewujung, line 166: Missing an `else` case to catch other possible inputs?

echopype/echodata/combine.py
```diff
@@ -118,6 +118,7 @@ def combine_echodata(echodatas: List[EchoData], combine_attrs="override") -> Ech
             )
         elif sonar_model is None:
             sonar_model = echodata.sonar_model
+            result.sonar_model = sonar_model
```
leewujung, line 121: in #437 this was at the end of `combine_echodata`. Which one should we use?

echopype/calibrate/calibrate_base.py
```diff
@@ -1,4 +1,10 @@
 import abc
+from typing import Dict, List, Optional
+
+import numpy as np
+import scipy.interpolate
+import xarray as xr
+from typing_extensions import Literal
 
 ENV_PARAMS = ("temperature", "salinity", "pressure", "sound_speed", "sound_absorption")
```
leewujung, line 9: Are these still used somewhere?

echopype/echodata/combine.py
leewujung: Would be a good idea to merge `dev` to make sure other changes are in here also.

echopype/calibrate/calibrate_base.py
imranmaj: Yes I was planning to use this for the `"organized"` case

echopype/echodata/combine.py
imranmaj: I deleted mine, I forgot this was fixed already

echopype/calibrate/calibrate_base.py
imranmaj: I think this is just for converting `env_params` into a dictionary format. I believe the accepted values are `EnvParams`, `None`, or a dictionary, so if it isn't `EnvParams` or `None` then it should be a dictionary which is OK

leewujung: What if people put in `env_params=""` ?

imranmaj: Is that something they can do? I didn’t think that was allowed

leewujung: I was thinking that people won't necessary read the docstring, so it would be useful to a thing to catch those and print out an error message. Maybe I missed something?

echopype/calibrate/calibrate_base.py
```diff
@@ -1,19 +1,216 @@
...
+        ----------
+        env_params : xr.Dataset
+            The environmental parameters to use for calibration. This data will be interpolated with
+            a provided `EchoData` object.
+
+            When `data_kind` is `"stationary"`, env_params must have a coordinate `"ping_time"`.
+            When `data_kind` is `"mobile"`, env_params must have coordinates `"latitude"`
+            and `"longitude"`.
+            When `data_kind` is `"mobile"`, env_params must have coordinates `"time"`,
+            `"latitude"`, and `"longitude"`.
```
leewujung, line 50: Typo: this is the "organized" case?

echopype/calibrate/calibrate_base.py
```diff
@@ -1,19 +1,216 @@
...
+            The type of the environmental parameters.
+
+            `"stationary"`: environmental parameters from a fixed location
+            (for example, a single CTD).
+            `"mobile"` environmental parameters from a moving location (for example, a ship).
+            `"organized"`: environmental parameters from many fixed locations
+            (for example, multiple CTDs).
+        interp_method: {"linear", "nearest", "zero", "slinear", "quadratic", "cubic"}
+            Method for interpolation of environmental parameters with the data from the
+            provided `EchoData` object.
```
leewujung, line 61: Link to `scipy.interpolate` and point users to see details of the interp methods there.

echopype/calibrate/calibrate_base.py
leewujung: Also it would be good add that the "organized" case is not supported yet -- it would be redundant to what you have below, but many people don't read docs carefully...

echopype/calibrate/calibrate_base.py
leewujung: Just noticed this again -- could you add a TODO here so that the intention is clear?

echopype/calibrate/calibrate_base.py
```diff
@@ -1,19 +1,216 @@
...
+        else:
+            raise ValueError("invalid data_kind")
+
+        env_params = self.env_params
+
+        if self.data_kind == "mobile":
+            # compute_range needs indexing by ping_time
+            interp_plat = echodata.platform.interp(
+                {"location_time": echodata.platform["ping_time"]}
+            )
```
leewujung, line 103: This should use `ping_time` from the Beam group since ultimately we want the env variables to be aligned with the acoustic data, not the GPS data.

echopype/tests/calibrate/test_calibrate.py
```diff
@@ -218,3 +219,47 @@ def test_compute_Sv_ek80_CW_complex_BB_complex():
...
+        coords={
+            "ping_time": np.arange("2017-06-20T01:00", "2017-06-20T01:50", np.timedelta64(30, "s"), dtype="datetime64[ns]")
+        }
+    )
+    env_params = EnvParams(env_params_data, "stationary")
+    converted_env_params = env_params._apply(ed)
+    for var in converted_env_params.values():
+        assert np.all(np.diff(var) > 0)
+        assert np.all(0 <= var)
+        assert np.all(var < 100)
```
leewujung, line 246: These test for the range of values to see if they are reasonable. It will be better to test for the exact interpolated values so that we know things are really done correctly. To avoid potential problems with numerical precisions, it may be helpful to substitute the exact ping_time you need to interpolate to.

echopype/tests/calibrate/test_calibrate.py
```diff
@@ -218,3 +219,47 @@ def test_compute_Sv_ek80_CW_complex_BB_complex():
...
+        coords={
+            "latitude": ("time", rng.random(size=100) + 44),
+            "longitude": ("time", rng.random(size=100) - 125),
+        }
+    )
+    env_params = EnvParams(env_params_data, "mobile")
+    converted_env_params = env_params._apply(ed)
+    for var in converted_env_params.values():
+        assert np.all(0 <= var[~np.isnan(var)])
+        assert np.all(var[~np.isnan(var)] < 100)
```
leewujung, line 265: Same as the above: you should test for the actual interpolated values, not just the range of values.

echopype/tests/calibrate/test_calibrate.py
```diff
@@ -218,3 +219,47 @@ def test_compute_Sv_ek80_CW_complex_BB_complex():
     assert isinstance(ds_Sv, xr.Dataset)
     ds_Sv = ep.calibrate.compute_Sv(ed, waveform_mode="BB", encode_mode="complex")
     assert isinstance(ds_Sv, xr.Dataset)
+
+def test_env_params():
+    """
+    Tests EnvParams interpolation
```
leewujung, line 225: Please also add tests for the extrapolation part of data. You can force that by substituting the ping_time to extrapolate to also.

leewujung:
Thanks @imranmaj for the edits! I put in additional comments and requested changes for the docstring and code, and also a few things to change for the tests. It'll be great for us to discuss this and the new EnvParams class as a group in the next meeting. 😄

PR #471, opened 2021-11-01
Update zenodo link/badge to use general doi

### LINKED ISSUE #469: Zenodo badge leads to a very old version (0.4.1).
The current zenodo DOI leads to a very old version of 0.4.1. Rather than the latest 0.5.4. This needs to be changed. Echopype does have a `concept doi` that leads to the latest version of the software everytime. Maybe we can use that instead, or create a way to update the zenodo badge at each release.

**Zenodo Concept DOI**

<a href="https://doi.org/10.5281/zenodo.3906999">
    DOI
</a>

leewujung:
This resolves #469


leewujung:
Oops this should point to `main`. Updating it.

leewujung:
Actually, I just made a mess... it should be patch from `main` to PR to `main`... let me just redo it... :(

PR #472, opened 2021-11-01
update zenodo link and badge to use general doi

### LINKED ISSUE #469: Zenodo badge leads to a very old version (0.4.1).

leewujung:
This resolves #469

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
       - id: black
 
   - repo: https://github.com/PyCQA/flake8
-    rev: 3.9.2
+    rev: 4.0.1
     hooks:
       - id: flake8
 
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 # Echopype
 
 <div>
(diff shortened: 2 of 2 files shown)


lsetiawan (approved):

leewujung:
Ok, so the RTD problem would cause the build to fail on this one, but this would become quite messy if we start to pull later changes into `main` this way. Should we just wait until the next release?

lsetiawan:
If you merge this now, then the only place that this zenodo DOI will be shown is the github repo README. Maybe we should just wait till the next release or just point this to dev.

leewujung:
Yeah I think pointing this to dev and wait for the next release would be the way to go. People (hopefully!) are not installing by downloading the zip file from Zenodo...!

PR #474, opened 2021-11-04
Pinning s3fs to avoid the aiobotocore dependency issues

lsetiawan:
## Overview

Seems like this is needed for s3fs. Currently, the dependency resolution in inconsistent, across installs, so I think it's better for us to pin the s3fs version, which pins the aiobotocore and fsspec.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ scipy
 xarray
 zarr
 fsspec
-s3fs
+s3fs==2021.10.1
 requests
 aiohttp
 typing-extensions~=3.10


lsetiawan:
Connecting this to https://github.com/OSOceanAcoustics/echopype-examples/pull/17

lsetiawan:
This issue can be seen on echopype install test: https://github.com/OSOceanAcoustics/echopype/runs/4099611300?check_suite_focus=true#step:4:47

emiliom:
Looks good. Plus, I just created a conda env locally with the s3fs pinning included in this PR, and I confirmed it had the desired effect.

emiliom (approved):

PR #476, opened 2021-11-04
EchoData.update_platform: use location_time instead of time2; clip location_time range to EchoData.beam["ping_time"]; created utils/coding.py to centralize encodings

### LINKED ISSUE #460: Clip external platform data to ping_time range when using update_platform; and change time2 to location_time
The external platform data passed to `echodata.update_platform` may span a time range greater than the one in the echodata object being operated on. Currently, all the external data will be added ([echodata/echodata.py#L319](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/echodata/echodata.py#L319)). I think the "extra data" is unnecessary and potentially confusing. I suggest that it'd be best to clip the time range of the external platform data based on the time range in the beam `ping_time`

imranmaj:
* Uses `location_time` instead of `time2` as the index for the incoming platform data in `EchoData.update_platform`
* Limits the range of the incoming `location_time` to between the minimum and maximum values of `EchoData.beam["ping_time"]`

Fixes #460

### DIFF
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -6,7 +6,8 @@ import numpy as np
 import xarray as xr
 from _echopype_version import version as ECHOPYPE_VERSION
 
-from .set_groups_base import DEFAULT_CHUNK_SIZE, SetGroupsBase, set_encodings
+from ..utils.coding import set_encodings
+from .set_groups_base import DEFAULT_CHUNK_SIZE, SetGroupsBase
 
 
 class SetGroupsEK60(SetGroupsBase):
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -375,6 +375,15 @@ class EchoData:
             extra_platform_data = extra_platform_data.drop_vars(trajectory_var)
             extra_platform_data = extra_platform_data.swap_dims({"obs": time_dim})
 
+        extra_platform_data = extra_platform_data.sel(
+            {
+                time_dim: np.logical_and(
+                    self.beam["ping_time"].min() < extra_platform_data[time_dim],
+                    extra_platform_data[time_dim] < self.beam["ping_time"].max(),
+                )
+            }
+        )
+
         platform = self.platform
         platform = platform.drop_dims("location_time", errors="ignore")
         platform = platform.assign_coords(
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -331,7 +331,8 @@ class EchoData:
             - `"latitude"`
             - `"longitude"`
             - `"water_level"`
-        The data inserted into the Platform group will be indexed by a dimension named `"time2"`.
+        The data inserted into the Platform group will be indexed by a dimension named
+        `"location_time"`.
 
         Parameters
         ----------
@@ -374,10 +375,12 @@ class EchoData:
             extra_platform_data = extra_platform_data.drop_vars(trajectory_var)
             extra_platform_data = extra_platform_data.swap_dims({"obs": time_dim})
 
-        platform = self.platform.assign_coords(
-            time2=extra_platform_data[time_dim].values
+        platform = self.platform
+        platform = platform.drop_dims("location_time", errors="ignore")
+        platform = platform.assign_coords(
+            location_time=extra_platform_data[time_dim].values
         )
-        platform["time2"].attrs[
+        platform["location_time"].attrs[
             "long_name"
         ] = "time dimension for external platform data"
 
@@ -405,7 +408,7 @@ class EchoData:
         self.platform = platform.update(
             {
                 "pitch": (
-                    "time2",
+                    "location_time",
                     mapping_search_variable(
                         extra_platform_data,
                         ["pitch", "PITCH"],
@@ -413,7 +416,7 @@ class EchoData:
                     ),
                 ),
                 "roll": (
-                    "time2",
+                    "location_time",
                     mapping_search_variable(
                         extra_platform_data,
                         ["roll", "ROLL"],
@@ -421,7 +424,7 @@ class EchoData:
                     ),
                 ),
                 "heave": (
-                    "time2",
+                    "location_time",
                     mapping_search_variable(
                         extra_platform_data,
                         ["heave", "HEAVE"],
@@ -429,7 +432,7 @@ class EchoData:
                     ),
                 ),
                 "latitude": (
-                    "time2",
+                    "location_time",
                     mapping_search_variable(
                         extra_platform_data,
                         ["lat", "latitude", "LATITUDE"],
@@ -437,7 +440,7 @@ class EchoData:
                     ),
                 ),
                 "longitude": (
-                    "time2",
+                    "location_time",
                     mapping_search_variable(
                         extra_platform_data,
                         ["lon", "longitude", "LONGITUDE"],
@@ -445,7 +448,7 @@ class EchoData:
                     ),
                 ),
                 "water_level": (
-                    "time2",
+                    "location_time",
                     mapping_search_variable(
                         extra_platform_data,
                         ["water_level", "WATER_LEVEL"],
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -470,8 +470,13 @@ class EchoData:
         )
         if extra_platform_data_file_name:
             history_attr += ", from file " + extra_platform_data_file_name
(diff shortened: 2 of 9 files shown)


leewujung:
Recap from meeting: 
- expand the clipped range of `extra_platform_data` to be 1 index before and 1 index after the time period spanned by `ping_time`, so that the first and last `ping_time` has something to interpolate lat/lon to.
- @imranmaj could you also add a test for the time ranges?

Thanks!

emiliom:
> expand the clipped range of extra_platform_data to be 1 index before and 1 index after the time period spanned by ping_time, so that the first and last ping_time has something to interpolate lat/lon to.

I agree with that suggestion.

echopype/echodata/echodata.py
```diff
@@ -374,10 +375,21 @@ def update_platform(self, extra_platform_data: xr.Dataset, time_dim="time"):
...
+
+        platform = self.platform
+        platform = platform.drop_dims("location_time", errors="ignore")
+        platform = platform.assign_coords(
+            location_time=extra_platform_data[time_dim].values
         )
-        platform["time2"].attrs[
+        platform["location_time"].attrs[
             "long_name"
         ] = "time dimension for external platform data"
```
emiliom, line 395: I have a couple of comments about the `location_time` attribute assignments, going beyond `long_name`. Broadly, these should follow exactly what is done elsewhere when `location_time` is created during the conventional convert process (via open_raw):
- Exactly the [same attributes and attribute values as in ek60](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/convert/set_groups_ek60.py#L206), including `long_name`
- End with a [`set_encodings` (or some variation of it) as done in ek60](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/convert/set_groups_ek60.py#L321)

Currently, `long_name` is assigned a name that tries to inform users about the provenance of this time coordinate ("time dimension for external platform data"). I suggest that that provenance info be added instead via the [CF-recommended `history` attribute](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#description-of-file-contents). Here's what the CF convention says about `history`:
> Provides an audit trail for modifications to the original data. Well-behaved generic netCDF filters will automatically append their name and the parameters with which they were invoked to the ~~global~~ history attribute of an input netCDF ~~file~~ (variable). We recommend that each line begin with a timestamp indicating the date and time of day that the program was executed.

I suggest something like this: "2021-11-08T11:13:00 Created from external platform data". Is the file name for the external platform data recorded in the provenance somewhere? It could also be added in the Platform group via a global `history` attribute.

cc @leewujung

leewujung: Ok, finally got to this!

@emiliom I agree with your suggestion on using the `history` attribute for recording the time this operation happens and adding the external platform filename.

> - Exactly the [same attributes and attribute values as in ek60](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/convert/set_groups_ek60.py#L206), including `long_name`

For setting the attributes, perhaps using the same strategy as the one you did for the #480 ? 
This is nice, I feel we are now well on the way to address convention adherence issues! 😃 

> - End with a [`set_encodings` (or some variation of it) as done in ek60](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/convert/set_groups_ek60.py#L321)

Agreed!

emiliom: Thanks! I'll start working on this later today

emiliom:
I've pushed a commit that addresses the improvements I pointed out about location_time attributes, including time encodings. `test_echodata.test_update_platform` works, but I should generate a netcdf file and examine the `location_time` attributes directly.

HOWEVER ... I ran into ["circular imports" errors](https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/) (my first time ever, I think). Sigh. This was cause by trying to import `set_encodings` from `convert.set_groups_base`, in `echodata.echodata`. The conflict is due to this import statement found in `convert.api`:

```python
from ..echodata.echodata import XARRAY_ENGINE_MAP, EchoData
```

I tried a couple of options, with no luck. I wanted to at least be able to test my changes, and the only solution I found, *temporarily*, was to copy a whole block of declarations and functions from `convert.set_groups_base` in order to eliminate the need to import that module. It's very ugly, but it works.

@leewujung give a look to the enhancements, but ignore that ugly duplication.

@lsetiawan can you help!? I fear that some larger restructuring may be needed.

(Not addressed yet: "expand the clipped range of extra_platform_data to be 1 index before and 1 index after the time period spanned by ping_time, so that the first and last ping_time has something to interpolate lat/lon to." Or " add a test for the time ranges")

emiliom:
Turns out my fix was premature :cry: . There were some issues. I'm still working on it, and running into some problems.

In the process of going over this again, I realize that there are other problems with variable attributes: for all the variables that are overwritten ("pitch", "roll", "heave", "latitude", "longitude", "water_level"), all attributes are dropped and need to be added again.

@lsetiawan the part about having to duplicate code to avoid the circular import error is still true.

emiliom:
Alright, I've pushed my latest stuff.

To summarize where I'm at: I add the location_time attributes and encodings correctly, and they show up as expected in the echodata platform group:

image

But after saving to netcdf, they're gone and only the encoding is left:

```
group: Platform {
  dimensions:
  	mru_time = 1 ;
  	location_time = 3581 ;
  variables:
  	double mru_time(mru_time) ;
  		mru_time:_FillValue = NaN ;
  		mru_time:axis = "T" ;
  		mru_time:long_name = "Timestamps for MRU datagrams" ;
  		mru_time:standard_name = "time" ;
  		mru_time:units = "seconds since 1900-01-01T00:00:00+00:00" ;
  		mru_time:calendar = "gregorian" ;
  	double location_time(location_time) ;
  		location_time:_FillValue = NaN ;
  		location_time:units = "seconds since 1900-01-01T00:00:00+00:00" ;
  		location_time:calendar = "gregorian" ;
  	double pitch(location_time) ;
  		pitch:_FillValue = NaN ;
  	double roll(location_time) ;
  		roll:_FillValue = NaN ;
  	double heave(location_time) ;
  		heave:_FillValue = NaN ;
  	double latitude(location_time) ;
  		latitude:_FillValue = NaN ;
  	double longitude(location_time) ;
  		longitude:_FillValue = NaN ;
  	double water_level(location_time) ;
  		water_level:_FillValue = NaN ;
```

echopype/convert/set_groups_ek60.py
```diff
@@ -6,12 +6,12 @@
...
-    DEFAULT_CHUNK_SIZE,
-    SetGroupsBase,
-    set_encodings,
-)
+from ..utils.coding import set_encodings
+
+# fmt: off
+from .set_groups_base import DEFAULT_BEAM_COORD_ATTRS, DEFAULT_CHUNK_SIZE, SetGroupsBase
+
+# fmt: on
```
emiliom, line 14: What does `fmt` off and on mean?

emiliom:
Thanks @lsetiawan ! I've only inspected your changes visually, but they definitely seem to be what was needed to address the circular import errors. They'll also set us on a better path for supporting encoding more consistently.

I haven't tested any of it. I'll try to do that tonight or tomorrow.

More broadly, this PR keeps growing and growing in scope! I think we should get it to a point where it does most, but not all, of what it set out to do; then merge it and open new follow up (smaller) PR's as needed.

emiliom:
This PR passes all tests. I also ran a test to assess the results of `update_platform` both as the in-memory EchoData object and in the netcdf file generated by `to_netcdf`. The location_time range clipping seems to be working as designed (though it still needs the suggested enhancement to expand the range by one index), and the location_time attributes & encodings in the EchoData object are also as designed. The location_time attributes & encodings in the netcdf file are still not working as expected.

I propose we merge the PR now, and open other issues and PR's to address the remaining tasks. @leewujung do you agree?

leewujung:
Thanks! Yes I agree that we should merge this and create new issues to address the 2 remaining tasks.

Before you merge, could you also change the PR title to capture the addition of `utils/coding.py`?

emiliom:
> Before you merge, could you also change the PR title to capture the addition of utils/coding.py?

Done. 
Thanks for chiming in. I'll merge now.

leewujung:
@emiliom I have captured the clipping range and test needs in #485 .
I'll leave it to you to open up an issue for the location_time attributes & encodings in the netcdf file.

One thing I think we should consider is to add tests specifically for the attributes and encodings at the end of various operations, especially those that "touch" the organized Datasets in the groups. The ones I can think of now are combine_echodata, update_platform, read from results of to_netcdf/zarr. 

Thanks!

Issue #463, opened 2021-10-28
Document optional parameters for calibration functions


leewujung:
Currently parameters `waveform_mode` and `encode_mode` for calibration EK80 data are not documented. We should add them in in the not very distant future, especially since we're now diving back in to make sure everything is done correctly.

emiliom:
@leewujung can we close this issue now that #477 is merged?

leewujung:
Sounds good. Closing this now.

Issue #458, opened 2021-10-25
Allow feeding in frequency-dependent gain factor for broadband calibration


leewujung:
While working on #446 it became obvious that there is a need for users to be able to put in frequency-dependent gain factors externally and not relying on what is or isn't in the data. This is especially true since a lot of files don't already have the frequency-dependent gain factors stored (i.e. the `Vendor` group does not have a data variable called `gain` with dimension `cal_channel_id` and `cal_frequency`).

This seems can be handled in 2 ways:
1. Similar to how we update the platform group to add missing GPS data using `echodata.update_platform()`
2. Feeding in a `gain` xarray DataArray as part of the input argument `cal_params` when calling `compute_Sv`

Pinging @emiliom here since the route we choose is related to what we consider should be an integral part of the data file. 
Perhaps both should be supported? Users may want to apply different gain factors from different calibration sessions for comparison, so route 2 is useful.

emiliom:
Are there specific, convention defined variables for gain factors? Can you point me to them (name(s) and group)

leewujung:
It's a little involved...

There is a `gain_correction` data variable under the `Beam` group, which I decided to use to store the gain factor for CW mode data. There is another `gain_correction` data variable in the `Vendor` group that I decided to store, because it contains info that is more comprehensive than that stored under the `Beam` group (basically this is a look up table with coordinate dimension `frequency` and `pulse_length` -- the `gain_correction` in the `Beam` group is the result of indexing using the actual `pulse_length`)

For broadband data there could be additional calibration results (frequency-dependent gain factors) saved in the config XML, which are currently stored under data variable `gain` in the `Vendor` group with coordinate dimensions `cal_channel_id` and `cal_frequency`. The data variable name `gain` here should probably be changed to something like `gain_correction_broadband` or something similar to be consistent but distinguishable from the `gain_correction` defined in the convention.

Note for BB data files the narrowband gain (the ones used for CW data) still exist, probably because the echosounder can function in both modes so it is more convenient for them to just be included regardless.

leewujung:
This is addressed in #955.

PR #477, opened 2021-11-04
Clean up `compute_Sv/Sp` accepted values and add docstrings

leewujung:
This PR addresses #463 by adding docstrings for `compute_Sv` and `compute_Sp`. I set the target to the `dev` branch because I feel more confident that with fixes in #446 we are handling the different types of EK80 files better.

To update the documentation would take a bit more effort, because I think it requires perhaps a separate section to explain the transmit waveform options and data recording options for EK80. So, I only added these to the docstring for now.

@emiliom : thoughts?


echopype/calibrate/api.py
```diff
@@ -34,11 +36,101 @@ def _compute_cal(
...
+    Compute volume backscattering strength (Sv) from raw data.
+
+    The calibration routine varies depending on the sonar type.
+    Currently this operation is supported for the following ``sonar_model``:
+    EK60, AZFP, EK80 (see below for detail).
+
+    The EK80 echosounder can be configured to transmit
+    either broadband (``waveform_mode="BB"``) or narrowband (``waveform_mode="CW"``) signals.
+    When transmitting in broadband mode, the returned echoes are
+    encoded as complex samples (``encode_mode="complex"``).
```
emiliom, line 50: So, if waveform_mode="BB", encode_mode is unnecessary b/c compute_Sv can set it to "complex", right?

leewujung: Yes, but I don't remember if we currently allow waveform_mode="BB" and encode_mode=None to work...

echopype/calibrate/api.py
```diff
@@ -34,11 +36,101 @@ def _compute_cal(
...
+    waveform_mode : {"CW", "BB"}, optional
+        Type of transmit waveform:
+
+        - `"CW"` for narrowband transmission,
+          returned echoes recorded either as complex or power/angle samples
+        - `"BB"` for broadband transmission samples,
+          returned echoes recorded as complex samples
+
+        This parameter is only used if `sonar_model` is `"EK60"` or `"EK80"`,
+        and in those cases it must be specified.
```
emiliom, line 70: ?? Is this saying that for EK60 and EK80 'waveform_mode` must be specified, or `sonar_model`? The sentence is confusing

echopype/calibrate/api.py
leewujung: Hmm, upon looking into this further, it's a bit of a sprawling situation and I need to work on the code itself before updating the docstrings. I put the docstring in `calibrate/api.py` because that's the ones that RTD will see, but the actual kwargs are passed under the hood to the right CALIBRATOR depending on the sonar model. 

The current default and if/else checks are somewhat problematic. As a note for myself, this is where the fixes should go: 
https://github.com/OSOceanAcoustics/echopype/blob/main/echopype/calibrate/calibrate_ek.py#L727

@emiliom I will ping you on this PR again when it is ready for review. I'll also update the PR title to reflect changes made in the code.

echopype/calibrate/api.py
```diff
@@ -34,11 +36,101 @@ def _compute_cal(
...
+
+    encode_mode : {"complex", "power"}, optional
+        Type of encoded return echo data:
+
+        - `"complex"` for complex samples
+        - `"power"` for power/angle samples
+
+        This parameter is only used if `sonar_model` is `"EK80"`.
```
emiliom, line 78: ... and if EK80 and `waveform_mode = "BB"`, only `encode_mode = "complex"` is accepted. This is kind of weird, but I guess it needs to be stated.

echopype/calibrate/api.py
```diff
@@ -34,11 +36,101 @@ def _compute_cal(
...
+    When transmitting in broadband mode, the returned echoes are
+    encoded as complex samples (``encode_mode="complex"``).
+    When transmitting in narrowband mode, the returned echoes can be encoded
+    either as complex samples or as power/angle combinations in a format similar to
+    those recorded by EK60 echosounders (``encode_mode="power"``).
+
+    The current calibration implemented for EK80 broadband complex data
+    uses band-integrated Sv with the gain computed at the center frequency
+    of the transmit signal.
+
```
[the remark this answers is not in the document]
emiliom: I might suggest moving the preceding paragraph to Notes, too.

echopype/calibrate/api.py
```diff
@@ -34,11 +36,101 @@ def _compute_cal(
...
+    Parameters
+    ----------
```
emiliom, line 60: Add `env_params` and `cal_params`, right? https://echopype.readthedocs.io/en/stable/process.html#calibration-parameters

echopype/calibrate/api.py
```diff
@@ -34,11 +36,101 @@ def _compute_cal(
...
+        The calibrated Sv dataset, including calibration and environmental variables
+        used in the calibration operations.
+    """
     return _compute_cal(cal_type="Sv", echodata=echodata, **kwargs)
 
 
 def compute_Sp(echodata: EchoData, **kwargs):
-    """Compute point backscattering strength (Sp) from raw data."""
+    """
+    Compute point backscattering strength (Sp) from raw data.
```
emiliom, line 91: See my comments for `compute_Sv`. I think all those comments apply here too.

echopype/echodata/echodata.py
```diff
@@ -143,35 +143,48 @@ def compute_range(
...
             does not need to be specified.
+
         azfp_cal_type : {"Sv", "Sp"}, optional
+
             - `"Sv"` for calculating volume backscattering strength
             - `"Sp"` for calculating point backscattering strength.
+
             This parameter is only used if `sonar_model` is `"AZFP"`,
             and in that case it must be specified.
+
```
emiliom, line 169: Same comment as in `compute_Sv` about the ambiguous statement. What does "it" in "it must be specified" refer to? `sonar_model`? If that's the case, can't `sonar_model` be read directly from the echodata metadata?

emiliom:
Alright, I've added a bunch of inline comments. The only additional, general comment I have is more of my own confusion about `.reST` syntax. In some cases you use double backquote ("``") and in other cases single backquotes. I don't remember which one is correct (though from the numpydoc examples, it looks like a single backquote is fine), or even if both will generate the same formatting result. But it's worth checking.

Otherwise: Great, thanks! This will be a great improvement to usability.

leewujung:
@emiliom : After looking into this some more, I don't think we are there yet to fully document this and complete the docstrings. I think we should delay this until later, especially the components involving EK80. For env_params, @imranmaj added a new EnvParams class in #464 -- I requested some changes and explanations before I will be able to review everything. For cal_params updates for EK80 BB data is not yet implemented (#458). Tests are not ready yet for EK80 data -- I'll be meeting with @gavinmacaulay next week to talk about implementation details. This goes on and on...

Linking a comment above so that it doesn't get lost for later ref: https://github.com/OSOceanAcoustics/echopype/pull/477#discussion_r743297332

emiliom:
Yikes. Still, this is progress, even if it's not actually ready to go. Good luck.

leewujung:
> In some cases you use double backquote ("``") and in other cases single backquotes. I don't remember which one is correct (though from the numpydoc examples, it looks like a single backquote is fine), or even if both will generate the same formatting result.

Yes, actually I am quite confused about this. I kept single backquotes in places where I see existing docstrings using single backquotes, and used double backquotes in places where I need inline code rendering.

They actually don't generate the same outputs...  It may be related to our sphinx configuration: the single quotes generates italicized text and the double quotes generates inline code rendering.

Which one should we use? 🤯

emiliom:
>Which one should we use?

:exploding_head: me too. Ask me again when the PR is actually closer to prime time! I don't want to dig deeper right now :stuck_out_tongue_winking_eye:

leewujung:
To make it better (or worth): I will put in an issue re. potentially standardize what we do in docstrings....

echopype/calibrate/api.py
leewujung: actually the current text here came from there with some modifications. I think an overhaul is in order!!

emiliom: Overhaul of the doc page? If so, can you open a new issue with the specifics?

leewujung: Not the doc pages, just overhaul of the `compute_range` docstring.

leewujung:
@emiliom : ok, I think this PR is again ready for review...

leewujung:
@emiliom : added a small bug fix that's required for #487. Once we have both in we are very much ready for releasing v0.5.5 in the coming week. This also serves as a good point for discussing @imranmaj 's #464, including an addition of the `EnvParams` class.

echopype/calibrate/api.py
leewujung: the docstrings were overhauled

echopype/echodata/echodata.py
leewujung: docstrings were overhauled

echopype/calibrate/api.py
leewujung: docstrings were overhauled

echopype/calibrate/api.py
leewujung: docstrings were overhauled

docs/source/convert.rst
```diff
@@ -37,11 +37,16 @@ File conversion for different types of echosounders is achieved by
...
-For data files from EK60, EK80 and  EA640 echosounders,
-use the parameter ``sonar_model`` to indicate the echosounder type,
-since there is no specific information in the extension ``.raw``
-that includes information about the echosounder type.
-In this example, ``open_raw`` is used to convert a raw EK80 file,
+Use the parameter ``sonar_model`` to indicate the echosounder type:
+    - ``EK60``: Kongsberg Simrad EK60 echosounder
+    - ``ES70``: Kongsberg Simrad ES70 echosounder
+    - ``EK80``: Kongsberg Simrad EK80 and Kongsberg EA640 echsoounders
+    - ``ES80``: Kongsberg Simrad ES80 echsoounder
```
emiliom, line 44: echosounder is mispelled in the 2 lines above

echopype/calibrate/api.py
```diff
@@ -34,11 +58,159 @@ def _compute_cal(
...
+    The calibration routine varies depending on the sonar type.
+    Currently this operation is supported for the following ``sonar_model``:
+    EK60, AZFP, EK80 (see Notes below for detail).
+
+    Parameters
+    ----------
+    echodata : EchoData
+        An `EchoData` object created by using `open_raw` or `open_converted`
+
+    env_params : dict
```
emiliom, line 74: env_params and cal_params are also optional, so for consistency with your use of the qualifier "optional" in the other parameters, I'd add that qualifier to these two. More generally, we're not currently consistent across the board (across all echopype functions/classes) in having a parameter "optional" qualifier string. But that's a broader topic.

echopype/calibrate/api.py
```diff
@@ -34,11 +58,159 @@ def _compute_cal(
...
+          `"sa_correction"`, `"gain_correction"`, `"equivalent_beam_angle"`
+        - for AZFP echosounder, allowed parameters include:
+          `"EL"`, `"DS"`, `"TVR"`, `"VTX"`, `"equivalent_beam_angle"`, `"Sv_offset"`
+
+        Passing in calibration parameters for other echosounders
+        are not currently supported
+
+    waveform_mode : {"CW", "BB"}, optional
+        Type of transmit waveform.
+        Required only for data from the EK80 echosounder.
```
emiliom, line 100: Is waveform_mode *used* at all for other echosounders? That is, is it a parameter that only applies to EK80? If that's what you mean, the wording maybe you should be changed to, say: "Required for the EK80 echosounder; not used with any other echosounder"

echopype/calibrate/api.py
```diff
@@ -34,11 +58,159 @@ def _compute_cal(
...
+
+        - `"complex"` for complex samples
+        - `"power"` for power/angle samples, only allowed when
+          the echosounder is configured for narrowband transmission
+
+    Returns
+    -------
+    xr.Dataset
+        The calibrated Sp dataset, including calibration parameters
```
emiliom, line 118: Change to "Sv"

echopype/calibrate/api.py
```diff
@@ -34,11 +58,159 @@ def _compute_cal(
...
+    or narrowband (``waveform_mode="CW"``) signals.
+    When transmitting in broadband mode, the returned echoes are
+    encoded as complex samples (``encode_mode="complex"``).
+    When transmitting in narrowband mode, the returned echoes can be encoded
+    either as complex samples (``encode_mode="complex"``)
+    or as power/angle combinations (``encode_mode="power"``) in a format
+    similar to those recorded by EK60 echosounders.
+
+    The current calibration implemented for EK80 broadband complex data
+    uses band-integrated Sv with the gain computed at the center frequency
```
emiliom, line 213: Should that be Sp rather than Sv? (I don't know enough about Sp to know)

echopype/calibrate/calibrate_ek.py
```diff
@@ -311,7 +320,7 @@ def __init__(self, echodata, env_params, cal_params, waveform_mode, encode_mode)
         # self.range_meter computed under self._compute_cal()
         # because the implementation is different depending on waveform_mode and encode_mode
 
-    def get_env_params(self, env_params, waveform_mode=None, encode_mode="complex"):
+    def get_env_params(self, env_params, waveform_mode="BB", encode_mode="complex"):
         """Get env params using user inputs or values from data file.
 
         EK80 file by default contains sound speed, temperature, depth, salinity, and acidity,
```
emiliom, line 326: acidity, really?? Requiring a pH sensor seems surprising

emiliom (commented):
I've reviewed much of it, mainly the doc strings, and added some comments. I'm not in a position to review the accepted values and the tests for them.

There's a lot of doc string text that's repeated in multiple functions. I didn't review all of it.

emiliom:
Needless to say, the doc strings now vs what was there previously is a night-and-day situation! Thanks!

echopype/calibrate/calibrate_ek.py
gavinmacaulay: The sound speed equations require acidity and the EK80 allows the users to provide it. Most (all?) people just set it to 8.0.

emiliom: Interesting, thanks

docs/source/convert.rst
leewujung: Good catch - fixed!

echopype/calibrate/api.py
leewujung: > More generally, we're not currently consistent across the board (across all echopype functions/classes) in having a parameter "optional" qualifier string. 

Could you add this to #478 ?

echopype/calibrate/api.py
leewujung: Yes, this is only required for EK80. The same is true for encode_mode also. I've changed both.

echopype/calibrate/api.py
leewujung: Done.

echopype/calibrate/api.py
leewujung: Yes, changed.

echopype/calibrate/api.py
leewujung: the "optional" qualifier is added to both env_params and cal_params docstrings

leewujung:
Self-merging this since everything checked out!

Issue #373, opened 2021-05-29
Add explanation of `range_bin` to docs and attributes to the coordinate


leewujung:
In the raw converted data we have the coordinate `range_bin` which is just sample numbers (0, 1, 2, ...). This coordinate is then converted to `range` or `depth` in physical units (meters) when the data is calibrated. This can be confusing to the users and we should expand this in the docs.

This is related to #259 -- we should provide a couple simple helper function for this conversion.

emiliom:
Related to this documentation, also need to add variable attributes to the `range_bin` coordinate variable, which currently doesn't have any. I can do this while adding to the documentation.

For reference, for EK60 here's where the code to add attributes would go (in the `dev` branch): [echopype/convert/set_groups_ek60.py#L528](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/convert/set_groups_ek60.py#L528)

emiliom:
@leewujung do you have a suggestion for where the `range_bin` description should go in the documentation? I'm thinking, for now, it belongs in https://echopype.readthedocs.io/en/stable/data-format.html#modifications-to-sonar-netcdf4

That section also needs to be rewritten, borrowing heavily from text and figures in the paper.

leewujung:
Yes I think it'll work well to have the description in the part you proposed, and agree with updating the whole thing using materials from the paper.

One thing it'd be nice to have is probably some printouts from an `EchoData` object to show how things look like exactly.

PR #480, opened 2021-11-06
Generalize assignment of beam coordinate var attributes

emiliom:
Unified where the attributes for the core beam coordinate variables (`frequency`, `ping_time`, `range_bin`) are set. I've created a new `DEFAULT_BEAM_COORD_ATTRS` dict in `set_groups_base.py` that declares these common attributes and their values, then import that variable in the various set_groups_x.py modules. This PR implements that change for ek60, ek80, azfp.This abstraction/centralization will reduce the potential for variability across sensors and make it a tiny, tiny bit easier to roll out support for new sensors in a consistent fashion.

Also added a `long_name` attribute to `range_bin`, partly addressing #373

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -4,7 +4,7 @@ Class to save unpacked echosounder data to appropriate groups in netcdf or zarr.
 import numpy as np
 import xarray as xr
 
-from .set_groups_base import SetGroupsBase, set_encodings
+from .set_groups_base import DEFAULT_BEAM_COORD_ATTRS, SetGroupsBase, set_encodings
 
 
 class SetGroupsAZFP(SetGroupsBase):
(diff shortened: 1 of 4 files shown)


leewujung:
Thanks @emiliom, this looks great! 

One thing I thought of when going through this is that `DEFAULT_BEAM_COORD_ATTRS` should probably live somewhere else, something like `core.py` but for convention side of things. Perhaps that's something to be resolved in a more top-down form, related to the pydantic work. For now it seems fine for that to live under `set_group_base.py`.

On an _unrelated_ note, I noticed that `DEFAULT_CHUNK_SIZE` is only used for encoding EK60 data. Properly chunking is arguably going to be more important for EK80. I'll submit an issue about that so that we don't forget about that when we get to working on performance-related issues..

leewujung (approved):

emiliom:
> On an unrelated note, I noticed that DEFAULT_CHUNK_SIZE is only used for encoding EK60 data. Properly chunking is arguably going to be more important for EK80. I'll submit an issue about that so that we don't forget about that when we get to working on performance-related issues..

Continuing the unrelated note, I think I noticed that DEFAULT_CHUNK_SIZE is actually declared in two different modules, the same way, But yeah, that's for another issue

emiliom:
> One thing I thought of when going through this is that `DEFAULT_BEAM_COORD_ATTRS` should probably live somewhere else, something like `core.py` but for convention side of things. Perhaps that's something to be resolved in a more top-down form, related to the pydantic work. For now it seems fine for that to live under `set_group_base.py`.

Let's revisit that thought for sure, once we're creating the convention-compliance architecture for 0.6.0

emiliom:
Thanks for the review. I'll self merge now

Issue #454, opened 2021-10-11
Preprocess module does not have tests against "real" data.


lsetiawan:
Currently, the preprocess module is not tested against any "real" data. Tests against toy data can be found at https://github.com/OSOceanAcoustics/echopype/blob/main/echopype/tests/old/test_preprocess.py.

leewujung:
It is kinda bizarre that `test_preprocess.py` is under `tests/old`! 
Shouldn't it be under `tests/preprocess`? 
I guess this calls for a PR to both add tests running on real data and fix the folder structure.

PR #487, opened 2021-11-11
Move preprocess test and test against "real" data; perform a little cleanup.

lsetiawan:
## Overview

This PR addresses #454. Preprocess test is now moved under preprocess module. There's a `test_old.py` that is now moved under the "old" folder in tests, and extra test_summary_statistics is now deleted.


lsetiawan:
@leewujung The preprocess mvbs is going to fail in one of the EK80 data. Might need your help to figuring out why. Thanks! 😄

leewujung:
Roger that @lsetiawan. Thanks, I'll take a look!

echopype/tests/preprocess/test_preprocess.py
```diff
@@ -230,3 +281,28 @@ def test_compute_MVBS():
...
+
+
+@pytest.mark.parametrize(param_args, param_testdata)
+def test_preprocess_mvbs(
+    filepath,
+    sonar_model,
+    azfp_xml_path,
+    range_kwargs,
+):
+    # TODO: Need to figure out how to compare the actual rendered plots
```
leewujung, line 293: This line is leftover from another test?

leewujung:
@lsetiawan : the test file `echopype-test-D20211004-T235930.raw` appears to not contain the gain values needed for calibration, causing the test to fail. There is a change better addressed in #477 so that will require pulling those changes here.

leewujung:
Ok, I just fixed that bug and resolved a merge conflict there. Tests passed locally so everything should be fine here once those changes are pulled in there. :)

echopype/tests/preprocess/test_preprocess.py
leewujung: I removed this and added a docstring.

Issue #488, opened 2021-11-12
Verify correct location_time attributes and encoding from update_platform


emiliom:
Replaces part of #460 

Ensure that the `location_time` attributes and encodings that have been added (see [#476](https://github.com/OSOceanAcoustics/echopype/pull/476#pullrequestreview-800621745)) are saved into the netcdf file (and zarr?) when exporting via `to_netcdf`. See the status update [here](https://github.com/OSOceanAcoustics/echopype/pull/476#issuecomment-965661370)
 
Additional note: I think there's some linkage between this task and #479. See the recent PR #486

emiliom:
After #486, I've rerun my tests to check that the assigned `location_time` attributes and encoding are preserved after saving to netcdf. They are! But also realized that I wasn't using `overwrite=True` in my recent `to_netcdf` tests, so there's a cance I actually had fixed this problem earlier and missed seeing that :sob:.

Anyways, this is fixed. But I'll leave the issue open for now to focus on adding attributes to the additional variables added by `update_platform`:  "pitch", "roll", "heave", "latitude", "longitude", "water_level". For these variables, all attributes are dropped and need to be added again.

emiliom:
This issue has been addressed, including this: adding attributes to the additional variables added by update_platform: "pitch", "roll", "heave", "latitude", "longitude", "water_level". For these variables, all attributes are dropped and need to be added again.

Issue #490, opened 2021-11-12
Parsing error when redundant frequencies are found in EK60 data


oftfrfbf:
Scanning through some miscellaneous EK60 files I have been finding issues when redundant frequencies are packaged in the raw data.

Example code:
```python
import echopype as ep

raw_file_s3path = f"s3://noaa-wcsd-pds/data/raw/Oscar_Dyson/DY1002/EK60/DY1002_EK60-D20100318-T023008.raw"
ed = ep.open_raw(raw_file_s3path, sonar_model='EK60', storage_options={'anon': True})
```
Returns:
```python
ValueError: cannot reindex or align along dimension 'frequency' because the index has duplicate values
```
Looking at the same data with pyEcholab:
```python
from echolab2.instruments import EK60

ek60 = EK60.EK60()
ek60.read_raw("DY1002_EK60-D20100318-T023008.raw")
print(ek60)
```
I can see that there are 6 channels reported including two 70 kHz channels. My understanding is that physically the same hardware was present throughout the cruise — the likely cause is that the transceiver number in the ER60 software changed for whatever reason ("3-1" to "3-2").
```python
<class 'echolab2.instruments.EK60.EK60'> at 0x7fe5e3322850
    EK60 object contains data from 6 channels:
        1:GPT  18 kHz 009072034d45 1-1 ES18-11
        2:GPT  38 kHz 009072033fa2 2-1 ES38B
        3:GPT  70 kHz 009072058c6c 3-1 ES70-7C
        4:GPT  70 kHz 009072058c6c 3-2 ES70-7C
        5:GPT 120 kHz 00907205794e 4-1 ES120-7C
        6:GPT 200 kHz 0090720346a8 5-1 ES200-7C
    data start time: 2010-03-18T02:30:08.779
      data end time: 2010-03-18T02:47:01.549
    number of pings: 609
```
The error occurs in set_groups_ek60.py when "ds" and "ds_backscatter" are merged.

https://github.com/OSOceanAcoustics/echopype/blob/1548f46cf7f163188b2858e9a6838290fd25b7ec/echopype/convert/set_groups_ek60.py#L651

Merging the two channels like below works fine because they contain exclusive data but consolidating that back into the main dataset is giving me some trouble. I'll be working on this to see if I can figure out a solution.
```python
[...]

PR #492, opened 2021-11-13
update_platform add missing variable attributes

emiliom:
Completes additional task added to #488 

> adding attributes to the additional variables added by `update_platform`: "pitch", "roll", "heave", "latitude", "longitude", "water_level". For these variables, all attributes are dropped and need to be added again.

### DIFF
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -10,10 +10,11 @@ from ..utils.coding import COMPRESSION_SETTINGS, set_encodings
 
 DEFAULT_CHUNK_SIZE = {"range_bin": 25000, "ping_time": 2500}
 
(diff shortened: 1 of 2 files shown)


leewujung:
Looks good! I'll merge this now so that #495 can incorporate these changes. 

In case this got buried: let's discuss the Beam group dimension `frequency` when we are ready to take on the convention-related overhaul -- see [here](https://github.com/OSOceanAcoustics/echopype/issues/490#issuecomment-971651101) for my thoughts on a potential option when there are more than 1 channel at the same frequency (under #490) -- although one needs to be really careful in making sure the approach would work with other echosounder models also. For example, what should be used as the equivalent of channel_id in AZFP?

This is probably better suited in a discussion thread rather than a merged PR, so I'll open one up.

PR #495, opened 2021-11-19
Expand extracted time range in EchoData.update_platform by 1 in both directions; add test for extracted time range

### LINKED ISSUE #485: Expand extracted time range in EchoData.update_platform
This is a leftover from #476.

Currently, `EchoData.update_platform` extracts the relevant timestamps by clipping the range _between_ the first and the last `ping_time` of the Beam group data. The clipped range should be expanded to be 1 index before and 1 index after the time period spanned by ping_time, so that the first and last ping_time has something to interpolate lat/lon to. An explicit test to check the extracted `location_time` is also needed.

#### Tasks
- [x] Expand clipping range
- [x] Add explicit test for timestamp

imranmaj:
* Expands the extracted time range from external platform data to 1 less than min of `EchoData.beam["ping_time"]` and 1 more than max of `EchoData.beam["ping_time"]`
    * Was originally between min of `EchoData.beam["ping_time"]` and max of `EchoData.beam["ping_time"]`
* Adds test for extracted time range of updated platform data

Fixes #485

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -404,19 +404,20 @@ class EchoData:
 
         # clip incoming time to 1 less than min of EchoData.beam["ping_time"] and
         #   1 greater than max of EchoData.beam["ping_time"]
-        # account for unsorted time
+        # account for unsorted external time by checking whether each time value is between
+        #   min and max ping_time instead of finding the 2 external times corresponding to the
+        #   min and max ping_time and taking all the times between those indices
         sorted_external_time = extra_platform_data[time_dim].data
         sorted_external_time.sort()
         min_index = max(
             np.searchsorted(
-                sorted_external_time, self.beam["ping_time"].min(), side="right"
-            )
-            - 1,
+                sorted_external_time, self.beam["ping_time"].min(), side="left"
+            ) - 1, # noqa
             0,
         )
         max_index = min(
             np.searchsorted(
-                sorted_external_time, self.beam["ping_time"].max(), side="left"
+                sorted_external_time, self.beam["ping_time"].max(), side="right"
             ),
             len(sorted_external_time) - 1,
         )
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -409,12 +409,14 @@ class EchoData:
         #   min and max ping_time and taking all the times between those indices
         sorted_external_time = extra_platform_data[time_dim].data
         sorted_external_time.sort()
+        # fmt: off
         min_index = max(
             np.searchsorted(
                 sorted_external_time, self.beam["ping_time"].min(), side="left"
-            ) - 1, # noqa
+            ) - 1,
             0,
         )
+        # fmt: on
         max_index = min(
             np.searchsorted(
                 sorted_external_time, self.beam["ping_time"].max(), side="right"
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -403,6 +403,8 @@ class EchoData:
             extra_platform_data = extra_platform_data.drop_vars(trajectory_var)
             extra_platform_data = extra_platform_data.swap_dims({"obs": time_dim})
 
+        # clip incoming time to 1 less than min of EchoData.beam["ping_time"] and
+        #   1 greater than max of EchoData.beam["ping_time"]
         # account for unsorted time
         sorted_external_time = extra_platform_data[time_dim].data
         sorted_external_time.sort()
@@ -416,8 +418,7 @@ class EchoData:
         max_index = min(
             np.searchsorted(
                 sorted_external_time, self.beam["ping_time"].max(), side="left"
-            )
-            + 1,
+            ),
             len(sorted_external_time) - 1,
         )
         extra_platform_data = extra_platform_data.sel(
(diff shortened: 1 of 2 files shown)


echopype/echodata/echodata.py
```diff
@@ -403,12 +403,29 @@ def update_platform(
             extra_platform_data = extra_platform_data.drop_vars(trajectory_var)
             extra_platform_data = extra_platform_data.swap_dims({"obs": time_dim})
 
-        # TODO: Expand the clipping range by one (or 2?) indices
+        # clip incoming time to 1 less than min of EchoData.beam["ping_time"] and
+        #   1 greater than max of EchoData.beam["ping_time"]
+        # account for unsorted time
```
leewujung, line 408: > `# account for unsorted time`

Which unsorted time are you referring to? is it EchoData.beam."ping_time" or the time in the external platform location time? From code the below it seems to be both? Would be great to add that to improve code readability.

echopype/echodata/echodata.py
```diff
@@ -403,12 +403,29 @@ def update_platform(
...
+        sorted_external_time = extra_platform_data[time_dim].data
+        sorted_external_time.sort()
+        min_index = max(
+            np.searchsorted(
+                sorted_external_time, self.beam["ping_time"].min(), side="right"
+            )
+            - 1,
```
leewujung, line 415: Why use "right" instead of "left" and then -1? 
For example, in the case when you have [1,2,3,3,3,4,5], to find the index to insert a new 3, wouldn't it be side="left" and then -1? Otherwise you get another 3 in the outcome.

echopype/echodata/echodata.py
```diff
@@ -403,12 +403,29 @@ def update_platform(
             extra_platform_data = extra_platform_data.drop_vars(trajectory_var)
             extra_platform_data = extra_platform_data.swap_dims({"obs": time_dim})
 
-        # TODO: Expand the clipping range by one (or 2?) indices
+        # clip incoming time to 1 less than min of EchoData.beam["ping_time"] and
+        #   1 greater than max of EchoData.beam["ping_time"]
+        # account for unsorted time
+        sorted_external_time = extra_platform_data[time_dim].data
+        sorted_external_time.sort()
```
leewujung, line 410: Could you move .sort to the top of that you have `sorted_external_time = extra_platform_data[time_dim].data.sort()` ?

echopype/echodata/echodata.py
leewujung: Also, a small nitpicking thing: could you combine the -1 to the line above it? The code is more readable that way.

echopype/echodata/echodata.py
```diff
@@ -403,12 +403,29 @@ def update_platform(
...
+            np.searchsorted(
+                sorted_external_time, self.beam["ping_time"].min(), side="right"
+            )
+            - 1,
+            0,
+        )
+        max_index = min(
+            np.searchsorted(
+                sorted_external_time, self.beam["ping_time"].max(), side="left"
+            ),
```
leewujung, line 421: I have the same question as the above: why wouldn't it be using side="right"?

echopype/tests/echodata/test_echodata.py
```diff
@@ -168,3 +168,12 @@ def test_update_platform():
...
         assert not np.isnan(ed.platform[variable].values).all()
+
+    # check times are > min(ed.beam["ping_time"]) - 2s
+    assert (ed.platform["location_time"] > ed.beam["ping_time"].min() - np.timedelta64(2, "s")).all()
+    # check there is only 1 time < min(ed.beam["ping_time"])
+    assert np.count_nonzero(ed.platform["location_time"] < ed.beam["ping_time"].min()) == 1
+    # check times are < max(ed.beam["ping_time"]) + 2s
+    assert (ed.platform["location_time"] < ed.beam["ping_time"].max() + np.timedelta64(2, "s")).all()
+    # check there is only 1 time > max(ed.beam["ping_time"])
+    assert np.count_nonzero(ed.platform["location_time"] > ed.beam["ping_time"].max()) == 1
```
leewujung, line 179: Where does the 2s come from? Is that the temporal interval of the data in `extra_platform_data`? It would be good to add the explanation to the comments so people know why there's a hard-code value. (I have not tried to look into what's in the extra_platform_data and ed.beam.ping_time.)

echopype/echodata/echodata.py
imranmaj: The sort is in place, .sort() doesn't return anything

echopype/tests/echodata/test_echodata.py
imranmaj: Yes that is the interval of the data

echopype/echodata/echodata.py
leewujung: Ah ok, somehow I missed that. Thanks!

leewujung:
@imranmaj : Could you also pull in the changes from #492 so that the attributes are set in the updated platform group? Thanks!

echopype/echodata/echodata.py
imranmaj: I think I mixed up left and right

leewujung:
The changes look great. I'll merge this now. Thanks @imranmaj !

PR #498, opened 2021-12-02
Reduce EK80 test data files

leewujung:
The real changes in this PR will not be reflected in the repo because it involves reducing the size of the EK80 test files in the CI_test_data folder.

The changes are described in test_data/README.


leewujung:
There's still another set of files to reduce but that requires having access to EchoView dongle to generate data to test against. Merging this for now since it'll be a couple of months before we can do that. Will submit a new PR then.

PR #503, opened 2021-12-06
Docs echodata HTML repr test

emiliom:
I'm trying to include in the docs the interactive EchoData HTML repr. Here's my best attempt yet; it has problems, but I wanted to get your input and help. ~~It's in the ["Data format" page](https://echopype--503.org.readthedocs.build/en/503/data-format.html).~~ ["Open converted files" page](https://echopype--503.org.readthedocs.build/en/503/open-converted.html), which already contained a small section describing the EchoData object.

The main problem is that the CSS styles it brings in interfere with the page styles. They're changing font sizes and colors, the spacing between sections (above section header), etc. Also, it'd be great if we could increase the font size for the Echodata group description header line.

@lsetiawan any chance you could take a look at this and see if these first-order issues could be addressed?? I've exhausted my rusty and very out-of-date css+html chops trying a number of things, and I'm sure it'll be much easier for you to zoom in to where the problems are being generated and what the solutions might be.

FYI, I generated the HTML "fragment" that includes the echodata HTML repr by setting a minimal notebook (included in this PR), exporting and to an HTML file (`echodata_html_repr.html`), removing portions of the the HTML elements, and using this [reST directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html#raw-data-pass-through):
```rest
.. raw:: html
   :file: ../echodata_html_repr.html
```

I tried to find a cleaner way to reuse the xarray dataset HTML repr (lots of Googling ...), but came empty handed.

cc @leewujung


lsetiawan:
Due to the trickiness of CSS, I've updated the code to use iframe. I hope that works out okay! I created the html by running `_repr_html_()` on the echodata object. Then going through and removing the original header and updating the `xr-wrap` max-width to `600px` since this is the width of the document page.

emiliom:
Thank you @lsetiawan !!! This looks perfect as is! 

> I created the html by running `_repr_html_()` on the echodata object. Then going through and removing the original header and updating the `xr-wrap` max-width to `600px` since this is the width of the document page.

So, you ran `_repr_html_()` on the echodata object in the notebook, and what it returned was the HTML (as a string?) that you used to create `docs/source/_static/echodata_sample.html`? Did you update the notebook, too?

lsetiawan:
> So, you ran _repr_html_() on the echodata object in the notebook, and what it returned was the HTML (as a string?) that you used to create docs/source/_static/echodata_sample.html? Did you update the notebook, too?

Yepp I just write out the string to file using pathlib. I didn't update the notebook with those steps.

emiliom:
I've revised and finalized this PR. I moved the EchoData object example to the ["Open converted files" page](https://echopype--503.org.readthedocs.build/en/503/open-converted.html), which already contained a small section describing the EchoData object. I regenerated that HTML repr myself based on @lsetiawan's instructions, just to make sure I had the workflow in place.

 I've also removed the "Draft" mode. @leewujung can you review and merge if it looks good?

Note that while I didn't add text to explicitly explain `range_bin`, it does describe it via its new `long_name` attribute in the EchoData example. So, I'd say we can close #373. I'm leaving an explicit treatment to an upcoming rewrite of the "Data format" page based on the text and figures in the echopype paper.

leewujung:
@emiliom @lsetiawan : Great to see the rendered EchoData object in the docs! 😄 

@emiliom : I think that having the EchoData description under the "Open converted files" page is a little awkward, since EchoData object is also returned by `open_raw` introduced in the "Convert raw files" page, so it is out of sequence. But I think we can roll with the current version and leave the re-organization to some time later.

I think the next re-organization could probably take the following form:
- have a first-level section "Working with raw data" or something similar, and have 3 second-level items:
   - `open_raw`-related materials
   - `open_converted` 
   - the `EchoData` object

This will become useful once we have more pre-processing and processing functionality in place.

Also, there is a section "Non-uniform data" under "Convert raw files", but a better place for this material is likely in the to-be-rewritten first-level heading `Data format`.

I can move the above elaboration into a new issue. Just want to make this thread more readable.

emiliom:
I agree that a substantial re-organization is needed next. Please do open a new issue with the suggestions you've made, and let's discuss at the next echopype meeting!

In the end I chose to put the rendered echodata object in "Open converted files" mainly b/c there was already an echodata section in there, so it was the least disruptive.

PR #513, opened 2021-12-13
Change imports to all relative for viz

lsetiawan:
## Overview

This PR should fix the circular imports found.


leewujung:
cool, let's give this a try!

leewujung:
This is not the problem. :/  Seems that relative import of EchoData has cause this circular import problem a couple times now and we should look into this more carefully.

PR #521, opened 2021-12-16
Update s3fs to 2021.11.0 and add link to issue

lsetiawan:
This PR updates s3fs pinning to 2021.11.0 for now until we asses if the latest works.


emiliom:
Thanks!
Do you actually need a review of this? The tests are passing, so that's that. I suggest you self merge.

Issue #502, opened 2021-12-04
Add ES70/80 test files to CI


leewujung:
This is leftover from #409 where code snippets were added to enable parsing ES70/80 files.

### Tasks
- [ ] add the below test files to among convert tests:
   - `es80/WBT-D20210620-T012250.raw`
   - `es80/WBT-but-internally-marked-as-EK80-D20210710-T204029.raw`
   - `es70/D20151202-T020259.raw`
- [ ] update `test_data/README.md`

Note that file `es80/WBT-GPT-D20181216-T111107.raw` should not be added, because it has extra complications (see #459)

Issue #501, opened 2021-12-03
Update EK80 convert tests using new files with combination of parameters


leewujung:
We have a new collection of EK80 test files with different combinations of configuration settings from 3 transducers. Let's incorporate them into the convert testing to ensure robustness.

Issue #244, opened 2021-01-28
Tests take too long to run


lsetiawan:
## Overview

The current tests takes way too long to run, especially when we have to interact with an external server.

## Suggestions

I was thinking maybe we pick a few sample data and then stand up a local http server that serve those datasets to test against, rather from an external data repository. This way we still can test for the http/s3 read functionality without actually hitting a live server, which sometimes can timeout.

leewujung:
Yes I was going to raise this issue. :)  I think it's a good idea to go with a local server as you suggestion. Also I wonder if we can move those files currently in GitLFS there too, so that we don't run into the bandwidth limit (I found out that what I referred to before was the GitLFS limit and not GitHub actions).

emiliom:
Are the latencies primarily on the external server end + download, or is some of it due to the size of the raw files (ie, processing time, not network latency)?

What are the smallest raw files we can use for meaningful testing? As I mentioned in https://github.com/OSOceanAcoustics/echopype/pull/240#issuecomment-765560547, we can use a 25 MB Hake Survey EK60 raw file. Do we *need* to test against a much larger OOI raw file, at least for EK60?? It doesn't seem like it. I haven't worked with AZFP files, so I have no clue there. And for broadband EK80, I can imagine that getting a small-ish raw file might be challenging.

Regardless, a "local" http server for testing seems like a good idea, assuming it's not a substantial new effort (and assuming I understand what that means).

lsetiawan:
> Are the latencies primarily on the external server end + download, or is some of it due to the size of the raw files (ie, processing time, not network latency)?

I think there are sometimes network latency or timeouts for the data straight from OOI. For example:
https://github.com/OSOceanAcoustics/echopype/runs/1788810887?check_suite_focus=true#step:12:1916

I'm not sure about the others, but I think local network is better than outside when testing code, that way we don't depend on them and changes can be managed.

> we can use a 25 MB Hake Survey EK60 raw file
Is this the one from `ncei-wcsd-archive` bucket? If so, I am using that at the moment.

**EK60** https://github.com/OSOceanAcoustics/echopype/blob/fa081884d0de3a6a41fa79813be6245e656de2a0/echopype/tests/test_convert.py#L324-L335

**EK80** https://github.com/OSOceanAcoustics/echopype/blob/fa081884d0de3a6a41fa79813be6245e656de2a0/echopype/tests/test_convert.py#L472-L479

[...]

PR #523, opened 2021-12-18
Update tests

lsetiawan:
This PR updates the whole tests to use new files and perform some clean up.

## Todo
- [x] #244
- [x] #501 
- [x] #502
- [x] #544


leewujung:
@lsetiawan -- sorry for mission creep on you, but it's small ;) 
Could you add a conversion test for an EA640 file?
I've put in the 10.3 MB file @cywhale provided into the CI_test_data folder below:
https://drive.google.com/drive/u/0/folders/1qGiUpD205K5PV89m6nmUA5go7hhfeU3f

Thought it's better for you to do it since it should be straight forward with the fixture?

Another reason is that #539 does overlap with the fix you have for allowing ES80 files in sonar_model, so when you merge `dev` there will be some small conflicts.

Thanks! I'll review everything in one go after that.

echopype/tests/conftest.py
```diff
@@ -1,13 +1,47 @@
...
+    ek80_files = ek80_new_path.glob("**/*.raw")
+    if "ek80_file" in metafunc.fixturenames:
+        metafunc.parametrize(
+            "ek80_file", ek80_files, ids=lambda f: str(f.name)
+        )
+
+
+@pytest.fixture(scope="session")
+def ek80_file(request):
+    return request.param
```
leewujung, line 21: Since `ek80_file` is only used in `test_convert_ek80.py`, it seems that these can be moved to that file? 

Also, could you change `ek80_file` to `ek80_new_file` just so that it is clear which set of files are being referred to here?

echopype/tests/convert/test_convert.py
```diff
@@ -354,3 +443,168 @@ def test_convert_ek80(
...
+        ipath = es80_input_paths[0]
+
+    input_storage_options = (
+        common_storage_options if ipath.startswith("s3://") else {}
+    )
+    if output_save_path and output_save_path.startswith("s3://"):
+        output_storage_options = common_storage_options
+
+    # Only using one file
+    ec = open_raw(
```
leewujung, line 523: Change to `ec` to `echodata` so that it is consistent with other tests. Please change the other occurrences in this file also.

echopype/tests/convert/test_convert.py
```diff
@@ -114,43 +281,18 @@ def test_convert_time_encodings(sonar_model, raw_file, xml_path):
...
-)
 def test_convert_ek60(
-    model,
-    input_path,
+    ek60_input_paths,
     export_engine,
     output_save_path,
     minio_bucket,
+    model="EK60",
 ):
```
leewujung, line 290: Not trying to make this over complicated, but the content of `test_convert_*` (* is an allowable sonar_model) is identical except for the AZFP case where an XML file is needed. It seems that these can be consolidated also using the power of fixtures so later changes are easier to maintain?

leewujung:
@lsetiawan : this `pytest.fixture` is awesome!! Really make things a lot cleaner. Thanks for putting in the effort to upgrade the tests!

I added some inline comments above.

A couple more overall points:
- I think it'd be a good idea to rename `test_convert.py` to something like `test_convert_source_target_locs` to make it obvious the kind of tests within this module (vs the other test_convert_SONAR_MODEL.py are testing for the actual values or files with irregular parameters along time or depth dimensions).
-

leewujung (commented):
@lsetiawan : this `pytest.fixture` thing is awesome!! Really make things a lot cleaner. Thanks for putting in the effort to upgrade the tests!

I added some inline comments above. In addition, I think it'd be a good idea to rename `test_convert.py` to something like `test_convert_source_target_locs` to make it obvious the kind of tests within this module -- in contrast with the other `test_convert_SONAR_MODEL.py` modules that are testing for the actual values or files with irregular parameters along time or depth dimensions).

leewujung:
@lsetiawan : The changes all look good! Re my last comment, let me know if you'd like that to be in this PR or a separate one. We can merge this if the latter. Thanks!

lsetiawan:
Which last comment are you referring to? I renamed the test file.

leewujung:
Oh! Somehow I just now saw that one. Then we are good to go now! 🚀

PR #525, opened 2021-12-30
Overhaul and bug fixes for utils.uwa

leewujung:
The `utils.uwa` subpackage contains some bugs and the docstring and formatting also need an overhaul. Tests are also needed. Since the changes are relatively granular and standalone, current goal is to do this in 1 PR.

## Tasks
- [x] fix bug FG sound absorption calculation
- [x] overhaul docstring
- [x] overhal code formatting
- add tests for uwa functions
   - [x] absorption
   - [x] sound speed

### DIFF
--- a/echopype/tests/utils/test_utils_uwa.py
+++ b/echopype/tests/utils/test_utils_uwa.py
@@ -35,7 +35,9 @@ def test_absorption(frequency, temperature, salinity, pressure, pH, tolerance, t
 
     # AZFP values are an order of magnitude larger than the other 2
     assert np.all(
-        np.abs([abs_dB_m["AM"] - abs_dB_m["AZFP"], abs_dB_m["FG"] - abs_dB_m["AZFP"]]) < tolerance_AZFP
+        np.abs(
+            [abs_dB_m["AM"] - abs_dB_m["AZFP"], abs_dB_m["FG"] - abs_dB_m["AZFP"]]
+        ) < tolerance_AZFP
     )
 
 
--- a/echopype/tests/utils/test_utils_uwa.py
+++ b/echopype/tests/utils/test_utils_uwa.py
@@ -2,7 +2,7 @@ import pytest
 
 import numpy as np
 
-from echopype.utils.uwa import calc_absorption
+from echopype.utils.uwa import calc_absorption, calc_sound_speed
 
 
 # Tolerance used here are set empirically
@@ -32,3 +32,27 @@ def test_absorption(frequency, temperature, salinity, pressure, pH, tolerance):
             formula_source=fm,
         )
     assert np.abs(abs_dB_m["AM"] - abs_dB_m["FG"]) < tolerance
+
+
+# These tests are just to make sure the code runs
+# and the difference between formula is not too dramatic,
+# instead of trying to verify the actual numbers
+# as we don't know the source of the AZFP formula.
+@pytest.mark.parametrize(
+    "temperature, salinity, pressure, tolerance",
+    [
+        (27, 35, 10, 0.07),
+        (27, 35, 100, 0.07),
+        (5, 35, 3500, 0.5),
+    ],
+)
+def test_sound_speed(temperature, salinity, pressure, tolerance):
+    c = dict()
+    for fm in ["Mackenzie", "AZFP"]:
+        c[fm] = calc_sound_speed(
+            temperature=temperature,
+            salinity=salinity,
+            pressure=pressure,
+            formula_source=fm
+        )
+    assert np.abs(np.diff(list(c.values()))) < tolerance
--- a/echopype/tests/utils/test_utils_uwa.py
+++ b/echopype/tests/utils/test_utils_uwa.py
@@ -1,0 +1,34 @@
+import pytest
+
+import numpy as np
+
+from echopype.utils.uwa import calc_absorption
+
+
+# Tolerance used here are set empirically
+# so the test captures a sort of current status
+@pytest.mark.parametrize(
+    "frequency, temperature, salinity, pressure, pH, tolerance",
+    [
+        (18e3, 27, 35, 10, 8, 3e-5),
+        (18e3, 27, 35, 100, 8, 3e-5),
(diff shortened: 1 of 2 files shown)


emiliom:
I'll take a look today or tomorrow. FYI I haven't used these functions previously, but they look straightforward.

emiliom:
FYI I'm preparing a commit or two that I will push to your PR. In the meantime, I'll make some in-line comments

echopype/utils/uwa.py
```diff
@@ -56,73 +64,91 @@ def calc_sound_speed(
...
+            / (1 + 0.0018 * (salinity - 35))
+        )
         P3 = 1.0 - 3.83e-5 * pressure + 4.9e-10 * pressure * pressure
         if temperature < 20:
             A3 = (
                 4.937e-4
                 - 2.59e-5 * temperature
-                + 9.11e-7 * temperature ** 2
-                - 1.5e-8 * temperature ** 3
+                + 9.11e-7 * temperature * temperature
```
emiliom, line 135: Why change the formula format to x * x rather than x ** 2? I know it's a matter of style. IMHO using exponents is cleaner and easier to grasp. But beyond style preference, exponents are also used in calc_sound_speed, so making the change here will make the code style inconsistent.
I'll revert your change in my commit, and will also change pressure * pressure in the above lines to pressure ** 2 while I'm at it. You can revert/reject those changes if you feel strongly about it.

echopype/utils/uwa.py
```diff
@@ -56,73 +64,91 @@ def calc_sound_speed(
...
+
+    The accuracy of the simplified formula from Ainslie & McColm 1998
+    compared with the original complicated formula from Francois & Garrison 1982
+    was demostrated between 100 Hz and 1 MHz.
     """
-    # fmt: off
     if formula_source == "FG":
         f = frequency / 1000.0
-        d = distance / 1000.0
+        d = 1 / 1000.0  # ref to 1 m
```
emiliom, line 118: > # ref to 1 m
Expand a bit to make this clearer. "Referenced to 1 m"?

echopype/utils/uwa.py
```diff
@@ -56,73 +64,91 @@ def calc_sound_speed(
...
+                - 6.5e-10 * temperature * temperature * temperature
             )
         a = (
             A1 * P1 * f1 * f * f / (f1 * f1 + f * f)
             + A2 * P2 * f2 * f * f / (f2 * f2 + f * f)
             + A3 * P3 * f * f
         )
-        # convert to db/m from db/km
-        sea_abs = -20 * np.log10(10 ** (-a * d / 20.0)) / 1000
+        sea_abs = a * d
```
emiliom, line 150: Wow, that's way simpler! I had to convert the old formula by hand to confirm that they're the same -- except for the `/1000` that you've dropped, but it looks like it all balances out by the change to the `d` formula

echopype/tests/utils/test_utils_uwa.py
```diff
@@ -0,0 +1,58 @@
...
+    for fm in ["AM", "FG", "AZFP"]:
+        abs_dB_m[fm] = calc_absorption(
+            frequency=frequency,
+            temperature=temperature,
+            salinity=salinity,
+            pressure=pressure,
+            pH=pH,
+            formula_source=fm,
+        )
+    assert np.abs(abs_dB_m["AM"] - abs_dB_m["FG"]) < tolerance
```
emiliom, line 34: This assertion only tests AM vs FG; AZFP doesn't come into play. So, you're not testing AZFP, per se, just confirming that it runs (per the comment below)?

echopype/tests/utils/test_utils_uwa.py
```diff
@@ -0,0 +1,58 @@
...
+def test_sound_speed(temperature, salinity, pressure, tolerance):
+    c = dict()
+    for fm in ["Mackenzie", "AZFP"]:
+        c[fm] = calc_sound_speed(
+            temperature=temperature,
+            salinity=salinity,
+            pressure=pressure,
+            formula_source=fm
+        )
+    assert np.abs(np.diff(list(c.values()))) < tolerance
```
emiliom, line 58: This is the same as `assert np.abs(c["Mackenzie"] - c["AZFP"]) < tolerance`, right? I think that form would be both easier to read and consistent with what's done in `test_absorption`

emiliom (commented):
I'm done with my review, including the commit I've pushed with suggested changes.

echopype/tests/utils/test_utils_uwa.py
leewujung: Yep it is the same, in fact that was the original form there, until I found that the AZFP one is very different from the other two. I've changed this.

echopype/tests/utils/test_utils_uwa.py
leewujung: Oh good catch! I don't know how I forgot to add that AZFP parts. The values are about an order of magnitude different from the other two. I'll add the AZPF parts and add a comment about the difference as well.

echopype/utils/uwa.py
leewujung: I didn't actually dive into the code (not from me!) but took the lazy route to fix bugs via testing... Looking at this again, I'd rather things as close to the original paper as possible instead of this fix at the end. The original output (in the paper) was in units [dB / km] so it the scaling should just be at the end. I've fix this one and will simplify the formula too: it is pretty awful to have things like `temperature * temperature * temperature`.

leewujung: Revised!

echopype/utils/uwa.py
leewujung: Ended up removing `d` as it's just a scaling factor that can be applied at the end.

echopype/tests/utils/test_utils_uwa.py
leewujung: Done!

emiliom:
Merge whenever you'd like

leewujung:
Alright, thanks for reviewing this @emiliom! I'll merge it now.

Issue #526, opened 2022-01-12
Make matplotlib an optional dependency


emiliom:
`matplotlib` is listed in the [requirements.txt](https://github.com/OSOceanAcoustics/echopype/blob/main/requirements.txt#L2). I suggest we replace it with the lighter weight [matplotlib-base](https://stackoverflow.com/questions/61325817/differences-between-matplotlib-and-matplotlib-base) (matplotlib minus `pyqt`)

Any objections?

leewujung:
Sounds good!

emiliom:
After realizing that `matplotlib-base` is only available as a conda package (not as a PyPI package), running into build errors in PR #527, and discussing this further, we've decided to go in a different direction: to make `matplotlib` an optional dependency. We'll add matplotlib to https://github.com/OSOceanAcoustics/echopype/blob/main/setup.py#L16, like this:
```python
EXTRA_REQUIRES = {"dev": dev_reqs, "plot": ["cmocean", "matplotlib"]}
```

emiliom:
Addressed in #529

PR #527, opened 2022-01-13
replace matpplotlib with matplotlib-base in requirements

emiliom:
Addresses #526

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 dask[array]
-matplotlib
+matplotlib-base
 netCDF4
 numpy
 pynmea2


emiliom:
Hmm. It looks like `matplotlib-base` is only available on conda, not on pypi (pip). But much/all of the test infrastructure creates an environment using pip, so all tests (and readthedocs) fail at that point.

This issue & PR are not important. It just seemed like a very easy, nice improvement to the packaging. But I can withdraw the PR unless someone has an easy solution. It's not worth spending much time on this, IMHO.

emiliom:
We've decided to go in a different direction, to make matplotlib an optional dependency. We'll add matplotlib to https://github.com/OSOceanAcoustics/echopype/blob/main/setup.py#L16, like this:
```python
EXTRA_REQUIRES = {"dev": dev_reqs, "plot": ["cmocean", "matplotlib"]}
```

@lsetiawan is that correct?

lsetiawan:
Yepp that is correct 👍🏽

emiliom:
Thanks. I'll close this PR and create a new one by tomorrow.

PR #530, opened 2022-01-14
Remove the extra space before colon in parameters docstring

leewujung:
Currently in the API references, the parameter data types are not rendered correctly due to an extra space immediately after the parameter and before the ":" -- this is somewhat unexpected since that format "parameter : datatype" is what's in the [style guide](https://numpydoc.readthedocs.io/en/latest/format.html), but perhaps `automodapi` has its own mind?

### DIFF
--- a/echopype/calibrate/api.py
+++ b/echopype/calibrate/api.py
@@ -68,10 +68,10 @@ def compute_Sv(echodata: EchoData, **kwargs) -> xr.Dataset:
 
     Parameters
     ----------
-    echodata : EchoData
+    echodata: EchoData
         An `EchoData` object created by using `open_raw` or `open_converted`
 
-    env_params : dict, optional
+    env_params: dict, optional
         Environmental parameters needed for calibration.
         Users can supply `"sound speed"` and `"absorption"` directly,
         or specify other variables that can be used to compute them,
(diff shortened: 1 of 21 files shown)


emiliom:
Hmm. I've compared the RTD build resulting from this PR, vs the current, production build (default/stable version). Here's a comparison for `uwa.sound_speed`:
- PR build: https://echopype--530.org.readthedocs.build/en/530/api/echopype.utils.uwa.calc_sound_speed.html#echopype.utils.uwa.calc_sound_speed
- Production (stable): https://echopype.readthedocs.io/en/v0.5.5/api/echopype.utils.uwa.calc_sound_speed.html#echopype.utils.uwa.calc_sound_speed

I don't see a significant difference in how parameters and parameter types are handled. There are other style differences, and I actually like the current style better (I prefer that the "Parameters" and "Returns" headings use a style that's different from the one used by the parameters themselves).

Am I missing something??

emiliom:
I'm getting more confused by the minute ...

First, I made a mistake above when I listed the url for the production/stable version. I listed the **v0.5.5** url instead! The correct stable version url is https://echopype.readthedocs.io/en/stable/api/echopype.utils.uwa.calc_sound_speed.html#echopype.utils.uwa.calc_sound_speed. And here, I see that the parameter names and type are combined into a single string (wrong, for sure), and the font style for the "Parameters" and "Returns" heading is the same as for the parameters. Huh?! A different style and parameter parsing is being used for the tagged-release versions (and `latest`) compared to `stable`. There other differences too, such as `stable` -- incorrectly -- not using lists in the front-page TOC whereas the other versions do.

Something bigger is going on here :exploding_head:

leewujung:
@b-reyes : this is the PR I mentioned which looks into the docstring problem. You can see that @emiliom did some more comparison and found some weird discrepancy -- this is why I think it's better to discuss this after you've read through this.

@emiliom : this came up after you logged off, @b-reyes thought that he can help with this based on his experience with sphinx. 

Thanks both!

Issue #531, opened 2022-01-19
open_raw of EA640 encounter self.parser_obj.environment["temperature"]], KeyError: 'temperature'


cywhale:
I have some .raw files from two ships, both by EA640. For one ship open_raw is ok, but fails for the other.
I use echopype version 0.5.5
```
ed = ep.open_raw(
    raw_file="xxx.raw", 
    sonar_model="EK80", 
)
``` 
Error message:
```
KeyError                                  Traceback (most recent call last)
/tmp/ipykernel_70694/522230037.py in <module>
      1 #echodata = ep.open_raw(raw_file=ek_nor3, sonar_model="ek80")
----> 2 ed0 = ep.open_raw(
      3     raw_file=rawfiles[0],
      4     sonar_model="EK80",
      5 ) #files[0] #just debug

~/py3.8/lib/python3.8/site-packages/echopype/convert/api.py in open_raw(raw_file, sonar_model, xml_path, convert_params, storage_options)
    441             sonar_model=sonar_model, date_created=parser.ping_time[0]
    442         )
--> 443     echodata.environment = setgrouper.set_env()
    444     echodata.platform = setgrouper.set_platform()
    445     if sonar_model in ["EK60", "ES70", "EK80", "ES80"]:

~/py3.8/lib/python3.8/site-packages/echopype/convert/set_groups_ek80.py in set_env(self, env_only)
     29                 "temperature": (
     30                     ["ping_time"],
---> 31                     [self.parser_obj.environment["temperature"]],
     32                 ),
     33                 "depth": (["ping_time"], [self.parser_obj.environment["depth"]]),

KeyError: 'temperature'
```

Is there something wrong with the .raw? or any way to prevent this error if I just want to parse the NMEA log out from the raw file. Thanks.

cywhale:
I just found a much smaller .raw that can be tested for this issue and uploaded to my repo:
https://github.com/cywhale/ODB/blob/ff7aa49e11403ae10d17729d211598631dbe8691/EK/test.raw

```
echodata = ep.open_raw(raw_file="test.raw", sonar_model="ek80")
```
and got KeyError: 'temperature'
FYI

leewujung:
@cywhale : Awesome, thanks for reporting this and providing the test file!
I wonder whether it is just that that key does not exist in the file so the parsed dict does not content that. Let me take a look.

leewujung:
[...]

PR #537, opened 2022-01-22
Allow flexible environment variables in EA640/EK80 files

leewujung:
This PR changes `SetGroupsEK80.set_env` to dynamically generates the dataset content. This addresses #531.

Still need to add a corresponding test to the convert tests.
@cywhale: If you have access to the recording software, could you try to see if you could further reduce the test file size?

### DIFF
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -24,20 +24,14 @@ class SetGroupsEK80(SetGroupsBase):
         ping_time = np.array([ping_time.astype("datetime64[ns]")])
 
         # Collect variables
+        tmp_env = self.parser_obj.environment.copy()
+        tmp_env.pop("timestamp")
+        tmp_env.pop("xml")
+        dict_env = dict()
+        for k, v in tmp_env.items():
+            dict_env[k] = (["ping_time"], [v])
         ds = xr.Dataset(
-            {
-                "temperature": (
-                    ["ping_time"],
-                    [self.parser_obj.environment["temperature"]],
-                ),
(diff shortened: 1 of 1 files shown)


cywhale:
> @cywhale: If you have access to the recording software, could you try to see if you could further reduce the test file size?

@leewujung No, I cannot access. But I have other .raw in disks, but usually larger than 100M. I uploaded another EA640 .raw of anther ship ~ 14Mb in  https://github.com/cywhale/ODB/blob/master/EK/test_v3.raw

leewujung:
Got it. No worries. We can probably include the 10 MB file for now, and if you could try to ask whether it's possible that folks operating the instrument next time collect a couple very small files, that would be great!

Tagging @lsetiawan here since we're talking about test file size.

echopype/convert/set_groups_ek80.py
```diff
@@ -24,20 +24,13 @@ def set_env(self, env_only=False) -> xr.Dataset:
...
-                    [self.parser_obj.environment["temperature"]],
-                ),
-                "depth": (["ping_time"], [self.parser_obj.environment["depth"]]),
-                "acidity": (["ping_time"], [self.parser_obj.environment["acidity"]]),
-                "salinity": (["ping_time"], [self.parser_obj.environment["salinity"]]),
-                "sound_speed_indicative": (
-                    ["ping_time"],
-                    [self.parser_obj.environment["sound_speed"]],
-                ),
-            },
```
emiliom, line 40: See my comment elsewhere: https://github.com/OSOceanAcoustics/echopype/pull/539#pullrequestreview-860237299

Issue #538, opened 2022-01-22
Allow setting `sonar_model="EA640"` properly


leewujung:
Right now we have a discrepancy in the codebase where in `core.py` we set EA640 to be one of the sonar models, but actually did not include this in `open_raw` properly. Instead, in the docstring we ask users to use `sonar_model="EK80"` for EA640 files. This results in the `keywords` variable in the top level group being set to "EK80" instead of "EA640".

Additionally, we have a duplicated line in the `open_raw` docstring for EK80/EA640 files.

I found out about these while looking into #531.

PR #539, opened 2022-01-22
Make "EA640" an allowable `sonar_model` input to `open_raw`

leewujung:
This PR makes it possible to use `sonar_model="EA640"` in `open_raw`. The docstring is also updated.

This PR addresses #538.

### DIFF
--- a/echopype/convert/api.py
+++ b/echopype/convert/api.py
@@ -432,7 +432,7 @@ def open_raw(
     # Set up echodata object
     echodata = EchoData(source_file=file_chk, xml_path=xml_chk, sonar_model=sonar_model)
     # Top-level date_created varies depending on sonar model
-    if sonar_model in ["EK60", "ES70", "EK80", "ES80"]:
+    if sonar_model in ["EK60", "ES70", "EK80", "ES80", "EA640"]:
         echodata.top = setgrouper.set_toplevel(
             sonar_model=sonar_model, date_created=parser.config_datagram["timestamp"]
         )
@@ -442,7 +442,7 @@ def open_raw(
         )
     echodata.environment = setgrouper.set_env()
     echodata.platform = setgrouper.set_platform()
-    if sonar_model in ["EK60", "ES70", "EK80", "ES80"]:
+    if sonar_model in ["EK60", "ES70", "EK80", "ES80", "EA640"]:
         echodata.nmea = setgrouper.set_nmea()
     echodata.provenance = setgrouper.set_provenance()
     echodata.sonar = setgrouper.set_sonar()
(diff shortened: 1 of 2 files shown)


echopype/convert/api.py
```diff
@@ -338,8 +338,8 @@ def open_raw(
 
         - ``EK60``: Kongsberg Simrad EK60 echosounder
         - ``ES70``: Kongsberg Simrad ES70 echosounder
-        - ``EK80``: Kongsberg Simrad EK80 and Kongsberg EA640 echsoounders
-        - ``EK80``: Kongsberg Simrad EK80 and Kongsberg EA640 echsoounders
+        - ``EK80``: Kongsberg Simrad EK80 echsoounder
```
emiliom, line 341: Spelling mistake, echsoounder

echopype/convert/api.py
```diff
@@ -338,8 +338,8 @@ def open_raw(
...
+        - ``EA640``: Kongsberg EA640 echsoounder
```
emiliom, line 342: Spelling mistake, echsoounder

echopype/convert/set_groups_ek80.py
```diff
@@ -24,20 +24,13 @@ def set_env(self, env_only=False) -> xr.Dataset:
...
-                    ["ping_time"],
-                    [self.parser_obj.environment["temperature"]],
-                ),
-                "depth": (["ping_time"], [self.parser_obj.environment["depth"]]),
-                "acidity": (["ping_time"], [self.parser_obj.environment["acidity"]]),
-                "salinity": (["ping_time"], [self.parser_obj.environment["salinity"]]),
-                "sound_speed_indicative": (
-                    ["ping_time"],
-                    [self.parser_obj.environment["sound_speed"]],
-                ),
```
emiliom, line 39: Note that your new, more compact scheme doesn't work for sound speed, b/c the input and output variable names are different: `sound_speed` vs `sound_speed_indicative`.  That is, unless the original code was wrong; but in that case I suggest you make a note that you're also correcting an error.

echopype/convert/api.py
leewujung: Hah, thanks for catching this, I guess I wasn't actually reading! :D

echopype/convert/set_groups_ek80.py
leewujung: This with the last one clearly showed that I wasn't awake... -_-

It should be `sound_speed_indicative` as defined in the convention. If there's absorption it should be `absorption_indicative`. I'll change that.

leewujung: I'll include this as part of #537.

echopype/convert/api.py
leewujung: Done!

PR #543, opened 2022-01-28
Upgrade echopype docs to use jupyter book

lsetiawan:
## Overview

This PR completely upgrades the echopype docs to the jupyter book (jb) system.
1. All sphinx and jb configurations are now made in `_config.yaml`
2. Table of contents for sidebar is now configured within `_toc.yaml`
3. Pre-commit hook has been setup so that `conf.py` is generated and committed along with changes so that RTD can read it.
4. Now we can use a combination of `.ipynb`, `.rst`, `.md`, etc to create our documentation. Example is I've ported over `open-converted.rst` to `open-converted.ipynb` and use the code from `echodata_html_repr.ipynb` to generate the beautiful EchoData object representation! 🥳 

## Next steps

Once we know for sure this works, maybe we should add in the examples within the docs to have one cohesive documentation! 😄


leewujung:
@lsetiawan : This looks great based on a quick peek! I like how it provides a very clean look. The fact that it is seamless with the original setup doubles the incentive. 

Let's see how executable code blocks are handled in this, or alternatively we can just write things as notebooks.

Is the location of the version menu configurable? (I haven't tried looking into that)  It is on the right and the one from RTD before was on the left. Just wondering since people might miss it.

The correct rendering of docstring (including both the colon problem and headings like Parameters and Returns) takes care of #530, so we can just close that PR. 🤓  

Tagging @emiliom @b-reyes

lsetiawan:
> Is the location of the version menu configurable? (I haven't tried looking into that) It is on the right and the one from RTD before was on the left. Just wondering since people might miss it.

I don't think it is. I tried looking into the css, but not sure where the update should be 😞 I think this is pretty standard now with this theme, so I'm thinking if we want something custom, probably better to discuss within the upstream repo: https://github.com/executablebooks/sphinx-book-theme/

lsetiawan:
> This means we would no longer need to use [skip ci] for docs testing, right?

That is correct.

emiliom (approved):
Looks good

b-reyes (approved):

Issue #483, opened 2021-11-09
echodata.compute_range() returns full matrix while some backscatter entries are NaN-padded


gavinmacaulay:
The EK80 test file, D20170912-T234910.raw, has two channels (70 and 200 kHz) with different sample rates (about a x3 difference). When using the echodata.compute_range() function, the returned range vector for the 70 kHz channel is incorrect - it has the same number of range values as the 200 kHz channel, when it should have about a third less. Due to how the range vector is calculated, the actual ranges in the 70 kHz vector are, per-sample, a third of what they should be. The dropna() method used when extracting the Sv at 70 kHz doesn't work for the 70 kHz range vector due to this.

Code, derived from test_calibrate.py, that shows this is:

```
from pathlib import Path
import echopype as ep

ek80_path = Path('./test_data/ek80')
ek80_raw_path = ek80_path.joinpath('D20170912-T234910.raw')

echodata = ep.open_raw(ek80_raw_path, sonar_model='EK80')
ds_Sv = ep.calibrate.compute_Sv(echodata, waveform_mode='BB', encode_mode='complex')

Sv_70k = ds_Sv.Sv.isel(frequency=0, ping_time=0).dropna('range_bin').values

Sv_range = echodata.compute_range(env_params=[], ek_waveform_mode='BB')
Sv_70k_range = Sv_range.isel(frequency=0, ping_time=0).values
Sv_200k_range = Sv_range.isel(frequency=1, ping_time=0).values

print(f'Backscatter data at 70 kHz has {Sv_70k.shape[0]} samples')
print(f'But range vector at 70 kHz has {Sv_70k_range.shape[0]} samples')
print(f'Range vector at 200 kHz has {Sv_200k_range.shape[0]} samples')

print(f'Range at max sample for 70 kHz is {Sv_70k_range[-1]:.2f} m')
print(f'Range at max sample for 200 kHz is {Sv_200k_range[-1]:.2f} m')
```

leewujung:
@gavinmacaulay : I looked into this and think the calculation itself is correct. The actual range vector (in meters) in the computed Sv dataset is stored in the variable `range`, and it is calculated using echodata.compute_range() under the hood. When `range` is used in conjunction with the computed Sv data, only the first 8284 range samples will be used, since there is no data after that for the 70kHz channel (NaN-padded dataarray).

[...]

PR #547, opened 2022-01-31
Set range entries with no backscatter data to `NaN` in output of `echodata.compute_range()`

leewujung:
This PR sets entries in the `range` data variable in the output dataset from `compute_Sv` that does not have corresponding backscatter entries (also set to NaN) to NaN. Some backscatter entries are NaN due to changes in echosounder settings, such as data collection range, transmit signal duration, etc. 

The change addresses the potential confusion #483.

## Tasks
- [x] set `range` entries with no backscatter data to `NaN` in `echodata.compute_range()`
- [x] update tests to test for NaN entries


leewujung:
@lsetiawan : In our current testing setup, since this PR only "touches" `echodata.py`, only tests under `tests/echodata` were run. In practice, the method (`compute_range()`) being updated here is used in the calibration functions, so tests under `tests/calibrate` should also be run. Right now the entire set of tests are only run when merged to `main` so the calibration tests will only be run then.

What's the best practice here? I think we were tripped a few times by this (i.e., additional fixes needed when merging `dev` to `main` due to the adaptive test not catching bugs that are downstream from the code being modified in the PR).

Can we set up something that pulls out the downstream pieces that uses the changed functions  and run the corresponding tests? If we do this, at what stage this should be run -- when it is merged to `dev`, or as a label such as what is done with putting on the label `Needs Complete Testing`?

leewujung:
@lsetiawan : I merged your #523 and used the fixture structure for the test here. Also added an EK60 file (`test_data/ek60/Winter2017-D20170115-T150122.raw`) that contains NaN entries in addition to the EK80 file (`test_data/ek80/D20170912-T234910.raw`) that we already have there.

lsetiawan:
> What's the best practice here? I think we were tripped a few times by this (i.e., additional fixes needed when merging dev to main due to the adaptive test not catching bugs that are downstream from the code being modified in the PR).

The CI is now setup to run all tests when merged to dev as shown here: https://github.com/OSOceanAcoustics/echopype/blob/0f699cce311c5b86a897f167496ab852f7d2fbb1/.github/workflows/ci.yaml#L91-L94

So as a first step, as long as the isolated module test works you can merge to dev for a full test. Or you can always just use the `Needs complete testing` label with the PR if you know that particular function touches many different things. In practice, most software CI actually runs everything anyways to make sure your changes doesn't break anything, we just decided to make this custom, brittle, intelligence 😜

emiliom:
> Or you can always just use the `Needs complete testing` label with the PR if you know that particular function touches many different things.

And whenever you have a need for that mechanism but don't remember how to use it, go to the docs: https://echopype.readthedocs.io/en/stable/contributing.html#continuous-integration-github-actions

leewujung:
Hmm, we should revisit this soon, in terms of how/what we want the tests to be set up. The past few times tests failed when merging to `main` I think was after the supposedly complete test when merging to `dev`, and we opted for fixing the bugs for obvious reasons then.

In theory we can get out of the current brittle setup once most big test data files are updated to a smaller size. I updated 2 major ones, but ran out of time to do the last couple files before I had to return a hardware dongle needed to generate data we test against -- the next time I can do this is likely late March - April. 

BUT regardless, it is unlikely we'll be able to reduce all files to be KB size like the new EK80 test files I collected, since we don't have access to many instrument variations that were used to collect the diverse test files.

lsetiawan (approved):
LGTM! Thanks for this awesome work. I need to think about the whole testing scheme and see how we can improve it 😄 Would you mind, pasting all of the thoughts from here to an issue so we can discuss there? Thanks @leewujung

lsetiawan:
> The past few times tests failed when merging to main I think was after the supposedly complete test when merging to dev, and we opted for fixing the bugs for obvious reasons then.

Just checked this. The tests failed the past few times on main due to the publish to PyPi action issues, which is more of a packaging issue rather than code. I think I need to figure out how to activate that within the dev branch also so it imitates main. In terms of full CI testing PR #411 fixed the missing testing since the test for all testing are actually run directly with pytest rather than the small script I created.

leewujung:
> Just checked this. The tests failed the past few times on main due to the publish to PyPi action issues, which is more of a packaging issue rather than code. 

Thanks for checking @lsetiawan , I must have confused that and the other earlier ones! Great that there was no unexpected behavior!

And yes I'll paste the thoughts we've had here to a new issue. :)

PR #555, opened 2022-02-07
Enable the handling of nans for visualize of echodata and Sv

lsetiawan:
## Overview

This PR handles the `nan` values when trying to plot `backscatter_r` or `Sv` with `get_range=True`. It basically drops the nans and plot each frequency individually. The `create_echogram` API now returns a list rather than just a facet grid or quadmesh.

### DIFF
--- a/echopype/tests/visualize/test_plot.py
+++ b/echopype/tests/visualize/test_plot.py
@@ -183,7 +183,6 @@ def test_plot_mvbs(
     mvbs = echopype.preprocess.compute_MVBS(Sv, ping_time_bin='10S')
 
     plots = []
-    plots = echopype.visualize.create_echogram(mvbs)
     try:
         plots = echopype.visualize.create_echogram(mvbs)
     except Exception as e:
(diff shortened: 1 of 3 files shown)


leewujung:
Thanks @lsetiawan ! 
The change from just facet grid or quadmesh to a list of them looks fine, but I do have a complaint about the gigantic output figures for each individual frequency. :P 
Could you please change it to have the original multi-panel layout? like with each row a channel. It is also more useful for users.

leewujung:
> It does have the multi-panel layout... I guess I just have to divide the sizing when it loops over each frequency 😛

When I ran the tests I got multiple plots (each for one frequency) instead of multiple panels. I did get multiple panels (by column) when it is EK80 complex data since it has 4 quadrants. Are these not the expected behavior?

leewujung:
> I think for now we should just provide users an MVP for quick plotting and in the future we can give a more thorough plotting library. I do return the individual FacetGrids so that someone that have the matplotlib chops can still customize their plot.

The easiest is probably to just set up multiple axes depending on the number of frequency channels and the number beams/quadrant and plot there. But I like the suggestion to push this to another PR later, so that we can get this release out first, while still maintaining the basic ability.

PR #556, opened 2022-02-08
Update test_convert_source_target_locs to enable parallel testing

lsetiawan:
## Overview

This PR updates the creation of temporary folders within `test_convert_source_target_locs` so that the tests can run in parallel without any issues.

## Current dev local test

Testing our current `dev` branch locally resulted in all tests being run within ~5 minutes rather than the current time of ~26 min.

Screenshot from 2022-02-08 10-14-44

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,7 +91,7 @@ jobs:
       - name: Running All Tests
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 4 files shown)


lsetiawan:
**UPDATE**: I realized that locally, the 5min was because the ad2cp tests aren't actually run since the files are already run. With the actual run of ad2cp tests. We got ` ====== 2 failed, 518 passed, 8 xfailed, 489 warnings in 808.85s (0:13:28) ======` based on the last run within Github Actions.

leewujung:
This is awesome!! Looking forward to merging this after #555.

PR #557, opened 2022-02-09
Implement codespell and fix spellings

lsetiawan:
## Overview

This PR adds codespell for checking misspellings and fixes it as suggested by @leewujung 😄


leewujung:
Thanks @lsetiawan !

PR #559, opened 2022-02-09
Make visualize module fully optional

### LINKED ISSUE #558: Make visualize module optional
## Overview

Based on the packaging test: https://github.com/OSOceanAcoustics/echopype/runs/5132878099?check_suite_focus=true#step:6:148

Seems like `visualize` gets imported in the root of echopype `__init__.py`. This is a problem because it causes an error. Need fixing.

lsetiawan:
## Overview

This PR fixes #558 by removing direct import of visualize module within the root echopype `__init__.py` file. This means that user now has to specifically import this module by doing: `from echopype import visualize` or `import echopype.visualize`. In that case if the `plot` dependency extra reqs isn't installed, it will throw the `No module named 'matplotlib'` error.

### DIFF
--- a/echopype/__init__.py
+++ b/echopype/__init__.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 from _echopype_version import version as __version__  # noqa
 
-from . import calibrate, preprocess, visualize
+from . import calibrate, preprocess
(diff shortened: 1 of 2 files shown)


leewujung:
This looks fine so I'll merge it now. Thanks @lsetiawan !

Issue #542, opened 2022-01-27
Port echodata/convention/attrs.py to a reusable yaml file


emiliom:
I recently created a centralized module that stores attribute values (and related info) from SONAR-netCDF4 v1.0: [echodata/convention/attrs.py](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/echodata/convention/attrs.py). We've decided that it should be in a yaml file rather than a Python module, so it can be reused more easily across the board, specially by [echopydantic](https://github.com/OSOceanAcoustics/echopydantic)

lsetiawan:
This suggestion came from https://github.com/OSOceanAcoustics/echopype/pull/493#issuecomment-974294527.

emiliom:
@lsetiawan I've started working on this. I decided to reuse [convention/1.0.yml](https://github.com/OSOceanAcoustics/echopype/blob/main/echopype/echodata/convention/1.0.yml) and [convention/\_\_init__.py](https://github.com/OSOceanAcoustics/echopype/blob/main/echopype/echodata/convention/__init__.py) by porting [convention/attrs.py](https://github.com/OSOceanAcoustics/echopype/blob/main/echopype/echodata/convention/attrs.py) into `1.0.yml`. That won't impact anything else currently in use, as far as I can tell. However, it means that the yml file would have to be read (via `_get_convention`) every time some portion of it is needed. That seems inefficient. Is it possible (and a good practice) to find a way to "pre load it" always just once, automatically? Or do you think having to read the yml several times is not a big deal?

lsetiawan:
> "pre load it" always just once, automatically? Or do you think having to read the yml several times is not a big deal?

I think pre loading it is best, since it seems to me that these metadata will be used in several places. I also agree that your `attrs.py` content can go to the 1.0.yml. This is a great place for that. Thanks for working on this @emiliom.

emiliom:
Even though #565 has addressed the core thrust of this issue, I'm inclined to leave it open until we're wrapping up 0.6.0. I think there are related issues that may come up that we can discuss here.

PR #565, opened 2022-02-15
Consolidate convention specs into single yml file pre loaded via import

emiliom:
This is a first attempt at addressing #542. The tests are passing.

Let's not merge it. Mainly, I'd like @lsetiawan to take a good look at my use of `importlib.resources` to handle the loading of the consolidated yml content using `import` machinery, including caching after the first `import`. 

BTW, I'm not terribly happy with some names I picked (for a module, methods, etc), but I wanted to get this out for review before trying to make it cleaner.


echopype/echodata/convention/1.0.yml
```diff
@@ -44,3 +44,51 @@ groups:
...
+      standard_name: sound_frequency
+      units: Hz
+    valid_min: 0.0
+    ping_time:
+      long_name: Timestamp of each ping
+      standard_name: time
+      axis: T
+    range_bin:
+      long_name: Along-range bin (sample) number, base 0
+  platform_coord_default:
```
lsetiawan, line 61: Can all of this be put within the group itself? For example:

```yaml
  platform:
    name: Platform
    description: contains information about the platform on which the sonar is installed.
    ep_group: Platform
    attributes:
      coords:
        location_time:
          axis: T
          long_name: Timestamps for NMEA datagrams
          standard_name: time
      data_vars:
        latitude:
          long_name: Platform latitude
          standard_name: latitude
          units: degrees_north
          valid_range: (-90.0, 90.0)
        longitude:
          long_name: Platform longitude
          standard_name: longitude
          units: degrees_east
          valid_range: (-180.0, 180.0)
        pitch:
          long_name: Platform pitch
          standard_name: platform_pitch_angle
          units: arc_degree
          valid_range: (-90.0, 90.0)
        roll:
          long_name: Platform roll
          standard_name: platform_roll_angle
          units: arc_degree
          valid_range: (-90.0, 90.0)
        heave:
          long_name: Platform heave
          standard_name: platform_heave_angle
          units: arc_degree
          valid_range: (-90.0, 90.0)
        water_level:
          long_name: z-axis distance from the platform coordinate system origin to the sonar transducer
          units: m
```

echopype/echodata/convention/sonarnetcdf_1.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self):
+        """Prepare to read the convention yaml file"""
+        self._yaml_dict = {}
+        # Hardwired to 1.0, for now
+        self.version = "1.0"
+
+    @property
+    def yaml_dict(self):
+        """Read data from disk"""
+        if self._yaml_dict:  # Data has already been read, return it directly
```
lsetiawan, line 18: It is cool how things get cached! Nice!

echopype/convert/set_groups_ek60.py
```diff
@@ -6,18 +6,19 @@
...
-)
+from ..echodata.convention import sonarnetcdf_1
 from ..utils.coding import set_encodings
 
 # fmt: off
 from .set_groups_base import DEFAULT_CHUNK_SIZE, SetGroupsBase
 
 # fmt: on
 
+_varattrs = sonarnetcdf_1.conv.yaml_dict["variable_and_varattributes"]
```
lsetiawan, line 17: Do you see this `sonarnetcdf_1` to have more than the `conv` object? It seems a little awkward to call `.conv` after already getting the `sonarnetcdf_1`. I would expect more of a `yaml_dict` call on that to get to the convention yaml right away.

echopype/echodata/convention/1.0.yml
emiliom: We definitely can. I was trying to be the least disruptive in this first step. I'll have to see if there could be any negative consequences. There might be some redundancy involved.

echopype/convert/set_groups_ek60.py
emiliom: > Do you see this sonarnetcdf_1 to have more than the conv object?

Not at this time. I agree, it's awkward to have `sonarnetcdf_1.conv.yaml_dict`. I ran out of patience trying to figure out how to try to accomplish that via the `__init__.py` after many failed attempts. Can you suggest a solution?

emiliom:
One other thing. By defining module-scope variables (eg, [this](https://github.com/OSOceanAcoustics/echopype/pull/565/files#diff-310c3c66163a9ec9a8ba3606b681cdf1ceb358326744d75c9d42311605dc9301R18-R20)) wherever the convention yaml content is used, I'm forcing the reading of the yaml file during an import as simple as `import echopype` (but only on the first one). It creates a small but noticeable lag. Using module-scope variables minimizes redundancy. But we could eliminate that initial lag and make the yaml reading happen only when it's first needed by moving those variables into the classes / functions where they're used. 

Thoughts?

emiliom:
I've made changes such that it now reads the convention yaml within classes rather than as module-scope assignments.

echopype/convert/set_groups_azfp.py
```diff
@@ -8,13 +8,15 @@
...
-DEFAULT_BEAM_COORD_ATTRS = _varattrs["beam_coord_default"]
-
 
 class SetGroupsAZFP(SetGroupsBase):
     """Class for saving groups to netcdf or zarr from AZFP data files."""
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self._varattrs = sonarnetcdf_1.conv.yaml_dict["variable_and_varattributes"]
```
lsetiawan, line 18: I like how this works @emiliom. Though one question I have is how come this is not inserted within `SetGroupsBase`? It seems that `self._varattrs` is used in each SetGroups.

lsetiawan:
I think the question above is my last comment on this PR. Overall, I think that you did an awesome job @emiliom consolidating the convention specs with the minimal amount of changes and impact. 😄

echopype/convert/set_groups_azfp.py
emiliom: I didn't think of that! It sounds like a great suggestion. I'll look into it soon.

emiliom:
@lsetiawan I'm working on your suggestion to move `self._varattrs = sonarnetcdf_1.conv.yaml_dict["variable_and_varattributes"]` to `SetGroupsBase`. If all goes well, I'll merge the PR. 

But, in the meantime, do you have a suggestion for how to shorten the `sonarnetcdf_1.conv.yaml_dict` to `sonarnetcdf_1.yaml_dict`? As I mentioned above, I tried to figure out a solution using the `__init__.py` but was very unsuccessful. If you can figure it out in time for this PR, great. Otherwise, it could be in a follow up PR. Thanks!

emiliom:
The move to `SetGroupsBase` worked w/o any issues (all tests ran). That's definitely a better approach! I've pushed the changes to the PR.

lsetiawan:
> But, in the meantime, do you have a suggestion for how to shorten the sonarnetcdf_1.conv.yaml_dict to sonarnetcdf_1.yaml_dict? As I mentioned above, I tried to figure out a solution using the __init__.py but was very unsuccessful

I just pushed a solution. Let me know if that works. Also, thanks for moving the `self._varattrs` to the `SetGroupsBase` 😄

lsetiawan (approved):
LGTM when everything passes! Nice work 🎉

emiliom:
Very nice, @lsetiawan. Thanks!

leewujung:
Awesome work!! 🎊

PR #575, opened 2022-03-09
Add file checking before deleting

lsetiawan:
## Overview

As I was working in trying to parallelize hake data conversion I run into file not found errors. During file permission check echopype attempts to create a permission test file. In the case of parallel run, this file might not always be there since it might have gotten deleted by another process. So this PR fixes that by checking whether the file is still there, if it is, then delete.


lsetiawan:
This PR is not quite ready... not sure why CI is failing.

lsetiawan:
Now with the new parallel way to convert, this doesn't seem necessary atm.

Issue #582, opened 2022-03-13
Add variable Sv attributes to output dataset from compute_Sv


emiliom:
The Sv DataArray generated by `calibrate.compute_Sv` has no variable attributes; see an example [here](https://osoceanacoustics.github.io/echopype-examples/ms_OOI_EK60_mooringtimeseries.html#basic-echo-processing). Add the usual set:
- `long_name`: "Volume backscattering strength (Sv)"
- `units`: "dB"
- `actual_range`: Generated from the data min & max values (`valid_min` and `valid_max` are not applicable)

There doesn't seem to be a CF `standard_name` for Sv.

#576 and #481 are somewhat related, though focused on MVBS (mean Sv)

emiliom:
@leewujung could you clarify what the units of `Sv` are? I've seen different versions. In this issue I wrote `dB`, from something I saw I forget where; but in other recent echopype inline comments I've seen "dB re 1 m<sup>-1</sup>". BTW, what's "re" in those units?

I assume the units for MVBS identical, but please confirm.

leewujung:
@emiliom : yes, the units for Sv is "dB re 1 m<sup>-1</sup>"
The "re" here is "in reference to," since dB is a relative scale, like this:
![dB_ref](https://user-images.githubusercontent.com/15334215/159042982-b2468908-b0c0-423b-8a06-fabaa9580675.png)
That's why people always write "re" when using dB as units, like sound pressure level (SPL) is in units "dB re 1 uPa."

And yes the units for MVBS is identical to Sv since it is mean Sv.

emiliom:
Thanks! I'm adding this reference to the "re" notation: https://en.wikipedia.org/wiki/Decibel#Suffixes_and_reference_values. I wasn't familiar with that notation. FYI, it doesn't look like there's support for it in the [udunits](http://www.unidata.ucar.edu/software/udunits/) package that CF refers to for units strings, though I didn't look very deeply. [The CF FAQ has this entry about decibel and reference "levels"](http://cfconventions.org/faq.html#udunits_missing): "decibel: a logarithmic measure of relative acoustic or energy intensity; symbol dB, db, or dbel (the reference level, needed for logarithmic units, is specified in the standard names that use this canonical unit)"

emiliom:
[...]

Issue #576, opened 2022-03-10
MVBS variable name is Sv


lgarzio:
I think the MVBS variable name is incorrectly labeled Sv after doing the following (using echopype version 0.5.6):

ed = ep.open_raw(raw_file, sonar_model='EK60')
ds_Sv = ep.calibrate.compute_Sv(ed)
ds_MVBS = ep.preprocess.compute_MVBS(ds_Sv, range_meter_bin=5, ping_time_bin='20S')

When I open ds_MVBS in xarray I get the following. I think the data variable is probably just mis-labeled as Sv instead of MVBS.

<xarray.Dataset>
Dimensions:    (frequency: 3, ping_time: 113, range: 120)
Coordinates:
  * ping_time  (ping_time) datetime64[ns] 2018-01-16T03:08:20 ... 2018-01-16T...
  * frequency  (frequency) float64 3.8e+04 1.2e+05 2e+05
  * range      (range) float64 0.0 20.0 40.0 60.0 ... 2.34e+03 2.36e+03 2.38e+03
Data variables:
    Sv         (frequency, ping_time, range) float64 ...
Attributes:
    binning_mode:          physical units
    range_meter_interval:  20m
    ping_time_interval:    20S

leewujung:
Thanks @lgarzio -- MVBS is _mean_ volume backscattering strength, so it is indeed still Sv. 
Perhaps there should be some explanations somewhere to elaborate this? Do you have suggestions for this??
The pre-processing functions are not really in the documentation at the moment (limited bandwidth...) but we do want to improve it.

I can see that this is also a problem with not having provenance in the output data after processing steps, which we should also improve.

lgarzio:
@leewujung I was very confused with my processed files because I did all of the processing a while back, then when I went to plot today I couldn't figure out why Sv was in my MVBS files, hahaha. But yes this makes sense, and I think just adding a comment to the attributes that clarify this is mean volume backscattering strength would definitely suffice. Thanks!

emiliom:
[...]

Issue #481, opened 2021-11-06
Add missing coordinate attributes to MVBS dataset from compute_MVBS


emiliom:
In the MVBS xarray dataset produced by `preprocess.compute_MVBS`, `ping_time` and `range` are missing user-facing attributes (I think `ping_time` must have proper time encoding attributes). For example, see https://osoceanacoustics.github.io/echopype-examples/echopype_tour.html. I don't know if they're missing in the output of `compute_Sv` as well.

emiliom:
This issue has been addressed. Closing

Issue #221, opened 2020-12-22
Add provenance to MVBS and noise removed Sv outputs


leewujung:
get_MVBS and remove_noise output should also include provenance on which Sv file was used for calculation and the bin size along range and ping sequence.

leewujung:
This is related to #210.

PR #594, opened 2022-03-24
Add missing Sv and coordinate attributes for computed Sv and MVBS

emiliom:
Addresses #481, #582, and parts of #576


echopype/preprocess/api.py
```diff
@@ -64,18 +64,32 @@ def _freq_MVBS(ds, rint, pbin):
...
+    ds_MVBS = (
+        ds_Sv.groupby("frequency")
+        .apply(_freq_MVBS, args=(range_interval, ping_time_bin))
+        .to_dataset()
     )
-
     # Attach attributes
-    MVBS.attrs = {
+    ds_MVBS["ping_time"].attrs = {
+        "long_name": "Timestamp of each ping",
```
emiliom, line 74: @leewujung this MVBS `ping_time` long_name doesn't seem quite right, given the MVBS gridding. But I couldn't come up with something better. Plus, is keeping `ping_time` as the time dimension name even appropriate given the gridding?

leewujung: Hmm... I think it should still be `ping_time` since conceptually is still the same thing as before the averaging operation. But you are right, the long_name should be change to something else that reflects what it really is.

"Leading timestamp of the averaging time window"  ??

emiliom: > Hmm... I think it should still be ping_time since conceptually is still the same thing as before the averaging operation.

Yeah. I guess I was getting hung up on the fact that with MVBS we're not dealing with individual pings any more.

> "Leading timestamp of the averaging time window" ??

It's getting there, and it looks factually correct. But, also, yuck :laughing:. Maybe we should take a look at the [CF concept of "cells"](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#_data_representative_of_cells) and [cell "methods"](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#cell-methods).

Come to think of it, the same logic would apply to `range` (stating that it's based on an averaging window).

leewujung: Agree with corresponding changes to `range`.  In a way this has the "provenance" in the dataset also. But of course we will need something actually added to the `Provenance` group (#221 ).

echopype/calibrate/api.py
```diff
@@ -53,18 +53,34 @@ def _compute_cal(
...
             waveform_mode=waveform_mode, encode_mode=encode_mode
         )
+        # Add attributes
+        sv_dataset["Sv"].attrs = {
+            "long_name": "Volume backscattering strength (Sv)",
+            "units": "dB re 1 m-1",
+            "actual_range": [
+                round(float(sv_dataset["Sv"].min().values), 2),
+                round(float(sv_dataset["Sv"].max().values), 2),
+            ],
```
leewujung, line 66: I didn't know about this specification. Cool!

leewujung (commented):
Thanks @emiliom : This is ready to go once we resolve the long_names for `ping_time` and `echo_range`.

But inevitably, changes of `range_bin` to `range_sample` would have to be merged here after #595 is merged. 😅

echopype/preprocess/api.py
emiliom: I've reviewed the CF [cell methods](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#cell-methods) section, including the sub-sections [Statistics for more than one axis](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#statistics-more-than-one-axis) and [Recording the spacing of the original data and other information](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#recording-spacing-original-data). I'll start with a question / suggestion: `ping_time` is really just a timestamp, isn't it? The original clock (the source of the timestamps) is of course the transducer's clock, but for MVBS we're dealing with gridded/binned Sv and the time dimension can be generalized to simply `time`, with a generic `long_name` ("time"?). Do you see a problem with doing that?

Then, for the MVBS variable (`Sv`), I recommend adding some iteration of the following attribute, where in this example time binning is 20 sec and range binning is 10 m:
```
cell_methods = "time: mean (interval: 20 second comment: time is the interval start) echo_range: mean (interval: 20 meter comment: echo_range is the interval start)"
```

CF specifies that when using `cell_methods`, cell bound on the corresponding dimensions should also be included as ancillary variables. For expedience (we have a lot of other work to do for 0.6.0!) I suggest we skip that effort right now, and work on adding those variables in the future.

emiliom:
> But inevitably, changes of range_bin to range_sample would have to be merged here after #595 is merged.

Yup

echopype/preprocess/api.py
emiliom: I've implemented the recommendations I described above, so you can assess them more easily. Here's a screenshot of a resulting MVBS dataset:

image

leewujung: Thanks @emiliom! I agree with the additions, and that it is nice that the `cell_methods` give very concise and precise description. Could you create an issue to make sure that we add in the cell bounds?

I think though we should keep `ping_time` instead of changing that to `time`. It is more intuitive for acoustics people, explicit that this MVBS is just a mean of Sv so has the same coordinate, and would avoid confusion with time1, time2, etc in the other groups.

emiliom: > I think though we should keep ping_time instead of changing that to time

Will do

emiliom:
I'll investigate the test failures.

echopype/preprocess/api.py
emiliom: @leewujung  I've reverted the time coord name to `ping_time`. All tests ran successfully, locally.

Note: once we're good with this, it looks like we should also add attributes to `compute_MVBS_index_binning`. I've never used this function before, so I wanted to hold off until we're in full agreement.

emiliom: Sigh, the tests are failing again. I think I finally see what's going on. I'll get to it

emiliom:
Woo-hoo, NOW the CI tests passed!

echopype/preprocess/api.py
leewujung: Sounds good. Do you want to handle adding attribute to `compute_MVBS_index_binning ` in this PR? It seems better to have that in a separate PR so we can get this fully tested first.

emiliom: Tackling compute_MVBS_index_binning attributes in a separate PR sounds good. I'm all for completing this PR soon!

leewujung:
I merged `dev` into this locally and running the fully suite of tests locally now. Will push the merge commit if everything passes or report failure on what I found.

emiliom:
Thanks for merging dev!!! All tests passed, including pre-commit. We're ready to merge, right?

leewujung:
Yes, please go ahead and merge! Thanks for all the work @emiliom !

PR #599, opened 2022-03-28
Upgrade typing-extensions to version 4 and use conda-compatible requirements syntax

imranmaj:
`typing-extensions~=3.10` is valid syntax with pip, but conda does not support it yet, so we can use the equivalent `>=3.10,==3.*` syntax instead

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ fsspec
 s3fs==2021.11.0
 requests
 aiohttp
-typing-extensions~=3.10
+typing-extensions>=3.10,==3.*
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ fsspec
 s3fs==2021.11.0
 requests
 aiohttp
-typing-extensions>=3.10,==3.*
+typing-extensions>=4.1,==4.*


emiliom:
Given that we're trying to see if echopype will work with current `typing-extensions` versions, I suggest removing all version pinning altogether. If problems arise, we can add version pinning.

BTW, I see that the CI tests have already passed for Python 3.7 - 3.9, and they ran quickly (in 4 min). I'm guessing that only a subset of the tests were run. But I also see that you did add the "Needs Complete Listing". But given that it was added ~ 6 minutes after the PR was created, it's likely that the CI tests had already been spawned with a limited set of tests.

leewujung:
Good point @emiliom. I push the re-run all jobs button, let's see what happens!

leewujung:
@imranmaj : it appeared that the "Need complete testing" label needs to be applied when the PR is opened for it to take effect. So, I closed and re-opened this PR for all tests to run through the full set of tests. 

I agree with @emiliom's suggestion to remove all version pinning.
> Given that we're trying to see if echopype will work with current typing-extensions versions, I suggest removing all version pinning altogether. If problems arise, we can add version pinning.

The CI pulled typing_extensions=4.1.1 as in my local environment and all test passed.

Could you remove the version pinning? Thanks!

emiliom:
> it appeared that the "Need complete testing" label needs to be applied when the PR is opened for it to take effect. 

I wasn't sure if that subtlety was already captured in the docs, but I'm glad to confirm it is: ["To have the CI execute the entire test suite, add the GitHub label `Needs Complete Testing` to the PR before submitting it."](https://echopype.readthedocs.io/en/latest/contributing.html#continuous-integration-github-actions)

leewujung:
Alright all tests passed. I'll merge this now.

Issue #506, opened 2021-12-10
Remove old API


leewujung:
Currently we issue deprecation warnings for the old API. Let's remove it completely in the next major release.

leewujung:
(Note for myself)
Also remove obvious unused attributes:
https://github.com/OSOceanAcoustics/echopype/blob/a18ec08c543292693e4bbff1ae02cce9e3b3a5d5/echopype/convert/parse_base.py#L43

leewujung:
Remove unused comments:
https://github.com/OSOceanAcoustics/echopype/blob/a18ec08c543292693e4bbff1ae02cce9e3b3a5d5/echopype/convert/set_groups_ek80.py#L601-L605

emiliom:
I suggest we prioritize this issue before convention-related changes in 0.6.0. Removing the old API's will make the convention-related changes a bit less burdensome, because there will be less code to go through. We've already run into the drag of having to modify the code involving the deprecated API's.

leewujung:
Sounds good. Let me get on it as soon as my environment is back... 😅

Issue #552, opened 2022-02-04
Rethink and upgrade testing scheme


leewujung:
In our current testing setup, when PR to `dev` the test is only run on the modified module and not on the downstream tests. The entire set of tests are run when the PR is merged to `dev` (or `main`). So what we have right now is a little brittle even though bugs would be caught and fixed eventually. It'll be good for us to rethink how we can upgrade this and still with reasonable the CI running time.

Below are conversions from discussions re testing in #547:

@lsetiawan : 
> In practice, most software CI actually runs everything anyways to make sure your changes doesn't break anything, we just decided to make this custom, brittle, intelligence 😜

@leewujung :
> In theory we can get out of the current brittle setup once most big test data files are updated to a smaller size. I updated 2 major ones, but ran out of time to do the last couple files before I had to return a hardware dongle needed to generate data we test against -- the next time I can do this is likely late March - April.
> 
> BUT regardless, it is unlikely we'll be able to reduce all files to be KB size like the new EK80 test files I collected, since we don't have access to many instrument variations that were used to collect the diverse test files.

Also related on the current status:
@lsetiawan :
> as long as the isolated module test works you can merge to dev for a full test. Or you can always just use the Needs complete testing label with the PR if you know that particular function touches many different things

@emiliom :
> And whenever you have a need for that mechanism but don't remember how to use it, go to the docs: https://echopype.readthedocs.io/en/stable/contributing.html#continuous-integration-github-actions

lsetiawan:
I think the solution may be speeding up the test and like you said in #553, run all tests every time. We have not utilized the parallel testing... I know there were some issues there, but I'll get back to it.

leewujung:
From @lsetiawan in #601:
> Thinking about it, I wonder if we can add like specific connected modules that should also be run for that module if it uses functions from those connected modules. Does that make sense? 😆 This way, we can have an "extensive" test without having to run the whole test!

leewujung:
Noted down some 3.9 and 3.10 test failures for investigation:
https://github.com/OSOceanAcoustics/echopype/actions/runs/2153214186/attempts/1
Similar test failures have occurred a few times in the v0.5.6-->v0.6.0 dev period.

leewujung:
Let's close this one and open a new issue on whether we want to switch to always running all tests regardless of which files are changed in a PR.

leewujung:
Seems like we can close this now that @lsetiawan has put in quite a bit of work to make the tests more robust?

[...]

PR #601, opened 2022-03-28
Remove old API (those prior to v0.5.0)

leewujung:
This PR removes the old API `Convert` and `Process` prior to v0.5.0. 

The changes include:
- deletion of the subpackage `process`, `convert/convert.py`, `convert/convert_combine.py`
- deletion of related test
- revision of documentation

This addresses #506.


.ci_helpers/run-test.py
```diff
@@ -101,8 +100,6 @@
                 f"echopype/{module}/*",
                 f"echopype/tests/{module}/*",
             ]
-        if "extra_globs" in mod_extras:
-            file_globs = file_globs + mod_extras["extra_globs"]
```
lsetiawan, line 105: This if section should be okay. This is in case there are exceptions like these in the future, if not then the dict above can be changed to just a list, really.

lsetiawan (approved):
LGTM! Feel free to either bring back the if section I mentioned above or leave your changes alone 😄

leewujung:
Thanks @lsetiawan ! I will leave that section in for future use. :)

.ci_helpers/run-test.py
leewujung: awesome, this is exactly the place i wasn't sure if I should delete!

lsetiawan:
> Thanks @lsetiawan ! I will leave that section in for future use. :)

Thinking about it, I wonder if we can add like specific connected modules that should also be run for that module if it uses functions from those connected modules. Does that make sense? 😆

leewujung:
> Thinking about it, I wonder if we can add like specific connected modules that should also be run for that module if it uses functions from those connected modules. Does that make sense? 😆 This way, we can have an "extensive" test without having to run the whole test!

Definitely worth discussion as part of #552! This seems would fall in line with the integration test.

docs/source/convert.rst
```diff
@@ -63,8 +63,7 @@ netCDF file named ``FILENAME.nc`` saved to the directory path
 
    - Prior to version 0.5.0, conversion was carried out through the
      `"Convert" interface <https://echopype.readthedocs.io/en/v0.4.1/usage.html#conversion-operation>`_.
-     This interface is still available but will be deprecated in
-     a future version.
+     This interface is deprecated in version 0.6.0.
```
emiliom, line 66: Come to think of it, we previously should've said that the interface was *deprecated* [(ie, discouraged, no longer recommended or actively developed, but still available)](https://en.wikipedia.org/wiki/Deprecation) and would be *removed* in a future version. What we're doing now is removing it, the next step after deprecation. So, replace with: "This deprecated interface was removed in version 0.6.0."

Alternatively, at this stage how about just removing all references to it from the docs? That would be true also for the references to raw2nc and raw2zarr, below, though I see that those methods are not being removed in this PR (I'll comment about that elsewhere).

docs/source/process.rst
```diff
@@ -2,9 +2,8 @@ Data processing
 ===============
 
 .. attention::
-   Starting with version 0.5.0, the data processing interface ``Process``
-   is deprecated. Attempts to use ``Process`` will still
-   work at the moment but will no longer be available in the future.
+   The previous data processing interface ``Process`` was deprecated
+   starting from version 0.5.0.
```
emiliom, line 6: See my comment above in `convert.rst` about deprecation vs removal.

emiliom:
How about also removing [raw2nc](https://github.com/OSOceanAcoustics/echopype/blob/main/echopype/echodata/echodata.py#L718) and [raw2zarr](https://github.com/OSOceanAcoustics/echopype/blob/main/echopype/echodata/echodata.py#L737) in this PR, too? That would take care of all the old API's, I think.

leewujung:
@emiliom : thanks for finding the `raw2nc` and `raw2zarr`, somehow I saw them in the docs but it did not register for me to remove them. I'll push up new commits to remove them.

Also, I think it's fine to remove the reference to old API altogether. Our docs are versioned so folks can hopefully find what they need if they really wanted to use the old versions. I'll make these changes in the docs also.

docs/source/process.rst
leewujung: removed!

docs/source/convert.rst
leewujung: I removed references to the old API completely from the docs. I agree that is a better approach at this stage.

emiliom (approved):

emiliom:
I think this is ready to merge. Thanks for pursuing this! Good riddance :laughing:

PR #604, opened 2022-03-28
Upgrade python minimum to 3.8 and tidy up packaging

### LINKED ISSUE #603: Sync up with xarray python support to only >=3.8?
Right now xarray requirement for it's latest versions for python is `>=3.8`. We are right now at `>=3.6`. I think we should sync up with xarray's requirements to keep ourselves up to date and in case we do use future features and improvements of xarray. Otherwise, we might not be able to utilize the latest versions. This would also get rid of the need for `typing-extensions`.

lsetiawan:
## Overview

This PR updates the python minimum requirement to `3.8` syncing up with xarrays. This resolves #603. Also, some unneeded packages has been deleted from requirements. The `setup.cfg` has now been filled with all the package requirements even the optional ones. Now any `requirements-x.txt` files are pretty much redundant.


emiliom:
Thanks for doing this!

> Now any `x-requirements.txt` files are pretty much redundant.

Did you mean `requirements-x.txt`? Currently we only have `requirements.txt` and `requirements-dev.txt`. While `requirements-dev.txt` may now be redundant for the CI, it's still required for [building local dev conda environments](https://echopype.readthedocs.io/en/stable/contributing.html#installation-for-echopype-development). Given that, having dev requirements listings duplicated in `setup.cfg` and `requirements-dev.txt` sounds like a recipe for divergence in the future. Do we lose anything significant by having dev requirements only in `requirements-dev.txt`?

lsetiawan:
> Do we lose anything significant by having dev requirements only in requirements-dev.txt?

For dev development, one can do `pip install -e .[dev]` to get all the required packages as well as the dev dependency. I mean, we can just remove this optional extra packages configuration from setup.cfg and just allow make users do an install from requirements-dev.txt as shown in the docs or update the docs. I was thinking of cleaning up the `requirements-x.txt` files if we have all the configuration in setup.cfg. One of the "proper" use of the requirements.txt would be for a freezed environment for testing purposes, so maybe we can go that route for development?

lsetiawan:
Okay. I've updated this to be that dev related requirements are not in `setup.cfg` to make it cleaner as in what is the barebone package dependencies, so one has to install from `requirements-dev.txt` for anything that is related to dev and the `docs/requirements.txt` for docs stuff. Now everything is separated. Also I've moved some settings for black and flake8 to their proper places, now it seems like it's actually expanding the line length! We set this at `100`, is this still okay? PEP8 standard is actually 79, black style standard puts it at 88 by default, which I think what was happening.

emiliom:
Thanks!! We can discuss these tradeoffs (requirements-dev.txt vs setup.cfg vs ...) some other time. I like this being a narrower PR at this time, focused on the >=3.8 change.

On line length, I personally like a length of 100

emiliom:
Oh, and I can't fully review this PR yet (**42** files changed?! Though it looks like many of them are just line lengths). I can get to it maybe by tonight, otherwise tomorrow.

lsetiawan:
Yea sorry for the blow up in files... that was because of line length fix by pre-commit! Other than that, looks like all tests are passing 😄

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 black
 check-manifest
 codespell
-docker-compose
+docker-compose # Not available in py3.10
```
emiliom, line 4: So, to do local testing with py3.10, we'd have to install docker-compose separately (in the system, rather than via conda)?

requirements.txt
```diff
@@ -1,3 +1,7 @@
+# This file is redundant with setup.cfg;
```
emiliom, line 1: If it's redundant but it must exist, what's the advantage of duplicating the dependencies in setup.cfg? This duplication seems brittle, since we have to manually ensure the files are never out of sync.

setup.cfg
```diff
@@ -13,25 +14,36 @@ classifiers =
...
-known_first_party = echopype
-known_third_party = _echopype_version,setuptools
-multi_line_output = 3
-include_trailing_comma = True
-force_grid_wrap = 0
-combine_as_imports = True
-line_length = 100
+[options.extras_require]
+plot =
+    matplotlib
```
emiliom, line 46: In the conda  yml files for the CI, you're now using matplotlib-base. Is there a good reason for this one to be different?

setup.py
```diff
@@ -1,42 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-from setuptools import find_packages, setup
-
-# Long description will go up on the pypi page
-with open("README.md") as file:
-    LONG_DESCRIPTION = file.read()
-
-# Dependencies.
-with open("requirements.txt") as f:
```
emiliom, line 10: So, using setup.py allowed the reading of requirements from the requirements files, which minimized redundant dependency listings. But in the switch to setup.cfg, that capability is not longer there?

emiliom:
I ended up being able to review this now :smile: . I went through the line-length changes quickly (there are too many!). See my inline comments elsewhere.

leewujung:
I have a not substantial comment to add, that the PR title should change 😁  Now it is a wayyy bigger PR than what was originally intended.

emiliom:
I think this is ready to merge, right @leewujung ?

leewujung:
Yeah I think this is ready to merge. It would be nice to transfer some of the detailed descriptions from here to the Contributing to echopype page, or we can have an infrastructure page.

@emiliom there are some open questions you had above. Do you want those answered before merging? Or take them to another issue?

emiliom:
> @emiliom there are some open questions you had above. Do you want those answered before merging? Or take them to another issue?

Right, thanks! I'm not sure if they belong in a single issue. @lsetiawan could you reply to my unanswered inline comments when you have a chance?

requirements-dev.txt
lsetiawan: Yes sir. 😬

requirements.txt
lsetiawan: In future GitHub features hopefully this file won't be necessary

setup.cfg
lsetiawan: matplotlib-base is a conda package. It's only matplotlib in pypi

setup.py
lsetiawan: No, setup.cfg is static. Though we can still keep it dynamic with setup.py.

lsetiawan:
Sorry for the late replies! Just saw those inline comments. This setup is pretty flexible at this moment, I know there are some redundancy mainly with the requirements.txt. I can easily revert back to dynamic reading of that file. I'm simply trying to begin future proof the packaging, and I have no hard fast rule on this at the moment haha 🙈

setup.cfg
emiliom: Right, I keep forgetting that!

requirements-dev.txt
emiliom: I'll stick to doing local testing with Py 3.9

emiliom:
> This setup is pretty flexible at this moment, I know there are some redundancy mainly with the requirements.txt. I can easily revert back to dynamic reading of that file. I'm simply trying to begin future proof the packaging, and I have no hard fast rule on this at the moment haha :see_no_evil:

Thanks. It seems tough trying to balance multiple drivers, such as future proofing vs minimizing duplicated specifications of the same information. I'm not sure what to recommend, though I personally tend to worry more about duplicated information going out of sync and causing confusion.

docs/source/contributing.rst
```diff
@@ -76,12 +76,12 @@ Create a `conda <https://docs.conda.io>`_ environment for echopype development
...
 .. code-block:: bash
 
-    conda create -c conda-forge -n echopype --yes python=3.9 --file requirements.txt --file requirements-dev.txt
+    conda create -c conda-forge -n echopype --yes python=3.9 --file requirements-dev.txt
     conda activate echopype
     # ipykernel is recommended, in order to use with JupyterLab and IPython
     # to aid with development. We recommend you install JupyterLab separately
     conda install -c conda-forge ipykernel
-    pip install -e .
+    pip install -e .[all]
```
leewujung, line 84: I think we should have some comments here to let users know (and as a reminder for ourselves) that the core requirements will be installed at this time, otherwise this can get confusing for folks seeing it for the first time.

Also some pointers for contributors know what `[all]` does.

leewujung:
I added one comment above.

> I personally tend to worry more about duplicated information going out of sync and causing confusion.

I feel the same way. I guess I don't see clear advantage of having everything static in `setup.cfg` since right now for GH to build the dependency graph `requirements.txt` still has to exist. The potential out of sync problem seems to make this approach less robust.

Where does the future proof come in here?

docs/source/contributing.rst
emiliom: I agree. But, based on our comments below, it looks like we're converging on retaining `requirements.txt` w/o duplicating that dependency listing in `setup.cfg`. In that case, this additional explanation/comments won't be needed.

docs/source/contributing.rst
```diff
@@ -76,12 +76,14 @@ Create a `conda <https://docs.conda.io>`_ environment for echopype development
...
-    conda create -c conda-forge -n echopype --yes python=3.9 --file requirements.txt --file requirements-dev.txt
+    conda create -c conda-forge -n echopype --yes python=3.9 --file requirements-dev.txt
     conda activate echopype
     # ipykernel is recommended, in order to use with JupyterLab and IPython
     # to aid with development. We recommend you install JupyterLab separately
     conda install -c conda-forge ipykernel
-    pip install -e .
+    # plot is an extra set of requirements that can be used for plotting.
+    # the command will install all the dependencies along with plotting dependencies.
+    pip install -e .[plot]
```
lsetiawan, line 86: @leewujung and @emiliom I added comment here about `[plot]` let me know if it makes sense.

setup.py
lsetiawan: Okay. I've updated this now so that we have the dynamic generation. Let me know what you think.

lsetiawan:
I've put back the dynamic generation of requirements. I still kept everything about the extra plotting requirement within the `setup.cfg`. Now I think it makes a lot of sense to make this optional, and remove the `all` altogether since there's only one optional dependencies at the moment. Let me know if there's anything else. Otherwise, feel free to merge if all good. Thanks again for both of your inputs 😄

docs/source/contributing.rst
emiliom: Thanks. Yeah, the development environment should include the plot dependencies. Your comment looks good

emiliom:
@lsetiawan thank you for your flexibility as well as for helping us stay on top of current trends. Let's revisit the `setup.cfg` vs `requirements.txt` periodically, to assess when the shift to having the dependencies in `setup.cfg` alone starts to make sense.

emiliom (approved):

docs/source/contributing.rst
leewujung: Thanks @lsetiawan ! this looks good.

PR #605, opened 2022-03-29
Rename Sonar beam dimension to beam_group

emiliom:
Using `beam` as the dimension name for the new variables in the `Sonar` group was a bad decision, since it'll cause confusion with the `Beam_groupX` `beam` dimension defined in the SONAR-netCDF4. They are different concepts.

This conflict was pointed out in https://github.com/OSOceanAcoustics/echopype/pull/598#discussion_r836960802


leewujung:
This looks good to be merged now.

Issue #606, opened 2022-03-30
Backward/forward compatibility wrt v0.6.0 group structure/variable name change


leewujung:
## Context 
We are coming up against multiple planned breaking changes in v0.6.0. The changes include netCDF group structures as well as coordinates and variable names and attributes.

We know that many users, including both OOI and NCEI have converted some non-trivial volume of data into zarr or nc using echopype v0.5.x, so it is imperative that all processing can accommodate these already converted data.

In light of the group structure changes, we also want to ease the transition for ourselves such that we do not disturb the entire codebase significantly all at once (so that we can focus on convention compliance for v0.6.0).

## Problems and Approaches
There are two types of changes we need to deal with:
1. group structure changes: more involved
2. variable name changes: more straightforward

### Group structure changes
After discussions (#567) we concluded that a cleaner approach is for us to use DataTree functionality to allow accessing the data like a dictionary or like in the netcdf4 library, such as `EchoData["Sonar/Beam_groupX"]`, because our current attribute setup would require `EchoData.sonar.beam_group1` which can be quite involved to implement. The downside to this are:
- DataTree could be not yet stable, though it seems pretty much on track now
- we would need to change how data in different groups are accessed for all the downstream functions
- if we go ahead and directly use `open_tree` on v0.5.x data, the groups will _not_ be in the right place

The last one requires some more discussion, so I wonder if an intermediate step we can take in v0.6.0 is to avoid this (for now) by making our code _forward_ compatible to the v0.6.0 data:
Basically, we would hook up:
- `EchoData.beam` to `root/Sonar/Beam_group1`
- `EchoData.beam_power` (if exists) to `root/Sonar/Beam_group2`
- `EchoData.sonar` still to `root/Sonar`  (no action)
- `EchoData.nmea` still to `root/Platform/NMEA`  (no action)
This way all the downstream functions do not have to change in v0.6.0, and can be dealt with later once we figure out how to use DataTree to handle v0.5.x data.

### Variable name changes
For these changes, I propose that instead of taking the forward compatibility route as above, here we actually change the variable names in the code once the correct groups are read into xarray datasets. 

For example, we would:
- read in `range_bin` --> immediately change that to `range_sample` --> change all the downstream processing that uses `range_bin` also to `range_sample`
- read in `frequency` --> immediately change that to `channel` --> change all the downstream processing that uses `range_bin` also to `range_sample`

## Decision?
[...]

Issue #567, opened 2022-02-24
Access `Beam_groupX` as subgroups of the `Sonar` group


b-reyes:
In issue #519 we are proposing to move the beam group to a sonar subgroup. If we make this change, we need to consider what is the best way for a user to easily understand this change and more importantly how we should code this. This spawns several questions:

1. Say we have two beam groups: narrowband and broadband, then how should the user expect to access these groups? If given the EchoData object "ed", the intuitive way to access these groups is to do: "ed.sonar.narrowband" or "ed.sonar.broadband". The output here would then be an Xarray Dataset that is similar to the current beam group.

   Now, say the user wants to access the metadata of the sonar system, then they would do "ed.sonar". From a coding perspective, how could we accomplish this? It seems like "sonar" would have to be both a variable and a class at the same time. 

2. If we resolve how to code item 1., how can we make it such that a user knows what sonar subgroups are available within "ed.sonar"?

3. Once this change is made, how do we reflect this change within the documentation? Currently, we have a very nice display of the EchoData object at the bottom of our [documentation](https://echopype.readthedocs.io/en/stable/open-converted.html) that is useful for new users. However, once we make this move, this depiction will no longer be accurate. Is it possible to maintain this portion of the documentation, and if so, how?

leewujung:
In general I think we should do it in a way that is consistent with the convention ver 1.

> 1. Say we have two beam groups: narrowband and broadband, then how should the user expect to access these groups? If given the EchoData object "ed", the intuitive way to access these groups is to do: "ed.sonar.narrowband" or "ed.sonar.broadband". The output here would then be an Xarray Dataset that is similar to the current beam group.

For this I think we should have the access pattern `ed.sonar.beam_group1` and `ed.sonar.beam_group2` that is consistent with the convention, so new users understanding the convention will be able to use this easily. This will also allow us to open and process files that in theory can be written to netcdf directly from the instruments once manufacturer made that change.

Note that we do expect power users who will want to access the details of the data, so our code should support that.

> 2. If we resolve how to code item 1., how can we make it such that a user knows what sonar subgroups are available within "ed.sonar"?

As we discussed today (and what @emiliom @gavinmacaulay and I discussed on Monday), in the `Sonar` group (`ed.sonar`) we can have a data variable recording which beam_group contain what type of data. In the form similar to the below:
```python
# say we 2 beam groups as in your example

# coordinate of length 2
beam_group_sequence = ("beam_group1", "beam_group2")

[...]

PR #611, opened 2022-03-31
Implement DataTree for EchoData

lsetiawan:
This PR implements [`xarray-datatree`](https://github.com/xarray-contrib/datatree) for the underlying EchoData structure to ease access. This work is part of issue #567 and #606

### TODO
- [x] Make root accessible via `echodata['Top-level']` or `echodata['/']`.
- [x] Make repr to show the tree relatioships
- [x] Make repr_html to have dropdowns for each tree
- [X] Add a way to access echopype software version
- [X] ~Add group paths listing with `/`~
- [x] Tweak group texts (Don't include paths!)


echopype/echodata/echodata.py
```diff
@@ -59,12 +61,19 @@ def __init__(
         self.xml_path: Optional["PathHint"] = xml_path
         self.sonar_model: Optional["SonarModelsHint"] = sonar_model
         self.converted_raw_path: Optional["PathHint"] = None
+        self._tree: DataTree = None
 
         self.__setup_groups()
-        self.__read_converted(converted_raw_path)
+        # self.__read_converted(converted_raw_path)
```
lsetiawan, line 67: @b-reyes Note that I commented this out. This uses the old `_load_file` functionality.

echopype/echodata/echodata.py
```diff
@@ -133,6 +142,82 @@ def __read_converted(self, converted_raw_path: Optional["PathHint"]):
...
+    def tree(self):
+        return self._tree
+
+    def __getitem__(self, __key: str) -> Any:
+        if self._tree:
+            if __key == "Top-level":
+                return self._tree.ds
+            return self._tree[__key].ds
+        else:
+            raise ValueError("Datatree not found!")
```
lsetiawan, line 207: This is used for the new access patter of `echodata['/path/to/group']` I add a special key of `Top-level` to access the root of the tree to make it still very explicit in terms of accessing as the convention states.

echopype/echodata/echodata.py
```diff
@@ -133,6 +142,84 @@ def __read_converted(self, converted_raw_path: Optional["PathHint"]):
...
+                else:
+                    ds = self._tree[value["name"]].ds
+            except ChildResolverError:
+                # Skips group not found errors for EK80 and ADCP
+                ...
+            if group == "top" and hasattr(ds, "keywords"):
+                self.sonar_model = ds.keywords.upper()  # type: ignore
+
+            if isinstance(ds, xr.Dataset):
+                setattr(self, group, ds)
```
lsetiawan, line 195: This is similar as _load_file for now to not break any current Echodata functionalities, though it grabs the dataset from the tree instead of loading it one by one.

echopype/echodata/echodata.py
```diff
@@ -133,6 +142,118 @@ def __read_converted(self, converted_raw_path: Optional["PathHint"]):
...
+        tree = open_datatree(
+            converted_raw_path,
+            engine=XARRAY_ENGINE_MAP[suffix],
+            **echodata.open_kwargs,
+        )
+        # Change root name to top
+        tree.name = "Top-level"
+
+        echodata._set_tree(tree)
+
```
lsetiawan, line 167: @b-reyes I'm thinking that whatever internal restructuring method for `0.5.x` to `0.6.x` can go here. I suggest modifying the internal `self._tree` in the echodata object, since the actual `setattr` doesn't get called until later within `_load_tree`. This way you can still use the `version_info` and `group_paths`.

lsetiawan:
## Demo

<details><summary>Open Raw</summary>
<p>

```python
from echopype import open_raw
echodata = open_raw('echopype/test_data/ek60/ncei-wcsd/Summer2017-D20170615-T190214.raw', sonar_model='EK60')

In [5]: print(repr(echodata))
EchoData: standardized raw data from Internal Memory
  > top: (Top-level) contains metadata about the SONAR-netCDF4 file format.
  > environment: (Environment) contains information relevant to acoustic propagation through water.
  > platform: (Platform) contains information about the platform on which the sonar is installed.
  > nmea: (Platform/NMEA) contains information specific to the NMEA protocol.
  > provenance: (Provenance) contains metadata about how the SONAR-netCDF4 version of the data were obtained.
  > sonar: (Sonar) contains specific metadata for the sonar system.
  > beam: (Sonar/Beam_group1) contains backscatter data (either complex samples or uncalibrated power samples) and other beam or channel-specific data, including split-beam angle data when they exist.
  > vendor: (Vendor specific) contains vendor-specific information about the sonar and the data.

In [6]: print(echodata)
EchoData: standardized raw data from Internal Memory
DataTree('Top-level', parent=None)
│   Dimensions:  ()
│   Data variables:
│       *empty*
│   Attributes:
│       conventions:                 CF-1.7, SONAR-netCDF4-1.0, ACDD-1.3
│       keywords:                    EK60
│       sonar_convention_authority:  ICES
│       sonar_convention_name:       SONAR-netCDF4
│       sonar_convention_version:    1.0
│       summary:                     
│       title:                       
│       date_created:                2017-06-15T19:02:14Z
│       survey_name:                 
├── DataTree('Environment')
│   Dimensions:                 (frequency: 3, ping_time: 19)
│   Coordinates:
│     * frequency               (frequency) float64 1.8e+04 3.8e+04 1.2e+05
│     * ping_time               (ping_time) datetime64[ns] 2017-06-15T19:02:14.20...
│   Data variables:
│       absorption_indicative   (frequency, ping_time) float64 0.002226 ... 0.04069
│       sound_speed_indicative  (frequency, ping_time) float64 1.507e+03 ... 1.50...
├── DataTree('Platform')
│   │   Dimensions:        (location_time: 72, frequency: 3, ping_time: 19)
│   │   Coordinates:
│   │     * location_time  (location_time) datetime64[ns] 2017-06-15T19:02:15.4450001...
│   │     * frequency      (frequency) float64 1.8e+04 3.8e+04 1.2e+05
│   │     * ping_time      (ping_time) datetime64[ns] 2017-06-15T19:02:14.206000128 ....
│   │   Data variables:
│   │       latitude       (location_time) float64 dask.array<chunksize=(72,), meta=np.ndarray>
│   │       longitude      (location_time) float64 dask.array<chunksize=(72,), meta=np.ndarray>
│   │       sentence_type  (location_time) <U3 dask.array<chunksize=(72,), meta=np.ndarray>
│   │       pitch          (frequency, ping_time) float64 dask.array<chunksize=(3, 19), meta=np.ndarray>
│   │       roll           (frequency, ping_time) float64 dask.array<chunksize=(3, 19), meta=np.ndarray>
│   │       heave          (frequency, ping_time) float64 dask.array<chunksize=(3, 19), meta=np.ndarray>
│   │       water_level    (frequency, ping_time) float64 dask.array<chunksize=(3, 19), meta=np.ndarray>
│   └── DataTree('NMEA')
│       Dimensions:        (location_time: 688)
│       Coordinates:
│         * location_time  (location_time) datetime64[ns] 2017-06-15T19:02:14.2059996...
│       Data variables:
│           NMEA_datagram  (location_time) <U73 '$SDVLW,2.084,N,2.084,N' ... '$INHDT,...
│       Attributes:
│           description:  All NMEA sensor datagrams
├── DataTree('Provenance')
│   Dimensions:  ()
│   Data variables:
│       *empty*
│   Attributes:
│       conversion_software_name:     echopype
│       conversion_software_version:  0.5.6.dev42+gebb160b.d20220331
│       conversion_time:              2022-03-31T23:48:28Z
│       src_filenames:                echopype/test_data/ek60/ncei-wcsd/Summer201...
│       duplicate_ping_times:         0
├── DataTree('Sonar')
│   │   Dimensions:           (beam_group: 1)
│   │   Dimensions without coordinates: beam_group
│   │   Data variables:
│   │       beam_group_name   (beam_group) <U11 'Beam_group1'
│   │       beam_group_descr  (beam_group) <U131 'contains backscatter power (uncalib...
│   │   Attributes:
│   │       sonar_manufacturer:      Simrad
│   │       sonar_model:             ER60
│   │       sonar_serial_number:     
│   │       sonar_software_name:     
│   │       sonar_software_version:  2.4.3
│   │       sonar_type:              echosounder
│   └── DataTree('Beam_group1')
│       Dimensions:                         (frequency: 3, ping_time: 19,
│                                            range_sample: 3888)
│       Coordinates:
│         * frequency                       (frequency) float64 1.8e+04 3.8e+04 1.2e+05
│         * ping_time                       (ping_time) datetime64[ns] 2017-06-15T19:...
│         * range_sample                    (range_sample) int64 0 1 2 ... 3886 3887
│       Data variables: (12/30)
│           channel_id                      (frequency) <U37 'GPT  18 kHz 009072058c8...
│           beam_type                       (frequency) int64 1 1 1
│           beamwidth_receive_alongship     (frequency) float64 10.3 6.8 7.3
│           beamwidth_receive_athwartship   (frequency) float64 10.3 6.8 7.2
│           beamwidth_transmit_alongship    (frequency) float64 10.3 6.8 7.3
│           beamwidth_transmit_athwartship  (frequency) float64 10.3 6.8 7.2
│           ...                              ...
│           data_type                       (frequency, ping_time) float64 3.0 ... 3.0
│           count                           (frequency, ping_time) float64 3.888e+03 ...
│           offset                          (frequency, ping_time) float64 0.0 ... 0.0
│           transmit_mode                   (frequency, ping_time) float64 0.0 ... 0.0
│           angle_athwartship               (frequency, ping_time, range_sample) float64 ...
│           angle_alongship                 (frequency, ping_time, range_sample) float64 ...
│       Attributes:
│           beam_mode:              vertical
│           conversion_equation_t:  type_3
└── DataTree('Vendor')
    Dimensions:           (frequency: 3, pulse_length_bin: 5)
    Coordinates:
      * frequency         (frequency) float64 1.8e+04 3.8e+04 1.2e+05
      * pulse_length_bin  (pulse_length_bin) int64 0 1 2 3 4
    Data variables:
        sa_correction     (frequency, pulse_length_bin) float64 0.0 -0.83 ... -0.34
        gain_correction   (frequency, pulse_length_bin) float64 20.3 23.35 ... 26.62
        pulse_length      (frequency, pulse_length_bin) float64 0.000512 ... 0.00...
```

</p>
</details>

<details><summary>Open Converted</summary>
<p>

```python
from echopype import open_raw
echodata = open_converted('echopype/test_data/ek60/ncei-wcsd/Summer2017-D20170615-T190214.zarr')

# Note that without the backward compatibility current version missed the beam group!!!
In [18]: print(repr(echodata))
EchoData: standardized raw data from /home/lsetiawan/GitRepos/GitHub/echopype/echopype/test_data/ek60/ncei-wcsd/Summer2017-D20170615-T190214.zarr
  > top: (Top-level) contains metadata about the SONAR-netCDF4 file format.
  > environment: (Environment) contains information relevant to acoustic propagation through water.
  > platform: (Platform) contains information about the platform on which the sonar is installed.
  > nmea: (Platform/NMEA) contains information specific to the NMEA protocol.
  > provenance: (Provenance) contains metadata about how the SONAR-netCDF4 version of the data were obtained.
  > sonar: (Sonar) contains specific metadata for the sonar system.
  > vendor: (Vendor specific) contains vendor-specific information about the sonar and the data.

In [19]: print(echodata)
EchoData: standardized raw data from /home/lsetiawan/GitRepos/GitHub/echopype/echopype/test_data/ek60/ncei-wcsd/Summer2017-D20170615-T190214.zarr
DataTree('Top-level', parent=None)
│   Dimensions:  ()
│   Data variables:
│       *empty*
│   Attributes:
│       conventions:                 CF-1.7, SONAR-netCDF4-1.0, ACDD-1.3
│       date_created:                2017-06-15T19:02:14Z
│       keywords:                    EK60
│       sonar_convention_authority:  ICES
│       sonar_convention_name:       SONAR-netCDF4
│       sonar_convention_version:    1.0
│       summary:                     
│       survey_name:                 
│       title:                       
├── DataTree('Beam')
│   Dimensions:                         (frequency: 3, ping_time: 19,
│                                        range_bin: 3888)
│   Coordinates:
│     * frequency                       (frequency) float64 1.8e+04 3.8e+04 1.2e+05
│     * ping_time                       (ping_time) datetime64[ns] 2017-06-15T19:...
│     * range_bin                       (range_bin) int64 0 1 2 3 ... 3885 3886 3887
│   Data variables: (12/30)
│       angle_alongship                 (frequency, ping_time, range_bin) float64 ...
│       angle_athwartship               (frequency, ping_time, range_bin) float64 ...
│       angle_offset_alongship          (frequency) float64 ...
│       angle_offset_athwartship        (frequency) float64 ...
│       angle_sensitivity_alongship     (frequency) float64 ...
│       angle_sensitivity_athwartship   (frequency) float64 ...
│       ...                              ...
│       transducer_offset_y             (frequency) float64 ...
│       transducer_offset_z             (frequency) float64 ...
│       transmit_bandwidth              (frequency, ping_time) float64 ...
│       transmit_duration_nominal       (frequency, ping_time) float64 ...
│       transmit_mode                   (frequency, ping_time) float64 ...
│       transmit_power                  (frequency, ping_time) float64 ...
│   Attributes:
│       beam_mode:              vertical
│       conversion_equation_t:  type_3
├── DataTree('Environment')
│   Dimensions:                 (frequency: 3, ping_time: 19)
│   Coordinates:
│     * frequency               (frequency) float64 1.8e+04 3.8e+04 1.2e+05
│     * ping_time               (ping_time) datetime64[ns] 2017-06-15T19:02:14.20...
│   Data variables:
│       absorption_indicative   (frequency, ping_time) float64 ...
│       sound_speed_indicative  (frequency, ping_time) float64 ...
├── DataTree('Platform')
│   │   Dimensions:        (frequency: 3, ping_time: 19, location_time: 72)
│   │   Coordinates:
│   │     * frequency      (frequency) float64 1.8e+04 3.8e+04 1.2e+05
│   │     * location_time  (location_time) datetime64[ns] 2017-06-15T19:02:15.4450001...
│   │     * ping_time      (ping_time) datetime64[ns] 2017-06-15T19:02:14.206000128 ....
│   │   Data variables:
│   │       heave          (frequency, ping_time) float64 ...
│   │       latitude       (location_time) float64 ...
│   │       longitude      (location_time) float64 ...
│   │       pitch          (frequency, ping_time) float64 ...
│   │       roll           (frequency, ping_time) float64 ...
│   │       sentence_type  (location_time) <U3 ...
│   │       water_level    (frequency, ping_time) float64 ...
│   └── DataTree('NMEA')
│       Dimensions:        (location_time: 688)
│       Coordinates:
│         * location_time  (location_time) datetime64[ns] 2017-06-15T19:02:14.2059996...
│       Data variables:
│           NMEA_datagram  (location_time) <U73 ...
│       Attributes:
│           description:  All NMEA sensor datagrams
├── DataTree('Provenance')
│   Dimensions:  ()
│   Data variables:
│       *empty*
│   Attributes:
│       conversion_software_name:     echopype
│       conversion_software_version:  0.4.1.dev438+g7aa2cd0.d20210409
│       conversion_time:              2021-04-16T17:48:07Z
│       src_filenames:                ./echopype/test_data/ek60/ncei-wcsd/Summer2...
├── DataTree('Sonar')
│   Dimensions:  ()
│   Data variables:
│       *empty*
│   Attributes:
│       sonar_manufacturer:      Simrad
│       sonar_model:             ER60
│       sonar_serial_number:     
│       sonar_software_name:     
│       sonar_software_version:  2.4.3
│       sonar_type:              echosounder
└── DataTree('Vendor')
    Dimensions:           (frequency: 3, pulse_length_bin: 5)
    Coordinates:
      * frequency         (frequency) float64 1.8e+04 3.8e+04 1.2e+05
      * pulse_length_bin  (pulse_length_bin) int64 0 1 2 3 4
    Data variables:
        gain_correction   (frequency, pulse_length_bin) float64 ...
        pulse_length      (frequency, pulse_length_bin) float64 ...
        sa_correction     (frequency, pulse_length_bin) float64 ...

```

</p>
</details>

echopype/convert/api.py
```diff
@@ -427,28 +429,43 @@ def open_raw(
...
+        # make into list
+        beam_groups = [beam_groups]
+
+    for idx, beam_group in enumerate(beam_groups, start=1):
+        tree_dict[f"Sonar/Beam_group{idx}"] = beam_group
+
+    tree_dict["Vendor"] = setgrouper.set_vendor()
+
+    # Create tree and echodata
+    # TODO: make the creation of tree dynamically generated from yaml
```
b-reyes, line 465: Just curious about this TODO. Is your idea here to replace the above code with a dynamic approach?

echopype/echodata/echodata.py
b-reyes: So, after `echodata._set_tree(tree)` I can add/remove a child/parent from the tree? Similarly, can I rename coords/variables and add/delete them or does this need to happen after `setattr`?

requirements.txt
```diff
@@ -8,3 +8,4 @@ xarray
 zarr
 fsspec
 s3fs
+xarray-datatree==0.0.4
```
b-reyes, line 11: Is this equality because this is being actively developed?

echopype/echodata/echodata.py
```diff
@@ -133,6 +142,118 @@ def __read_converted(self, converted_raw_path: Optional["PathHint"]):
...
+        if isinstance(converted_raw_path, fsspec.FSMap):
+            # Convert fsmap to Path so it can be used
+            # for retrieving the path strings
+            converted_raw_path = Path(converted_raw_path.root)
+        echodata.converted_raw_path = converted_raw_path
+        echodata._load_tree()
```
b-reyes, line 173: It looks like `_load_tree()` is only necessary because we are currently trying to temporarily preserve the way we are accessing groups (e.g. ed.beam). Is this correct?

echopype/echodata/echodata.py
```diff
@@ -133,6 +142,82 @@ def __read_converted(self, converted_raw_path: Optional["PathHint"]):
...
+        return echodata
```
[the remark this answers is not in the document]
b-reyes: I think that makes sense and is more understandable. I know the first time I looked at this code I had to "hunt down" what function was actually loading the file.

echopype/convert/api.py
lsetiawan: Yepp, this is from a source of truth somewhere? I'm not sure to be honest.

requirements.txt
lsetiawan: Yepp, it will be developed heavily, so I wanna make sure nothing breaks on our end unexpectedly.

echopype/echodata/echodata.py
lsetiawan: Yes sir. `_load_tree` creates those dynamically generated attributes.

lsetiawan:
From slack conversations, access need to change to be like `['/Groupname']` for subgroups from root aka 'Top-level'. Ongoing discussion about whether to access `Top-level` as `echodata['Top-level']` or not.. 

@b-reyes  asked:
> I understand why you have it. From what I understand, that is not how netcdf python does it (please correct me if I am wrong). I believe that ed would give the root group if we were following netcdf access patterns. Is that possible and if so, is this better?

My current thoughts:

> The thing is "EchoData" is essentially a wrapper object for the Sonarnetcdf. The problem with having Echodata (ed) as this root group means that we would have to put a wrapper to the Xarray Dataset since essentially when you are fetching these groups, you get the actual underlying dataset. So in my mind it makes sense to have this separation and call the actual groups with ['Group'] and leave EchoData as our own object that we work with that we can highly customize with various functions, etc.

> Right now when you retrieve a group such as ['/Sonar'] Ideally I'd like the repr to have the header of the name of that group, description, and any children, but not sure how to do that since our return value is an xarray Dataset... wonder if somehow I can put a wrapper to the ds repr :thinking_face:

echopype/convert/api.py
b-reyes: Sounds good!

requirements.txt
b-reyes: Makes sense!

echopype/echodata/echodata.py
b-reyes: Great!

emiliom:
Regarding netcdf4 group path references, including (especially!) the "root group":

First, a distinction. The encoding in `1.0.yml`, the SONAR-netCDF4 convention, and our own discussions are dealing with two parallel but not identical entities: a **label** (string) we may use verbally to conveniently refer (not access) to a group vs the group **path** (string) used to actually access a group (and we can add a third: if the group path is "parent/child", the group **name** is "child"). In `1.0.yml`, the group label is the group item's `name` attribute and the group path is the `ep_group` attribute. Here's a fragment from that file:

```yaml
groups:
  top:
    name: Top-level
    description: contains metadata about the SONAR-netCDF4 file format.
    ep_group:
...

  beam_power:
    name: Sonar/Beam_group2
    description: >-
      contains backscatter power (uncalibrated) and other beam or channel-specific data,
      including split-beam angle data when they exist.
      Only exists if complex backscatter data they already in Sonar/Beam_group1
    ep_group: Sonar/Beam_group2
```

Second, the root group a *special* group. It can have all the elements of a group (variables, dimensions, other groups), but it's path is simply "/". There's no group name per se. When you open a netCDF4 file with the Python netcdf4 package using `Dataset`, what's returned is the root group. Its `groups` property lists all the groups under it, but not itself. From the Python netcdf4 package documentation:

- [A `Dataset` creates a special group, called the 'root group', which is similar to the root directory in a unix filesystem.](https://unidata.github.io/netcdf4-python/#groups-in-a-netcdf-file)
- [`path` shows the location of the `Group` in the `Dataset` in a unix directory format (the names of groups in the
hierarchy separated by backslashes). A `Dataset` instance is the root group, so the path is simply `'/'`](https://unidata.github.io/netcdf4-python/#Dataset)

So, at one level, the root group is just another group. At another, it's a very special group with no name other than "/". SONAR-netCDF4 describes it in a way that's completely parallel to all other groups. The conventions gives *labels* to all groups, b/c it's convenient, and the label for the root group is "Top-level".

As you can see in netcdf4 an explicit analogy is made between group paths and unix directory paths. But I'd say its implementation is a bit loose. In xarray `open_dataset` you can interchangeably use or omit the initial "/" when using the `group` parameter to specify a group. Same difference. You can get the root group in 3 ways: by omitting the `group` parameter, or by passing either an empty string "" or a slash "/". In the netcdf4 package, slashes are not a part of the group name, but of course it's part of the path string. Hence, SONAR-netCDF4's reference to, say, "/Sonar" as the name of a group is somewhat pedantic and maybe confusing. It'd be necessary only if a "Sonar" group could occur somewhere else other than at the`/Sonar` path. 

Whew.

The `EchoData` object follows the SONAR-netCDF4 of listing all groups equally, including the root group. The convention label for the root group is "Top-level". Check. Because the `EchoData` object is our own customized thing that exists for convenience, we get to define its features. The implicit decisions we've made in the past are that each group returns an xarray dataset. xarray datasets don't know anything about groups. That's ok, we choose to stick with xarray datasets for their usability. So, `ed["Top-level"]` is perfectly fine as a group like any other which returns an xarray dataset and doesn't know anything about group hierarchies. `ed["Sonar"]` is also an xarray dataset that doesn't know the netcdf4 `Sonar` group has `Beam_groupX` children. The group hierarchy will only be found in the `EchoData` getters and the `group_paths` attribute @lsetiawan has created.

Could this arrangement cause confusion? Sure. But does have its own self consistency and clear motivations and benefits? Absolutely.

One possible change we could make, to be more explicit, is to set the top-level `ep_group` to "/" rather than None, and for all other groups make sure they start with "/". It's not necessary, but it might add clarity. If we make this change, we'll have to also change other modules and tests.

lsetiawan:
Thanks @emiliom for your thoughts on this and your research! This is great.

> You can get the root group in 3 ways: by omitting the group parameter, or by passing either an empty string "" or a slash "/".

I guess in our case, we can have 4 ways: `ed['Top-level']`, `ed[None]`, `ed['']`, or `ed['/']`

> One possible change we could make, to be more explicit, is to set the top-level ep_group to "/" rather than None, and for all other groups make sure they start with "/". It's not necessary, but it might add clarity. If we make this change, we'll have to also change other modules and tests.

In terms of using `/` datatree already supports this option so one can either use `/Sonar` to be explicit or `Sonar`, assuming that user know that it is a subgroup under `Top-level`. So, when they initiate Echodata object from either `open_raw` or `open_converted` a clear repr to show this relationship is critical. Right now I do like how the group description is laid out and maybe a change can be showing the group path:

```
EchoData: standardized raw data from Internal Memory
  Top-level (Top-level): contains metadata about the SONAR-netCDF4 file format.
  |--Provenance (/Provenance): contains metadata about how the SONAR-netCDF4 version of the data were obtained.
  |--Sonar (/Sonar): contains specific metadata for the sonar system.
  |  |--Beam_group1 (/Sonar/Beam_group1):  contains backscatter data (either complex samples or uncalibrated power samples) and other beam or channel-specific data, including split-beam angle data when they exist.
  |--Vendor specific (/Vendor): contains vendor-specific information about the sonar and the data.
```

echopype/convert/api.py
```diff
@@ -427,28 +429,43 @@ def open_raw(
...
+        tree_dict["Platform/NMEA"] = setgrouper.set_nmea()
+    tree_dict["Provenance"] = setgrouper.set_provenance()
+    tree_dict["Sonar"] = setgrouper.set_sonar()
+
+    # Set multi beam groups
+    beam_groups = setgrouper.set_beam()
+    if isinstance(beam_groups, xr.Dataset):
+        # if it's a single dataset like the ek60,
+        # make into list
+        beam_groups = [beam_groups]
```
lsetiawan, line 457: @emiliom Do you think there's a way that `set_beam()` always return a list of `xr.Dataset` even if there's only one?

echopype/echodata/echodata.py
```diff
@@ -59,12 +61,19 @@ def __init__(
...
+        # self.__read_converted(converted_raw_path)
 
         self._varattrs = sonarnetcdf_1.yaml_dict["variable_and_varattributes"]
 
+    def __str__(self) -> str:
+        fpath = "Internal Memory"
+        if self.converted_raw_path:
+            fpath = self.converted_raw_path
+        return f"EchoData: standardized raw data from {fpath}\n{str(self._tree)}"
+
```
lsetiawan, line 76: I'll sync up `repr` stuff in an another PR after this is merged.

echopype/echodata/echodata.py
lsetiawan: This function just puts the tree into `self._tree`. You can add/remove a child/parent from the tree, but seems like this is not implemented upstream in datatree yet. So you'd have to think about overwriting this tree after you get it to how you want it internally.

b-reyes:
> Right now I do like how the group description is laid out and maybe a change can be showing the group path:

I guess this could be helpful for very large trees, but I don't think our tree is large enough to where you can't see both the parent and the child at the same time. For this reason, I am somewhat against putting the group paths as it adds clutter.

> I guess in our case, we can have 4 ways: ed['Top-level'], ed[None], ed[''], or ed['/'] though I like the very explicit 'Top-level' best.

I think that `'Top-level'` is good (because it is explicit) and I guess when it comes down to it, we probably are not accessing this information too often. So, the length isn't that big of a deal.

echopype/convert/api.py
emiliom: @lsetiawan I've looked at `set_groups_ek80.py`, which already returns a list of `xr.Dataset`, vs ek60 and azfp which return a single `xr.Dataset`. I don't see any problem with changing `set_beam()` so it always returns a list. I like that suggestion. Though it's not necessarily for this PR, right?

echopype/echodata/echodata.py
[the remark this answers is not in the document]
emiliom: It looks fine.

lsetiawan:
> I think that 'Top-level' is good (because it is explicit) and I guess when it comes down to it, we probably are not accessing this information too often. So, the length isn't that big of a deal.

Okay. If everyone else agrees, then I'll keep only to one option of `['Top-level']` to access the root to keep this consistency among the other groups.

> I guess this could be helpful for very large trees, but I don't think our tree is large enough to where you can't see both the parent and the child at the same time. For this reason, I am somewhat against putting the group paths as it adds clutter.

Sounds good. I will not include the paths, and I think we just need a good doc to explain how to get at the path. I'm sure most people are already familiar on how to figure this out as long as we can show the relationships. 

Thank you @b-reyes and @emiliom for your inputs. I'll make my next changes soon and let you know when it's ready for a full review! This is so exciting! 🥳

emiliom:
Regarding `Top-level` and root: give me a bit more time. I'm actually working on it at this very instant!

lsetiawan:
No rush. I will not be working on this today. 😄

emiliom:
> No rush. I will not be working on this today.

Got it. I read your comment as implying you were going to do it, say, this morning!

echopype/convert/api.py
emiliom: > I'll keep this check for now, but if you are planning to change this

Nope, no plan to do this. I don't want to expand the scope of this PR too much.

emiliom:
Sorry, but I'll go philosophical again .... This issue of `Top-level` as just another group vs as the netcdf4 root group (a special group) is tricky.

The problem is that we're dealing with two presentations that don't fully match up. The netcdf4 format description is pretty unambiguous about presenting the hierarchical group structure, including root being a special group that's not accessed quite the same way as all other child groups (BTW, `datatree` follows that structure nicely). But the SONAR-netCDF4 v1 convention doesn't present it that way. Instead, it never describes `Top-level` as being the root under which all the other groups reside and can be accessed. It includes it in the same flat list as Sonar, Platform, etc, and simply describes it like this: "2.10.1 Top-level group. The top-level group contains metadata about the SONAR-netCDF4 file, represented as attributes in the group". That's it! The rare user who would scan the SONAR-netCDF4 v1 convention doc to learn about it would not clearly come away with the "root" interpretation. In addition, unlike all other groups defined by the convention, the group name "Top-level" is a construct that only exists conceptually. A netcdf/zarr file created by echopype will not contain a group called "Top-level", because it can't. In netcdf4 and zarr, root doesn't have a label; it's just root. Someone opening an echopype-generated file using *any* generic netcdf library (xarray, netcdf4, or a library in any other language) will never see a reference to "Top-level", b/c it's not there.

Our `EchoData` object and especially its repr is a nice, customized mechanism that presents a user-friendly view of the SONAR-netCDF4 dataset that aligns with the SONAR-netCDF4 v1 presentation. echopype *adds* the "Top-level" label that's not actually there in the source nc/zarr file or the xarray datasets. And now we're considering adding to the "Top-level" group the nc4 root group concept that's not actually present in the SONAR-netCDF4 v1 convention. I don't think that makes things clearer or easier to all users. The netcdf4-savvy users may be puzzled about the "Top-level" group that doesn't exist in the file, and the less technical user who's familiar with the SONAR-netCDF4 v1 convention may be puzzled by the root hierarchy presentation that's not present in SONAR-netCDF4.

**My recommendation:**
- In the `EchoData` object, stick to the SONAR-netCDF4 view of the world. Treat `Top-level` as just another group with no child subgroups. Its group path (`ep_group`) will be "/", for clarity. `ed["Top-level/Sonar"]` will not be supported and will give an error.
- The EchoData group getter can either support just the group label (eg, "Top-level") string, or both the group label and the group path (eg, "Top-level" and "/"). I could go either way but lean slightly towards the latter. For "Top-level" getter access, let's not support `ed[None]` and `ed['']`; I think that's confusing.

emiliom:
Just wanted to add that I've tested this PR as it stands right now (or last night), locally. I ran the complete test suite and also explored the results of `open_raw` and `open_converted` fairly extensively. Everything worked great. Great job @lsetiawan !

That doesn't mean it's ready to merge. We just need to arrive at a consensus about `Top-level`, group getters and `repr` (and HTML `repr`) details; I'll add some comments on the latter later today.

b-reyes:
> The EchoData group getter can either support just the group label (eg, "Top-level") string, or both the group label and the group path (eg, "Top-level" and "/"). I could go either way but lean slightly towards the latter. For "Top-level" getter access, let's not support ed[None] and ed['']; I think that's confusing.

I like the latter too. I think both "Top-level" and "/" are natural options that one might choose.

echopype/echodata/echodata.py
```diff
@@ -133,6 +142,118 @@ def __read_converted(self, converted_raw_path: Optional["PathHint"]):
...
+    @property
+    def group_paths(self) -> Set[str]:
+        paths = set()
+        root_path = self._tree.pathstr
+        paths.add(root_path)
+        for desc in self._tree.descendants:
+            child_path = desc.pathstr.replace(root_path + "/", "")
+            paths.add(child_path)
+        return paths
```
b-reyes, line 220: I think you can use Datatree's built-in group function for this:
```
def group_paths(self) -> Set[str]:
    root_path = self._tree.pathstr
    return set([i.replace(root_path + "/", "") for i in self._tree.groups])
```

gavinmacaulay:
Some comments on how the root group was presented in SONAR-netCDF4 v1: 
- The absence of an explanation of the hierarchal nature of the groups was an oversight and I suggest not reading too much into that absence.
- `Top-level` was intended as a more descriptive phrase for `/` or `root`.
- It was intended that `Top-level` be the netCDF4 root group that all other groups are subgroups of, not that there be a group called `Top-level`.

echopype/echodata/echodata.py
emiliom: That does seem to work just as well, and it's more compact and easier to understand.

emiliom:
@gavinmacaulay thanks for that background on the intent for the root / `Top-level` group.

echopype/echodata/echodata.py
lsetiawan: Thanks @b-reyes for this suggestion :) Have you tried the direct github code suggestion feature? You should try it so your suggestion can be credited to you! 😄

lsetiawan:
Consensus from echopype meeting:
- Top-level underlying tree name should change to default to "root"
- `ed.group_paths` is fine for listing the Sonar netcdf groups pathings to pass on access to echodata
- Top-level should be accessed with either `['Top-level']` or `['/']`.
- Repr should have hierarchy of the tree:
  ```
  EchoData: standardized raw data from Internal Memory
  Top-level (Top-level): contains metadata about the SONAR-netCDF4 file format.
  |--Provenance (/Provenance): contains metadata about how the SONAR-netCDF4 version of the data were obtained.
  |--Sonar (/Sonar): contains specific metadata for the sonar system.
  |  |--Beam_group1 (/Sonar/Beam_group1):  contains backscatter data (either complex samples or uncalibrated power samples) and other beam or channel-specific data, including split-beam angle data when they exist.
  |--Vendor specific (/Vendor): contains vendor-specific information about the sonar and the data.
  ```

emiliom:
Thanks for the nice summary, @lsetiawan !

lsetiawan:
## Demo
Below is the demo of the echodata representations

### repr or str
Screenshot from 2022-04-11 11-29-32

### html repr
Screenshot from 2022-04-11 11-29-20

emiliom:
That's awesome, @lsetiawan ! Both repr's look good. My only comment is about indentation, especially for the Beam_groupX groups. For both the text and HTML repr's, b/c the group description is long the text wraps into a new line. But the wrapped line is not indented, so it looks awkward and breaks the visual arrangement a bit. For the html repr is a small, minor effect, b/c there's still an overall indentation. For the text repr it's a bigger impact.

lsetiawan:
> it looks awkward and breaks the visual arrangement a bit.
Yea... that's because of my small screen... I'm not sure how to fix that. If you or brandon have any suggestion, please let me know. I'm at a loss on that one 😅

emiliom:
> Yea... that's because of my small screen...

Well, the small screen *exposes* the wrapping issue. But the wrapping issue is there. 

Looking at the example of the current (0.5.x) text repr you [included earlier](https://github.com/OSOceanAcoustics/echopype/pull/611#issuecomment-1085245760), the long text doesn't wrap; it just extends beyond the size of the window. Is that just the default markdown block behavior vs the behavior in a terminal?

One alternative is to clip the line at a certain maximum length. That's done in the DataTree text repr (you also included an example above last week), where the line gets clipped and a "..." string is added. Anyway, I'm not saying I think that's a better option. Just an alternative to consider.

lsetiawan:
I'll take a look at the css and see if I can make adjustments there. I think the `...` is good.

> Is that just the default markdown block behavior vs the behavior in a terminal?

That example is the default behaviour of the markdown block. If you try this on a terminal, it will wrap.

emiliom:
> I'll take a look at the css and see if I can make adjustments there. I think the ... is good.

Maybe wait to see what @b-reyes and @leewujung think? I think I prefer the ... behavior, but I could go either way.

leewujung:
I don't think the lack of indent is a problem in the HTML repr since the arrowhead is clear on what the expected behavior is. I agree this is a bigger problem for the text repr, but I think terminal in general has that behavior and it is expected by the user. Going with `...` is fine with me also. 

I would advocate for getting this merged and shelve the text repr "problem" to a separate issue that would potentially to be fixed in the future.

b-reyes:
@lsetiawan what happens if you throw a newline in the string?

lsetiawan:
> I would advocate for getting this merged and shelve the text repr "problem" to a separate issue that would potentially to be fixed in the future.

Yea. I think this will take more thinking in terms of the repr behavior. And probably better to put it to a separate PR since this PR is getting really big.

lsetiawan:
Yea... looks like even default `xarray` and `datatree` repr wraps funky in a small enough screen.
Screenshot from 2022-04-11 14-18-13

emiliom:
I just realized that when I generated a screenshot of the spacing between groups in the HTML repr just now, I ended up running my set of "tests" on this PR! That's because I ran the notebook where I use the datatree accessors in different ways. SO, I can say that this PR is good to go!

emiliom (approved):
Woo-hoo!!

b-reyes (changes requested):
@lsetiawan thank you for all of your great work Don! 

Two items that I think need to be reviewed: 

* There are a couple of places that are trying to import from echopype. However, these references seem to be incorrect and will attempt to import using whatever echopype is in your environment. Please see my suggested changes. 

* In a past meeting, I think we talked about changing the name `group_paths` to `group_names` or something like that.

echopype/echodata/convention/utils.py
```diff
@@ -0,0 +1,16 @@
+from echopype.echodata.convention import sonarnetcdf_1
```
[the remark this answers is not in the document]
lsetiawan: The above method that I'm using is more explicit, but I can go with your suggestion.

echopype/echodata/echodata.py
```diff
@@ -133,6 +109,117 @@ def __read_converted(self, converted_raw_path: Optional["PathHint"]):
...
+    def version_info(self) -> Tuple[int]:
+        if self["Provenance"].attrs.get("conversion_software_name", None) == "echopype":
+            version_str = self["Provenance"].attrs.get("conversion_software_version", None)
+            if version_str is not None:
+                version_num = version_str.split(".")[:3]
+                return tuple([int(i) for i in version_num])
+        return None
+
+    @property
+    def group_paths(self) -> Set[str]:
```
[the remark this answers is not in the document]
lsetiawan: I thought we concluded on group paths?

echopype/echodata/widgets/widgets.py
```diff
@@ -0,0 +1,58 @@
...
+
+HERE = Path(__file__).parent
+STATIC_DIR = HERE / "static"
+
+TEMPLATE_PATHS = [HERE / "templates"]
+
+STATIC_FILES = (
+    "echodata/widgets/static/html/icons-svg-inline.html",
+    "echodata/widgets/static/css/style.css",
+)
```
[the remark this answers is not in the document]
lsetiawan: this is no longer in the top folder, so your suggestion is incorrect.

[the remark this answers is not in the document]
lsetiawan: Oh wait.. it connects to your suggestion below! Okay. cool!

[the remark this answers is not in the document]
b-reyes: Ah, yes. I should have mentioned that! Sorry for the confusion.

echopype/echodata/convention/utils.py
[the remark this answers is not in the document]
b-reyes: If you use this more explicit solution, then it will not work if you are trying to run straight from the codebase.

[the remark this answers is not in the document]
lsetiawan: I actually like the more explicit. So I'm going to resolve this conversation.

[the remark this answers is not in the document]
lsetiawan: It is actually recommended to be more explicit. https://peps.python.org/pep-0008/#imports

Things should work. You have to do `pip install -e .`

echopype/echodata/echodata.py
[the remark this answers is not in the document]
lsetiawan: Based on my summary comment: https://github.com/OSOceanAcoustics/echopype/pull/611#issuecomment-1092001292. It is `group_paths`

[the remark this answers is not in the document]
emiliom: I had suggested `group_names` or something other than `group_paths`. I think in the end we left it at `group_paths`. "Consensus through exhaustion" :laughing:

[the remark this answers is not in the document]
b-reyes: Sounds good haha. I think we can go ahead and resolve this now then.

echopype/echodata/convention/utils.py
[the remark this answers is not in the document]
emiliom: I think what you're discussing here is absolute vs relative imports, not "explicit" vs less explicit. @lsetiawan thanks for that helpful pep8 link. What I'm reading there is that explicit is always preferred (no surprise), absolute imports are recommended, but relative imports are an acceptable alternative. Also, I think relative imports is the most common practice in echopype, so for code style consistency I would suggest relative imports (@b-reyes 's suggestion)

lsetiawan:
I think we are good to go now. Once I have @b-reyes seal of approval. I will merge 😄

echopype/echodata/convention/utils.py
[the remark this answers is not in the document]
lsetiawan: Alright. 😛 Though I think it would be good for echopype to update this in the near future to be a more absolute imports for better error messaging, as PEP8 nicely put it:
> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages)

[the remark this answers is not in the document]
b-reyes: I agree, we should clean this up in the future.

b-reyes (approved):

Issue #586, opened 2022-03-15
Addition of unit tests for the variable water_level


b-reyes:
Currently, it looks like there are not any unit tests for checking whether or not `water_level` in the `platform` group of the EchoData object is properly converted and stored. It seems like the proper place to add these unit tests is in `./echopype/tests/calibrate`. I suggest we add these tests soon.

leewujung:
This was from #579.

Issue #516, opened 2021-12-14
Include water_level or range_offset in Sv dataset


emiliom:
From @leewujung, https://github.com/OSOceanAcoustics/echopype/pull/436#event-5750581750:

One thing I noticed which we'll need to do something about (also requested in #259 ) is to include the addition of water_level in the Sv dataset (since right now even if users put in `water_level=True` if the input is an Sv dataset no water_level offset is added.  A more accurate way to think about it is probably that we are actually plotting "depth" in the case when we have range_meters + water_level.

The more general description is that people are plotting the range _offset_ from some platform. Let's discuss how that coordinate should be named...  Hence I am wondering if `water_level` should be `range_offset` instead, so that it is more general. 

Pinging @emiliom here since this is related to the whole convention and attribute conversation.

b-reyes:
1. To allow for the inclusion of the water level in the Sv dataset, I propose that we make this an option in the calibrate function [compute_Sv](https://echopype.readthedocs.io/en/stable/api/echopype.calibrate.compute_Sv.html). For example, we could have the function parameter "include_range_offset" with a default value of False. 

In regards to actually implementing this, I believe the simplest place to do this is in the if statement i.e. line 56 in the code below:

https://github.com/OSOceanAcoustics/echopype/blob/17a86e4c87f28d1457c582062e3ac59367b37348/echopype/calibrate/api.py#L54-L58

this will allow us to skip duplicate code in the calibrated classes for each of the echosounders. 

2. In the returned Xarray Dataset, I also propose that we call the data variable "range_offset" to allow for the generality previously mentioned.

emiliom:
> Hence I am wondering if `water_level` should be `range_offset` instead, so that it is more general.

I'm only now paying close attention to the convention context for this change (sorry!), and realizing that `water_level` is specified in the convention but `range_offset` isn't. That's also the case for the vers. 2 draft of the convention. While I see the logic for this generalization, @leewujung could you flesh out your argument for deviating from the convention?

leewujung:
Sigh, yes, in commenting on your comment to #579 I just went back to the convention again and found the same thing. I'm copying my comment there below:

[...]

PR #615, opened 2022-04-06
Add attributes for Sp and MVBS from index binning

emiliom:
Addresses remaining tasks in #582 and #481

### DIFF
--- a/echopype/calibrate/api.py
+++ b/echopype/calibrate/api.py
@@ -77,8 +77,8 @@ def _compute_cal(
             sv_dataset["water_level"] = echodata.platform.water_level
         return sv_dataset
     else:
-        sp_dataset = cal_obj.compute_Sp(waveform_mode=waveform_mode, encode_mode=encode_mode)
-        add_attrs("Sp", sp_dataset)
+        sp_dataset = cal_obj.compute_TS(waveform_mode=waveform_mode, encode_mode=encode_mode)
+        add_attrs("TS", sp_dataset)
         return sp_dataset
 
 
@@ -171,9 +171,9 @@ def compute_Sv(echodata: EchoData, **kwargs) -> xr.Dataset:
     return _compute_cal(cal_type="Sv", echodata=echodata, **kwargs)
 
 
-def compute_Sp(echodata: EchoData, **kwargs):
+def compute_TS(echodata: EchoData, **kwargs):
     """
-    Compute point backscattering strength (Sp) from raw data.
+    Compute target strength (TS) from raw data.
 
     The calibration routine varies depending on the sonar type.
     Currently this operation is supported for the following ``sonar_model``:
@@ -234,7 +234,7 @@ def compute_Sp(echodata: EchoData, **kwargs):
     Returns
     -------
     xr.Dataset
-        The calibrated Sp dataset, including calibration parameters
+        The calibrated TS dataset, including calibration parameters
         and environmental variables used in the calibration operations.
 
     Notes
@@ -250,7 +250,7 @@ def compute_Sp(echodata: EchoData, **kwargs):
     similar to those recorded by EK60 echosounders.
 
     The current calibration implemented for EK80 broadband complex data
-    uses band-integrated Sp with the gain computed at the center frequency
+    uses band-integrated TS with the gain computed at the center frequency
     of the transmit signal.
     """
-    return _compute_cal(cal_type="Sp", echodata=echodata, **kwargs)
+    return _compute_cal(cal_type="TS", echodata=echodata, **kwargs)
--- a/echopype/calibrate/api.py
+++ b/echopype/calibrate/api.py
@@ -51,7 +51,7 @@ def _compute_cal(
     )
 
     def add_attrs(cal_type, ds):
-        """ Add attributes to backscattering strength dataset.
+        """Add attributes to backscattering strength dataset.
         cal_type: Sv or Sp
         """
         ds["range_sample"].attrs = {"long_name": "Along-range sample number, base 0"}
--- a/echopype/calibrate/api.py
+++ b/echopype/calibrate/api.py
@@ -49,29 +49,34 @@ def _compute_cal(
         waveform_mode=waveform_mode,
         encode_mode=encode_mode,
     )
+
+    def add_attrs(cal_type, ds):
+        """ Add attributes to backscattering strength dataset.
+        cal_type: Sv or Sp
+        """
+        ds["range_sample"].attrs = {"long_name": "Along-range sample number, base 0"}
+        ds["echo_range"].attrs = {"long_name": "Range distance", "units": "m"}
+        long_name_cal_type = {"Sv": "Volume", "Sp": "Point"}
+        ds[cal_type].attrs = {
+            "long_name": f"{long_name_cal_type[cal_type]} backscattering strength ({cal_type})",
+            "units": "dB re 1 m-1",
+            "actual_range": [
+                round(float(ds[cal_type].min().values), 2),
+                round(float(ds[cal_type].max().values), 2),
+            ],
+        }
+
     # Perform calibration
     if cal_type == "Sv":
         sv_dataset = cal_obj.compute_Sv(waveform_mode=waveform_mode, encode_mode=encode_mode)
-        # Add attributes
-        sv_dataset["Sv"].attrs = {
-            "long_name": "Volume backscattering strength (Sv)",
-            "units": "dB re 1 m-1",
-            "actual_range": [
-                round(float(sv_dataset["Sv"].min().values), 2),
-                round(float(sv_dataset["Sv"].max().values), 2),
-            ],
-        }
-        sv_dataset["range_sample"].attrs = {"long_name": "Along-range bin (sample) number, base 0"}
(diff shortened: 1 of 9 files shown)


echopype/preprocess/api.py
```diff
@@ -156,29 +170,32 @@ def compute_MVBS_index_binning(ds_Sv, range_sample_num=100, ping_num=100):
...
-    # Attach attributes
-    ds_out.attrs = {
-        "binning_mode": "index",
-        "range_sample_num": range_sample_num,
-        "ping_num": ping_num,
-    }
+    _set_MVVBS_attrs(ds_MVBS)
+    ds_MVBS["Sv"] = ds_MVBS["Sv"].assign_attrs(
+        {
+            "comment": "MVBS binned on the basis of range_sample and ping number specified as index numbers",
```
emiliom, line 191: @leewujung pay attention to this addition. I couldn't come up with a `cell_methods` alternative. I was getting hung up with the specification of bins based on index numbers rather than physical units.
FYI, `comment` is a CF attribute.

emiliom, line 191: @leewujung pay attention to this addition. I couldn't come up with a `cell_methods` alternative. I was getting hung up with the specification of bins based on index numbers rather than physical units.
FYI, `comment` is a CF attribute.

echopype/preprocess/api.py
```diff
@@ -16,6 +16,30 @@ def _check_range_uniqueness(ds):
     ).all()
 
 
+def _set_MVVBS_attrs(ds):
```
leewujung, line 19: noticed a small typo "MVVBS" here

emiliom: Good eye. I'll fix it

echopype/preprocess/api.py
leewujung: Based on what you had for the case when MVBS is computed based on physical units (actual range in meters and ping time), how about something like below?
```
{
            "cell_methods": (
                f"ping_time: mean (interval: {ping_num} pings "
                "comment: ping_time is the interval start) "
                f"echo_range: mean (interval: {range_sample_num} samples along range "
                "comment: echo_range is the interval start)"
            ),
            "binning_mode": "sample number",
            "range_meter_interval": str(range_sample_num) + "samples along range",
            "ping_time_interval": str(ping_num) + "pings",
}
```

echopype/preprocess/api.py
emiliom: It's fixed.

echopype/preprocess/api.py
emiliom: Adding link to where I discussed CF `cell_methods`: https://github.com/OSOceanAcoustics/echopype/pull/594#pullrequestreview-921175073

emiliom: A challenge with your suggestion is the interval units string. [Per CF](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#recording-spacing-original-data): 
> The syntax is (`interval`: *value unit*), where *value* is a numerical value and *unit* is a string that can be recognized by UNIDATA’s Udunits package [[UDUNITS]](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#UDUNITS).

So, strictly speaking, using "pings" and "samples along range" would not be compliant. The difficulty is in trying to specify a constant interval that's based on the index number rather than the dimension units. Though CF also says this:

> The *unit* will usually be dimensionally equivalent to the unit of the corresponding dimension, but this is not required (which allows, for example, the interval for a standard deviation calculated from points evenly spaced in distance along a parallel to be reported in units of length even if the zonal coordinate of the cells is given in degrees).

Let me chew on this.

emiliom:
```
"range_meter_interval": str(range_sample_num) + "samples along range",
"ping_time_interval": str(ping_num) + "pings",
```
This seems confusing, since the value after "range_meter_interval" will not be in meters, nor will the value after "ping_time_interval" be in time units. It seems like the goal of these custom echopype attributes was to record the function parameters used, as is. If we find a way to settle on using `cell_methods`, the linkage between `ping_time` and `echo_range` and the binning method (mean) & interval values used would be explicit, so I'd then vote for retaining the original attribute names of "range_sample_num" and "ping_num".

leewujung: I am not entirely understanding what you are proposing, but when proposing the `cell_methods` I actually forgot to check the attribute names. It should be
```
"range_sample_interval": str(range_sample_num) + "samples along range",
"ping_interval": str(ping_num) + "pings",
```
ie "range_**sample**_interval" instead of "range_meter_interval", "ping_~time~_interval" instead of "ping_time_interval"

emiliom:
Pushed new commits:
- Replace Sp with TS. See https://github.com/OSOceanAcoustics/echopype/issues/582#issuecomment-1095258029
- Update dB reference levels and symbology. See https://github.com/OSOceanAcoustics/echopype/issues/582#issuecomment-1094504660
- Add MVBS_index_binning cell_methods attribute and update other provenance attributes

leewujung:
Great, thanks @emiliom !
The failed pre-commit check are just 2 lines that are too long:
https://github.com/OSOceanAcoustics/echopype/blob/dd369c1282abfdbea6cc479f6247e591cab69678/echopype/preprocess/api.py#L131
and
https://github.com/OSOceanAcoustics/echopype/blob/dd369c1282abfdbea6cc479f6247e591cab69678/echopype/preprocess/api.py#L197
For readability an `#noqa` may be in order.

echopype/calibrate/api.py
```diff
@@ -49,29 +49,36 @@ def _compute_cal(
...
-        sv_dataset["range_sample"].attrs = {"long_name": "Along-range bin (sample) number, base 0"}
-        sv_dataset["echo_range"].attrs = {"long_name": "Range distance", "units": "m"}
 
+    # Perform calibration
+    if cal_type == "Sv":
+        sv_dataset = cal_obj.compute_Sv(waveform_mode=waveform_mode, encode_mode=encode_mode)
+        add_attrs("Sv", sv_dataset)
         if "water_level" in echodata.platform.data_vars.keys():
             # add water_level to the created xr.Dataset
             sv_dataset["water_level"] = echodata.platform.water_level
```
leewujung, line 77: @emiliom : Not sure if this should be in another PR: 

Looking at this now, I realized the check and addition of `water_level` should be in the outputs for both compute_Sv and compute_TS so that users can use this to convert `echo_range` to `depth` easily. 

Also in the realm of this PR, the attributes of `water_level` can be set in `_set_MVBS_attrs` below under echopype/preprocess/api.py.

leewujung: Related to #516 #583 #586

emiliom: It could be a separate PR. But this looks like an easy change w/o additional ramifications that will lead to protracted discussions. It's also in line with the broader improvements to `compute_MVBS_index_binning`. So, I'll just add it to this PR.

leewujung: Sounds good. Thank you @emiliom !

emiliom: > Also in the realm of this PR, the attributes of water_level can be set in _set_MVBS_attrs below under echopype/preprocess/api.py.

Actually, `water_level` retains its original attributes from the echodata `Platform` group. So, no attribute assignment is taking place in `calibrate/api.py`.

I've implemented the addition of `water_level` to `compute_TS` (and in the process it made the code quite a bit simpler!). I'll push it to the PR in a bit.

emiliom:
> For readability an #noqa may be in order.

Done

leewujung:
I think you had a typo on noqa 🤪

emiliom:
> I think you had a typo on noqa

:weary: Sigh. I'll fix it now.

emiliom:
pre-commit is now passing. Nice!

leewujung:
Thanks @emiliom ! I'll add a small thing about TS applicability in the docstring later today, and then merge this.

emiliom:
I think what remains an open question is the `cell_methods` implementation. As I wrote earlier (above):

> A challenge with your suggestion is the interval units string. [Per CF](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#recording-spacing-original-data): 
>> The syntax is (`interval`: *value unit*), where *value* is a numerical value and *unit* is a string that can be recognized by UNIDATA’s Udunits package [[UDUNITS]](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#UDUNITS).
>
> So, using "pings" and "samples along range" would not be CF compliant. The difficulty is in trying to specify a constant interval that's based on the index number rather than the dimension units. 

I don't have a solution. I think we should just merge this PR (after your additions), then come back to it in the future.

leewujung:
@emiliom : I think you should make a decision here since I am not familiar with the CF conventions here. 

I will merge this after my addition of TS-related docstring.

leewujung:
@emiliom : do you wanna capture this leftover dilemma in a new issue? Thanks!

emiliom:
> @emiliom : do you wanna capture this leftover dilemma in a new issue? Thanks

Done. #629

Issue #518, opened 2021-12-15
Rename time dimensions and coordinates in Platform group


emiliom:
Rename time dimensions and coordinates in `Platform` group, to adhere to the convention naming guideline. `location_time` does not comply; it should be `time1`, `time2`, etc (though location_time *is* more intuitive!)

~~**TODO:** Add a link/reference to the SONAR-netCDF4 convention section that discusses this~~

emiliom:
### SONAR-netCDF4 v1.0 and echopype implementation

`Platform` time dimensions and coordinates are discussed in the [SONAR-netCDF4 v1.0 report](https://www.ices.dk/sites/pub/Publication%20Reports/Cooperative%20Research%20Report%20(CRR)/CRR341.pdf) in section 2.10.4, page 8:

Dimensions
Description | Obligation | Comment
----------------|---------------|--------------
`time1` = unlimited | | Can be of fixed or unlimited length, as appropriate.
`time2` = unlimited | | Can be of fixed or unlimited length, as appropriate. Use as many time-dimensions as necessary with names of `timeX`, where X is a positive integer.

Corresponding Coordinate variables:   
`uint64 time1(time1)`, `uint64 time2(time2)`. In the examples provided, the `long_name` entries used for illustration are "Timestamps for position data" and "Timestamps for gyrocompass data".

**echopype** currently names the Platform time dimension and coordinate variable as `location_time` for the most common platform data from GPS. Additional platform sensor data have been introduced, currently using a similarly hard-wired and intelligible dimension and coordinate name.

emiliom:
### SONAR-netCDF4 Version 2 draft

The [rendered (updated daily, I think) version of the SONAR-netCDF4 Version 2 draft](https://htmlpreview.github.io/?https://github.com/ices-publications/SONAR-netCDF4/blob/master/Formatted_docs/crr341.html) includes significant changes to the `Platform` group. They are partly summarized in [7.1. Significant changes from version 1 to version 2](https://htmlpreview.github.io/?https://github.com/ices-publications/SONAR-netCDF4/blob/master/Formatted_docs/crr341.html#_significant_changes_from_version_1_to_version_2). However, the [description of the Platform group](https://htmlpreview.github.io/?https://github.com/ices-publications/SONAR-netCDF4/blob/master/Formatted_docs/crr341.html#_platform_group), particularly the introduction, is a much better and clearer source of these changes.

[...]

PR #616, opened 2022-04-07
Add additional EK80 environment variables

### LINKED ISSUE #540: Include additional environment variables stored in EK80 files to converted dataset
EK80 files contain a lot more environmental variables in addition to the 5 ("temperature", "depth", "acidity", "salinity", "sound_speed") that we currently save in the converted dataset. For example, the current test file `D20170912-T234910.raw` contains the following:
Screen Shot 2022-01-22 at 4 54 50 AM

We should add those that are of the nature of environmental variables (such as "sound_velocity_profile", "sound_velocity_source") to the output variables into the Environment group, and others, such as "water_level_draft" and "water_lebel_draft_is_manual" into the Platform group, so on and so forth.

I am not sure why there is a "latitude" variable in this...

This is related to #531 .

imranmaj:
Includes additional EK80 environment variables in the converted dataset.

Resolves #540


imranmaj:
@leewujung I didn't include the `latitude` field in the raw file which (according to the discussion in #540) is used to calculate sound speed; should that be included also? I wasn't sure if it was necessary since the latitude is already in the Platform group and also the `sound_speed`, which I assume uses the `latitude`, has already been calculated and included as `sound_speed_indicative`, so there isn't any use for it anymore.

leewujung:
@leewujung will look into why there is no depth specification attached to the profile.

leewujung:
https://github.com/OSOceanAcoustics/echopype/issues/540#issuecomment-1091948048 and the following comments clarified this.

@imranmaj : Once we settle down the variable name and attributes re https://github.com/OSOceanAcoustics/echopype/issues/540#issuecomment-1091948048 and https://github.com/OSOceanAcoustics/echopype/issues/540#issuecomment-1093369009 please add those to this PR. Thanks!

echopype/convert/set_groups_ek80.py
```diff
@@ -39,6 +39,31 @@ def set_env(self, env_only=False) -> xr.Dataset:
...
+                    "units": "m/s",
+                    "valid_min": 0.0,
+                    "comment": "parsed from raw data files as (depth, sound_speed) value pairs",
+                },
+            )
+
+        vars = {
+            "sound_velocity_source": "sound_velocity_source",
+            "transducer_name": "sonar_name",
+            "transducer_sound_speed": "sonar_sound_speed",
```
leewujung, line 58: @emiliom : here do you think we should keep the original names `transducer_*` instead of changing to `sonar_*`?  Since it is not entirely clear what this is other than my conjectured use of line of sight time delay to measure sound speed, it seems better to keep its original name. This is not in the convention.

leewujung:
@imranmaj : As we discussed last week and I have specified in https://github.com/OSOceanAcoustics/echopype/issues/540#issuecomment-1099364682, all the environmental variables have a time dimension that come with the Environment datagram -- it is NOT `ping_time` for EK80 because the data are not part of the Raw datagram (which has the acoustic data). Please add a time dimension. 

@emiliom : would `environment_time` be a good name? Or do you think we should use `time1` as in the Platform group?

echopype/convert/set_groups_ek80.py
gavinmacaulay: The EK80 has a `transducer_sound_speed` value separate from water column sound speed because some of their systems potentially do a correction for transducer parameters based on the sound speed at the transducer face (e.g., beamwidth changes with sound speed across the transducer face).

leewujung: Interesting... Thanks @gavinmacaulay for explaining where this is from!
Is this something that's done automatically based on the transducers? or something the users can toggle on/off?

gavinmacaulay: I've never actually seen a Simrad system that uses this, but think that the bathymetric systems may have a use for it.

leewujung:
Thanks @imranmaj ! All the changes look great. I was going to approve it but realized that I forgot to ask for tests... could you please add tests to these also just like in #631? Just so that we have our base covered for the v0.6.0 endeaver.

echopype/convert/set_groups_ek80.py
```diff
@@ -181,6 +242,12 @@ def set_platform(self) -> xr.Dataset:
                         "standard_name": "time",
                     },
                 ),
+                "time3": (
+                    ["time3"],
+                    [self.parser_obj.environment["timestamp"]]
+                    if "timestamp" in self.parser_obj.environment
+                    else np.datetime64("NaT"),
+                ),
```
emiliom, line 250: Add attributes to time3. Use the same set of attributes as with the other time coordinates:
```python
{
    "axis": "T",
    "long_name": "Timestamps for NMEA datagrams",
    "standard_name": "time",
},
```
But change long_name to whatever is appropriate for the source of this time.

echopype/convert/set_groups_ek80.py
```diff
@@ -50,7 +71,26 @@ def set_env(self, env_only=False) -> xr.Dataset:
                         "long_name": "Timestamp of each ping",
                         "standard_name": "time",
                     },
-                )
+                ),
+                "time3": (
+                    ["time3"],
+                    [self.parser_obj.environment["timestamp"]]
+                    if "timestamp" in self.parser_obj.environment
+                    else np.datetime64("NaT"),
```
emiliom, line 79: Add attributes to time3. Use the same set of attributes as with the other time coordinates:
```python
{
    "axis": "T",
    "long_name": "Timestamps for NMEA datagrams",
    "standard_name": "time",
},
```
But change long_name to whatever is appropriate for the source of this time.

emiliom:
> Q1: For the variables that are saved into the `Platform` group, I am thinking we can use `time3` so that the current `location_time` and `mru_time` can become `time1` and `time2` (#518). What do you think?

Yeah, `time3` makes sense, since it's the third time dimension used.

> Q2: For the variables that are saved in the `Environment` group, should we use `environment_time` or `time3`?  The latter would make it consistent across groups.

IMHO it's fine but somewhat weird to use `time3` in the `Environment` group. The 3 only makes sense within the `Platform` context. I think I'd lean towards something more specific like `environment_time` or something closer to the instrument type. But, frankly, it's a toss up! Consistency with `Platform` is an advantage, too.

> Q3: What do you think about adding `comment` or `long_name` on where the timestamps come from?

A `long_name` attribute should ALWAYS be included in general, and ALWAYS^10 for coordinate variables. I've added inline comments to that effect. The same `long_name` string should be used in both `Platform` and `Environment`, if they're really the same time sources/arrays.

I suggest adding a `comment` attribute that looks something like this:
```
"comment": "Platform.time3 and Environment.environment_time are identical time coordinates from the same sensor"
```

leewujung:
Thanks @emiliom !

@imranmaj : please use the following:
- `time3` in the `Platform` group
- `environment_time` in the `Environment` group
- in both groups: attributes for `time3` and `environment_time`
  ```
  {
      "axis": "T",
      "long_name": "Timestamps for Environment XML datagrams",
      "standard_name": "time",
      "comment": "Platform.time3 and Environment.environment_time are identical time coordinates from the same datagrams",
  },

Note that I changed the end of the `comment` attribute from "the same sensors" to "the same datagrams" to be more general.

leewujung (commented):
@imranmaj : two comments:
- Since the blocks you have are in the same file and identical, how about consolidating them to a function?
- You are currently testing for the existence but not the values. Could you add tests for values too?

imranmaj:
> * Since the blocks you have are in the same file and identical, how about consolidating them to a function?

Is this referring to the tests? Is there a way to add functions to test files without running them as tests? I’ve mostly found answers like this: https://stackoverflow.com/questions/53063852/pytest-how-do-you-test-a-function-and-not-run-the-rest-of-the-file

leewujung:
Oh I wasn't actually thinking that much, I though we already did something like this in https://github.com/OSOceanAcoustics/echopype/blob/a9c19f76fa8638c36063dfa2a42809bc1e29d0ae/echopype/tests/preprocess/test_preprocess.py#L201

Issue #520, opened 2021-12-15
Add beam dimension to Beam group


emiliom:
*(Updated 2022-3-30)*. Add a `beam` dimension to `Beam_groupX` groups. We've decided to always include the `beam` dimension, instead of relying on an implicit length-1 dimension.

- [x] Currently for EK60 and AZFP, a `beam` dimension is not used explicitly and is effectively length 1. We'll need to add it.
- [x] #619  For EK80 there is already a dimension called `quadrant` that is conceptually equivalent (for our purposes) to the `beam` dimension as defined in SONAR-netCDF4 v1. We will need to rename it to `beam`.

**Older notes (just for reference):**

- For single beam or split-beam, it could be a length-1 explicit dimension or an implicit coordinate (scalar `beam` )
- For multi-beam, an explicit dimension is required. Currently the EK80 parser generates such a structure, but the dimension is named `quadrant` (WJ: there is an extra dimension quadrant — you can call them separate beams, but since Simrad use quadrant i used that for our first shot though I think it should be probably sector since it is not always 4)
- EM: It occurs to me now that, in terms of match-up with SONAR-netCDF4, the Beam group in AZFP and EK60 echopype nc/zarr files could be described as having an implicit, single-value `beam` dimension. Single-value dimensions can be implemented interchangeably in a netcdf file either explicitly with the dimension or implicitly (refer to the CF convention). In SONAR-netCDF4, since there is an explicit `beam` dimension, single-beam data would be expected to be stored with `beam` as a dimension.
- issue of convention use of `beam` dimension (ie, >1 beams per Beam subgroup? Based on "beam mode" alone?)
- In the convention, under 2.10.6 (Sonar groups), it says: "Variable definitions for data from split-aperture systems are not currently specified"

emiliom:
Note about a related change in the draft SONAR-netCDF4 version 2: Under [7.1. Significant changes from version 1 to version 2](https://htmlpreview.github.io/?https://github.com/ices-publications/SONAR-netCDF4/blob/master/Formatted_docs/crr341.html#_significant_changes_from_version_1_to_version_2):

**Beam:** Added variables for the description and storage of data from split-aperture beams via a `subbeam` dimension added in the Sonar group.

emiliom:
[...]

PR #619, opened 2022-04-09
Rename ek80 `quadrant` dimemsions and coordinates to `beam`

emiliom:
Addresses the first, easy part of #520:
> For EK80 there is already a dimension called `quadrant` that is conceptually equivalent (for our purposes) to the `beam` dimension as defined in SONAR-netCDF4 v1. We will need to rename it to beam.

This PR involved just simple search-and-replace steps.


b-reyes (approved):
@emiliom I don't see any issue with your proposed changes.

emiliom:
Thanks @b-reyes ! I'll merge.

Issue #620, opened 2022-04-09
Change `src_filenames` string attribute to `source_filenames` list-of-strings variable


emiliom:
This kept falling through the cracks. echopype uses `src_filenames` (in the `Provenance` group), and sets it as a string when it's a single file name (ie, from a simple, single raw file conversion). But in the convention its name is `source_filenames` and it's actually defined as a *variable*, not a global attribute. Also, it's a "vector of datafile names that were used ...", not a single string.

In echopype  `src_filenames` is used only in just a couple of places at this time. BTW, one of those places, in `SetGroupsEK60.set_provenance`, seems overly redundant of `SetGroupsBase.set_provenance`

leewujung:
> in the convention its name is source_filenames and it's actually defined as a variable, not a global attribute. Also, it's a "vector of datafile names that were used ...", not a single string.

This is great, thanks for discovering this.
In `combine_echodata` @imranmaj @lsetiawan and I actually arrived at the conclusion to have `src_filenames` as a data variable, and as a vector also. 
https://github.com/OSOceanAcoustics/echopype/blob/185cc1e6f444538a70996578b5006047e2a51565/echopype/echodata/combine.py#L27-L39
so we can now standardize this according to the convention! :D

emiliom:
Closed by #674

Issue #618, opened 2022-04-09
Add echopype software and version provenance attributes to Sv and MVBS datasets


emiliom:
An easy first step would be to adapt the SONAR-netCDF4 v1 `Provenance` attributes, to be inserted into Sv and MVBS datasets as global attributes. Here's a sample. Instead of "conversion", the prefix might be "calibration" or something more generic ("processing").

From the `Provenance` group of a converted file:
```
conversion_software_name : echopype
conversion_software_version : 0.5.4
conversion_time : 2021-10-29T00:20:18Z
src_filenames : s3://ncei-wcsd-archive/data/raw/Bell_M._Shimada/SH1707/EK60/Summer2017-D20170728-T195219.raw
```

Should this go into version 0.6.0?

(Note: see #620 vis a vis `src_filenames`)

Added from #221: `remove_noise` output should also include provenance on which Sv file was used for calculation and the bin size along range and ping sequence.

leewujung:
> Instead of "calibration", the prefix might be "calibration" or something more generic ("processing").

Somehow "calibration_software_name" feels a bit weird, "processing_software_name" seems more natural?

emiliom:
> Somehow "calibration_software_name" feels a bit weird, "processing_software_name" seems more natural?

I don't know about the weirdness (I was going off the fact that echopype calls the creation of Sv a "calibration"). But I like "processing" better b/c it's more reusable and we could use the attribute names for both Sv and MVBS.

leewujung:
Ok, let's go with using "processing_*" for these attributes then. This is also something that higher processing levels can use.

emiliom:
Just realized that this issue is related to a pretty old issue, #221. Maybe we could fold the `remove_noise` aspect of #221 into this one, then close it?

leewujung:
> Maybe we could fold the remove_noise aspect of https://github.com/OSOceanAcoustics/echopype/issues/221 into this one, then close it?

Sounds good!

emiliom:
#621 closes this issue

PR #621, opened 2022-04-10
Generalize echopype software provenance and apply to Sv and MVBS

emiliom:
Addresses #618. Let's keep it as a draft PR for now, for discussion.

I've abstracted the assignment of the current `Provenance` software attributes into a function in the new module `utils.prov.py`. This function now handles both `conversion_*` and `processing_*` attributes in a single place. Also eliminated existing duplication of attribute assignments.

Note the additional (not from the convention) global attribute I've added as a suggestion for Sv and MVBS: `processing_function`. It would allow us to record the actual function used.

Doesn't address the move from `src_filenames` global attribute to a `source_filenames` variable (#620)


echopype/utils/prov.py
```diff
@@ -0,0 +1,31 @@
...
+        Source file path. A list of files is not currently supported
+    """
+    prov_dict = {
+        f"{process_type}_software_name": "echopype",
+        f"{process_type}_software_version": ECHOPYPE_VERSION,
+        f"{process_type}_time": dt.utcnow().isoformat(timespec="seconds") + "Z",  # use UTC time
+    }
+    if source_files:
+        # TODO: src_filenames will be replaced with a new variable, source_filenames
+        #   Also, come to think of it, source files is not "echopype provenance" info per se
```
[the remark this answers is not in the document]
emiliom: What I meant with "source files is not "echopype provenance" info per se"  was that all the other attributes I've included in `echopype_prov_attrs` are about the echopype software itself, but source files is not. Though, come to think of it, strictly speaking, `conversion_time` is not either .... Hmm. We could rename `echopype_prov_attrs` to something more general. Its scope is the software used (name & version) and basic info about its application (timestamp when it was run, and input files used -- though the latter will be moved to a variable and will no longer be an attribute)

emiliom:
> Note the additional (not from the convention) global attribute I've added as a suggestion for Sv and MVBS: `processing_function`. It would allow us to record the actual function used.

@leewujung this is what I was referring to.

emiliom:
TODO: Add software provenance attributes to `remove_noise` output dataset. I'll add a commit to this PR.

emiliom:
TODO: For EK80, add `waveform_mode` and `encode_mode` parameters passed to `compute_Sv` and `compute_TS` (and pass them to MVBS too?). These parameters are not used in any other sensors at this time, so no attributes will be added.

leewujung:
Yes please pass these to MVBS also, as part of the chain of provenance (is there a specific term for this chain?).

emiliom:
@leewujung I've pushed several commits that include multiple improvements, some of which are not about provenance per se.

- Removed the `sv` variable being added in `compute_MVBS_index_binning` by mistake
- `remove_noise` Add provenance global attributes, and add variable attributes to `Sv_corrected` and `Sv_noise`
- Add `actual_range` attribute to `Sv_corrected`, `Sv_noise` and MVBS `Sv`

I decided not to try to implement "provenance chain", where, say, parameters used in computing Sv are passed downstream to MVBS. I think that's a larger topic. Let's revisit it *after* 0.6.0 is out, at some point when we focus on provenance again.

echopype/utils/prov.py
```diff
@@ -0,0 +1,31 @@
+from datetime import datetime as dt
+from typing import Dict
+
+from _echopype_version import version as ECHOPYPE_VERSION
+from typing_extensions import Literal
+
+ProcessType = Literal["conversion", "processing"]
+
+
+def echopype_prov_attrs(process_type: ProcessType, source_files: str = None) -> Dict[str, str]:
```
leewujung, line 10: Looking at this again and seeing how `processing_function` is currently separately done in `calibrate.api._compute_cal`, I wonder if we could rename this to `echopype_ver_time_prov` so that it is clear what this function does, for when we come back to provenance in the future.

echopype/convert/set_groups_ek60.py
```diff
@@ -109,15 +108,8 @@ def __init__(self, *args, **kwargs):
...
-        prov_dict = {
-            "conversion_software_name": "echopype",
-            "conversion_software_version": ECHOPYPE_VERSION,
-            "conversion_time": dt.utcnow().isoformat(timespec="seconds") + "Z",  # use UTC time
-            "src_filenames": self.input_file,
-            "duplicate_ping_times": 1 if self.old_ping_time is not None else 0,
-        }
-        # Save
+        prov_dict = echopype_prov_attrs(process_type="conversion", source_files=self.input_file)
+        prov_dict["duplicate_ping_times"] = 1 if self.old_ping_time is not None else 0
```
leewujung, line 112: Shall we change 1 to True and 0 to False to make it clear what this is for?
This is a temporary change for sure, just thought our intention to indicate that there was a change done here would be more clear:
```suggestion
        prov_dict["duplicate_ping_times"] = True if self.old_ping_time is not None else False
```

echopype/preprocess/noise_est.py
```diff
@@ -113,13 +113,22 @@ def remove_noise(self, noise_max=None, SNR_threshold=3):
...
                 "SNR_threshold": SNR_threshold,
                 "noise_max": noise_max,
             }
-        )
+
+        self.ds_Sv["Sv_noise"] = self.Sv_noise
+        add_attrs("noise", self.ds_Sv["Sv_noise"])
+
+        self.ds_Sv["Sv_corrected"] = Sv_corr
+        add_attrs("corrected", self.ds_Sv["Sv_corrected"])
```
leewujung, line 134: Should we also add like what you have for `compute_MVBS`?
```suggestion
        add_attrs("corrected", self.ds_Sv["Sv_corrected"])
        
        prov_dict = echopype_prov_attrs(process_type="processing", source_files=None)
        prov_dict["processing_function"] = "preprocess.remove_noise"
        ds_Sv = ds_Sv.assign_attrs(prov_dict)
```

echopype/preprocess/api.py
```diff
@@ -136,9 +137,17 @@ def _freq_MVBS(ds, rint, pbin):
...
             "range_meter_interval": str(range_meter_bin) + "m",
             "ping_time_interval": ping_time_bin,
+            "actual_range": [
+                round(float(ds_MVBS["Sv"].min().values), 2),
+                round(float(ds_MVBS["Sv"].max().values), 2),
+            ],
         }
     )
 
+    prov_dict = echopype_prov_attrs(process_type="processing", source_files=None)
```
leewujung, line 147: It seems weird to me that `source_files=None` here. But it is true that the source of the Sv files users open up is not accessible by use. 

I can see that if we implement this as part of a pipeline we can add this info there (the same chain of provenance discussion that is _not now_), but I wonder if there's a way for us to remind users to add this. Perhaps can only make it as explicit as possible in the docs?

echopype/preprocess/noise_est.py
leewujung: I have the same `source_files=None` thoughts as in the `ds_MVBS` section.

leewujung (approved):
I pushed a change to remove the unused datetime import since GH doesn't allow me to comment on unchanged code.

I added a few suggestions, but otherwise this looks ready to go!

FYI- I pondered a bit about the `long_name` you added for `Sv_noise` and `Sv_corrected`, but decided that what you have is appropriate.

echopype/utils/prov.py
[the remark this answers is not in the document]
emiliom: Sure, sounds good. It sounds like you're also saying that we keep the source_files provenance functionality in this function, even though it'll be changed to add xarray variables rather than an attribute. Correct?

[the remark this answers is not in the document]
leewujung: > It sounds like you're also saying that we keep the source_files provenance functionality in this function, even though it'll be changed to add xarray variables rather than an attribute. Correct?

I didn't think that far... ha. I think you are right that since it will be a data variable not an attribute, it should not be here. In that case, the function name would become `echopype_ver_time_prov`.

echopype/convert/set_groups_ek60.py
leewujung: geez, I didn't think about this, and it does not work!
https://github.com/pydata/xarray/issues/1500
I tested it out and it spilled the same error as stated in this open issue. Let's discard my suggestion!

echopype/preprocess/noise_est.py
emiliom: > Should we also add like what you have for compute_MVBS?

Heh. That provenance info is already being added, in [`preprocess/api.py`, line 283](
https://github.com/OSOceanAcoustics/echopype/pull/621/files/c1ebfb36478b5c7a3f52a1f817a00abd9d23a1af#diff-2848209b968526da06d5475e3df089d5a7138b944681119fbe9f64f9aad81603R283). ie, downstream. This was trickier than for `compute_MVBS`, b/c there's an external function (`preprocess/noise_est.py`) doing the heavy lifting.

leewujung: Ah! I missed that. Thanks!

echopype/preprocess/api.py
emiliom: > It seems weird to me that source_files=None here. But it is true that the source of the Sv files users open up is not accessible by use.

Exactly. See the contortion I implemented for this provenance info for calibration functions in `calibrate/api.py`:
```python
    add_attrs(cal_type, cal_ds)
    # Provenance source files may originate from raw files (echodata.source_files)
    # or converted files (echodata.converted_raw_path)
    if echodata.source_file is not None:
        source_file = echodata.source_file
    elif echodata.converted_raw_path is not None:
        source_file = echodata.converted_raw_path
    else:
        source_file = "SOURCE FILE NOT IDENTIFIED"
    prov_dict = echopype_prov_attrs(process_type="processing", source_files=source_file)
```

For downstream (pre)processing functions, it becomes harder to track unambiguously.

> I can see that if we implement this as part of a pipeline we can add this info there (the same chain of provenance discussion that is not now), but I wonder if there's a way for us to remind users to add this. Perhaps can only make it as explicit as possible in the docs?

Reminding users of adding desirable but optional attributes is tricky business. I think that should all be part of post-0.6.0 provenance discussions. Not now.

echopype/preprocess/noise_est.py
emiliom: > I have the same source_files=None thoughts as in the ds_MVBS section.

See my reply there

echopype/preprocess/api.py
leewujung: Yes, definitely. Let's merge this and worry about that later.

emiliom:
> I pushed a change to remove the unused datetime import since GH doesn't allow me to comment on unchanged code.

Thanks!

> FYI- I pondered a bit about the long_name you added for Sv_noise and Sv_corrected, but decided that what you have is appropriate.

Great. And yeah, I struggled a bit with that.

> I added a few suggestions, but otherwise this looks ready to go!

Great! But since I've made some follow-up comments to your inline comments, I'll give you more time to react. Let's plan to merge today, if possible.

leewujung:
I think reacted to everything, haha. I can't wait to see this merged!! 😎

emiliom:
Thank you!! Merging now

Issue #592, opened 2022-03-18
Add missing required variables in `Platform` groups


leewujung:
This was discovered in #516 discussion and is better captured in its own issue:

@emiliom:
> Hmm. After reading and rereading the definitions of `water_level` and `vertical_offset`, plus the Coordinate Systems chapter in SONAR-netCDF v1 (ch. 4), here are some observations -- some of which I'm sure are obvious, but it helps me to state them:
> 1. The platform coordinate system is the grounding for all other coordinates. The origin of that coordinate system is arbitrary.
> 2. There are 3 types of x-y-z offsets defined: position (GPS), MRU and transducer.
> 3. `water_level` and `vertical_offset` are variables that refer exclusively to the vertical (z) axis and are there to enable the generation of depth values relative to actual water level
> 
> The generalization to `range_offset` implies an along-range ("transducer orientation", in my layman's term) axis relative to the platform origin, or relative to actual or nominal water level (I can't remember which). But regardless of the orientation of the transducer, its offsets x-y-z relative to platform origin are already specified by the transducer offsets, and `water_level` and `vertical_offset` together provide that vertical referencing to either a constant water level (I assume that's what nominal water level is?) or actual, time varying water level. It's not clear to me what value the generalized, along-range `range_offset` actually provides when all the other offsets as well as `water_level` and `vertical_offset` are available. I *think* what's potentially missing is more clarity and specificity about the coordinate system for platforms other than ships. This is apparently being addressed in the convention v2 draft (see #517).

## Action items
In view of the above above and in the mind set of convention v1:

- [ ] review the convention and echopype's handling of various position and offset variables (water_level, vertical_offset, transducer_offset_x/y/z, MRU_offset_x/y/z
- [ ] ensure compliance of required variables in v0.6.0

emiliom:
Thanks for this assessment @imranmaj!

> MRU_offset_x/y/z: could not find in the datagram

Could you add which EK60 raw file(s) you tested to arrive at this finding? If you know where it came from (say, Hake survey), that'd be great too.

imranmaj:
I looked through our EK60 parser implementation and the [EK60 manual](https://www.kongsberg.com/globalassets/maritime/simrad/echo-sounders/ek60/operation/164692ad_ek60_reference_manual_english_lores.pdf), in particular through the raw file specification (beginning page 194) and the NMEA specification (beginning page 177).

leewujung:
[...]

PR #623, opened 2022-04-12
Rename `heave` to `vertical_offset`

imranmaj:
Part of #592 

Renames `heave` to `vertical_offset` in accordance with the [SONAR-netCDF4 convention 1.0](https://www.ices.dk/sites/pub/Publication%20Reports/Cooperative%20Research%20Report%20(CRR)/CRR341.pdf):
image

Note that only the variable in the converted dataset will be renamed; the parsers will still use the name `heave` to match the raw data specification.

### DIFF
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -153,10 +153,7 @@ class SetGroupsEK80(SetGroupsBase):
                 "vertical_offset": (
                     ["mru_time"],
                     np.array(self.parser_obj.mru.get("heave", [np.nan])),
-                    {
-                        "long_name": "Platform vertical offset from nominal",
-                        "units": "m",
-                    },
+                    self._varattrs["platform_var_default"]["vertical_offset"],
                 ),
                 "latitude": (
                     ["location_time"],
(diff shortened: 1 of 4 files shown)


echopype/convert/set_groups_ek80.py
```diff
@@ -150,14 +150,12 @@ def set_platform(self) -> xr.Dataset:
...
+                "vertical_offset": (
                     ["mru_time"],
                     np.array(self.parser_obj.mru.get("heave", [np.nan])),
                     {
-                        "long_name": "Platform heave",
-                        "standard_name": "platform_heave_angle",
-                        "units": "arc_degree",
-                        "valid_range": (-90.0, 90.0),
+                        "long_name": "Platform vertical offset from nominal",
+                        "units": "m",
```
leewujung, line 158: For this part why wouldn't the same `self._varattrs` approach like in EK60 work?

emiliom: Agreed

emiliom:
This PR looks good, other than @leewujung's comment/suggestion. Thanks.

emiliom (approved):
Looks good to go. Thanks!

PR #647, opened 2022-04-25
Rename Platform location_time and mru_time

emiliom:
Implements #518

### DIFF
--- a/echopype/utils/coding.py
+++ b/echopype/utils/coding.py
@@ -24,7 +24,7 @@ DEFAULT_ENCODINGS = {
     "ping_time_echosounder_raw": DEFAULT_TIME_ENCODING,
     "ping_time_echosounder_raw_transmit": DEFAULT_TIME_ENCODING,
     "location_time": DEFAULT_TIME_ENCODING,
-    "mru_time": DEFAULT_TIME_ENCODING,
+    "time2": DEFAULT_TIME_ENCODING,
 }
 
 
@@ -56,7 +56,7 @@ def set_encodings(ds: xr.Dataset) -> xr.Dataset:
     for var, encoding in DEFAULT_ENCODINGS.items():
         if var in new_ds:
             da = new_ds[var].copy()
-            if "_time" in var:
+            if "_time" in var or ("time" in var and len(var) == 5):
                 new_ds[var] = xr.apply_ufunc(
                     _encode_dataarray,
                     da,
--- a/echopype/utils/coding.py
+++ b/echopype/utils/coding.py
@@ -23,7 +23,7 @@ DEFAULT_ENCODINGS = {
     "ping_time_echosounder": DEFAULT_TIME_ENCODING,
     "ping_time_echosounder_raw": DEFAULT_TIME_ENCODING,
     "ping_time_echosounder_raw_transmit": DEFAULT_TIME_ENCODING,
-    "location_time": DEFAULT_TIME_ENCODING,
+    "time1": DEFAULT_TIME_ENCODING,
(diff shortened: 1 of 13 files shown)


emiliom:
I just realized I didn't change the documentation (`.rst`) files. I'll do that next.

echopype/utils/coding.py
```diff
@@ -56,7 +56,7 @@ def set_encodings(ds: xr.Dataset) -> xr.Dataset:
     for var, encoding in DEFAULT_ENCODINGS.items():
         if var in new_ds:
             da = new_ds[var].copy()
-            if "_time" in var:
+            if "_time" in var or ("time" in var and len(var) == 5):
```
leewujung, line 59: what is this `len(var) == 5` thing for here?

emiliom: The second test condition (in parenthesis) is testing for names like `time1`, `time2` ... `time9`. Actually, I'm wondering if we can just change the whole if-statement to `if "time" in var:`. I don't remember if we were testing for "_time" rather than "time" for a specific reason. I was being cautious in this PR.

leewujung: > Agreed about the intention being more explicit.

It is also that testing a length of 5 would now allow more than one digit.

> It would work. It's testing for the presence of the "time" string in var, not an exact match. 

Oh right it is "in" rather than "is" in the if statement. 

I think the intention was trying to test everywhere time encodings occur. With that in mind using `time` would work. But, I think for testing it'd be better to be more specific by allowing `*_time` and `time*` patterns.

emiliom: > I think the intention was trying to test everywhere time encodings occur. With that in mind using time would work. But, I think for testing it'd be better to be more specific by allowing *_time and time* patterns.

I see two options. 

Given that we intend to capture all variables that include the "time" substring (right?), my preference is the simple `in` test plus a comment. The code is super clean, and the comment conveys the intent.

```python
# Match all variable names containing a "time" substring, including *_time and time*
if "time" in var:
```

Here's one option using a regex. The complexity of the regex depends on how strict you want to be. While the regex makes the intent precisely coded, I think it makes it less immediately readable. Plus using regex adds a slight performance hit (probably trivial, though):

```python
from re import search

# Match all variable names matching the patterns *_time or time*
if bool(search(r".+_time$|^time[\d]+", var)):

# an alternative that might be more readable:
if bool(search(r".+_time$", var)) or if bool(search(r"^time[\d]+", var)):
```

I think the crux is whether we deliberately want to limit the test to `*_time` and `time*`

emiliom:
> I just realized I didn't change the documentation (`.rst`) files. I'll do that next.

`location_time` and `mru_time` are not currently found in any of the `.rst` files (except for release notes in "What's new", which doesn't count). So, we're good!

echopype/utils/coding.py
leewujung: Thanks @emiliom for putting these together. My vote is to limit the tests to `*_time` and `time*`, to avoid accidental error, if there happen to be some variables that has "time" in it but is actually not a _time_ per se.

Also, I just realized that somehow I thought this was in tests. These are not, so all the more important to be accidental-error proof!

emiliom: I've pushed a solution. FYI, AD2CP uses some time variable that must be of the form `*_time*`, so my final regex looks like this:
```python
bool(search(r"_time|^time[\d]+$", var))
```
The first part is basically the same as what we had previously (equivalent to `"_time" in var`)

leewujung:
Thanks @emiliom for this and putting in the patterns for time variable checking. I'll merge this now.

PR #651, opened 2022-04-26
Update setup-services.py to include images & volumes subtleties

lsetiawan:
This PR makes `--tear-down` to also clean up images and volumes.

### DIFF
--- a/.ci_helpers/docker/setup-services.py
+++ b/.ci_helpers/docker/setup-services.py
@@ -58,6 +58,8 @@ def run_commands(commands: List[Dict]) -> None:
             subprocess.run(cmd)
         elif callable(cmd):
             cmd(args)
+        else:
+            raise ValueError(f"command of {type(cmd)} is invalid.")
 
 
 if __name__ == "__main__":
--- a/.ci_helpers/docker/setup-services.py
+++ b/.ci_helpers/docker/setup-services.py
@@ -4,10 +4,19 @@ Script to help bring up docker services for testing.
 
 """
 import argparse
-import os
+import logging
 import shutil
+import subprocess
 import sys
 from pathlib import Path
+from typing import Dict, List
+
+logger = logging.getLogger("setup-services")
+streamHandler = logging.StreamHandler(sys.stdout)
+formatter = logging.Formatter("%(message)s")
+streamHandler.setFormatter(formatter)
+logger.addHandler(streamHandler)
+logger.setLevel(level=logging.INFO)
 
 HERE = Path(".").absolute()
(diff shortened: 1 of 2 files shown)


emiliom:
Make this optional? Often I tear-down to not have active containers, but I'd like to have the ability to restart from the images w/o downloading everything (deploy) again.

lsetiawan:
Just to add more details on this PR:
1. The issue of volumes blowing was because every time someone runs `--deploy` docker-compose will recreate the container, and specifically, the `minio` container will create *NEW* volume reference with a random id, which allocates space within the host machine. Now in the `docker-compose.yaml`. I've added a specific volume for the minio service called `echopype_miniodata`. This would then only create this volume one time, and it will not be recreated again, unless it gets deleted from running `--tear-down`.
2. Now `--deploy` can be paired with `--no-pull`, which skips pulling the latest images for both minio and http, by *default it will pull*.
3. Last, there's now `--images` flag that can be paired during `--tear-down` to also delete the docker images during tear down. By default, the images will be kept around. However, now the volumes for the docker containers will be removed during tear down.

emiliom:
> 2. Now `--deploy` can be paired with `--no-pull`, which skips pulling the latest images for both minio and http, by *default it will pull*.

Perfect

> 3. Last, there's now `--images` flag that can be paired during `--tear-down` to also delete the docker images during tear down. By default, the images will be kept around. However, now the volumes for the docker containers will be removed during tear down.

Perfect

.ci_helpers/docker/setup-services.py
```diff
@@ -18,17 +27,44 @@
 def parse_args():
     parser = argparse.ArgumentParser(description="Setup services for testing")
     parser.add_argument("--deploy", action="store_true", help="Flag to setup docker services")
+    parser.add_argument(
+        "--no-pull",
+        action="store_true",
+        help="Optional flag to pull the latest images",
```
emiliom, line 33: I think the help message should actually be something like this: "Optional flag to skip pulling the latest images from dockerhub"

lsetiawan: Oops that was from my initial iteration. Thanks for catching that!

lsetiawan, line 33:
```suggestion
        help="Optional flag to skip pulling the latest images from dockerhub",
```

emiliom: I'm still new to PR "suggestions"! Thanks for doing that.

emiliom (approved):

Issue #569, opened 2022-02-25
 echopype.visualize.create_echogram got AttributeError: module 'echopype' has no attribute 'visualize' for 0.5.6


cywhale:
The same .raw, when using echopype 0.5.6,  ep.visualize.create_echogram got Attribute Error.
But got correct result by using 0.5.5

```
ed = ep.open_raw(
    raw_file="xxx.raw", 
    sonar_model="EK80", 
)

fg = ep.visualize.create_echogram(ed, frequency=200000, 
        get_range=True, robust=True, 
        range_kwargs={"encode_mode":"complex", "waveform_mode":"CW"})
```

Result figs:
- version 0.5.5

https://github.com/cywhale/ODB/blob/master/EK/create_echogram_0_5_5.png

- version 0.5.6

https://github.com/cywhale/ODB/blob/master/EK/create_echogram_0_5_6.png

-- Error message

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/home/cmd/ek/test_openraw.ipynb Cell 7' in <module>
----> 1 fg = ep.visualize.create_echogram(ed, frequency=200000, 
      2         get_range=True, robust=True, 
      3         range_kwargs={"encode_mode":"complex", "waveform_mode":"CW"},figsize=(16,48),ylim=[150,0])

AttributeError: module 'echopype' has no attribute 'visualize'
```

I used previous testing raw file ( https://github.com/cywhale/ODB/blob/master/EK/test_v3.raw ) in ver 0.5.6 got the same Attribute Error, but cannot get correct plot when using version 0.5.5. In aboving testing, the raw file is just too big. But seems that this error always occurs on any other raw files when using version 0.5.6

leewujung:
I think this is related to our decision in v0.5.6 to slim down the default requirements of echopype so that now you actually have to do `import echopype.visualize` in order to use the viz function (see [here](https://github.com/OSOceanAcoustics/echopype/blob/17a86e4c87f28d1457c582062e3ac59367b37348/echopype/tests/visualize/test_plot.py#L1-L2)).
Could you give it a try?

@lsetiawan @emiliom : ugh, a breaking change we didn't intend... let's make sure to add this to the documentation to avoid this issue!

cywhale:
@leewujung Yes. But it raises a Value Error if using get_range=True:

```
ValueError                                Traceback (most recent call last)
/home/cmd/ek/test_openraw.ipynb Cell 7' in <module>
[...]

PR #653, opened 2022-04-27
Fixing nan filter for echodata and Sv

lsetiawan:
This PR fixes the nan filtering issues that causes plotting to break in #569. Additionally, single ping time wasn't caught when the input data is echodata object; this also address that. Plotting appears to work now with some of the given datasets. Seems like some of the dataset contain errors/warning when converting with echopype.

### Dataset details

- **https://github.com/cywhale/ODB/blob/master/EK/test_v3.raw**:  echopype `open_raw` converted the file in about ~323ms. Plotting attempt results in ValueError since there's only a single ping time. With this PR this error is now caught and raised.
  ```
  ValueError: Ping time must be greater or equal to 2 data points.
  ```

- **from_taiwan_NOR1/test_v2_D20211013-T014141.raw**: WARNING FOUND and TERMINATED ... echopype `open_raw` spits out a warning, and after a bit blew up my memory and jupyter lab terminated the process, so I can't plot this dataset or even convert it.
   ```
  15:19:11  parsing file test_v2_D20211013-T014141.raw, time of first ping: 2021-Oct-13 01:41:41
  WARNING: The water_level_draft was not in the file. Value set to NaN.
   ```

- **from_taiwan_NOR1/test_v3_D20200611-T224608.raw**: WARNING FOUND and HANGS ... echopype `open_raw` doesn't seem to be working.. it just hangs for a long time until I stop the process... in which case, I can't plot anything since I can't get the data converted.
  ```
  Datagram failed size check:  65424 != -3473309 @ (142974980, 14971)
  Skipping to next datagram...
  Attempting to find next valid datagram...
  ```
- **from_taiwan_NOR1/test_v3_D20200719-T052111.raw**: ERROR FOUND... echopype `open_raw` fails in ~599ms after execution with a `DatagramReadError`. I can't plot this dataset at all
  ```
  Datagram 0 (@0) shorter than expected length:  537277772 < 941471481
  Attempting to find next valid datagram...
  
  ...
  DatagramReadError: Short read while getting dgram size 4 != 0 @ (537277776L, 0)
  ```
- **from_taiwan_NOR1/test_v3_D20211108-T033024.raw**: echopype `open_raw` converted the file in about ~12.38s. Plotting with the commands below works with this PR.
  ```
  from echopype import visualize as epviz
  range_kwargs = {"encode_mode":"complex", "waveform_mode":"CW"}
  plots = epviz.create_echogram(
      ed, robust=True, get_range=True, range_kwargs=range_kwargs
  )
  ```
Screenshot from 2022-04-27 15-17-46


echopype/visualize/api.py
```diff
@@ -171,6 +173,11 @@ def create_echogram(
     )
 
 
+def _check_ping_time(ping_time):
+    if ping_time.shape[0] < 2:
+        raise ValueError("Ping time must be greater or equal to 2 data points.")
```
leewujung, line 178: How about the following?
```suggestion
        raise ValueError("Ping time must have a length that is greater or equal to 2")
```

leewujung:
@lsetiawan : thanks for investigating this and fixing the bug!

Geez, these are really big files. I tested the last one and it plotted without problem. For the others I ran into the same problems with `open_raw`. It seems that the files are either corrupted themselves at some point, or have heavy memory load as we've found that echopype couldn't handle well at the moment.

I suggest that we merge this fix for plotting and look into what the parsing problems are for these other files separately. 

@cywhale: If you have a channel to talk with the people who collect these data, please let them know that limiting the max file size for these instruments would be a good idea, and that it does not interrupt continuous data collection.

leewujung (approved):

Issue #649, opened 2022-04-25
Inconsistency in Platform variable dimensions amongst the sensors


b-reyes:
In PR #631 variables were added to the Platform group and if they did not exist, then they were set to NaN.

I believe that a couple of items need to be discussed further. 

* The variables `transducer_offset_x/y/z`, `MRU_offset_x/y/z`, `MRU_rotation_x/y/z`, `position_offset_x/y/z`, and `water_level` were added to EK60 and EK80. Should we also add these variables to AZFP?
* In EK60, the variables `transducer_offset_x/y/z`, `MRU_offset_x/y/z`, `MRU_rotation_x/y/z`, and `position_offset_x/y/z` have the dimension `(frequency)`. However, in EK80 `MRU_offset_x/y/z`, `MRU_rotation_x/y/z`, and `position_offset_x/y/z` do not have a dimension associated with them. For consistency sake, should we make them have the same dimension? Similarly, should we do the same thing for AZFP? 
* In EK60 `water_level` has the dimensions `(frequency, ping_time)`. Again, for consistency, should we let EK80 and AZFP have these dimensions too? 
* The dimensions are not consistent from EK60 to EK80 for the variables `pitch`, `roll`, and `vertical_offset`. In EK60 they have dimensions `(frequency, ping_time)` and in EK80 they have `(mru_time)`. This is partially related to #647.

@leewujung @emiliom @imranmaj what do you think?

emiliom:
> * The variables `transducer_offset_x/y/z`, `MRU_offset_x/y/z`, `MRU_rotation_x/y/z`, `position_offset_x/y/z`, and `water_level` were added to EK60 and EK80. Should we also add these variables to AZFP?

I would think so.

> * In EK60, the variables `transducer_offset_x/y/z`, `MRU_offset_x/y/z`, `MRU_rotation_x/y/z`, and `position_offset_x/y/z` have the dimension `(frequency)`. However, in EK80 `MRU_offset_x/y/z`, `MRU_rotation_x/y/z`, and `position_offset_x/y/z` do not have a dimension associated with them. For consistency sake, should we make them have the same dimension? Similarly, should we do the same thing for AZFP? 

Per the convention, those variables are scalar and should not have any dimensions. So, I would think that they should be scalar for EK60 and AZFP, too.

> * In EK60 `water_level` has the dimensions `(frequency, ping_time)`. Again, for consistency, should we let EK80 and AZFP have these dimensions too? 

[...]

PR #654, opened 2022-04-28
Add missing convention-required Platform variables to AZFP

imranmaj:
- Adds missing convention-required Platform variables to AZFP
- Adds tests for those variables

Part of #649


leewujung:
LGTM! Thanks!

PR #655, opened 2022-04-28
Add visualize doc and ignore ipynb for pre-commit

lsetiawan:
This PR adds a small documentation about visualize subpackage.


emiliom:
> Can someone help me in adding the create_echogram function into the api reference dropdown list in the sidebar?? I'm not sure how to do that

I'll do this. But not right now :cry:

emiliom:
The new "Data visualization" page has instructions for including plotting dependencies during installation with pip install: `pip install echopype[plot]`. How about with conda? In the [Installation and Examples page](https://echopype.readthedocs.io/en/latest/installation.html), we have instructions for installation with both pip and conda, so we should have conda instructions here, too.

lsetiawan:
For conda, everything is included, so it already comes with the dependencies for plotting also.

lsetiawan:
> This is a Jupyter Book problem. In the original RTD Sphinx theme, the section navigation on the left was doing exactly what I expect. See this [example from v0.5.4](https://echopype.readthedocs.io/en/v0.5.4/api.html). This will require research.

Thank you so much @emiliom for looking into this in depth! Hmm... I was hoping it would be a simple change! I'll look into further. Appreciate your help 😄

emiliom:
>> This is a Jupyter Book problem. In the original RTD Sphinx theme, the section navigation on the left was doing exactly what I expect. See this [example from v0.5.4](https://echopype.readthedocs.io/en/v0.5.4/api.html). This will require research.
>
> Thank you so much @emiliom for looking into this in depth! Hmm... I was hoping it would be a simple change! I'll look into further. Appreciate your help smile

As the next step, *ideally* I would suggest trying to "turn off" the drop down men on the left sidebar and enabling 3rd-level section headers on the table of content on the right. In practice, it'll depend on how hard each of those tasks is!

emiliom:
See #686. I give up trying to tweak the behavior of the table of contents. I'll merge this (the new visualization page), and we can revisit the table of content issue in the future.

PR #658, opened 2022-05-01
Overhaul and fix Sonar beam_group variables

emiliom:
Working on something else, I realized there were some problems with my initial implementation of the `Sonar` `beam_group_name` and `beam_group_descr` variables and associated `SetGroups<SENSOR>._beamgroups` property. For reference, this implementation was part of PR #574 (see especially https://github.com/OSOceanAcoustics/echopype/pull/574#issuecomment-1062552195).

### Problems

1. While the `beam_group_` variables had a `beam_group` dimension, there was no corresponding coordinate variable.
2. For AD2CP, SetGroupsAd2cp._beamgroups` was present (added in PR #617), but the `beam_group_` variables were not being created.
3. For EK80, it assumed that the two Beam groups were always present. The `beam_group_` variables were being added as length 2, even when only one Beam group (`Beam_group1`) was actually created.

### In this PR

All 3 problems are addressed in this PR. For problem 1, I renamed `beam_group_name` to `beam_group` and made it the previously missing coordinate variable.

Problem 3 (EK80 variable-number-of-beam-groups issue) was more complicated. Previously I set the `SetGroupsEK80._beamgroups` property in the `__init__`, and it was fixed. I've moved the dict of the 2 possible beam groups to a new `SetGroupsEK80` *class* property, `beamgroups_possible`. The decision of whether there is 1 or 2 beam groups present apparently can only be made after `set_beam` is executed. Therefore, I modified `convert/api.py` so that `set_sonar` is executed after `set_beam`, and for EK80 added a new `beam_group_count` argument to `set_sonar`.

Everything seems to work as expected now.

**TODO:** Add a test.

@leewujung : I added the `set_sonar` `beam_group_count` argument  to EK80 only. That keeps the other `SetGroups<SENSOR>` classes a bit cleaner, but with the downside that the `set_sonar` methods are no longer exactly identical across `SetGroups<SENSOR>` classes. Thoughts?

@lsetiawan : When flipping the order of execution in `convert/api.py` so that `set_sonar` happens after `set_beam`, a datatree error occurred. See below. I don't fully understand how it's happening, given that `tree_dict` assignments are just dictionary assignments. But I found that I could eliminate the error by preassigning `tree_dict["Sonar"] = None` before `tree_dict[f"Sonar/Beam_group{idx}"]` is assigned. Do you think this could be simplified a bit?

#### Error in `convert/api.py` when `tree_dict["Sonar"] = None` is not preassigned before `tree_dict[f"Sonar/Beam_group{idx}"]`:

```python
--> 477 tree = DataTree.from_dict(tree_dict)
    478 echodata = EchoData(source_file=file_chk, xml_path=xml_chk, sonar_model=sonar_model)
    479 echodata._set_tree(tree)

File ~/miniconda3/envs/echopype_dev/lib/python3.9/site-packages/datatree/datatree.py:179, in DataTree.from_dict(cls, data_objects, name)
    177         # Create and set new node
    178         new_node = cls(name=node_name, data=data)
--> 179         obj.set_node(
    180             relative_path,
    181             new_node,
    182             allow_overwrite=False,
    183             new_nodes_along_path=True,
    184         )
    185 return obj

File ~/miniconda3/envs/echopype_dev/lib/python3.9/site-packages/datatree/treenode.py:194, in TreeNode.set_node(self, path, node, new_nodes_along_path, allow_overwrite)
    191         del child
    192     else:
    193         # TODO should this be before we walk to the new node?
--> 194         raise KeyError(
    195             f"Cannot set item at {path} whilst that path already points to a "
    196             f"{type(parent.get_node(node_name))} object"
    197         )
    199 # Place new child node at this location
    200 parent.add_child(node)

KeyError: "Cannot set item at / whilst that path already points to a <class 'datatree.datatree.DataTree'> object"

### DIFF
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -195,22 +195,24 @@ class SetGroupsBase(abc.ABC):
         return time1, msg_type, lat, lon
 
     def _beam_groups_vars(self):
-        """Stage beam_group_name and beam_group_descr variables sharing a common dimension,
-        beam_group, to be inserted in the Sonar group"""
+        """Stage beam_group coordinate and beam_group_descr variables sharing
+        a common dimension, beam_group, to be inserted in the Sonar group"""
         beam_groups_vars = {
-            "beam_group_name": (
-                ["beam_group"],
-                [di["name"] for di in self._beamgroups],
-                {"long_name": "Beam group name"},
-            ),
             "beam_group_descr": (
                 ["beam_group"],
                 [di["descr"] for di in self._beamgroups],
                 {"long_name": "Beam group description"},
             ),
         }
+        beam_groups_coord = {
+            "beam_group": (
+                ["beam_group"],
+                [di["name"] for di in self._beamgroups],
+                {"long_name": "Beam group name"},
+            ),
+        }
 
-        return beam_groups_vars
+        return beam_groups_vars, beam_groups_coord
 
     def _add_beam_dim(
         self, ds: xr.Dataset, beam_only_names: Set[str], beam_ping_time_names: Set[str]
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -196,7 +196,8 @@ class SetGroupsEK80(SetGroupsBase):
                     ["channel"],
                     list(self.parser_obj.config_datagram["configuration"].keys()),
                     self._varattrs["beam_coord_default"]["channel"],
-                )
+                ),
+                "beam_group": beam_groups_coord["beam_group"],
             },
             attrs={"sonar_manufacturer": "Simrad", "sonar_type": "echosounder"},
         )
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -43,22 +43,22 @@ class SetGroupsEK80(SetGroupsBase):
         "beamwidth_twoway_athwartship",
     }
 
+    beamgroups_possible = [
+        {
+            "name": "Beam_group1",
+            "descr": "contains complex backscatter data and other beam or channel-specific data.",  # noqa
+        },
+        {
+            "name": "Beam_group2",
+            "descr": (
+                "contains backscatter power (uncalibrated) and other beam or channel-specific data,"  # noqa
+                " including split-beam angle data when they exist."
+            ),
+        }
+    ]
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-
-        self._beamgroups = [
-            {
-                "name": "Beam_group1",
-                "descr": "contains complex backscatter data and other beam or channel-specific data.",  # noqa
-            },
-            {
-                "name": "Beam_group2",
-                "descr": (
-                    "contains backscatter power (uncalibrated) and other beam or channel-specific data,"  # noqa
-                    " including split-beam angle data when they exist."
-                ),
-            },
-        ]
 
     def set_env(self, env_only=False) -> xr.Dataset:
         """Set the Environment group."""
@@ -150,7 +150,7 @@ class SetGroupsEK80(SetGroupsBase):
         )
         return set_encodings(ds)
 
-    def set_sonar(self) -> xr.Dataset:
+    def set_sonar(self, beam_group_count=1) -> xr.Dataset:
         # Collect unique variables
         params = [
             "transducer_frequency",
@@ -168,6 +168,7 @@ class SetGroupsEK80(SetGroupsBase):
         # Create dataset
         # beam_group_name and beam_group_descr variables sharing a common dimension (beam),
         # using the information from self._beamgroups
+        self._beamgroups = self.beamgroups_possible[:beam_group_count]
         beam_groups_vars = self._beam_groups_vars()
         sonar_vars = {
             "serial_number": (["frequency"], var["serial_number"]),
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -49,7 +49,7 @@ class SetGroupsEK80(SetGroupsBase):
             "descr": (
                 "contains backscatter data (either complex samples or uncalibrated power samples)"  # noqa
                 " and other beam or channel-specific data"
-            )
+            ),
         },
         {
             "name": "Beam_group2",
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -46,7 +46,10 @@ class SetGroupsEK80(SetGroupsBase):
     beamgroups_possible = [
         {
             "name": "Beam_group1",
-            "descr": "contains complex backscatter data and other beam or channel-specific data.",  # noqa
+            "descr": (
+                "contains backscatter data (either complex samples or uncalibrated power samples)"  # noqa
+                " and other beam or channel-specific data"
+            )
         },
         {
             "name": "Beam_group2",
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -197,7 +197,7 @@ class SetGroupsEK80(SetGroupsBase):
                     list(self.parser_obj.config_datagram["configuration"].keys()),
                     self._varattrs["beam_coord_default"]["channel"],
                 ),
-                "beam_group": beam_groups_coord["beam_group"],
+                **beam_groups_coord,
             },
             attrs={"sonar_manufacturer": "Simrad", "sonar_type": "echosounder"},
         )
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -166,10 +166,11 @@ class SetGroupsEK80(SetGroupsBase):
                 var[param].append(data[param])
 
         # Create dataset
-        # beam_group_name and beam_group_descr variables sharing a common dimension (beam),
-        # using the information from self._beamgroups
+        # Add beam_group and beam_group_descr variables sharing a common dimension
+        # (beam_group), using the information from self._beamgroups
         self._beamgroups = self.beamgroups_possible[:beam_group_count]
-        beam_groups_vars = self._beam_groups_vars()
+        beam_groups_vars, beam_groups_coord = self._beam_groups_vars()
+
         sonar_vars = {
             "serial_number": (["frequency"], var["serial_number"]),
             "sonar_model": (["frequency"], var["transducer_name"]),
@@ -185,7 +186,10 @@ class SetGroupsEK80(SetGroupsBase):
         }
         ds = xr.Dataset(
             {**sonar_vars, **beam_groups_vars},
-            coords={"frequency": var["transducer_frequency"]},
+            coords={
+                "frequency": (var["transducer_frequency"]),
+                **beam_groups_coord
+            },
             attrs={
                 "sonar_manufacturer": "Simrad",
                 "sonar_type": "echosounder",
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by lsetiawan** [comment]
> When flipping the order of execution in convert/api.py so that set_sonar happens after set_beam

I don't completely understand why this change has to happen? Isn't it more intuitive to set_sonar first then set_beam? I think datatree is erroring out because it's trying to create a sub-node before the upper-node. In this case `Sonar`. Since `Sonar` is a parent of `Beam_x`. This needs to exist first. It's not just a simple dictionary assignment. Datatree creates nodes on the fly, so your solution is the most simple that I see.

**Comment by emiliom** [comment]
> I don't completely understand why this change has to happen? Isn't it more intuitive to set_sonar first then set_beam?

I agree that it's more intuitive. But for EK80, it turns out it's not possible. In `convert/api.py`, you don't know whether there'll be 1 or 2 beam groups until `set_beam` is executed. `set_sonar` needs to know how many beam groups there are.

> Since Sonar is a parent of Beam_x. This needs to exist first. It's not just a simple dictionary assignment. Datatree creates nodes on the fly, so your solution is the most simple that I see.

`tree_dict` is initially defined as a dict, though. As far as I can see, it remains a dict until all the group `set_*` methods have been executed. So, that's why I don't understand the error. Anyway, like I said, my solution does work, though it seems inelegant.

**Comment by leewujung** [comment]
Related to #625.

**Comment by leewujung** [comment]
repr: https://github.com/OSOceanAcoustics/echopype/blob/7231b59111966de6b3595f0f07597262dc72a87f/echopype/echodata/widgets/utils.py#L26

**Comment by emiliom** [comment]
We'll go ahead and add the class property `beamgroups_possible` to all sensors `SetGroups<SENSOR>` classes, not just EK80. I'll make the change to the PR ASAP

**Comment by emiliom** [comment]
> add the class property `beamgroups_possible` to all sensors `SetGroups<SENSOR>` classes, not just EK80. 

Done. @b-reyes , this PR will need your review.

@lsetiawan : If you are good with the PR as is, can you approve it? Otherwise, comment or commit away :smiley_cat:

**Review by lsetiawan — APPROVED** [review]

**Comment by b-reyes** [comment]
> tree_dict is initially defined as a dict, though. As far as I can see, it remains a dict until all the group set_* methods have been executed. So, that's why I don't understand the error. Anyway, like I said, my solution does work, though it seems inelegant.

I looked into this a little bit because I too was confused. From my understanding, dictionaries since Python 3.6 preserve order. Datatree follows this order using `.items()` as you can see from their [from_dict()](https://github.com/xarray-contrib/datatree/blob/badb7bc468bb48b163947af9f0626ceda19e272b/datatree/datatree.py#L353). That's why you get an error if the parents and children are out of order in the dictionary. 

> I agree that it's more intuitive. But for EK80, it turns out it's not possible. In convert/api.py, you don't know whether there'll be 1 or 2 beam groups until set_beam is executed. set_sonar needs to know how many beam groups there are.

If we want to preserve order i.e. first set `Sonar` and then set `Sonar/Beam_groupX`, we could just look at the parsed data. For example, in EK80 we would say there are two beam groups if `power` and `complex` are in the keys of `self.parser_obj.ping_data_dict`.

File: echopype/convert/set_groups_ek80.py [hunk 1]
```diff
@@ -194,7 +196,8 @@ def set_sonar(self) -> xr.Dataset:
                     ["channel"],
                     list(self.parser_obj.config_datagram["configuration"].keys()),
                     self._varattrs["beam_coord_default"]["channel"],
-                )
+                ),
+                "beam_group": beam_groups_coord["beam_group"],
```
emiliom, line 200: Thanks for catching this! Looking at my source code, I think I dropped it when resolving the conflict with your frequency > channel PR. Here's what I had, using the ** pattern I use elsewhere. It's cleaner, though less explicit:
```suggestion
                **beam_groups_coord,
```

b-reyes: @emiliom I like your suggestion more. `**beam_groups_coord` is cleaner and is more consistent with the rest of the code.

echopype/convert/set_groups_base.py
```diff
@@ -187,22 +187,24 @@ def _parse_NMEA(self):
         return time1, msg_type, lat, lon
 
     def _beam_groups_vars(self):
```
b-reyes, line 189: @emiliom I don't think this function correctly accounts for the EK80 sensor. From my understanding, there do exist EK80 files that only have the power data. Thus, these types of files would have `Beam_group1` with the description "contains complex backscatter data and other beam or channel-specific data". If you want a file that exhibits this behavior, please see `echopype/test_data/ek80_new/D20211004-T234429.raw`.

emiliom: Can you elaborate? I *think* what you're saying is that the current group description for EK80 `Beam_group1` is inadequate. If that's the case, then the fix would be to that text in `SetGroupsEK80` (note that @leewujung composed that text). This function is generic and just sets up the beam_groups_descr variable and beam_group coord variable, based on the beam group information that's passed to it.

emiliom:
> tree_dict is initially defined as a dict, though. As far as I can see, it remains a dict until all the group set_* methods have been executed. So, that's why I don't understand the error. Anyway, like I said, my solution does work, though it seems inelegant.
>
> I looked into this a little bit because I too was confused. From my understanding, dictionaries since Python 3.6 preserve order. Datatree follows this order using .items() as you can see from their [from_dict()](https://github.com/xarray-contrib/datatree/blob/badb7bc468bb48b163947af9f0626ceda19e272b/datatree/datatree.py#L353). That's why you get an error if the parents and children are out of order in the dictionary.

Thanks for looking into it! I think that does explain the problem, if Datatree is in fact relying on the order in the dict passed to it. 

I don't see any real need to maintain the order of `set_sonar` and `set_beam` execution. So, what I've done should be fine for our purposes.

echopype/convert/set_groups_base.py
b-reyes: I think the description is fine, just how you are assigning it to the beam group is incorrect. This is a known possibility, in `1.0.yml` we have 
```
name: Beam_group1
    description: >-
      contains backscatter data (either complex samples or uncalibrated power samples)
      and other beam or channel-specific data, including split-beam angle data when they exist.
```
Note that we say either complex samples OR uncalibrated power samples

Basically, for EK80, your code should check if only one beam group exists and if only one beam group exists, then it should select the appropriate description. It seems like the issue here (at least for EK80) is that `beamgroups_possible` ties the description to a specific beam group i.e. `Beam_group1` is always for the complex backscatter data.

emiliom: Thanks. Let me chew on this. I'll get back to it mid-late afternoon, hopefully.

emiliom: Ok, I've looked at this more closely. For the sample file you listed (`test_data/ek80_new/D20211004-T234429.raw` -- thanks!), the data is written to `Beam_group1` and its description becomes the EK80 `Beam_group1` description text we have, "contains complex backscatter data and other beam or channel-specific data." I'm not sure if you were saying that that's what *should* happen but doesn't or what does happen but shouldn't, but I believe you're saying the latter. That's the intended behavior, and not something that was changed in this PR. I think @leewujung provided that EK80 beam group description text.

I still think that what you're getting at can be addressed by simply changing the description in `beamgroups_possible` that's  associated with `Beam_group1`. It sounds like something like this would be more appropriate: "contains backscatter data (either complex samples or uncalibrated power samples) and other beam or channel-specific data".

Also, just to narrow down the problem, the function `beam_groups_vars` itself would not be the source of the potential problem, or solution. It's very generic. It just writes out what's passed to it. The issue you're getting at is what's passed to it by `SetGroupsEK80.set_sonar`, which occurs here:
https://github.com/OSOceanAcoustics/echopype/blob/91dc06d2b42be14a938394f6aba9b84d3ef43f10/echopype/convert/set_groups_ek80.py#L171-L172

The code does check -- elsewhere -- if only one beam group exists; if that's the case, that group will be `Beam_group1` and the corresponding  dict from `beamgroups_possible` is passed to it. The description text for that group doesn't need to exactly specify whether the beam group is complex or power only; that *would* be nice, and we could look into it in the future, but it's not necessary. That description does need to be accurate and not misleading, though. It sounds like a change to that description along the lines of what I suggest would make it more accurate.

b-reyes: > Ok, I've looked at this more closely. For the sample file you listed (`test_data/ek80_new/D20211004-T234429.raw` -- thanks!), the data is written to `Beam_group1` and its description becomes the EK80 `Beam_group1` description text we have, "contains complex backscatter data and other beam or channel-specific data." I'm not sure if you were saying that that's what _should_ happen but doesn't or what does happen but shouldn't, but I believe you're saying the latter. That's the intended behavior, and not something that was changed in this PR. I think @leewujung provided that EK80 beam group description text.

Yes, I was saying the latter. For the file you reference (and I provided), there is no complex backscatter data, so the description is wrong.  

> I still think that what you're getting at can be addressed by simply changing the description in `beamgroups_possible` that's associated with `Beam_group1`. It sounds like something like this would be more appropriate: "contains backscatter data (either complex samples or uncalibrated power samples) and other beam or channel-specific data".

Although your change to the description does fix the issue I described, it somewhat conflicts with the `Beam_group2` description: "contains backscatter power (uncalibrated) and other beam or channel-specific data, including split-beam angle data when they exist." I am sure any user could infer that `Beam_group1` corresponds to "complex samples" when two beam groups exist. However, I liked the idea of being very specific and stating whether the beam group has complex or power only data. I think this would be simple to implement and would benefit the user by providing further clarity. If you think this change to a more specific description conflicts with the intent of this PR, I understand. I am willing to take on this issue, if you do not have the time for it. 

> Also, just to narrow down the problem, the function `beam_groups_vars` itself would not be the source of the potential problem, or solution. It's very generic. It just writes out what's passed to it. The issue you're getting at is what's passed to it by `SetGroupsEK80.set_sonar`

I see where you are coming from, maybe the function `beam_groups_vars` is not the most appropriate place to change this.

emiliom: > However, I liked the idea of being very specific and stating whether the beam group has complex or power only data. I think this would be simple to implement and would benefit the user by providing further clarity. If you think this change to a more specific description conflicts with the intent of this PR, I understand. I am willing to take on this issue, if you do not have the time for it.

If you really think it'd be *very* simple to implement, go ahead; otherwise my vote would be for leaving it to 0.6.1. While the impact of this addition is "minor" on the scope of this PR (and it does fit its intent), I think it should be done in a different PR, mainly so we can move on with this one.

b-reyes: Maybe "simple" was the wrong word. I would say it is doable haha. I am fine if you would like to leave this for 0.6.1. I think we have enough on our plate for 0.6.0. I will create a new issue for this. For this PR, we can just change the `Beam_group1` description to "contains backscatter data (either complex samples or uncalibrated power samples) and other beam or channel-specific data", as you suggested.

emiliom: > I am fine if you would like to leave this for 0.6.1. I think we have enough on our plate for 0.6.0. I will create a new issue for this.

That would be perfect, thank you. If you already had concrete ideas on how to accomplish this, I suggest you describe them at least briefly when creating the new issue, while they're fresh in your mind.

> For this PR, we can just change the Beam_group1 description to "contains backscatter data (either complex samples or uncalibrated power samples) and other beam or channel-specific data", as you suggested.

I'll go ahead and make the change now.

echopype/convert/set_groups_ek80.py
```diff
@@ -43,23 +43,23 @@ class SetGroupsEK80(SetGroupsBase):
         "beamwidth_twoway_athwartship",
     }
 
+    beamgroups_possible = [
+        {
+            "name": "Beam_group1",
+            "descr": "contains complex backscatter data and other beam or channel-specific data.",  # noqa
```
emiliom, line 49:
```suggestion
            "descr": (
                "contains backscatter data (either complex samples or uncalibrated power samples)"  # noqa
                " and other beam or channel-specific data"
            )
```

b-reyes (approved):
@emiliom this looks good to me!

Issue #664, opened 2022-05-05
Difference in Datatree and dot access to EchoData groups


b-reyes:
I found an interesting "bug" that happens due to this inbetween state of using Datatree. In `calibrate_ek.py`, we sometimes do actions such as `ed.beam = blah` (something that seems a little strange to me). After these types of actions, `ed.beam` will be different from `ed['Sonar/Beam_group1']`. This was obvious to me once I came across it, but it was definitely not something I had thought about. @lsetiawan if possible, maybe we could address this inconsistency? Is it possible to have say `ed.beam` get `ed._tree['Sonar/Beam_group1'].ds`?

b-reyes:
This issue has been fully addressed in PR #665.

PR #665, opened 2022-05-05
Add setattr functionality to echodata and more tests

lsetiawan:
This PR addresses #664 where when getting the attribute, say `.beam` on echodata object, the underlying dataset may differ from the tree dataset. Additionally, after setting the attribute, the dataset may differ. This PR now introduces `__setattr__` built-in method for us to fine tune the behavior of attribute setting such as `EchoData.beam = some_other_dataset`.

### DIFF
--- a/echopype/tests/echodata/test_echodata.py
+++ b/echopype/tests/echodata/test_echodata.py
@@ -2,6 +2,7 @@ from textwrap import dedent
 
 import fsspec
 
+from datatree import DataTree
 from zarr.errors import GroupNotFoundError
 
 import echopype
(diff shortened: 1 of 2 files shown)


echopype/tests/echodata/test_echodata.py
```diff
@@ -272,6 +275,70 @@ def test_repr_html(self, converted_zarr):
...
+        current_ed_top = ed.top
+        ed.beam = sample_data
+        ed.top = sample_data2
+
+        assert ed.beam.equals(sample_data) is True
+        assert ed.beam.equals(ed['Sonar/Beam_group1']) is True
+        assert ed.beam.equals(current_ed_beam) is False
+
+        assert ed.top.equals(sample_data2) is True
+        assert ed.top.equals(ed['Top-level']) is True
```
b-reyes, line 292: Do we need the `is True` here and on lines 287,288? I understand why 289 needs it.

b-reyes (commented):
@lsetiawan thank you for putting this together! Everything seems to work well. I only had one comment on the tests.

echopype/tests/echodata/test_echodata.py
lsetiawan: Yea. This ensures that we get `True`. Probably don't need it, but I like to be explicit 😄

lsetiawan:
If you think all is good. Could you put the approval and go ahead and merge? Thanks @b-reyes 😃

b-reyes (approved):
@lsetiawan I think this is good to go.

Issue #634, opened 2022-04-19
Beam group variable suffixes `_alongship/athwart` vs  `_major/minor`


b-reyes:
In issue #520 it was found that variables within the Beam group with the suffixes `_alongship/athwart` do not appear in the convention. It seems that their original intent was possibly to coincide with the variables `beamwidth_receive/transmit_major/minor` in the convention. Based on the convention beamwidth_receive_major/minor are listed as "M" and beamwidth_transmit_major/minor are listed as "MA" (mandatory when applicable).  Thus, this conflict with the convention needs to be addressed. In #520 @leewujung made the following comment in regards to the matter: 

> For the athwartship/alongship vs major/minor question:
> Thinking out loud here, it seems that we can do the following:
> 
> 1. use what @gavinmacaulay pointed out in v2 in our compliance in v1 and describe clearly in the `comment` attribute [of the `_alongship/athwart` variables what our interpretation of these variables are] and reference the sources.
> 2. set `*_major/minor` to NaN and use `*_athwartship/alongship` in the variables as in the raw data, and describe clearly that this is a decision we've made in echopype.
> 
> I am inclined slightly to option 2 so that there is no ambiguity in interpreting the variables on what they are. When we need to go to v2, connecting the right variables to according the v2 definitions is straightforward.

leewujung:
Hey @b-reyes : what I meant was that I will take the part about the 1-way and 2-way beamwidth business. For the overall handling I think you'll be better suited since you're already working on #520 .

emiliom:
> In issue #520 it was found that variables within the Beam group with the suffixes `_alongship/athwartship` do not appear in the convention. It seems that their original intent was possibly to coincide with the variables `beamwidth_receive/transmit_major/minor` in the convention.

To add to this: there is no doubt about this original intent! Otherwise, @leewujung can confirm :stuck_out_tongue_winking_eye: 

For echopype 0.6.0 we also agreed to add a `comment` attribute to each `_alongship/athwart` variable that is based on a convention variable, indicating the variable it broadly corresponds to. It could be something like this:
```
beamwidth_receive_athwartship
:comment: "Corresponds to beamwidth_receive_major in SONAR-netCDF4 vers 1, but introduced in echopype for echosounders"
```

The change @leewujung describes above will also be for 0.6.0. It sounds like it's fairly easy.

[...]

PR #668, opened 2022-05-06
Add comment attribute to all _alongship/_athwartship variables and use two-way beamwidth variables

leewujung:
This PR addresses #634.

I made some executive decisions on what echopype would do here to make sure there is no confusion on what the variables are and where they come from. 

In this PR:
- The beamdwidth variables are called `beamwidth_twoway_alongship/athwartship`: the convention defines 1-way beamwidth for transmit and receive separately. While I can split the two-way beamwidth into one-way beamwidth (by multiplying the value by 2), I think it is safer for users to see its original form without echopype altering the data.
- I've added a `comment` attribute to all `*_alongship`/`*_athwartship` variables, to say that we introduce this in echopype for Simrad echosounders, and the angle correspondence with [convention ver.2 definitions on p.72](http://nbviewer.org/github/ices-publications/SONAR-netCDF4/blob/v2.0/Formatted_docs/crr341.pdf). I know we are _not_ going to ver.2 any time soon, but since the angle definitions in ver.1  can be confusing, I think it is better to just include the correspondence with ver.2 where the angle definitions for split-beam echosounders are clearly stated.

### DIFF
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -414,10 +414,10 @@ class SetGroupsEK60(SetGroupsBase):
                         "units": "arc_degree",
                         "valid_range": (0.0, 360.0),
                         "comment": (
-                            "Introduced in echopype for Simrad echosounders to avoid potential confusion with convention definitions. " # noqa
-                            + "The alongship angle corresponds to the minor angle in SONAR-netCDF4 vers 2. " # noqa
-                            + "The convention defines one-way transmit or receive beamwidth, but Simrad echosounders record two-way beamwidth in the data." # noqa
-                        )
+                            "Introduced in echopype for Simrad echosounders to avoid potential confusion with convention definitions. "  # noqa
+                            + "The alongship angle corresponds to the minor angle in SONAR-netCDF4 vers 2. "  # noqa
+                            + "The convention defines one-way transmit or receive beamwidth, but Simrad echosounders record two-way beamwidth in the data."  # noqa
+                        ),
                     },
                 ),
                 "beamwidth_twoway_athwartship": (
@@ -429,10 +429,10 @@ class SetGroupsEK60(SetGroupsBase):
                         "units": "arc_degree",
                         "valid_range": (0.0, 360.0),
                         "comment": (
-                            "Introduced in echopype for Simrad echosounders to avoid potential confusion with convention definitions. " # noqa
-                            + "The athwartship angle corresponds to the major angle in SONAR-netCDF4 vers 2. " # noqa
-                            + "The convention defines one-way transmit or receive beamwidth, but Simrad echosounders record two-way beamwidth in the data." # noqa
-                        )
+                            "Introduced in echopype for Simrad echosounders to avoid potential confusion with convention definitions. "  # noqa
+                            + "The athwartship angle corresponds to the major angle in SONAR-netCDF4 vers 2. "  # noqa
+                            + "The convention defines one-way transmit or receive beamwidth, but Simrad echosounders record two-way beamwidth in the data."  # noqa
+                        ),
                     },
                 ),
                 "beam_direction_x": (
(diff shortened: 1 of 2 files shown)


echopype/convert/set_groups_ek60.py
```diff
@@ -405,46 +405,34 @@ def set_beam(self) -> xr.Dataset:
...
                     {"long_name": "type of transducer (0-single, 1-split)"},
                 ),
-                # TODO: check EK60 data spec:
-                #  the beamwidths provided are most likely 2-way beamwidth so below needs to change
-                "beamwidth_receive_alongship": (
+                "beamwidth_twoway_alongship": (
                     ["channel"],
                     beam_params["beamwidth_alongship"],
                     {
                         "long_name": "Half power one-way receive beam width along "
```
emiliom, line 412: Given the change to the variable name, shouldn't the `long_name` be changed as well?

echopype/convert/set_groups_ek60.py
```diff
@@ -405,46 +405,34 @@ def set_beam(self) -> xr.Dataset:
...
+                            + "The convention defines one-way transmit or receive beamwidth, but Simrad echosounders record two-way beamwidth in the data."  # noqa
+                        ),
                     },
                 ),
-                "beamwidth_receive_athwartship": (
+                "beamwidth_twoway_athwartship": (
                     ["channel"],
                     beam_params["beamwidth_athwartship"],
                     {
                         "long_name": "Half power one-way receive beam width along "
```
emiliom, line 427: Given the change to the variable name, shouldn't the `long_name` be changed as well?

echopype/convert/set_groups_ek60.py
```diff
@@ -405,46 +405,34 @@ def set_beam(self) -> xr.Dataset:
...
                     ["channel"],
                     beam_params["beamwidth_alongship"],
                     {
                         "long_name": "Half power one-way receive beam width along "
                         "alongship axis of beam",
                         "units": "arc_degree",
                         "valid_range": (0.0, 360.0),
+                        "comment": (
+                            "Introduced in echopype for Simrad echosounders to avoid potential confusion with convention definitions. "  # noqa
+                            + "The alongship angle corresponds to the minor angle in SONAR-netCDF4 vers 2. "  # noqa
```
emiliom, line 418: The + sign is unnecessary for string concatenation in this context. I'm totally fine with leaving it there for now, though I think the strings are a bit more readable w/o it (and I believe we typically don't use +'s in such multi-line strings in echopype).

This comments applies to all other `comment` multi-line strings in this PR.

emiliom:
In the `comment` strings where you make statements like "The alongship angle corresponds to the minor angle in SONAR-netCDF4 vers 2." (ditto for athwartship/major), shouldn't that be **vers 1**? I don't know how v2 uses major/minor, but [the correspondence we were trying to express was with v1.](https://github.com/OSOceanAcoustics/echopype/issues/634#issuecomment-1118776804)

echopype/convert/set_groups_ek60.py
emiliom: Also, should we include the (roughly) corresponding variable(s) from v1? ie, `beamwidth_receive_minor` and `beamwidth_transmit_minor`. eg, replace ""The convention defines one-way transmit or receive beamwidth, but Simrad echosounders record two-way beamwidth in the data." with ""The convention defines one-way transmit or receive beamwidth (beamwidth_receive_minor and beamwidth_transmit_minor), but Simrad echosounders record two-way beamwidth in the data."

This comment applies to other, related changse in this PR.

emiliom:
I didn't add any inline comments for EK80 specifically, but many of my inline comments for EK60 apply to EK80 too.

echopype/convert/set_groups_ek60.py
leewujung: changed.

leewujung:
@emiliom : I think yours and my interpretation of ver.1 definition were actually opposite (see https://github.com/OSOceanAcoustics/echopype/issues/520#issuecomment-1102170069), so I think it is a dangerous thing to refer to ver.1 for the case of scientific echosounders because people can get really confused. I think mentioning this correspondence with ver.2 is a safer thing to do -- I pointed to the exact page on ver.2 convention above for this definition.

echopype/convert/set_groups_ek60.py
leewujung: I replaced the original one with the new entry `"The convention defines one-way transmit or receive beamwidth (beamwidth_receive_minor and beamwidth_transmit_minor), but Simrad echosounders record two-way beamwidth in the data."`

leewujung:
@emiliom : I've put in all recommended changes except for the ver.1 vs ver.2 thing mentioned in my previous comment. Let me know what you think. Thanks!

emiliom:
@leewujung Note that there was a [test failure](https://github.com/OSOceanAcoustics/echopype/actions/runs/2302436812) during the merge. It looks like it had to with the echodata repr. That error didn't occur earlier. It's probably not worth looking into in the context of this PR.

PR #674, opened 2022-05-11
Change src_filenames string attribute to source_filenames list-of-strings variable

### LINKED ISSUE #620: Change `src_filenames` string attribute to `source_filenames` list-of-strings variable

emiliom:
Closes #620. Implemented using a new prov.py function, `source_files_vars`, that is reused wherever the `source_filenames` variable is needed, including in `assemble_combined_provenance` where it already existed. Addresses convention adherence.

### DIFF
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -7,7 +7,7 @@ import xarray as xr
 
 from ..echodata.convention import sonarnetcdf_1
 from ..utils.coding import COMPRESSION_SETTINGS, set_encodings
-from ..utils.prov import echopype_prov_attrs
+from ..utils.prov import echopype_prov_attrs, source_files_vars
 
 DEFAULT_CHUNK_SIZE = {"range_sample": 25000, "ping_time": 2500}
 
@@ -73,8 +73,8 @@ class SetGroupsBase(abc.ABC):
 
     def set_provenance(self) -> xr.Dataset:
         """Set the Provenance group."""
-        prov_dict = echopype_prov_attrs(process_type="conversion", source_files=self.input_file)
-        ds = xr.Dataset()
+        prov_dict = echopype_prov_attrs(process_type="conversion")
+        ds = xr.Dataset(source_files_vars(self.input_file))
         ds = ds.assign_attrs(prov_dict)
         return ds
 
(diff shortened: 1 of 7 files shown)


echopype/convert/set_groups_base.py
```diff
@@ -73,8 +73,8 @@ def set_toplevel(self, sonar_model, date_created=None) -> xr.Dataset:
 
     def set_provenance(self) -> xr.Dataset:
         """Set the Provenance group."""
-        prov_dict = echopype_prov_attrs(process_type="conversion", source_files=self.input_file)
-        ds = xr.Dataset()
+        prov_dict = echopype_prov_attrs(process_type="conversion")
+        ds = xr.Dataset(source_files_vars(self.input_file))
```
b-reyes, line 77: I see that previously we only stored the `raw_file` path from `open_raw` as an attribute. However, for AZFP we have both the `raw_file` and `xml_path` as inputs and the file names for these can be quite different. Should we make it so that we store both the `raw_file` and `xml_path` in the variable?

emiliom: Good point. I agree that storing/preserving `xml_path` would be a good idea, or at least something we should discuss at some point. But I think that's a new development that would be outside the scope of this PR. This PR doesn't change *what* is stored, just *how* it is stored.

b-reyes: I agree that it is outside the scope of this PR. If we do need to implement it, I think that would be a breaking change. So, we would need to fit this into v0.6.0, right?

emiliom: As a new feature, I would push this to >0.6.0. As usual: new issue!

b-reyes (approved):
@emiliom everything looks good here! I have opened up issues about my concerns that can be investigated further (after v0.6.0 has been released).

emiliom:
Thanks @b-reyes ! I'll merge.

PR #683, opened 2022-05-16
Renamed Vendor group to Vendor_specific

### LINKED ISSUE #675: Rename `Vendor` group to `Vendor_specific`
The convention specifies `Vendor_specific` as the group name

emiliom:
Closes #675

Also updated the two converted EK60 "_NEW" files in `test_data/ek60/ncei-wcsd`; updated a test to work with it; updated the files on the Google Drive test data folder; then manually ran the GH Action workflow that [builds the docker image](https://github.com/OSOceanAcoustics/echopype/actions/workflows/docker.yaml).

### DIFF
--- a/echopype/echodata/convention/1.0.yml
+++ b/echopype/echodata/convention/1.0.yml
@@ -46,9 +46,9 @@ groups:
       Only exists if complex backscatter data are already in Sonar/Beam_group1
     ep_group: Sonar/Beam_group2
   vendor:
-    name: Vendor specific
+    name: Vendor_specific
     description: contains vendor-specific information about the sonar and the data.
-    ep_group: Vendor
+    ep_group: Vendor_specific
 variable_and_varattributes:
   # Convention-based global, coordinate and variable attributes
   beam_coord_default:
--- a/echopype/tests/echodata/test_echodata.py
+++ b/echopype/tests/echodata/test_echodata.py
@@ -252,7 +252,7 @@ class TestEchoData:
             ├── Provenance: contains metadata about how the SONAR-netCDF4 version of the data were obtained.
             ├── Sonar: contains sonar system metadata and sonar beam groups.
             │   └── Beam_group1: contains backscatter data (either complex samples or uncalibrated power samples) and other beam or channel-specific data, including split-beam angle data when they exist.
-            └── Vendor specific: contains vendor-specific information about the sonar and the data."""
+            └── Vendor_specific: contains vendor-specific information about the sonar and the data."""
         )
         ed = EchoData.from_file(converted_raw_path=converted_zarr)
         actual = "\n".join(x.rstrip() for x in repr(ed).split("\n"))
@@ -317,7 +317,7 @@ class TestEchoData:
             'provenance': 'Provenance',
             'sonar': 'Sonar',
             'beam': 'Sonar/Beam_group1',
-            'vendor': 'Vendor',
+            'vendor': 'Vendor_specific',
         }
         for group, path in expected_groups.items():
             ds = getattr(ed, group)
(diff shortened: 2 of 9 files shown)


echopype/tests/echodata/test_echodata.py
```diff
@@ -317,17 +317,17 @@ def test_getattr(self, converted_zarr):
...
             ds = getattr(ed, group)
             assert ds.equals(ed[path])
 
     def test_setitem(self, converted_zarr):
         ed = EchoData.from_file(converted_raw_path=converted_zarr)
-        ed['Sonar/Beam_group1'] = ed['Sonar/Beam_group1'].rename({'frequency': 'channel'})
+        ed['Sonar/Beam_group1'] = ed['Sonar/Beam_group1'].rename({'beam': 'beam_newname'})
 
-        assert sorted(ed['Sonar/Beam_group1'].dims.keys()) == ['channel', 'ping_time', 'range_bin']
+        assert sorted(ed['Sonar/Beam_group1'].dims.keys()) == ['beam_newname', 'channel', 'ping_time', 'range_sample']
```
b-reyes, line 330: @emiliom did you add/modify these lines? I don't remember this being the current state of these tests. @lsetiawan's PR #665 that I approved is different here. Even stranger, somehow `range_bin` showed up here...

echopype/echodata/convention/1.0.yml
```diff
@@ -46,9 +46,9 @@ groups:
       Only exists if complex backscatter data are already in Sonar/Beam_group1
     ep_group: Sonar/Beam_group2
   vendor:
-    name: Vendor specific
+    name: Vendor_specific
```
b-reyes, line 49: Not a huge deal, but do we want an underscore here or a space? One impact that comes to mind is the repr, I believe `name` is what is currently used to set it.

echopype/tests/echodata/test_echodata.py
emiliom: This test failed, so I traced its origin. @lsetiawan added it on May 6. Line 328 was originally (in Don's PR) as:
```python
ed['Sonar/Beam_group1'] = ed['Sonar/Beam_group1'].rename({'frequency': 'channel'})
```
It turns out that the test was using one of the old (0.5.6), pre-converted test datasets (`*__NEW.zarr`) I mentioned above. That's why it referred to the `frequency` and `range_bin` dimension names that no longer exist. These files need to be updated to the current state of 0.6.0dev, so when I updated them, the test failed. Hence my modifications to the test.

echopype/echodata/convention/1.0.yml
emiliom: Unless something has changed, I'm pretty sure that the attribute that's critical is `ep_group`; that one does need to have the underscore. My understanding is that `name` is only used in the repr print out, and the content of the string doesn't impact anything other than the intelligibility of what users see. As for whether to use underscore or space, I think it's a toss up. I changed it to underscore just b/c for Beam_group1/2 we're also retaining the underscore.

echopype/tests/echodata/test_echodata.py
b-reyes: I see, ugh this is messy. I suppose this will be the case until we arrive at a final form for the structure.

echopype/echodata/convention/1.0.yml
b-reyes: I suppose it is better to go with the underscore as that is the actual value used to access it in echopype.

b-reyes (approved):
@emiliom since all the tests are passing and I didn't find any issues in the changes, I think this is good to go.

emiliom:
Thanks! I'll merge

PR #687, opened 2022-05-18
Squeeze and drop the redundant `beam` dimension in calibration outputs

### LINKED ISSUE #644: Squeeze out `beam` dimension in `compute_Sv/TS` outputs
With the addition of the `beam` dimension to a large number of variables in the `Beam_groupX` groups for compliance with the convention, now the outputs of `compute_Sv/TS` would also have a `beam` dimension. This dimension is of length=1 for AZFP, EK60 and EK80 power/angle data, but of length=4 for EK80 complex data.

The EK80 length=4 output does not make sense, since these are from a single split-beam transducers. There is a conceptual incongruence and potential confusion that takes place here. This is related to how we use the beam dimension to store data from each element of a split-beam transducer. (See https://github.com/OSOceanAcoustics/echopype/pull/638#issuecomment-1105807694 and https://github.com/OSOceanAcoustics/echopype/pull/638#issuecomment-1106744630 for detail)

The consensus was to squeeze out this extra `beam` dimension at the outputs of `computer_Sv/TS`.

We need to decide on our approach:
- option 1: modify the code so that the redundant `beam` dimension is dropped so that the variables used in the computation "makes sense"
- option 2: just carry through the redundant `beam` dimension toward the end of the computation and drop before returning the outputs.

Having done some of option 1 (those are "saved" in a local branch now), option 2 seems a cleaner approach. However it probably would have an impact computationally. 🤔

leewujung:
This PR closes #644.

I'm taking the approach to just squeeze and drop the redundant `beam` dimension at the end of the calibration operations. There may be efficiency advantage to take care of this extra dimension earlier in the computation, but let's delay this part to post-v0.6.0.

I've also noted the variables that have the redundant `beam` dimension (as comments) as I go through to make this change.

## Tasks
- [x] Squeeze out `beam` from `CalibrateEK._cal_power()`
- [x] Squeeze out `beam` from `CalibrateEK._cal_complex()`
- [x] Squeeze out `beam` from `CalibrateAZFP._cal_power()`
- [x] Fix all calibration tests

### DIFF
--- a/echopype/calibrate/calibrate_ek.py
+++ b/echopype/calibrate/calibrate_ek.py
@@ -752,7 +752,12 @@ class CalibrateEK80(CalibrateEK):
 
         # Derived params
         # TODO: right now squeeze out the single ping_time dimension
-        #       need to properly align based on actual time
+        #  Need to properly align based on actual time
+        #  This ping_time is actually renamed from time1 in Environment group,
+        #  so that in the computation of range_meter it can be aligned with echo data easily.
+        #  This however makes this part of code less traceable,
+        #  since the ping_time here does not come from the "pings"
+        #  Will resolve later.
         sound_speed = self.env_params["sound_speed"].squeeze(drop=True)
         absorption = self.env_params["sound_absorption"].sel(channel=chan_sel).squeeze(drop=True)
         range_meter = self.range_meter.sel(channel=chan_sel)
--- a/echopype/calibrate/calibrate_ek.py
+++ b/echopype/calibrate/calibrate_ek.py
@@ -181,7 +181,7 @@ class CalibrateEK(CalibrateBase):
             CSv = (
                 10 * np.log10(beam["transmit_power"])
                 + 2 * self.cal_params["gain_correction"]
-                + self.cal_params["equivalent_beam_angle"]
+                + self.cal_params["equivalent_beam_angle"]  # has beam dim
                 + 10
                 * np.log10(
                     wavelength**2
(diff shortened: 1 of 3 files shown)


echopype/calibrate/calibrate_ek.py
```diff
@@ -748,10 +750,17 @@ def _cal_complex(self, cal_type, waveform_mode) -> xr.Dataset:
...
+        # TODO: right now squeeze out the single ping_time dimension
+        #  Need to properly align based on actual time
+        #  This ping_time is actually renamed from time1 in Environment group,
+        #  so that in the computation of range_meter it can be aligned with echo data easily.
+        #  This however makes this part of code less traceable,
+        #  since the ping_time here does not come from the "pings"
+        #  Will resolve later.
+        sound_speed = self.env_params["sound_speed"].squeeze(drop=True)
+        absorption = self.env_params["sound_absorption"].sel(channel=chan_sel).squeeze(drop=True)
+        range_meter = self.range_meter.sel(channel=chan_sel)
```
emiliom, line 763: I'm curious why you removed the squeeze() here altogether. I assume it's all good, but I'll take a closer look.

leewujung: It is because `range_meter` would usually have `ping_time` with length >1 so `squeeze` here has no effect,  and even if the `ping_time` length is 1, `ping_time` actually should not be removed.

emiliom: Thanks!

emiliom:
The code looks good, the tests pass, and I ran it locally to verify that the beam dimension is squeezed out where it should. Looks good. Merging.

PR #636, opened 2022-04-20
Add development roadmap to docs

leewujung:
This is to address #602.

A couple of questions:
- Where do we want this page to go?
- Where should the data processing levels table go (I think it's too long to be in this page)?

This PR still has a few links to be fixed but the content is ready for review.

Note that right now this is PR to the `dev` branch but we will add this to `stable` once merged.

Issue #602, opened 2022-03-28
Add a "Development roadmap" page to docs


leewujung:
We should have a "Development roadmap" page to lay out what we envision in developing echopype in the long term. This becomes even more important now that we are also developing other related packages, such as echoshader. 

The roadmap of [xbatcher](https://xbatcher.readthedocs.io/en/latest/roadmap.html) provides a nice concise example. The roadmaps of [pandas](https://pandas.pydata.org/docs/development/roadmap.html) and [xarray](https://xarray.pydata.org/en/stable/roadmap.html) are much more substantial (just like the packages themselves!) and elaborated.

We should also have in the docs a place to talk about exciting new features and breaking changes (another type of excitement!) in a human-friendly version for users, in addition to the more mechanical list of bullet points in the [What's new](https://echopype.readthedocs.io/en/stable/whats-new.html) page. 

For example, the breaking changes we will have in v0.6.0 and the consequences can detailed in this page:
- major restructuring of netCDF groups/subgroups in files generated by echopype: moving Beam groups to under the Sonar group as defined in the convention ver.1 (#519)
- variable renaming in order to: increase adherence with with the convention (e.g., #518), handle exceptions due to transducer configuration (#490, #566), and clarify meaning of a variable (e.g., #570)

Perhaps these near term important changes can be part of the Roadmap page too, in a section upfront before the long term dev goals?

Pinging @emiliom @b-reyes @lsetiawan for inputs.

leewujung:
Done!

PR #688, opened 2022-05-18
Development roadmap in docs

emiliom:
Builds on PR #636. Addresses #602.

- Provides draft but "clean" Development roadmap page
- Moves roadmap page position in the left sidebar to "Help & reference", between "What’s new" and "Contributing to echopype"


emiliom:
A couple of things to note:

- This PR is against `dev`. We had talked about updating `stable` before the release is out. Depending on the timing of the 0.6.0 release, that may not buy us any extra public availability anymore
- The text says "in the upcoming 0.6.0 release" in a few places. We'll need to revise that language when 0.6.0 is released, or in preparation for it

leewujung:
Thanks @emiliom !  I pushed some small changes, and think this is ready to be merged!

The only thing I'd raise here is that the location of roadmap seems a little hidden? But maybe that's just me?

leewujung (approved):

emiliom:
> Thanks @emiliom ! I pushed some small changes, and think this is ready to be merged!

Great. I'll let you merge.

> The only thing I'd raise here is that the location of roadmap seems a little hidden? But maybe that's just me?

It is definitely less prominent. Personally, I think it does belong in that section aimed at developers, contributors and other interested parties; that's what xarray does. I tend to think that what's in front/top should be what users will look for to actually use the software.  But I can also see value in having it where you had it originally, in the first section. If you think it really belongs there, please go ahead and change it in `_toc.yml`, and let's go with that. We can always discuss it again down the road.

leewujung:
I think we can tune this later. Merging now!

PR #690, opened 2022-05-19
🛠️ Fix `combine_echodata` to result in a working echodata object

### LINKED ISSUE #676: Issues with `combine_echodata` and repr
While trying out the `combine_echodata` function, I noticed that the resulting `EchoData` object causes errors when we try to print out the object. 

The following is a simple example that produces the error:

```
path_to_file1 = 'test_data/ek60/ncei-wcsd/Summer2017-D20170615-T190214.raw'
path_to_file2 = 'test_data/ek60/ncei-wcsd/Summer2017-D20170615-T190843.raw'
ed1 = ep.open_raw(path_to_file1, sonar_model='EK60')
ed2 = ep.open_raw(path_to_file2, sonar_model='EK60')

ed_comb = ep.combine_echodata([ed1, ed2])

print(ed_comb)
```

I believe the reason this happens is because we used the dot accessor in `combine_echodata`, rather than directly working with the Datatree.

lsetiawan:
This PR fixes #676. `combine_echodata` will now create the proper echodata object with an internal DataTree object.

### DIFF
--- a/echopype/echodata/combine.py
+++ b/echopype/echodata/combine.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import Any, Dict, List
 
 import xarray as xr
+from datatree import DataTree
 
 from ..core import SONAR_MODELS
 from ..qc import coerce_increasing_time, exist_reversed_time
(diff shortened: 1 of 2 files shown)


lsetiawan:
@b-reyes I noticed that in #689 you also modified `combine_echodata`. I hope this won't break anything you're doing there. I don't think we are clashing, but let me know if we are! Thanks.

b-reyes (approved):
@lsetiawan everything looks good and I am no longer getting errors! I now understand why we were getting errors (we were not setting the tree). Although we were both working in `combine.py`, I don't really think our changes conflict. If they do, I can easily resolve them in #689.

Issue #681, opened 2022-05-13
Varying variables in `Sonar` group amongst sensors


b-reyes:
While reviewing the `Sonar` group across the sensors, I noticed that there is a slight inconsistency in what is an attribute and what is a variable. For EK80 we have the following variables in the `Sonar` group, which all have the dimension `(channel)`

- `frequency_nominal`, `serial_number`, `sonar_model`, `sonar_serial_number`, `sonar_software_name`, `sonar_software_version`, `beam_group_descr`

I know that for EK80 it absolutely makes sense to make these variables as they may vary amongst the channels. 

For EK60 we only have `beam_group_descr` as a variable, however, we do have the attributes `sonar_model`, `sonar_serial_number`, `sonar_software_name`, `sonar_software_version`, which are variables in EK80.

For AZFP we also only have `beam_group_descr` as a variable, but we do have the attributes `sonar_model`, `sonar_serial_number`, `sonar_software_name`, `sonar_software_version`, which are variables in EK80. 

I was curious if there was a specific reason why these were left as attributes or if this was simply something that was over looked. If we want to bring consistency to these variables, then I suggest that we make these attributes into variables with the dimension `channel` and add the variable `frequency_nominal` to the `Sonar` group (for EK60 and AZFP). 

Note that I am not considering AD2CP here.

emiliom:
Quick reply:

- All the `sonar_*` attributes are defined as such by the convention. For EK80, based on what I've seen, I assume @leewujung found it necessary to change them to a variable in order to support different entries per channel (likely a need not anticipated in the convention vers. 1). Changing the attributes on the other instruments to variables would bring consistency across echopype at the expense of adherence to the convention.
- `beam_group_descr` was introduced by us and is not from the convention. It's currently implemented identically across instruments.

b-reyes:
> Quick reply:
> 
> * All the `sonar_*` attributes are defined as such by the convention. For EK80, based on what I've seen, I assume @leewujung found it necessary to change them to a variable in order to support different entries per channel (likely a need not anticipated in the convention vers. 1). Changing the attributes on the other instruments to variables would bring consistency across echopype at the expense of adherence to the convention.

[...]

Issue #571, opened 2022-02-28
Populate or fix sonar_software_name and sonar_model for EK60 and maybe other instruments


emiliom:
(Originally discussed in https://github.com/OSOceanAcoustics/echopype/issues/210#issuecomment-822642399)

For EK60:

- Populate `Sonar/sonar_software_name` attribute. Right now it's left blank; see [convert/set_groups_ek60.py#L170](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/convert/set_groups_ek60.py#L170)
- `Sonar/sonar_model` attribute entry is "ER60". It should be EK60. Currently EK60 `Sonar/sonar_model` is set in [convert/set_groups_ek60.py#L168](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/convert/set_groups_ek60.py#L168) as `self.parser_obj.config_datagram['sounder_name']`, so that looks fine, unless the datagram `sounder_name` is parsed incorrectly.

We haven't examined these closely (AFAIK) for EK80, AZFP or other instruments.

These attributes could be examined in parallel with `Sonar/sonar_serial_number`, #212

emiliom:
I don't think I'm the right person for this issue, since I'm not well versed with the raw datagrams.

leewujung:
I assigned both you and me, because I need to consult you on the naming choice of attributes that we will need to create to take care of the things that may not be in the convention.

emiliom:
Ah! Thanks for that clarification. I didn't see that.

emiliom:
ER60 is in fact the name of the Simrad software. eg, see https://support.echoview.com/WebHelp/Reference/File_formats/Simrad_data_files.htm

leewujung:
We decided to review the detailed content of variables _after_ v0.6.0 now that we have the disruptive type of changes in for v0.6.0.

emiliom:
`sonar_model` was already addressed for v0.6.0. See https://github.com/OSOceanAcoustics/echopype/issues/681#issuecomment-1339962902.

In this issue, the only remaining topic is `sonar_software_name`. It may be more convenient to address that question in #681 and close this issue.

PR #692, opened 2022-05-19
Consistent use of Sonar.sonar_model attribute, also consistent with echodata.sonar_model

emiliom:
Addresses #681, specifically https://github.com/OSOceanAcoustics/echopype/issues/681#issuecomment-1130581845. 

- `sonar_model` is now a `Sonar` attribute for all sensors, including EK80.
- The value of that attribute is now assigned based on the echodata `sonar_model` property.
- The EK80 `Sonar.sonar_model` variable was  to `Sonar.transducer_name` (not a convention variable or attribute).

This PR also addresses #571 by setting `sonar_software_name` to `self.parser_obj.config_datagram["sounder_name"]`. This value is typically (always?) "ER60", which [is in fact the Simar sonar software name.](https://support.echoview.com/WebHelp/Reference/File_formats/Simrad_data_files.htm)

### DIFF
--- a/echopype/convert/set_groups_ad2cp.py
+++ b/echopype/convert/set_groups_ad2cp.py
@@ -378,10 +378,10 @@ class SetGroupsAd2cp(SetGroupsBase):
         beam_groups_vars, beam_groups_coord = self._beam_groups_vars()
         ds = xr.Dataset(beam_groups_vars, coords=beam_groups_coord)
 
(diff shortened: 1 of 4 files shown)


b-reyes (approved):
@emiliom All of your changes look good to me!

emiliom:
Thanks so much! Merging.

PR #660, opened 2022-05-03
Handle plotting of dimension frequency to channel

lsetiawan:
This PR follows #657. The visualize module can now handle the change from frequency to channels, and should behave the same way. However, I currently do not have the full way to test this, since I need the updated code for `compute_range` to check on the `get_range` behavior. In the mean time below is a demo of the result from a sample netCDF that @b-reyes has provided.

## Demo

```
import echopype
from echopype import visualize as epviz

ed = echopype.open_converted('v060_EK80_beam_1_2_with_channel_beam.nc')
range_kwargs = {"encode_mode":"complex", "waveform_mode":"CW"}
plots = epviz.create_echogram(
    ed, robust=True, get_range=False, range_kwargs=range_kwargs
)
```
Screenshot from 2022-05-03 10-44-53

PR #695, opened 2022-05-20
Improve selection of calibration parameters in Vender group

leewujung:
This PR improves the selection of EK CW mode calibration parameters in the `Vendor` group (`sa_correction`, `gain_correction`) as the current implementation is inefficient and not flexible.

There is an added goal to allow calibrating the example file with duplicated frequency channels that alternates in `ping_time`, which fails in the current implementation.

This PR builds on top of #660 as one of the plotting test uses this example file and is currently marked to not run.

Not ready for review yet.


leewujung:
Closing this PR because it is now reproduced in #697 without dependency on #660.

PR #696, opened 2022-05-20
Handle single pathlib Path in prov source_files_var

emiliom:
Addresses an oversight in PR #674


emiliom:
The tests failed, but I think they're due to "random" CI issues (including aiobotocore!) that don't tell us anything about the tests themselves. I ran the test suite locally, w/o any problems. Also, this PR is very, very simple and would not cause the CI test problems we're seeing.

leewujung:
I agree that we can just merge this. I re-triggered the tests and 2 of 3 passed.

PR #697, opened 2022-05-20
Improve selection of calibration parameters from `Vendor` group

leewujung:
This PR improves the selection operation of EK60 and EK80 CW mode calibration parameters in the `Vendor` group (`sa_correction`, `gain_correction`) as the current implementation is inefficient and inflexible.

This PR has an additional goal to allow calibrating the test file with duplicated frequency channels that alternates in `ping_time` (pinging sequence: ch3 - ch4 - ch3 -ch4 -...), which fails in the current implementation, like below:

Screen Shot 2022-05-19 at 7 06 44 PM

I also added a test in test_calibrate.py to run `compute_Sv/TS` on the duplicated frequency file.

The changes here should allow the test that's being xfailed at the moment in #660 to be run.

### DIFF
--- a/echopype/calibrate/calibrate_ek.py
+++ b/echopype/calibrate/calibrate_ek.py
@@ -53,8 +53,6 @@ class CalibrateEK(CalibrateBase):
         param : str {"sa_correction", "gain_correction"}
             name of parameter to retrieve
         """
-        # TODO: need to test with EK80 power/angle data
-        #  currently this has only been tested with EK60 data
         ds_vend = self.echodata.vendor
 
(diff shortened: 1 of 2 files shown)


leewujung:
@b-reyes @lsetiawan : requesting review from one of you -- not sure who is tuned in to #660 at the moment. Thanks!

lsetiawan (approved):
Just tested this. Sure enough it works! Thanks @leewujung! I think this is ready for merging

lsetiawan:
I'll finish testing out and fixing #660, once this is merged.

PR #698, opened 2022-05-20
Pinning CI s3fs version to latest

lsetiawan:
This PR pins the s3fs for the CI runs since there seems to be some weird conda dependency bug where randomly the version 0.6.0 gets installed for s3fs.


lsetiawan:
Self merging. Not code related changes.

PR #712, opened 2022-05-31
Fix test pypi action

lsetiawan:
This PR fixes the test pypi action from past configuration location for versioning.


lsetiawan:
Self merging, CI hotfix.

Issue #491, opened 2021-11-13
Error parsing split-beam EK60 file with problematic phase angle information


emiliom:
@lgarzio we are posting much of our email exchange here so it's much easier to track and resolve. I'm making small edits for clarity.

**(2021-6-18, from @lgarzio)**

I’ve been working with a small group of people here at Rutgers trying to develop tools to process our echosounder data – both shipboard data (EK60) and an AZFP mounted on a glider.

Right now I’m trying to do the shipboard data and I’m running into an issue with converting some of the raw files.

Screen shot of the main error:

![lorigarzio-1](https://user-images.githubusercontent.com/742403/141601115-f9066eb7-0b2f-446a-9a26-3c864ce3f053.png)

What is `ping_data_dict['angle']`? The best we can tell, it might be equivalent to ‘angular position of raw pings’ in Echoview but we weren’t entirely sure. This is what it looks like in a file that parses correctly:

![lorigarzio-2](https://user-images.githubusercontent.com/742403/141601123-e3acc9c5-d2d8-4207-b2d2-711ec5bde370.png)

I can provide more details when I raise the issue in github, but I first wanted to see if you know off the top of your head what that variable is.

emiliom:
**(2021-6-18, from @leewujung)**

This is the electric phase angle recorded by the split beam echosounder. It's curious that this errors out since we have been parsing EK files with both single and split beams without problem. Can you see what beam_type is in `self.parser_obj.config_datagram["transceivers"]` for the channel that errors out? 

You can see a TODO at line 532 in `set_groups_ek60.py` -- it is possible that your echosounder was set to collect power only even though it is a split beam. If you would consider submitting a PR that would be really great! 

**(2021-6-18, response from @lgarzio)**

I wasn’t on this cruise so I don’t know what the echosounder settings were, so I guess anything is possible.

The weird thing is that I have 2 different sets of files from this echosounder (corresponding with 2 different net tows). I am able to parse files from net tow 1, it’s just the files from net two 2 I can’t parse. My colleague is able to load and read these files in echoview if that helps.

I’ll work on submitting all the info to github, but in the meantime if you’re interested:

- Processing code: https://github.com/lgarzio/echosounder-processing/blob/master/RossSeaEAGER/EAGER_ship_acoustics_process.py
- Environment: https://github.com/lgarzio/echosounder-processing/blob/master/environment.yml

[...]

PR #718, opened 2022-06-05
Use `mode` to determine whether to save angle data for EK60

leewujung:
This PR addresses #491 by using `mode` in the parsed data to determine whether to save angle data for EK60. Previously it was assumed that angle data are always collected for split-beam setup, but the test data file @lgarzio provided had a split-beam transducer but only collected power data.

Note the test function is commented out because the test file is too large (100 MB) to be included in regular test set.

### DIFF
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -580,7 +580,6 @@ class SetGroupsEK60(SetGroupsBase):
         # Construct Dataset with ping-by-ping data from all channels
         ds_backscatter = []
         for ch in ch_ids:
-            data_shape = self.parser_obj.ping_data_dict["power"][ch].shape
             ds_tmp = xr.Dataset(
                 {
                     "backscatter_r": (
@@ -655,7 +654,7 @@ class SetGroupsEK60(SetGroupsBase):
                     ),
                     "range_sample": (
                         ["range_sample"],
-                        np.arange(data_shape[1]),
+                        np.arange(self.parser_obj.ping_data_dict["power"][ch].shape[1]),
                         self._varattrs["beam_coord_default"]["range_sample"],
                     ),
                 },
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -661,12 +661,10 @@ class SetGroupsEK60(SetGroupsBase):
                 },
             )
 
-            # TODO: below needs to be changed to use
-            #  self.convert_obj.ping_data_dict['mode'][ch] == 3
-            #  1 = Power only, 2 = Angle only 3 = Power & Angle
-            # Set angle data if in split beam mode (beam_type == 1)
-            # because single beam mode (beam_type == 0) does not record angle data
-            if self.parser_obj.config_datagram["transceivers"][ch]["beam_type"] == 1:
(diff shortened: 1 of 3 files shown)


leewujung:
@emiliom : I am a little puzzled why the 2 most recent merges to `dev` showed up on this PR... I created this branch from the most up-to-date `dev`. I think @b-reyes ran into this at least once in the v0.6.0 PRs.
Not crucial here since the files involved are completely different so easy to tell apart, but I just wanted to make a note because this is unexpected.

emiliom:
Gosh, I don't know. But I do have one question: are you deliberately trying to merge this PR into `main` rather than `dev`?? That seems odd.

leewujung:
> are you deliberately trying to merge this PR into `main` rather than `dev`??
😂 clearly that's what happened when one tries to do a PR on deep on Sat night... I'll change it to `dev`.

leewujung:
@emiliom : This PR has addressed the problem (https://github.com/OSOceanAcoustics/echopype/issues/491#issuecomment-1152816564) so we can either merge it directly or if you could take a quick look that'll be great too. Thanks!

echopype/convert/set_groups_ek60.py
```diff
@@ -655,18 +654,16 @@ def set_beam(self) -> xr.Dataset:
...
-            # TODO: below needs to be changed to use
-            #  self.convert_obj.ping_data_dict['mode'][ch] == 3
-            #  1 = Power only, 2 = Angle only 3 = Power & Angle
-            # Set angle data if in split beam mode (beam_type == 1)
-            # because single beam mode (beam_type == 0) does not record angle data
-            if self.parser_obj.config_datagram["transceivers"][ch]["beam_type"] == 1:
+            # Save angle data if exist based on values in self.parser_obj.ping_data_dict['mode'][ch]
+            # Assume the mode of all pings are identical
+            # 1 = Power only, 2 = Angle only 3 = Power & Angle
+            if np.all(np.array(self.parser_obj.ping_data_dict["mode"][ch]) != 1):
```
emiliom, line 666: Just for the record, this test (`!= 1`) is not the same thing as what you had in the previous "TODO" comment (`==3`). But I can see that !=1 does cover all cases that have angle data.

echopype/tests/convert/test_convert_ek60.py
```diff
@@ -193,3 +193,16 @@ def test_convert_ek60_duplicate_frequencies(ek60_path):
                        truth_freq_nom_vals, rtol=1e-05, atol=1e-08)
 
     assert np.all(ed['Sonar/Beam_group1'].channel.values == truth_chan_vals)
+
+
+def test_convert_ek60_splitbeam_no_angle(ek60_path):
+    """Convert a file from a split-beam setup but does not record angle data."""
```
emiliom, line 199: Change to this?
```suggestion
    """Convert a file from a split-beam setup that does not record angle data."""
```

emiliom (approved):
Looks good to me. See my very minor comment (no action, really) and minor suggested docstring edit

echopype/convert/set_groups_ek60.py
leewujung: Yes I believe so.

leewujung:
Thanks @emiliom ! I committed your suggestion and once the CI passes I'll merge this.

PR #719, opened 2022-06-05
Proposed reorganization of AD2CP data format

leewujung:
In v0.6.0 we left out AD2CP data because it takes quite a different flavor from all the other sonar models. Now we are in the position to reorganize the data format based on the previous experience.

In this PR I put in a markdown file that details my proposed changes of the data format. The file is put under `docs/source` but my intension is _not_ to merge this file but use this PR as a place where we can do line by line exchanges/discussions if needed.

@emiliom @imranmaj : I think it may work better if @emiliom and I meet first to settle down the proposed changes (which will be become new commits to this PR), and then @imranmaj and I meet to go through the required changes.

@emiliom: I can give you an overview of how this instrument operations and the main differences from the other sonar models. The variables are in such a way that can be grouped into a few conceptual groups, so even though there are a large number of them I think it won't be too hard for us to make decisions.

@imranmaj: There are a few questions that are independent of the format changes. We can discuss those in parallel.

### DIFF
--- a/docs/source/ad2cp_reorg.md
+++ b/docs/source/ad2cp_reorg.md
@@ -16,7 +16,7 @@
 # The original `Beam` group
 
 ## Which mode goes to which group
-Proposal/question: 
+Proposal/question:
 - use specific group name for different mode
 - this means that then some groups may not exist all together, i.e., there won't always be `Beam_group1` in the converted data file
 - this conflicts with what we have for EK80, where there will always be `Beam_group1` but the content changes depending on what are in the file (power/angle samples only, complex samples only, or both)
@@ -170,7 +170,3 @@ Coordinates:
 
 Data variable:
 - `magnetometer_raw`
-
-
-
-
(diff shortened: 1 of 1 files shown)


leewujung:
A new question as I test the new parser/organization from #731 using data from my fieldwork last year:
I think it'll be useful if we can save the entire configuration string as an attribute (or another appropriate form) in the converted data. This will ensure no loss of information, since right now the config info is parsed but not save. 

@imranmaj has a nice function to convert the config string to dict, and it is used in the parser for getting some specific info, but the config params are not saved in general.

@emiliom : Do you have recommendation on how this should be saved? I see this as similar to the config XML that is saved in its entity for EK80 in the `Vendor_specific` group:
https://github.com/OSOceanAcoustics/echopype/blob/12693a72badcf54a18408e694b72e7b0e6d48023/echopype/convert/set_groups_ek80.py#L988

emiliom:
Given that `config_xml` attribute precedent, it seems to me that what you're proposing makes sense. Plus, preserving config info is always a good thing, especially if there are no downsides. You could store the raw text file as is, or convert it to a dict then save it as a JSON string. The only comment I'd make about the format of that dict is that each of its items (eg, item key "BEAMCFGLIST") often has content in the form of a list of dicts, eg:
```python
"BEAMCFGLIST": [dict1, dict2, ...]
```

leewujung:
Given that:
1) Nortek does provide a path to go from `.ad2cp` to `.ntk` to `.nc`
2) we don't have the bandwidth to continue maintaining the AD2CP conversion
3) there are quite a bit to work out on how to store the variables (though IFREMER does use SONAR-netCDF4 to store ADCP data, though I haven't looked into the details)

It seems the best to discontinue AD2CP support, and consider creating a converter to go from Nortek-generated .nc to the echopype echodata format.

I'll close this now, and will follow up with a PR to remove the code and update the docs, and an issue to record the converter idea.

PR #721, opened 2022-06-09
Update pr action to use title [all tests ci]

### LINKED ISSUE #452: Change both CI actions flags to be based on PR titles
[Currently](https://echopype.readthedocs.io/en/stable/contributing.html#continuous-integration-github-actions), the CI can be skipped via the commit string "[skip ci]", and the PR label `Needs Complete Testing` is used to force the running of all CI tests.

For consistent usage and convenience, change this so that both are implemented as PR title strings: "[skip ci]" and "[all tests ci]".

From @lsetiawan: Looks like there's a "title" value that I can grab and check, https://docs.github.com/en/rest/reference/pulls

lsetiawan:
This PR moves triggers for running all tests to look for `[all tests ci]` flag or `[skip ci]` flag rather than just label. This resolves #452.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -11,7 +11,7 @@ jobs:
   test:
     name: ${{ matrix.python-version }}-build
     runs-on: ubuntu-20.04
-    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    if: "!contains(github.event.pull_request.title, '[skip ci]')"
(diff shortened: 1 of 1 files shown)


lsetiawan:
The scope of this PR regarding the title flag trigger works. Right now one of the test is failing due to the crashing of pytest workers that need to be figured out separately. With that said, I'm self-merging this.

emiliom:
Thanks @lsetiawan !

Issue #720, opened 2022-06-09
Empty filter coeffs in EK80 .raw file


martin-wegmann:
Hi everyone,

I am super new to the world of fish monitoring and acoustics, so treat me as an absolute amateur please.
I need your help regarding an issue reading in EK60 data. 

I am using version echopype==0.5.6 and Python 3.7.9 and was recently able to read-in EK60 example data in .raw format. The example data I used to play around with was 68MB large. 

Now I was switching to larger input EK60 data, yet still .raw data, and got an error message while not changing the commands.

I know for a fact that the model is an EK60. The file size is 525 MB. 

`import echopype as ep`
`ifrom echopype import open_raw`
`lexfish_folder="/Volumes/Lexplore/LéXFISH/"`
`ed = open_raw(lexfish_folder+"D20210727-T000128.raw", sonar_model='EK60')`

I got the following message that I didn't get before:

> 11:15:45  parsing file D20210727-T000128.raw, time of first ping: 2021-Jul-27 00:01:28

> KeyError                                  Traceback (most recent call last)
> <ipython-input-3-a3c1c69692e0> in <module>
> ----> 1 ed = open_raw(lexfish_folder+"D20210727-T000128.raw", sonar_model='EK60')

> ~/miniconda2/envs/temprec/lib/python3.7/site-packages/echopype/convert/api.py in open_raw(raw_file, sonar_model, xml_path, convert_params, storage_options)
>     428         output_path=None,
>     429         sonar_model=sonar_model,
> --> 430         params=_set_convert_params(convert_params),
>     431     )
>     432     # Set up echodata object

> ~/miniconda2/envs/temprec/lib/python3.7/site-packages/echopype/convert/set_groups_ek60.py in __init__(self, *args, **kwargs)
>      28         self.old_ping_time = None
>      29         # correct duplicate ping_time
> ---> 30         for ch in self.parser_obj.config_datagram["transceivers"].keys():
>      31             ping_time = self.parser_obj.ping_time[ch]
>      32             _, unique_idx = np.unique(ping_time, return_index=True)

> KeyError: 'transceivers'

Super thankful for any help regarding this issue. 

Cheers,
Martin

leewujung:
[...]

PR #724, opened 2022-06-14
Check filter coeffs existence in `SetGroupsEK80.set_vendor`

leewujung:
This PR fixes the bug that `SetGroupsEK80.set_vendor` attempts to save potentially non-existing filter coefficients in EK80 file. Addresses #720.

### DIFF
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -966,11 +966,14 @@ class SetGroupsEK80(SetGroupsBase):
         decimation_factors = dict()
         for ch in channels:
             # filter coeffs and decimation factor for wide band transceiver (WBT)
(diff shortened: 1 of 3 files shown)


emiliom:
@leewujung I can see that you've added tests for assignments to `coeffs` and `decimation_factors` based on `self.parser_obj.fil_coeffs` and `self.parser_obj.fil_df` not being `None`. I'm not in a position to say by just looking at the code that these tests address the issue found in #720, though it appears that way. I trust that the error found in #720 no longer appears with the test data.

emiliom (commented):
See my comment

leewujung:
@emiliom : I've added a test using the file @martin-wegmann provided and the tests all passed. Should I self-merge?

PR #731, opened 2022-06-15
AD2CP Data Format Reorganization

### LINKED ISSUE #711: Re-organize AD2CP data into different `Beam_groupX` groups
The v0.6.0 change of netCDF data structure did _not_ include the Nortek Signature ADCP (AD2CP) data, because the data is of substantially different format compared to the EK and AZFP echosounders. But now with the experience (and struggle!) with forcing data from the other models into a more compliant form for SONAR-netCDF4 ver.1 convention, I think we can take this step in the upcoming releases.

The main thing is that currently _all_ backscatter and velocity data are saved in a single `Beam` group, and because of it likely result into a large number of NaNs due to the different sampling time and resolution/range setting in the different "modes" the AD2CP supports. In v0.6.0 we moved the original `Beam` (and `Beam_power`) group to be `Sonar/Beam_group1` (and `Sonar/Beam_group2`) for the EK and AZFP echosounders. Taking the same approach, we can save the data from different AD2CP modes into different `Beam_groupX` so that it would be straightforward to find and use.

This issue will serve as the drafting and discussion space for this change.

imranmaj:
Reorganizes the AD2CP data format as was done with other formats for v0.6.0

Runs with datatree v0.0.4 on non-Windows machines (see #732)

Todo:

- [x] Move some variables from beam to vendor
- [x] Move `echosounder_raw` variables from vendor to beam
    - [x] Rename `echosounder_raw` variables
- [x] Change pulse compression to a vector
- [x] Separate beam group into multiple groups
- [X] Rename `sample` to `range_sample` and `sample_transmit` to `transmit_sample`
- [x] Rename `range_sample_X`'s to just `range_sample` within their own beam group
- [x] Rename combined `ping_time` to `time1`
- [x] Rename individual `ping_time_X`s to just `ping_time` within their own beam group
- [x] Move `magnetometer_raw` from platform to vendor
- [x] Add `descr` for each beam group
- [x] Add `long_name`, `comment`, and `units` for every variable
- [x] Update tests

Resolves #711, #733

See also: #719, [link to reorganization details](https://github.com/OSOceanAcoustics/echopype/blob/fc0ab2bbaffd48fc813c2444cf69ed10c35a8548/docs/source/ad2cp_reorg.md)


imranmaj:
It looks like echopype only expects 2 beam groups right now so it's erroring with `rawtest.076.00000.ad2cp` which has 3 (burst, echosounder, and echosounder raw)

leewujung:
> It looks like the EchoData repr only expects 2 beam groups right now so it's erroring with `rawtest.076.00000.ad2cp` which has 3 (burst, echosounder, and echosounder raw)

I think we can have up to 4. Tagging @lsetiawan here: could we change it to have arbitrary number of Beam_groupX?

lsetiawan:
> could we change it to have arbitrary number of Beam_groupX?

Probably, but right now I think we are still limited because of the convention yaml! I think this has to change. In order to do that I think we need to really pull away from the `.group` access and also not use convention yaml for repr. I'll need to look into this further. Should I be making changes in this PR for this or separate?

imranmaj:
@leewujung finally got the tests to pass so this should be ready to go

leewujung:
@imranmaj : thanks! I'll look into this tomorrow morning. I noticed that the requirements are not updated -- I thought updating datatree 0.0.6 is needed for you to run tests on your machine? (We do need users to be able to run on Windows)

imranmaj:
I was able to run the tests by manually editing the datatree library source code on my machine

echopype/convert/api.py
```diff
@@ -168,20 +169,20 @@ def _save_groups_to_file(echodata, output_path, engine, compress=True):
...
+                        "range_sample": DEFAULT_CHUNK_SIZE["range_sample"],
+                        "ping_time": DEFAULT_CHUNK_SIZE["ping_time"],
+                    }
+                ),
+                path=output_path,
+                mode="a",
+                engine=engine,
+                group="Sonar/Beam_group2",
+                compression_settings=COMPRESSION_SETTINGS[engine] if compress else None,
+            )
```
leewujung, line 185: Interesting, so the logic before was actually not correct, and it ran without problem just because `echodata.beam_power` only exist for EK80?

echopype/convert/set_groups_ad2cp.py
```diff
@@ -254,54 +246,166 @@ def set_platform(self) -> xr.Dataset:
...
-                "altimeter_raw_data_sample_distance": "altimeter_raw_data_sample_distance",
-                "altimeter_raw_data_samples": "altimeter_raw_data_samples",
-            }
-        )
-        ds = ds.assign_attrs({"pulse_compressed": self.pulse_compressed})
+        beam_groups = []
+        self._beamgroups = []
+
+        # burst
+        for packet in self.parser_obj.packets:
```
leewujung, line 257: This appears as a for loop but it doesn't actually loop through all packets and the looping through packets actually happens in `_make_dataset`. I wonder if it would be more explicit if you do a smaller loop to check if certain type of packets exist in the file (since the packet types to consider is finite), and then from that use the same `beam_groups.append` and `self._beamgroups.append` sequence to assemble the datasets and group sequence/descriptions.

Also, please change the sequence so that Average is the first before Burst, so the sequence of beam groups (whenever they exist) is: Average - Burst - Echosounder - Echosounder_raw.

leewujung (commented):
@imranmaj : thanks for the extensive changes! I have a couple of comments but only 1 is an actual suggestion/question. I am done with reviewing the code and logics, and will now go on to check the variables/attrs stuff using the test files and files from my fieldwork last summer.


Below are my notes on changes in the code for our later selves:
- In parse_ad2cp.py
  - Removed field_units from Field class def
  - Remove field units from whole bunch of variables in data record format (`HeaderOrDataRecordFormat`)
  - Add unit conversion for the following variables
    - magnetometer_temperature: now added unit conversion
    - ast_quality: now added unit conversion
- In set_groups_ad2cp.py
  - Add attrs to _make_dataset
  - Add logic to put different packet types into different beam groups
  - Construct pulse_compressed as a variable -- in the PR the data variable is called "echogram"

leewujung:
@emiliom : I removed you from the reviewer list -- I'll handle the review and merging of this PR, and you and I can pick up on the remaining variable movements that may have to happen.

echopype/convert/set_groups_ad2cp.py
```diff
@@ -17,20 +20,12 @@
...
-            ),
-        }
-    ]
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.pulse_compressed = self.parser_obj.get_pulse_compressed()
         self._make_time_coords()
+        with resources.open_text(convert, "ad2cp_fields.yaml") as f:
+            self.field_attrs: Dict[str, Dict[str, Dict[str, str]]] = yaml.safe_load(f)  # type: ignore # noqa
```
leewujung, line 28: I noticed that all the `comment` is missing a "." at the end before the sentence "Data was converted to the specified units." Could you put it in? Thanks!

echopype/convert/parse_ad2cp.py
```diff
@@ -100,22 +100,45 @@ class Dimension(Enum):
...
+    MIJ = auto()
+    XYZ = auto()
+    WXYZ = auto()
+
+    def dimension_name(self) -> str:
+        return DIMENSION_NAMES[self]
+
+
+DIMENSION_NAMES = {
+    Dimension.PING_TIME: "time1",
```
leewujung, line 125: Checking the converted results, I found that many variables in the `Beam_groupX` groups (except for `amplitude `, `velocity`,`correlation`) are with coordinate `time1`. These should actually be in `ping_time` since they are recorded with the same times as the pings. The `time1` coordinate name should not exist in `Beam_groupX` (and only in `Platform` or `Environment` groups). Could you change these?

Example see below print out from `rawtest.090.00001.nc`:

Also, what is the `velocity_range` variable in this? I couldn't find correspondence in the data spec, but it exists only in your `BURST_AVERAGE_VERSION2_DATA_RECORD_FORMAT` format and not in version 3.

Screen Shot 2022-06-19 at 12 29 45 PM

leewujung:
@imranmaj : 

A few other things needing changes:

## Status variables in `Vendor_specific` group
The `status` and `status0` variables are stored and not parsed, although your code does that in _postprocess_bitfield in the parser. Could you save these actual status as additional variables also in the `Vendor_specific` group? With a comment that "Derived from `status`" or "Derived from `status0`".

## `Vendor_specific` group variable dimensions/coordinates
Many variable in this group has dimensions `ping_time` AND `time1`. I don't understand why `ping_time` shows up here and it is actually not a coordinate -- these variables seem to have all come from format `BURST_AVERAGE_VERSION3_DATA_RECORD_FORMAT`, so in that case it would have a time dimension that is a combination all transmitted ping times in the average, burst, and echosounder (not raw), so that would be the same as `time1` in the `Environment` and `Platform` group. Is this correct? If so, could you change these?
  
Below is an example of what I see in "rawtest.090.00001.nc":
Screen Shot 2022-06-19 at 1 03 42 PM

## `Environment` group variable dimensions/coordinates
Simliar to the above, in the `Environment` there is an empty/unnecessary `ping_time` dimension:
Screen Shot 2022-06-19 at 1 05 40 PM

leewujung:
@imranmaj : I meant to check the actual values of the number of beams, physical beam used etc in the `Beam_groupX` groups. I'll do that when the `time1` vs `ping_time1` issue is resolved, since it is a bit messy to checked those now.

echopype/convert/api.py
imranmaj: `echodata.beam_power` aliases to the second beam group I think, even with AD2CP. Previously it wouldn't save the 3rd or 4th beam group

imranmaj: That if statement is not necessary for AD2CP because every beam group will be saved in the for loop that I wrote above

imranmaj:
> @imranmaj : I meant to check the actual values of the number of beams, physical beam used etc in the Beam_groupX groups. I'll do that when the time1 vs ping_time1 issue is resolved, since it is a bit messy to checked those now.

@leewujung All of the above issues, including that one, should be fixed now

leewujung:
@imranmaj : Thanks! I'll check in a couple hours on the changes.

echopype/convert/set_groups_ad2cp.py
```diff
@@ -252,79 +261,22 @@ def set_beam(self) -> List[xr.Dataset]:
...
-                        }
-                    )
-                )
+                beam_groups_exist.add(BeamGroup.AVERAGE)
+            elif packet.is_burst():
+                beam_groups_exist.add(BeamGroup.BURST)
+            elif packet.is_echosounder():
+                beam_groups_exist.add(BeamGroup.ECHOSOUNDER)
+            elif packet.is_echosounder_raw():
+                beam_groups_exist.add(BeamGroup.ECHOSOUNDER_RAW)
```
leewujung, line 271: Seems like you can break the look if all 4 possible groups already exist in the `beam_groups_exist` set -- though how much this helps on the speed depends on the actual number of packets in the file.

echopype/convert/set_groups_ad2cp.py
```diff
@@ -417,21 +414,55 @@ def set_beam(self) -> List[xr.Dataset]:
...
+                "ast_data_included": "ast_data_included",
+                "echosounder_data_included": "echosounder_data_included",
+                "ahrs_data_included": "ahrs_data_included",
+                "percentage_good_data_included": "percentage_good_data_included",
+                "std_dev_data_included": "std_dev_data_included",
+                "distance_data_included": "distance_data_included",
+                "figure_of_merit_data_included": "figure_of_merit_data_included",
                 "error": "error",
-                "status": "status",
                 "status0": "status0",
```
leewujung, line 444: here you kept `status0` and removed `status` -- is this intentional? I think it'll be fine to keep both of them, since the comments use that reference.

imranmaj: It's still there a few lines below

echopype/convert/set_groups_ad2cp.py
```diff
@@ -254,68 +252,220 @@ def set_platform(self) -> xr.Dataset:
...
-        if "ping_time" not in ds.dims:
-            ds = ds.expand_dims(dim="ping_time")
+        for i, ds in enumerate(beam_groups):
+            if "ping_time" not in ds.dims:
+                beam_groups[i] = ds.expand_dims(dim="ping_time")
 
-        return set_encodings(ds)
+        # remove time1 from beam groups
+        for i, ds in enumerate(beam_groups):
+            beam_groups[i] = ds.sel(time1=ds["ping_time"])
```
leewujung, line 422: Drop the `time1` coordinate after this selection.

Right now:
Screen Shot 2022-06-19 at 7 48 26 PM

leewujung:
This was embedded so highlighting it here since this variable is still in the beam groups from the latest parsed files.

> Also, what is the velocity_range variable in this? I couldn't find correspondence in the data spec, but it exists only in your BURST_AVERAGE_VERSION2_DATA_RECORD_FORMAT format and not in version 3.`

leewujung:
The `data_set_description` seems to encode different things for different beam groups. It seems that one is parsed and the other is unparsed. I think for `Sonar_group1` and `Sonar_group2` that's because you use the parsed values for the `beam` coordinate -- is this true? If so we need to do something to describe this difference of behavior.

Below again from `rawtest.090.00001.nc`
`Sonar_group1` and `Sonar_group2` look like this
Screen Shot 2022-06-19 at 8 04 24 PM


`Sonar_group3` and `Sonar_group4` look like this:
Screen Shot 2022-06-19 at 8 04 05 PM

echopype/convert/set_groups_ad2cp.py
leewujung: Oops I had missed that. Thanks!

imranmaj:
> I think for Sonar_group1 and Sonar_group2 that's because you use the parsed values for the beam coordinate -- is this true?

Yes, the beam coordinate comes from `data_set_description`

imranmaj:
> what is the velocity_range variable in this?

All the references to it in the spec pretty much just say this:

image

leewujung:
For `velocity_range`, what I meant was that in the data spec it only exists for `BURST_AVERAGE_VERSION2_DATA_RECORD_FORMAT` and not `BURST_AVERAGE_VERSION3_DATA_RECORD_FORMAT`, but I believe the data we use for testing are all from version 3, so what does `velocity_range` exist in the converted data?

leewujung:
For `echosounder_raw_echogram`:
Could you remind what the logic below (setting `echosounder_raw_echogram` in the previous packet to be the current index) is? If you make this simply using the current ping's data, does it work?

https://github.com/OSOceanAcoustics/echopype/blob/efbf9834383441b78892b96beba76077226e04ca/echopype/convert/parse_ad2cp.py#L742-L748

imranmaj:
> If you make this simply using the current ping's data, does it work?

No, `echosounder_index` does not exist in the echosounder raw data record format

imranmaj:
> so why does velocity_range exist in the converted data?

Currently all variables from all formats are always included regardless of whether they exist, I think because we decided that it would be confusing for variables to sometimes be included and sometimes not be included

leewujung:
All tests passed in this major reorganization work, so I'll merge this now. Leftover work captured in #732, #734.

Issue #710, opened 2022-05-27
Add utility function for swapping channel coordinate with frequency_nominal


emiliom:
When there are no duplicate frequencies, replacing the new (0.6.0) `channel` coordinate with the new `frequency_nominal` is a really helpful step (eg, see the [Echopype tour notebook](https://osoceanacoustics.github.io/echopype-examples/echopype_tour.html#compute-mean-volume-backscattering-strength-text-mvbs-and-plot-interactive-echograms-using-hvplot) in echopype-examples)). The steps are brief, but likely hard to figure out for xarray non-experts. Let's create a small utility function that does this. The core will look like this:

```python
def swap_chan(ds: xr.Dataset) -> xr.Dataset:
    """ Function to replace the channel dimension and coordinate with the 
        frequency_nominal variable containing actual frequency values. 
        Note that this step is possible only because there are 
        no duplicated frequencies present.
    """
    return (
        ds
        .set_coords('frequency_nominal')
        .swap_dims({"channel": "frequency_nominal"})
        .reset_coords('channel')
    )
```

But it should first test for duplicate frequencies and return an error if duplicates are found.

leewujung:
I think this can be closed now. I'll move the function to `preprocess` from `utils` in a separate PR before v0.6.1.

PR #735, opened 2022-06-21
Function to use `frequency_nominal` in place of `channel` for dimension/coordinate

leewujung:
Addresses #710.

## Tasks
- [x] function prototype
- [x] tests: on Sv and MVBS datasets

### DIFF
--- a/echopype/utils/common.py
+++ b/echopype/utils/common.py
@@ -25,10 +25,9 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
     # Only possible if no duplicated frequencies
     if np.unique(ds["frequency_nominal"]).size == ds["frequency_nominal"].size:
         return (
-            ds
-            .set_coords('frequency_nominal')
+            ds.set_coords("frequency_nominal")
             .swap_dims({"channel": "frequency_nominal"})
-            .reset_coords('channel')
+            .reset_coords("channel")
         )
     else:
         raise ValueError(
--- a/echopype/utils/common.py
+++ b/echopype/utils/common.py
@@ -1,0 +1,37 @@
+import numpy as np
+import xarray as xr
+
+
+def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
+    """
+    Use frequency_nominal in place of channel to be dataset dimension and coorindate.
+
+    This is useful because the nominal transducer frequencies are commonly used to
+    refer to data collected from a specific transducer.
+
+    Parameters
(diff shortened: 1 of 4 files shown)


echopype/utils/common.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+import xarray as xr
+
+
+def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
+    """
+    Use frequency_nominal in place of channel to be dataset dimension and coorindate.
```
emiliom, line 7:
```suggestion
    Use frequency_nominal in place of channel as dataset dimension and coordinate.
```

emiliom:
The function looks good. Thanks!

leewujung:
@emiliom : I added the tests and think this PR is ready to be merged. Let me know what you think.

In the latest commit I also modified the __init__ to allow using this function like the following:
```python
Sv_swap = ep.utils.swap_dims_channel_frequency(Sv)
```

We should definitely revisit putting this function in `./utils/common.py` and the import pattern when we discuss the processing chain.

emiliom:
In the test battery, I don't see what's gained by testing many types of raw data cases. Ultimately, the function `swap_dims_channel_frequency` operates only on `Sv` (and `MVBS`). The form of the resulting `Sv`, after conversion and calibration, should be identical regardless of the source raw data type; if it's not, the test responsible for catching such errors is `test_calibrate.py`, not this one.

I would pare it down to just one or two raw files.

Also, I don't see a specific test for the case of duplicate frequencies. It could be that one of the test files does have duplicated frequencies, but then `_check_swap` should fail. I'd suggest adding a new boolean param value that indicates if the raw file has duplicated frequencies or not, then using that param value to test for the expected outcome of `swap_dims_channel_frequency`.

I'll try to provide a modified version of the test incorporating my suggestions, in the next couple of days.

emiliom:
I've added a PR (https://github.com/leewujung/echopype/pull/5) to your PR, with the changes to the tests that I discussed above. Take a look.

leewujung:
> The form of the resulting `Sv`, after conversion and calibration, should be identical regardless of the source raw data type; if it's not, the test responsible for catching such errors is `test_calibrate.py`, not this one.

I agree -- it is also true that we should add a test to test for consistency of the `compute_Sv/TS` outputs. We currently only have tests to make sure the calibration routines are run through, and in a few limited cases the calibrated output values. I'll make a new issue for this.

emiliom (approved):

Issue #662, opened 2022-05-04
Generalize MVBS computation


leewujung:
The operation in `compute_MVBS` is basically to compute a 2D binned average of Sv along `echo_range` and `ping_time`. Note that the average should be done in the linear domain (s<sub>V</sub>) and converted back to the log domain (S<sub>V</sub>).

Currently error is raised when `echo_range` is different across `ping_time`, so that we can force the output `echo_range` to be the same for all pings. We should generalize it to compute this even if `echo_range` has different max value (related to the length of dimension `range_sample`) and step size (related to `sample_interval`).

leewujung:
This is now addressed in #878.

PR #736, opened 2022-06-21
Fix regression bug in computing MVBS in selecting `echo_range` for specific frequency

leewujung:
This PR fixes a few small things in related to `compute_MVBS` operation. Does _not_ address #662 😕

### Tasks
- [x] fix error that use `echo_range` from the first channel for all channels: this is a bug from the process of converting the data to be aligned with dimension `frequency` to `channel` in v0.6.0
- [x] re-orders the dimensions for `echo_range` in the AZFP Sv output to be consistent with data variable `Sv`
- [x] fix typing error: input `range_meter_bin` currently listed as int, but should be general as float

### DIFF
--- a/echopype/calibrate/calibrate_azfp.py
+++ b/echopype/calibrate/calibrate_azfp.py
@@ -154,6 +154,9 @@ class CalibrateAZFP(CalibrateBase):
         # Add env and cal parameters
         out = self._add_params_to_output(out)
 
+        # Order the dimensions
+        out["echo_range"] = out["echo_range"].transpose("channel", "ping_time", "range_sample")
+
         # Squeeze out the beam dim
         # doing it here because both out and self.cal_params["equivalent_beam_angle"] has beam dim
         return out.squeeze("beam", drop=True)
--- a/echopype/preprocess/api.py
+++ b/echopype/preprocess/api.py
@@ -86,9 +86,8 @@ def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
             .mean()
         )
         sv_groupby_bins.coords["echo_range"] = (["echo_range_bins"], rint[:-1])
-        sv_groupby_bins = (
-            sv_groupby_bins.swap_dims({"echo_range_bins": "echo_range"})
-            .drop_vars("echo_range_bins")
+        sv_groupby_bins = sv_groupby_bins.swap_dims({"echo_range_bins": "echo_range"}).drop_vars(
+            "echo_range_bins"
         )
         return 10 * np.log10(sv_groupby_bins)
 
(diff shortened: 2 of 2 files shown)


leewujung:
Should include a typing error fix in the input `range_meter_bin`. It is currently listed as `int`, but should be general as `float`.

echopype/calibrate/calibrate_azfp.py
```diff
@@ -154,6 +154,9 @@ def _cal_power(self, cal_type, **kwargs):
         # Add env and cal parameters
         out = self._add_params_to_output(out)
 
+        # Order the dimensions
+        out["echo_range"] = out["echo_range"].transpose("channel", "ping_time", "range_sample")
+
```
b-reyes, line 159: @leewujung I am wondering if this is the most appropriately place for this. This line is necessary because of how we create the EchoData object for AZFP, correct? If so, should we instead correct the order in `set_groups_azfp`? I am fine with leaving this as is, I just thought I would bring this up.

leewujung: Good point, I agree this is the best done in `set_groups_azfp`. In fact this was part of my test to see if we can do without `.drop("channel")`  in this line:
https://github.com/OSOceanAcoustics/echopype/blob/b6205e57a13fea2e50801f099ea4cea4c38855da/echopype/preprocess/api.py#L79

Somehow that extra dimension/coordinate only shows up for AZFP data, and I was wondering if it had anything to do with the order of the dimension, but that wasn't the reason.

I'll create an issue so we remember to do this in `set_groups_azfp` instead of here.

b-reyes: Sounds good!

leewujung: Oh actually, now that I looked at this again, the order of dimension to be changed is for `echo_range` and not `Sv` (which came from `backscatter_r`), so I guess nothing has to change for `set_groups_azfp`, but it is the AZFP part of `compute_range` that needs to change.

b-reyes (approved):
Everything looks good to me.

leewujung:
Thanks @b-reyes ! I'll merge now!

echopype/calibrate/calibrate_azfp.py
b-reyes: I was having a little trouble tracking down where that happened. The AZFP part of `compute_range` was using parts of the EchoData object, so I assumed that we needed to make corrections in `set_groups_azfp`. After looking into this a little further, I found that in the `EK` else statement of `compute_range` we do `range_meter = range_meter.transpose("channel", "ping_time", "range_sample")`, so maybe the fix is as simple as adding that to the AZFP if statement block.

leewujung: Oh right, cool that you tracked it down! I'll link your comment to that issue... v0.6.2... so that we don't have to do this now. :P

Issue #578, opened 2022-03-11
Add mechanism to convert `echo_range` to `depth` in calibrated datasets


leewujung:
This is derived from #259.

We decided that it is more flexible to have the `echo_range` and ~`range_offset`~ `water_level` to be separate in the calibrated dataset (the Sv). This is addressed in #579 .

However, it is necessary to have a mechanism for users to convert and combine these values into "depth." For example:
- for data from shipboard echosounders, `depth = echo_range + water_level`, where `water_level` can the depth of the transducer, or depth of the transducer plus some starting lag when the instrument started collecting data.
- for data from upward looking echosounders, `depth = sonar_range * cos(theta) + water_level `, where `theta` is some tilt angle of the instrument that may or may not be 0.

In these examples there's also the upside down flip depending on whether it is in a downward- or upward-looking configuration.

So, we need to figure out a mechanism to make it easy for users to get things into to the depth coordinate, something like a function with options to select pointing upward or downward, and whether there is a deviation from vertical, etc. To be discussed!

emiliom:
Note: PR #577 was closed and replaced by #579

valentina-s:
I see the value in the generality of keeping the `echo_range` and `range_offset` separate. However, I would like to see how one can envision a whole pipeline for processing files. Many analysis/science operations will require only `depth, channel, ping_time` (it might be interesting though also to brainstorm the cases when people need the offset/tilt). So, in this case if people convert bunch of files do they store them in the general 'Sv' format or they calculate the depth and store them with a depth dimension? If they do the first, each time they access the file they will need to do this conversion operation to plot the data or do something else. It can be facilitated with a function, but then users will need to install `echopype` just for that function, while in general a lot of follow up analysis can happen on Sv `.nc` file just with `xarray`.  If they directly store it with the `depth` dimension then then store it in a "non-standard Sv" format? I wonder if a general `Sv` standard can allow for one or the other one, with offset/tilt stored to convert between them. So they both are considered "compliant".

leewujung:
[...]

PR #738, opened 2022-06-22
Add `depth` as an additional variable in Sv dataset

leewujung:
This addresses #578.

The current prototype defaults to assuming that `water_level` exists in the dataset and that the transducers are mounted vertically, pointing downward, like in a ship-based survey.

Users have the option to set `water_level` and `tilt` via input directly, since many times these data are not in the data files generated by the instruments.

@valentina-s @emiliom : I'd appreciate comments on whether we should replace the original `echo_range` variable or keep `depth`. These 2 variables are of the same dimension as the echo data (e.g., `backscatter_r`) so it would blow up the size of the dataset -- maybe we should replace it?  Either way we'd need to be careful on the provenance to record the operations.

**Updates:**
- ~Note: I am keeping this as a draft for now because it builds on top of #735 and somehow the CI is failing repeated not due to fault of the tests but some bucket creation problems.~
- Due to the addition of the `consolidate` subpackage, this PR also includes:
  - removing a redundant `swap_dims_channel_frequency` previously in `preprocess`
  - moving `add_depth` and `swap_dims_channel_frequency` to the `consolidate` subpackage

### Tasks
- [x] add function
- [x] add tests


leewujung:
One thing that I think can be expanded on this function is to allow `tilt` (or ds["tilt"] ) to be a DataArray instead of a scalar, but I think we can add that expansion in a separate PR and get this simplistic but useful function in first.

echopype/consolidate/api.py
```diff
@@ -7,6 +7,111 @@
...
+    Create a depth data variable based on data in Sv dataset.
+
+    The depth is generated based on whether the transducers are mounted vertically
+    or with a polar angle to vertical, and whether the transducers were pointed
+    up or down.
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        Dataset for which the dimension will be swapped.
```
emiliom, line 61: This definition likely stems from an earlier version of what `add_depth` would do. Currently it doesn't swap dimensions
```suggestion
        Source Sv dataset to which a depth variable will be added.
```

echopype/consolidate/api.py
```diff
@@ -7,6 +7,111 @@
...
+        Must contain `echo_range`.
+    vertical : bool
+        Whether or not the transducers are mounted vertically.
+        Default to True.
+    downward : bool
+        Whether or not the transducers point downward.
+        Default to True.
+    tilt : float
+        Tilt angle (degree) if transducers are not mounted vertically (when `vertical=False`).
+        Required if `vertical=False` and `tilt` does not exist in the dataset.
```
emiliom, line 71: As we know, I'm less of a fan of redundant descriptions than you are :laughing: . So, here's a leaner description, for your consideration:
```suggestion
        Tilt angle (degree). Required if transducers are not mounted vertically
        (`vertical=False`) and `tilt` does not exist in the dataset.
```

echopype/consolidate/api.py
```diff
@@ -7,6 +7,111 @@
...
+        tilt = 0
+
+    # Multiplication factor depending on if transducers are pointing downward
+    if downward:
+        mult = 1  # no flip
+    else:
+        mult = -1  # flip upside down (closer to transducer is deeper)
+
+    # Compute depth
+    ds["depth"] = mult * ds["echo_range"] * np.cos(tilt / 180 * np.pi) + water_level
```
emiliom, line 110: Trivial tweak. I prefer to avoid multiplications after a denominator, for clarity.
```suggestion
    ds["depth"] = mult * ds["echo_range"] * np.cos(np.pi * tilt / 180) + water_level
```

emiliom: Also, some of the inherited attributes of the new depth variable should be changed:
- Change `long_name` to, say, "Depth"
- Add a `standard_name` with value "depth"

emiliom:
Regarding the interpretation of `water_level` and the depth calculation `ds["depth"] = mult * ds["echo_range"] * np.cos(tilt / 180 * np.pi) + water_level`: I believe (but haven't looked it up in the SONAR-netcdf4 manual) that we learned that your likely interpretation of `water_level` (depth below the water surface, right?) is not quite what's in the standard and nominally in echopype since 0.6.0. In the standard, `water_level` is the "z-axis distance from the platform coordinate system origin to the sonar transducer", and it requires `vertical_offset` ("Platform vertical offset from nominal") in order to calculate the actual depth below water surface.

In practice, if I remember correctly from our 0.6.0 development, `vertical_offset` is often not available and all that's available is a `water_level` variable that may or may not be heave. So, in practice, let's not worry about it for now.

A more important issue is that since 0.6.0, `water_level` is *not* a scalar. It's a vector with dimensions that don't match the `echo_range` dimensions. I *think* this vector is typically a constant, propagated value. So, an easy and cheap approach for this initial implementation could be to take the first `water_level` value and use it as a scalar.

emiliom:
Regarding `tilt`, that's a variable that's not found in the SONAR-netcdf4 v1 convention. I remember us discussing it, and I thought I had seen it in one of the OOI moored echosounders (the AZFP's, right?), but it's not in the EK60 example in https://osoceanacoustics.github.io/echopype-examples/ms_OOI_EK60_mooringtimeseries.html. Can tilt be calculated from other Platform orientation variables found in the standard, such that it could be calculated for that OOI EK60 example?

Assuming it *can* be calculated, we'd have the same issue as with `water_level`, of having to convert from a vector to a single scalar value. Again, things that could be deferred to the next release.

echopype/tests/consolidate/test_consolidate.py
```diff
@@ -1,6 +1,162 @@
...
+    sample_interval = 0.01
+    ds_Sv = _build_ds_Sv(channel, range_sample, ping_time, sample_interval)
+
+    # no water_level in ds
+    try:
+        ds_Sv_depth = ep.consolidate.add_depth(ds_Sv)
+    except ValueError:
+        ...
+
+    # user input water_level
```
emiliom, line 149: Just to be more explicit, maybe add a comment:
```
# Use defaults `vertical=True` and `downward=True`

emiliom:
I'm done reviewing this. I also ran `test_consolidate.py` locally, and all was fine. `add_depth` is really simple! That's great.

See my various comments. Most of them are minor or academic comments, except the ones about the expectation that the `water_level` dataarray (and `tilt`, but that's a rare case) found in the Sv dataset is a scalar. Your `test_add_depth` tests don't currently include a case with an Sv dataset that does have a `water_level` variable (as generated by `open_raw`), so that case is not covered.

echopype/consolidate/api.py
leewujung: Haha, sure!

echopype/consolidate/api.py
leewujung: > I prefer to avoid multiplications after a denominator, for clarity.

I can see the advantage, but here because the conversion is from degree to radians, breaking the sequence seems weird.

leewujung:
Alright, I made the following updates:
- simplified the function to only allow scalar inputs
- added notes to explain future expansion
- removed the unnecessary boolean option `vertical` -- since `tilt` says it all on whether or not something is vertical or not!

echopype/tests/consolidate/test_consolidate.py
leewujung: I removed the option `vertical`

echopype/consolidate/api.py
leewujung: attributes added!

leewujung:
I'll self-merge and get things ready for release.

Issue #740, opened 2022-06-25
Enhance update_platform to support more use cases and produce more consistent results


emiliom:
The use case that drove the development and testing of `update_platform` was a single set of Saildrone data files from the Pacific Hake survey, with one EK80 raw file with no Platform data at all and an associated external data file.

I illustrate and document some of the issues in [this notebook](https://gist.github.com/emiliom/9cddce2417713208f40800dc5ba0abb1).

### General comments about the current implementation of `update_platform`

- It fails with the use case of a single fixed location created as an xarray Dataset on the fly with no other variables
- It handles `latitude`, `longitude`, `pitch`, `roll`, `vertical_offset` and `water_level`
- It currently adds a `history` attribute only to `time1.history`. Here's an example: "2022-06-25 21:32:53.788634 +00:00. Added from external platform data". I think this same attribute string should also be added to all variables inserted by `update_platform`
- `time1` is used for *all* inserted variables. It is the only dimension assigned to these variables. That pattern predates echopype 0.6.0, and for variables other than `latitude` and `longitude` the resulting Platform group ends up being inconsistent with a Platform group generated using `open_raw` where the platform data are found in the raw file. Specifically, echopype 0.6.0 specifies the following dimensionality:
    ```
    pitch: (channel, time2)
    roll: (channel, time2),
    vertical_offset: (channel, time2)
    water_level: (channel, time3)
    ```
- It sets all missing target variables to nan except `water_level`, which is set to 0. I didn't find any valid reason for the `water_level` assignment of 0, and I believe the only reason was to prevent the failure of `tests/echodata/test_echodata.py::test_update_platform` as currently implemented.

### Some open questions about the *fixed-location use case*, such as from a mooring

- What should the timestamp on the location be? eg, the first or last `ping_time` value? When using a single timestamp (`time1` will have a length of 1), the assumption is that it applies to the entire dataset (all ping times).
- Would it be clearer to replicate the fixed location across more than one timestamp? For example, the first and last `ping_time`?

### Issues unrelated to `update_platform` per se

Comparing the outcome of `update_platform` for AZFP and EK60 raw files that don't have the variables handled by `update_platform` but do have other Platform variables exposed some remaining differences in how variables with empty values are being handled by the different `set_groups_*` classes. These differences are illustrated in my [sample notebook](https://gist.github.com/emiliom/9cddce2417713208f40800dc5ba0abb1), and they involve issues such as:
[...]

PR #741, opened 2022-06-25
Enhance `update_platform` to support more use cases, add more consistency

emiliom:
Starts to address #740. Initial updates only:
- Enables the use case of adding a simple lat-lon point as the external platform data. Previously `update_platform` failed with that use case.
- Attributes for the target variables are now set from `sonarnetcdf_1.yaml_dict["variable_and_varattributes"]["platform_var_default"]`
- Made some simplifications and generalizations in `update_platform`
- Previously `update_platform` set all missing target variables to nan except `water_level`, which was set to 0. I didn't find any reason for it, so I changed it to set it to nan like all other variables (that made the code cleaner). That broke `tests/echodata/test_echodata.py::test_update_platform`, but for bogus reasons. I fixed the test.

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -642,11 +642,12 @@ class EchoData:
                 }
             )
 
-        convention_varattrs = sonarnetcdf_1.yaml_dict["variable_and_varattributes"]
         for var in dropped_vars_target:
-            platform[var] = platform[var].assign_attrs(
-                **convention_varattrs["platform_var_default"][var]
-            )
+            var_attrs = self._varattrs["platform_var_default"][var]
+            # Insert history attr only if the variable was inserted with valid values
+            if not platform[var].isnull().all():
+                var_attrs["history"] = history_attr
+            platform[var] = platform[var].assign_attrs(**var_attrs)
 
         self.platform = set_encodings(platform)
 
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -586,7 +586,6 @@ class EchoData:
         )
 
         platform = self.platform
-        platform_vars_attrs = {var: platform[var].attrs for var in platform.variables}
         platform = platform.drop_dims(["time1"], errors="ignore")
         # drop_dims is also dropping latitude, longitude and sentence_type why?
         platform = platform.assign_coords(time1=extra_platform_data[time_dim].values)
@@ -599,14 +598,16 @@ class EchoData:
         }
         platform["time1"] = platform["time1"].assign_attrs(**time1_attrs)
 
-        dropped_vars_target = [
-            "pitch",
-            "roll",
-            "vertical_offset",
-            "latitude",
-            "longitude",
-            "water_level",
-        ]
+        platform_vars_sourcenames = {
+            "pitch": ["pitch", "PITCH"],
+            "roll": ["roll", "ROLL"],
+            "vertical_offset": ["heave", "HEAVE", "vertical_offset", "VERTICAL_OFFSET"],
+            "latitude": ["lat", "latitude", "LATITUDE"],
+            "longitude": ["lon", "longitude", "LONGITUDE"],
+            "water_level": ["water_level", "WATER_LEVEL"],
+        }
+
+        dropped_vars_target = platform_vars_sourcenames.keys()
         dropped_vars = []
         for var in dropped_vars_target:
             if var in platform and (~platform[var].isnull()).all():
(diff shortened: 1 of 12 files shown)


echopype/echodata/echodata.py
```diff
@@ -620,73 +621,32 @@ def update_platform(
...
-                    mapping_search_variable(
-                        extra_platform_data,
-                        ["water_level", "WATER_LEVEL"],
-                        default=platform.get("water_level", np.zeros(num_obs)),
-                    ),
-                ),
-            }
-        )
+        num_obs = len(extra_platform_data[time_dim])
+        for platform_var, sourcenames in platform_vars_sourcenames.items():
```
leewujung, line 631: This is so much cleaner! 🥳

leewujung:
I found it very bizarre that my changes from the already merged PR #714 show up in this PR, even though the target is `dev`. Also had a couple of comments on the initial changes.

emiliom:
> I found it very bizarre that my changes from the already merged PR https://github.com/OSOceanAcoustics/echopype/pull/714 show up in this PR, even though the target is dev. Also had a couple of comments on the initial changes.

Sigh. I have no idea why that's showing up. My PR originally contained just two changed files, `echodata.py` and `test_echodata.py`. Sometimes git and github are mysterious.

echopype/tests/echodata/test_echodata.py
```diff
@@ -472,7 +472,7 @@ def test_update_platform(update_platform_samples):
 
     ed = echopype.open_raw(raw_file, "EK80")
 
-    updated = ["pitch", "roll", "latitude", "longitude", "water_level"]
+    updated = ["pitch", "roll", "latitude", "longitude"]
     for variable in updated:
         assert np.isnan(ed.platform[variable].values).all()
```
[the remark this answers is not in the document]
leewujung: Ah, I see. Thanks for the explanation!

leewujung (approved):

emiliom:
Thanks for your review. I suggest we keep this PR unmerged for now, until we decide if anything else from #740 should make it into 0.6.1; for example, adding variable `history` attributes should be easy.

One other thing we (I?) could *probably* add in time for 0.6.1 is a test for the fixed-location use case.

emiliom:
This PR is now "clean" of the git/github mess! It no longer includes the extraneous files from a different PR, and it's back at just the two files being targeted.

emiliom:
Added commit that adds `history` attribute to variables that are inserted

emiliom:
Added commit that adds a test for the fixed-location use case. To do that, I overhauled the `test_update_platform` test, including the pytest fixture/parameters associated with it.

emiliom:
@leewujung could you take one last look at this PR, now that I've added:
- `history` attribute to variables that are inserted
-  A test for the fixed-location use case. To do that, I overhauled the `test_update_platform` test, including the pytest fixture/parameters associated with it.

I'm done with the enhancements I proposed for 0.6.1.

echopype/tests/echodata/test_echodata.py
```diff
@@ -466,17 +440,68 @@ def test_nan_range_entries(range_check_files):
...
+                        "saildrone",
+                        "SD2019_WCS_v05-Phase0-D20190617-T125959-0.raw",
+                ),
+                (
+                        "saildrone",
+                        "saildrone-gen_5-fisheries-acoustics-code-sprint-sd1039-20190617T130000-20190618T125959-1_hz-v1.1595357449818.nc",  #noqa
+                ),
+        ),
+        (
+                "fixedlocation",
```
leewujung, line 461: I just have a super minor suggestion here: `fixedlocation` --> `fixed-location` to be consistent with `external-trajectory`

leewujung, line 461:
```suggestion
                "fixed-location",
```
to be consistent with `external-trajectory`

echopype/tests/echodata/test_echodata.py
```diff
@@ -466,17 +440,68 @@ def test_nan_range_entries(range_check_files):
...
-
-    ed = echopype.open_raw(raw_file, "EK80")
+@pytest.mark.parametrize(
+    ["ext_type", "sonar_model", "updated", "path_model", "raw_path", "platform_data"],
+    [
+        (
+                "external-trajectory",
+                "EK80",
+                ("pitch", "roll", "longitude", "latitude"),
+                "EK80",
```
leewujung, line 450: It seems that the entire block is shifted 1 tab more than needed to the right. Not crucial at all, just the line overflows my screen.

echopype/tests/echodata/test_echodata.py
```diff
@@ -496,7 +521,7 @@ def test_update_platform(update_platform_samples):
         np.count_nonzero(
             ed.platform["time1"] < ed.beam["ping_time"].min()
         )
-        == 1
+        <= 1
```
leewujung, line 524: This is tied to this particular data file and not generally true, right?

leewujung (approved):
@emiliom : the changes look good. I like the new tests that are very specific on the use cases. I just have one question in the last comment. Otherwise this seems ready to go!

echopype/tests/echodata/test_echodata.py
```diff
@@ -1,3 +1,4 @@
+from pathlib import Path
```
leewujung, line 1: I think this is redundant.

echopype/tests/echodata/test_echodata.py
```diff
@@ -466,17 +440,68 @@ def test_nan_range_entries(range_check_files):
...
     for variable in updated:
         assert np.isnan(ed.platform[variable].values).all()
 
-    extra_platform_data = xr.open_dataset(extra_platform_data_file)
+    if ext_type == "external-trajectory":
+        extra_platform_data_file_name = platform_data[1]
+        extra_platform_data = xr.open_dataset(
+            test_path[path_model] / platform_data[0] / extra_platform_data_file_name
+        )
+    elif ext_type == "fixedlocation":
```
emiliom, line 493:
```suggestion
    elif ext_type == "fixed-location":
```

echopype/tests/echodata/test_echodata.py
emiliom: Yeah, I saw that. I think PyCharm did it. I figured pre-commit would've fixed it, but it looks like it didn't.

echopype/tests/echodata/test_echodata.py
emiliom: Right! It's not used. I'll remove it.

emiliom, line 1:
```suggestion
```

Issue #743, opened 2022-06-28
Errors/issues processing sequentially pinged multifrequency broadband EK80 files


brandynlucca:
Hi all. I didn't see this in either of the current open or closed issues, but I am running into a data processing issue with Simrad *.raw files containing broadband multiple frequency data. I followed the threads and chain of issues/pulls in both [#308](https://github.com/OSOceanAcoustics/echopype/issues/308#issue-870570268) and [#657](https://github.com/OSOceanAcoustics/echopype/pull/657), but those didn't seem to directly help address the issues I was encountering. I apologize if I missed a thread that does, in fact, address this exact issue. Namely, some of the issues addressed between 0.5.6 and 0.6.0 didn't appear to resolve errors involving changing data values for `transmit_power`, `slope`, `transmit_duration_nominal`, `frequency_start`, and `frequency_end`.

**Background**

**OS:** Windows 10.0.19043
**Python version:** 3.7.9 _and_ 3.8.10
**Echopype version:** 0.5.6 _and_ 0.6.0
**Simrad EK80 software version:** 2.0.0

```
ed["Provenance"]

<xarray.Dataset>
Dimensions:           (filenames: 1)
Dimensions without coordinates: filenames
Data variables:
    source_filenames  (filenames) <U57 'C:/Users/Brandyn/Downloads/NYOS2105-D...
Attributes:
    conversion_software_name:     echopype
    conversion_software_version:  0.6.0
    conversion_time:              2022-06-28T18:22:15Z
```

An aside: I ran `Echopype 0.5.6` in `Python 3.7.9` and `Echopype 0.6.0` in `Python 3.8.10`. For `Python 3.7.x`, there is an error in trying to install `Echopype 0.6.0` due to certain dependencies requiring `Python >= 3.8` (namely `xarray`). So the recommended Python version in the [Installation Guide](https://echopype.readthedocs.io/en/stable/installation.html) may need to be amended for Windows. This was detected via: 

`pip install git+https://github.com/OSOceanAcoustics/echopype.git#egg=echopype`

**Issue**
[...]

PR #717, opened 2022-06-04
Add sphinx-panels to docs/requirements.txt 

emiliom:
Add sphinx-panels to overcome new failure in readthedoc build

PR #744, opened 2022-06-29
Fix python version requirements in docs

leewujung:
This PR addresses the python version requirement discrepancy reported in #743 .

I also took this chance to address #717 for the `stable` branch. (Both `dev` and `main` have been updated.)

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,6 @@
 sphinx_rtd_theme
 sphinx-automodapi
+sphinx-panels
 sphinxcontrib-mermaid
 jupyter-book
 numpydoc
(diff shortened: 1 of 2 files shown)


emiliom:
You're trying to merge this into `stable`, right? But you selected `main`.

Also, I'm confused why it appears as if `sphinx-panels` is not present in `main`, when it clearly is:
https://github.com/OSOceanAcoustics/echopype/blob/12693a72badcf54a18408e694b72e7b0e6d48023/docs/requirements.txt#L3

But I guess that will be irrelevant once you change it to `stable`. `sphinx-panels` is definitely missing from `stable`.

Issue #725, opened 2022-06-15
Add mechanism to interpolate lat/lon to Sv data


leewujung:
Currently the GPS lat/lon data exist only in the converted raw file under the `Platform` group. However, for many science use cases, the calibrated Sv dataset should have the lat/lon positions of each of the pings attached. We should create a mechanism for this as we move data through the processing chain/pipeline.

leewujung:
This is addressed in #749.

PR #749, opened 2022-07-01
Add function to interpolate location to calibrated dataset

leewujung:
This addresses #725 .

### Tasks
- [x] add function to `utils.common` (temporary position until we decide where these higher-product-level utilities would go)
- [ ] add tests to Sv and MVBS data


leewujung:
@lsetiawan : When I added the typing of EchoData somehow I am getting circular import error!  Do you have ideas on where that may come from? I thought I have exactly the same thing (`from ..echodata import EchoData`)  as what we had in `calibrate/api.py`. Help!

echopype/utils/common.py
```diff
@@ -34,3 +39,50 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
...
+        return x_var.interp(time1=ds["ping_time"])  # time1 is always associated with location data
+
+    # Most attritbues are attached automatically via interpolation
+    # here we add the history
+    ds_copy = ds.copy()
+    ds_copy["latitude"] = sel_interp("latitude")
+    ds_copy["longitude"] = sel_interp("longitude")
+    history = (
+        f"{datetime.datetime.utcnow()} +00:00. " "Interpolated from Platform latitude/longitude."
+    )
```
leewujung, line 84: @emiliom : is this appropriate (for now)?

lsetiawan:
> When I added the typing of EchoData somehow I am getting circular import error! Do you have ideas on where that may come from? I thought I have exactly the same thing (from ..echodata import EchoData) as what we had in calibrate/api.py. Help!

Since common is imported under the utils `__init__.py` as such:
https://github.com/OSOceanAcoustics/echopype/blob/27a9b8266edd2da2e443783d58252088ea3002f9/echopype/utils/__init__.py#L1-L5

within `echodata.py` we have a utils import so it will try to import the function from common, which creates the circular import:
https://github.com/OSOceanAcoustics/echopype/blob/27a9b8266edd2da2e443783d58252088ea3002f9/echopype/echodata/echodata.py#L18

This is the issue. Probably shouldn't be importing EchoData class within utils common since, it is a "common" module that can be used in any of the other modules. Best to reserve this to module to not pull from other modules.

echopype/utils/common.py
```diff
@@ -34,3 +39,50 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
             "Duplicated transducer nominal frequencies exist in the file. "
             "Operation is not valid."
         )
+
+
+def add_location(ds: xr.Dataset, echodata: EchoData = None, nmea_sentence: Optional[str] = None):
```
lsetiawan, line 44: Should this be in the `preprocess` module instead? After all it enhances the spatial coherence of the data? 😄 

https://github.com/OSOceanAcoustics/echopype/blob/12693a72badcf54a18408e694b72e7b0e6d48023/echopype/preprocess/api.py#L2

leewujung: Oh that is very true! I agree with this and will move it.

leewujung:
> Probably shouldn't be importing EchoData class within utils common since, it is a "common" module that can be used in any of the other modules. Best to reserve this to module to not pull from other modules.

I agree with this! I wonder where these functions can go (swapping freq/channel, adding lat/lon, adding depth, etc -- basically higher level processing functions). I sticked `common.py` under `utils` because we hadn't discussed this, but now with this circular import/logic issue, we should move it out to something that is more sensible. Maybe these belong in `preprocess` also as you pointed out above? `preprocess.data_coherence` (just made up the name, in my mind these are functions that bring coherence across different "types" of data into the Sv dataset)

lsetiawan:
> I agree with this! I wonder where these functions can go (swapping freq/channel, adding lat/lon, adding depth, etc -- basically higher level processing functions).

I think higher level functions that can users will/can use should be in the `preprocess` module since it is a "preprocessing" step. I think of the `utils` module more for pure functions that we need for our purposes for either I/O or data formatting utilities more that information manipulations (processing).

echopype/utils/common.py
leewujung: I guess I had forgotten that I've written those sentences...

emiliom:
I'm submitting a PR (https://github.com/leewujung/echopype/pull/6) to your PR for your evaluation, with the following changes:
- Support for fixed-location case (a single lat-lon coordinate in the `Platform` group), by propagating that constant lat-lon pair along the ping_time dimension
- Add tests for missing or all-nan `Platform` coordinate variables
- Minor cosmetic tweaks

See what you think. Otherwise, I think it's ready to go. FYI I did a manual, visual evaluation of the lat-lon interpolation, comparing the results against the original lat-lon data. It looks good.

leewujung:
> Add tests for missing or all-nan Platform coordinate variables

This is kinda of a confusion/misnomer: what's added was the mechanism to test for missing or all-nan Platform coordinate variables, but not "tests" for them. We still need tests for these.

I will open up a new issue for completing the more comprehensive tests, and delegate that to someone, so that I can focus on adding the actual functions.

PR #753, opened 2022-07-04
Fix `compute_MVBS` bugs with NaN entries

leewujung:
This PR:
- fixes `compute_MVBS` bugs re different variations of NaN entries in the Sv dataset
  - empty values in the beginning of each ping for BB Sv
  - empty values at the end of each ping for channels with "shorter" pings
  - non-uniform accidental empty values only in some pings (this seems an instrument problem and only happens occasionally)
  - all-NaN pings due to alternating pinging sequence
- moves `swap_dims_channel_frequency` to under `preprocess`
  - i.e., removes `utils/common.py` since this is more of a preprocess function as @leewujung @lsetiawan discussed
  - related PRs that put new functions in `utils/common.py` (#749, #738) will need to move where the functions go as well
  - moves related tests


leewujung:
@b-reyes : Thanks a lot for adding the tests! I'll look into them and also try to see if what happens if one more channel gets added with different type of NaN entries. 

Reminder to myself: also test with example files from the recent PRs.

leewujung:
@b-reyes : I've gone through the tests and they look good. I think there's quite a bit of overall reworking we need for testing the MVBS computation, to make sure things are reliable, but I'll create a new issue for that and we can leave that for post v0.6.1.

One thing useful is that for multiple test cases using the same set of parameters, you could use @pytest.fixture with params/ids or @pytest.mark.parametrize that we have in other tests to reduce clutter and code repetition.

leewujung:
I'll merge this now and see if there are other problems needing work for v0.6.1.

Issue #750, opened 2022-07-01
Not showing echogram using ES80 sonar model


yalonzo:
Hello, 

I'm trying to show an echogram with Sonar ES80 data. However, first, I can't open the RAW data with this sonar model. Instead, I try to open with EK80 like parameter and I got success. I don't know if this trick is correct.

Now, I'm trying to show the echogram but I don't see anything (see the image)

![image](https://user-images.githubusercontent.com/50753972/176935400-9a30086f-691a-444b-8eb7-d66c8065a964.png)

Can you help with this please.

Thank you! 😎

emiliom:
Thanks for the report. Could you provide the error you got when you used `sonar_model="ES80"` in `open_raw`?

yalonzo:
There isn't a problem when I read the data using `open_raw `with `sonar_model="ES80"`. However, when I want to calibrate the data I get this error (see the image)

![image](https://user-images.githubusercontent.com/50753972/177201469-ebca2e09-68cf-43b3-a9c6-7d27b7c905e0.png)

So, this is the principal reason to use EK80 instead ES80 when I read raw data, but the echogram is empty using this trick.

What do I need to do to see the echogram?

emiliom:
Thanks for that information and screenshot. What echopype version are you using?

I think there may be two different issues going on. The first one is that it looks like `calibrate.compute_Sv` is not handling "ES80". I see that ES80 is not included here among the defined "calibrators", and it's also not handled explicitly in the blocks below that line.

https://github.com/OSOceanAcoustics/echopype/blob/a20ea8207484ed05c68e6556293c28b447fa0019/echopype/calibrate/api.py#L10

That's probably a relatively easy fix on our end.

The reason the `Sv` plot looks empty is a different problem. Do you have a small raw file you could share with us to test this? Preferably less than 50 MB or so.

yalonzo:
I understand. Ok.

I share with you the RAW file that I'm trying to plot.
[sample_raw.zip](https://github.com/OSOceanAcoustics/echopype/files/9047879/sample_raw.zip)

emiliom:
Thanks!

yalonzo:
So, can you plot the echogram with this RAW data? How can I plot the echogram now?

Also, I see that will be a new version of Echopype with new fixes, when will it be released?

Will I be able to plot the echogram with this new version?

Thanks in advance 😎

leewujung:
[...]

PR #755, opened 2022-07-05
Fix bug in calibration that results in empty Sv

leewujung:
The renaming of time coordinate in the `Environment` group (now called `time1` according to the convention) results in discrepancy in broadcasting the environment parameters during calibration, since the backscatter data is aligned with `ping_time`. In v0.6.0 `time1` was directly renamed to `ping_time`, which resulted in all non-matching timestamps in the backscatter data to be broadcasted to NaN. 

I think this is related to some parts of the problems seen in a number of recent issues #720 #743 #750. 

This is high priority and let's get the fix out in v0.6.1.

### Tasks
- [x] fix bugs related to time coordinate broadcasting
- [ ] add unit tests for the new function `EchoData._harmonize_env_param_time`
- [ ] add tests for catching unexpected NaN in calibrated data

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -349,6 +349,10 @@ class EchoData:
         if isinstance(env_params, EnvParams):
             env_params = env_params._apply(self)
 
+        # TODO:
+        # Right now this only works with Environment.time1 with length=1
+        # If length>1 then requires interpolation
+        # See CalibrateEK._harmonize_env_param_time for implementation
         def squeeze_non_scalar(n):
             if not np.isscalar(n):
                 n = n.squeeze()
@@ -364,9 +368,7 @@ class EchoData:
                 formula_source="AZFP" if self.sonar_model == "AZFP" else "Mackenzie",
             )
         elif self.sonar_model in ("EK60", "EK80") and "sound_speed_indicative" in self.environment:
-            sound_speed = squeeze_non_scalar(
-                self.environment["sound_speed_indicative"].rename({"time1": "ping_time"})
-            )
+            sound_speed = squeeze_non_scalar(self.environment["sound_speed_indicative"])
         else:
             raise ValueError(
                 "sound speed must be specified in env_params, "
@@ -470,6 +472,7 @@ class EchoData:
 
             # set entries with NaN backscatter data to NaN
             if "beam" in beam["backscatter_r"].dims:
+                # Drop beam because echo_range does not have beam dimension
                 valid_idx = ~beam["backscatter_r"].isel(beam=0).drop("beam").isnull()
             else:
                 valid_idx = ~beam["backscatter_r"].isnull()
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -2,7 +2,7 @@ import datetime
 import warnings
 from html import escape
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, Optional, Set, Tuple
+from typing import TYPE_CHECKING, Any, Dict, Optional, Set, Tuple, Union
 
 import fsspec
 import numpy as np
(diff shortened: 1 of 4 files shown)


leewujung:
Oops circular import. I'll work on that.

echopype/echodata/echodata.py
```diff
@@ -256,6 +256,42 @@ def __setattr__(self, __name: str, __value: Any) -> None:
...
 
+    def _harmonize_env_param_time(
+        self, p: Union[int, float, xr.DataArray], ping_time: Union[xr.DataArray, datetime.datetime]
+    ):
+        """
+        Harmonize time coordinate between Beam_groupX data and env_params to make sure
+        the timestamps are broacast correctly in calibration and range calculations.
+
+        If timestamp exist for env_params (i.e. it is of type xr.DataArray),
+        they cpme from EchoData["Environment"] with coordinate time1.
```
emiliom, line 267:
```suggestion
        If timestamps exist for env_params (i.e. it is of type xr.DataArray),
        they come from EchoData["Environment"] with coordinate time1.
```

echopype/echodata/echodata.py
```diff
@@ -256,6 +256,42 @@ def __setattr__(self, __name: str, __value: Any) -> None:
...
+        For EK60, Environment.time1 = Beam_group1.ping_time, both from the RAW0 datagram
+        so are directly interchangeable.
+        For EK80, Environment.time1 is from the Environment XML datagram,
```
emiliom, line 270: I assume that's true also for AZFP? Plus, that's how you've implemented it, below. 

How about ES70, ES80 and EA640?

echopype/echodata/echodata.py
```diff
@@ -256,6 +256,42 @@ def __setattr__(self, __name: str, __value: Any) -> None:
                         self._tree[group_path].ds = __value
         super().__setattr__(__name, attr_value)
 
+    def _harmonize_env_param_time(
+        self, p: Union[int, float, xr.DataArray], ping_time: Union[xr.DataArray, datetime.datetime]
```
emiliom, line 260:
```suggestion
        self, p: Union[int, float, xr.DataArray], ping_time: Union[xr.DataArray, datetime.datetime] = None
```

This change shouldn't have any effect, but it'll make it clearer that `ping_time` is in effect optional. It's not used (and therefore not passed) for EK60. Maybe you can add a test for `ping_time is not None` below, before the statement `p_new = p.interp(time1=ping_time)`

leewujung: Yes, thanks for catching this. I did the typing and forgot to add back the =None part.

And good point for testing for `ping_time` -- this interp part is not used at all in the current code (since our time1 is always length=1 from EK80), but it will surely come in later once we combine echodata. I will add a test for this too, to make sure the operations are done correctly (although it is arguable whether we should do `method="linear"` or `method="nearest"` or something else).

echopype/echodata/echodata.py
leewujung: ES70 in theory has the same format as EK60 but I have not actually handle an example file. ES80 and EA640 has the same format as EK80. I think we have ES80 and EA640 test files, and can try to run them through the calibration code to see what happens.

echopype/echodata/echodata.py
leewujung: I decided to use `method="linear"` and remove the check for `sonar_model` so that the function is general. I realized that even for EK60 it could be that users provide data arrays as inputs of the `env_params`, and in that case we can't just do the `time1` to `ping_time` renaming either.

echopype/echodata/echodata.py
leewujung: I've updated the function so that `sonar_model` is no longer used.

echopype/echodata/echodata.py
leewujung: Done!

PR #758, opened 2022-07-07
Release v0.6.1 [all tests ci]

leewujung:
This is a PR that contains all changes for v0.6.1 release.


lsetiawan:
I know what's going on here. The current PR uses the older pr.yaml github action, which skips CI when there's `[skip ci]` in the commit message, which there are! So that's why the CI is not running.

leewujung:
I see. So this has to be merged to `main` for this to run, is that true?

lsetiawan:
Yepp exactly. I think as long as dev passes. We should be good.

leewujung:
I'll merge this to `main` now. Everything passes locally and the failures are the usual intermittent ones. Packaging passed also.

PR #762, opened 2022-07-08
Overhaul access pattern [all tests ci]

### LINKED ISSUE #708: Remove old access pattern for `EchoData` object
In v0.6.0 we move to use the access pattern `EchoData["Platform"]`, `EchoData["Sonar/Beam_group1"]`, etc. for example to access different groups in the netCDF data model. The old access pattern `echodata.platform`, `echodata.sonar`, `echodata.beam`, etc. is deprecated and we should remove it soon.

I am setting it in v0.6.2 because I think we should prioritize the memory issues and get other bug fixes in for v0.6.1 soon, before doing this which would require working through many parts of the code.

lsetiawan:
This PR overhauls the old attribute based access pattern in favor of path key access pattern for EchoData.

Resolves #708

### DIFF
--- a/echopype/calibrate/calibrate_ek.py
+++ b/echopype/calibrate/calibrate_ek.py
@@ -1,7 +1,6 @@
 import numpy as np
 import xarray as xr
 from scipy import signal
-from zarr.errors import GroupNotFoundError
 
 from ..echodata import EchoData
 from ..utils import uwa
@@ -63,11 +62,8 @@ class CalibrateEK(CalibrateBase):
         if param not in ["sa_correction", "gain_correction"]:
             raise ValueError(f"Unknown parameter {param}")
 
-        try:
-            if waveform_mode == "CW":
-                beam = self.echodata["Sonar/Beam_group2"]
-        except GroupNotFoundError:
-            beam = self.echodata["Sonar/Beam_group1"]
+        if waveform_mode == "CW" and self.echodata["Sonar/Beam_group2"] is not None:
+            beam = self.echodata["Sonar/Beam_group2"]
         else:
             beam = self.echodata["Sonar/Beam_group1"]
 
@@ -106,11 +102,12 @@ class CalibrateEK(CalibrateBase):
         cal_params : dict
         """
 
-        try:
-            if encode_mode == "power" and waveform_mode == "CW":
-                beam = self.echodata["Sonar/Beam_group2"]
-        except GroupNotFoundError:
-            beam = self.echodata["Sonar/Beam_group1"]
+        if (
+            encode_mode == "power"
+            and waveform_mode == "CW"
+            and self.echodata["Sonar/Beam_group2"] is not None
+        ):
+            beam = self.echodata["Sonar/Beam_group2"]
         else:
             beam = self.echodata["Sonar/Beam_group1"]
 
@@ -341,11 +338,12 @@ class CalibrateEK80(CalibrateEK):
               the echosounder is configured for narrowband transmission
         """
 
-        try:
-            if encode_mode == "power" and waveform_mode == "CW":
-                beam = self.echodata["Sonar/Beam_group2"]
-        except GroupNotFoundError:
-            beam = self.echodata["Sonar/Beam_group1"]
+        if (
+            encode_mode == "power"
+            and waveform_mode == "CW"
+            and self.echodata["Sonar/Beam_group2"] is not None
+        ):
+            beam = self.echodata["Sonar/Beam_group2"]
         else:
             beam = self.echodata["Sonar/Beam_group1"]
 
@@ -924,8 +922,7 @@ class CalibrateEK80(CalibrateEK):
         #   power samples will be stored in echodata["Sonar/Beam_group2"]
         # When only one type of samples exist,
         #   all samples with be stored in echodata["Sonar/Beam_group1"]
-        try:
-            self.echodata["Sonar/Beam_group2"]
+        if self.echodata["Sonar/Beam_group2"] is not None:  # both power and complex samples exist
             # If both beam and beam_power groups exist,
             #   this means that CW data are encoded as power samples and in beam_power group
             if waveform_mode == "CW" and encode_mode == "complex":
@@ -940,8 +937,7 @@ class CalibrateEK80(CalibrateEK):
                 print(
                     "Only complex samples are calibrated, but power samples also exist in the raw data file!"  # noqa
                 )
-        except GroupNotFoundError:
-            # only power OR complex samples exist
+        else:  # only power OR complex samples exist
             if (
                 "backscatter_i" in self.echodata["Sonar/Beam_group1"].variables
             ):  # data contain only complex samples
--- a/echopype/calibrate/calibrate_ek.py
+++ b/echopype/calibrate/calibrate_ek.py
@@ -54,7 +54,7 @@ class CalibrateEK(CalibrateBase):
         param : str {"sa_correction", "gain_correction"}
             name of parameter to retrieve
         """
-        ds_vend = self.echodata.vendor
+        ds_vend = self.echodata["Vendor_specific"]
 
         if ds_vend is None or param not in ds_vend:
             return None
@@ -413,16 +413,18 @@ class CalibrateEK80(CalibrateEK):
             'coeff' or 'decimation'
         """
         if param_type == "coeff":
-            v = self.echodata.vendor.attrs[
+            v = self.echodata["Vendor_specific"].attrs[
                 "%s %s filter_r" % (channel_id, filter_name)
             ] + 1j * np.array(
-                self.echodata.vendor.attrs["%s %s filter_i" % (channel_id, filter_name)]
+                self.echodata["Vendor_specific"].attrs["%s %s filter_i" % (channel_id, filter_name)]
             )
             if v.size == 1:
                 v = np.expand_dims(v, axis=0)  # expand dims for convolution
             return v
         else:
-            return self.echodata.vendor.attrs["%s %s decimation" % (channel_id, filter_name)]
+            return self.echodata["Vendor_specific"].attrs[
+                "%s %s decimation" % (channel_id, filter_name)
+            ]
 
     def _tapered_chirp(
         self,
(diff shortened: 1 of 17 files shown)


echopype/calibrate/calibrate_ek.py
```diff
@@ -102,14 +106,13 @@ def get_cal_params(self, cal_params, waveform_mode, encode_mode):
...
-        ):
-            beam = self.echodata.beam_power
+        try:
+            if encode_mode == "power" and waveform_mode == "CW":
+                beam = self.echodata["Sonar/Beam_group2"]
+        except GroupNotFoundError:
+            beam = self.echodata["Sonar/Beam_group1"]
         else:
-            beam = self.echodata.beam
+            beam = self.echodata["Sonar/Beam_group1"]
```
leewujung, line 115: It seems that this should be the following?
```
if encode_mode == "power" and waveform_mode == "CW":
   try:
      beam = self.echodata["Sonar/Beam_group2"]
   except GroupNotFoundError:
      beam = self.echodata["Sonar/Beam_group1"]
else 
   beam = self.echodata["Sonar/Beam_group1"]
```

The logic I am following is that
- `Sonar/Beam_group2` only exists when there are both power and complex data, and in this case:
   - power data will be in `Sonar/Beam_group2`
   - complex data will be in `Sonar/Beam_group1`
- If only power or only complex data exist, they will be in `Sonar/Beam_group1`

Did I misunderstand this?

echopype/calibrate/calibrate_ek.py
```diff
@@ -54,18 +55,21 @@ def _get_vend_cal_params_power(self, param, waveform_mode):
...
-        if waveform_mode == "CW" and self.echodata.beam_power is not None:
-            beam = self.echodata.beam_power
+        try:
+            if waveform_mode == "CW":
+                beam = self.echodata["Sonar/Beam_group2"]
+        except GroupNotFoundError:
+            beam = self.echodata["Sonar/Beam_group1"]
         else:
-            beam = self.echodata.beam
+            beam = self.echodata["Sonar/Beam_group1"]
```
leewujung, line 72: See my comments for line 109-115 below.

echopype/calibrate/calibrate_ek.py
```diff
@@ -337,14 +341,13 @@ def get_env_params(self, waveform_mode=None, encode_mode="complex"):
...
-        ):
-            beam = self.echodata.beam_power
+        try:
+            if encode_mode == "power" and waveform_mode == "CW":
+                beam = self.echodata["Sonar/Beam_group2"]
+        except GroupNotFoundError:
+            beam = self.echodata["Sonar/Beam_group1"]
         else:
-            beam = self.echodata.beam
+            beam = self.echodata["Sonar/Beam_group1"]
```
leewujung, line 351: if my comment on line 109-115 above.

echopype/calibrate/calibrate_ek.py
```diff
@@ -879,22 +909,23 @@ def _compute_cal(self, cal_type, waveform_mode, encode_mode) -> xr.Dataset:
...
-        #   complex samples will be stored in echodata.beam
-        #   power samples will be stored in echodata.beam_power
+        #   complex samples will be stored in echodata["Sonar/Beam_group1"]
+        #   power samples will be stored in echodata["Sonar/Beam_group2"]
         # When only one type of samples exist,
-        #   all samples with be stored in echodata.beam
-        if self.echodata.beam_power is not None:  # both power and complex samples exist
+        #   all samples with be stored in echodata["Sonar/Beam_group1"]
+        try:
+            self.echodata["Sonar/Beam_group2"]
```
leewujung, line 928: would this line
```
self.echodata["Sonar/Beam_group2"]
```
result in something being printed out? I think we don't want that behavior?

echopype/echodata/combine.py
```diff
@@ -177,13 +178,19 @@ def combine_echodata(echodatas: List[EchoData], combine_attrs="override") -> Ech
     old_attrs: Dict[str, List[Dict[str, Any]]] = dict()
 
     for group, value in EchoData.group_map.items():
```
leewujung, line 180: Just for the sake of readability, seems a good idea to add a line of comment above this block to point to the file (`echodata/convention/1.0.yml`) where people will find the `group_path` names (e.g., `top`, `sonar`, `beam`, etc) corresponding with the actual group names ("Top-level", "Sonar", "Beam_group1", etc)

echopype/echodata/echodata.py
```diff
@@ -478,16 +458,15 @@ def compute_range(
...
+                try:
+                    if self.sonar_model == "EK80" and encode_mode == "power":
+                        # if both CW and BB exist and beam_power group is not empty
+                        # this means that CW is recorded in power/angle mode
+                        beam = self["Sonar/Beam_group2"]
+                except GroupNotFoundError:
+                    beam = self["Sonar/Beam_group1"]
                 else:
-                    beam = self.beam
+                    beam = self["Sonar/Beam_group1"]
```
leewujung, line 469: Same as comments in `calibrate_ek.py` for an identical block.

Seeing so many of these, I wonder if we should have a util function to do this triage.

echopype/tests/echodata/test_echodata.py
```diff
@@ -352,7 +335,7 @@ def test_compute_range(compute_range_samples):
         ek_encode_mode,
     ) = compute_range_samples
     ed = echopype.open_raw(filepath, sonar_model, azfp_xml_path)
-    print(ed.platform)
+    print(ed["Platform"])
```
leewujung, line 338: seems that we don't need this line? I wonder if this is a debugging leftover

echopype/tests/echodata/test_echodata_combine.py
```diff
@@ -107,54 +109,41 @@ def test_combine_echodata(raw_datasets):
...
+                data_vars=concat_data_vars.get(
+                    group_name, concat_data_vars["default"]
+                ),
+                coords="minimal",
+                combine_attrs="drop",
+            )
+            test_ds.attrs.update(union_attrs(eds_groups))
+            test_ds = test_ds.drop_dims(
                 [
+                    "concat_dim",
```
leewujung, line 146: I was a little puzzled by this "concat_dim" and then I found this line `# xarray inserts this dimension when concatenating along multiple dimensions` [here](https://github.com/OSOceanAcoustics/echopype/blob/6ebc9530eb762ad819cbcb338a2313d05b11e0ac/echopype/echodata/combine.py#L253). Would be good to add this line here to clarify that this is added by xarray, because initially I thought this is maybe a typo or something, since we definitely don't have dimensions named "concat_dim" in echopype. :)

leewujung (commented):
Hey @lsetiawan : thanks for the PR! Sorry it took me forever to get to this. 

The PR looks great, there's just this repeated block that handles whether `beam` is connected to `echodata["Sonar/Beam_group1"]` or `echodata["Sonar/Beam_group2"]` that I feel the if/else + try/except logic should be a little different. See what you think, it is possible that I misunderstood something.

echopype/echodata/echodata.py
lsetiawan: @leewujung Previously `ed.beam` and such results in `None`. Now we have `GroupNotFoundError` instead. I almost wonder if I should just make it `None` rather than error out?

echopype/tests/echodata/test_echodata.py
lsetiawan, line 338:
```suggestion
```

echopype/tests/echodata/test_echodata_combine.py
lsetiawan, line 146:
```suggestion
                    # xarray inserts this dimension when concatenating along multiple dimensions
                    "concat_dim",
```

echopype/calibrate/calibrate_ek.py
lsetiawan: See https://github.com/OSOceanAcoustics/echopype/pull/762#discussion_r941612868.. I think a lot of this `try/except` is kind of ugly haha

echopype/echodata/combine.py
lsetiawan: Done. See latest changes.

echopype/calibrate/calibrate_ek.py
lsetiawan: Resolved. Per discussion, `__getitem__` has change to return `None` so `try/except` is not needed and changes is simply on the access pattern.

echopype/calibrate/calibrate_ek.py
lsetiawan: Removed and revert back to checking of `self.echodata["Sonar/Beam_group2"] is None` since now groups that are not in echodata returns None when trying to be accessed.

lsetiawan:
@leewujung This is now ready for your final review hopefully 😄

echopype/tests/echodata/test_echodata.py
```diff
@@ -267,24 +267,25 @@ def test_getitem(self, converted_zarr):
...
         ed = EchoData.from_file(converted_raw_path=converted_zarr)
         ed['Sonar/Beam_group1'] = ed['Sonar/Beam_group1'].rename({'beam': 'beam_newname'})
 
         assert sorted(ed['Sonar/Beam_group1'].dims.keys()) == ['beam_newname', 'channel', 'ping_time', 'range_sample']
 
+        try:
+            ed['SomeRandomGroup'] = 'Testing value'
+        except Exception as e:
+            assert isinstance(e, GroupNotFoundError)
+
```
leewujung, line 288: I like this!

leewujung: @b-reyes : the `test_setitem` and `test_getitem` are the type of unit tests I was thinking for the changes you made in #774. Just making a note here as it came to my mind while reviewing this. We can circle back to that and your ideas in our tests-hackday!

echopype/tests/echodata/test_echodata_combine.py
```diff
@@ -112,38 +110,52 @@ def test_combine_echodata(raw_datasets):
...
+            data_vars=concat_data_vars.get(
+                group_name, concat_data_vars["default"]
+            ),
+            coords="minimal",
+            combine_attrs="drop",
+        )
+        test_ds.attrs.update(union_attrs(eds_groups))
+        test_ds = test_ds.drop_dims(
+            [
+                # xarray inserts this dimension when concatenating along multiple dimensions
```
leewujung, line 143:
```suggestion
                # xarray inserts "concat_dim" when concatenating along multiple dimensions
```
Just to make thing painfully clear

leewujung (approved):
@lsetiawan : I think this PR is ready to be merged! Just had a minor suggestion on a comment. It is nice to see how changing the getitem output to None makes the code so much cleaner to read.

PR #772, opened 2022-08-06
Add logging and optional printouts

### LINKED ISSUE #457: Provide control over echopype's output to the terminal
echopype outputs text when doing stuff, such as:

`15:40:36  parsing file D20180619-T183100.raw, time of first ping: 2018-Jun-19 18:31:00`

This can be distracting when using echopype, especially when it is used as part of a separate program. It would be useful if this could be turned off, or perhaps controlled using Python's logging functionality.

lsetiawan:
## Overview
This PR adds logging to echopype by removing print statements throughout the code and replacing them with `logger.info`.

1. There is now a new utility module under `echopype.utils` called `log`.
2. Users can now activate the print statements by calling `ep.verbose()`.
    ```python
    import echopype as ep
    ep.verbose()
    ```
3. The `ep.verbose` function contains argument `logfile` to specify where the log should be outputted to and `override` for overriding the verbosity value, so that users can turn off and on at any time. By default, echopype now hides the info statements.

This PR resolves #457

### DIFF
--- a/echopype/convert/utils/ek_raw_io.py
+++ b/echopype/convert/utils/ek_raw_io.py
@@ -6,18 +6,18 @@ by Rick Towler <rick.towler@noaa.gov> at NOAA AFSC.
 Contains low-level functions called by ./ek_raw_parsers.py
 """
 
-import logging
 import struct
 from io import SEEK_CUR, SEEK_END, SEEK_SET, BufferedReader, FileIO
 
 import fsspec
 from fsspec.implementations.local import LocalFileSystem
 
+from ...utils.log import _init_logger
 from . import ek_raw_parsers as parsers
 
 __all__ = ["RawSimradFile"]
 
-log = logging.getLogger(__name__)
+log = _init_logger(__name__)
 
 
 class SimradEOF(Exception):
@@ -590,7 +590,7 @@ class RawSimradFile(BufferedReader):
             dgram_size = self._read_dgram_size()
 
         except DatagramSizeError:
-            print("Error reading the datagram")
+            log.info("Error reading the datagram")
             self._seek_bytes(old_file_pos, SEEK_SET)
             raise
 
--- a/echopype/convert/utils/ek_raw_io.py
+++ b/echopype/convert/utils/ek_raw_io.py
@@ -17,7 +17,7 @@ from . import ek_raw_parsers as parsers
 
 __all__ = ["RawSimradFile"]
 
-log = _init_logger(__name__)
+logger = _init_logger(__name__)
 
 
 class SimradEOF(Exception):
@@ -265,7 +265,7 @@ class RawSimradFile(BufferedReader):
 
         #  check for invalid time data
         if (header["low_date"], header["high_date"]) == (0, 0):
-            log.warning(
+            logger.warning(
                 "Skipping %s datagram w/ timestamp of (0, 0) at %sL:%d",
                 header["type"],
                 str(self._tell_bytes()),
@@ -277,7 +277,7 @@ class RawSimradFile(BufferedReader):
         #  basic sanity check on size
         if header["size"] < 16:
             #  size can't be smaller than the header size
-            log.warning(
+            logger.warning(
                 "Invalid datagram header: size: %d, type: %s, nt_date: %s.  dgram_size < 16",
                 header["size"],
                 header["type"],
@@ -303,7 +303,7 @@ class RawSimradFile(BufferedReader):
 
         #  and make sure it checks out
         if bytes_read < header["size"]:
-            log.warning(
+            logger.warning(
                 "Datagram %d (@%d) shorter than expected length:  %d < %d",
                 self.tell(),
                 old_file_pos,
@@ -324,14 +324,14 @@ class RawSimradFile(BufferedReader):
         #  make sure they match
         if header["size"] != dgram_size_check:
             # self._seek_bytes(old_file_pos, SEEK_SET)
-            log.warning(
+            logger.warning(
                 "Datagram failed size check:  %d != %d @ (%d, %d)",
                 header["size"],
                 dgram_size_check,
                 self._tell_bytes(),
                 self.tell(),
             )
-            log.warning("Skipping to next datagram...")
+            logger.warning("Skipping to next datagram...")
             self._find_next_datagram()
 
             return self._read_next_dgram()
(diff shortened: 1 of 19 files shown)


lsetiawan:
Current CI errors seem to be something that is similar to the errors seen in `dev` branch.

leewujung:
Awesome! I'll review this today!

emiliom:
> Awesome! I'll review this today!

@leewujung I was planning to review it tonight. But if you will (or already have!), I won't :wink:. Let me know.

leewujung:
Actually if @emiliom if you could review this I could use the time for other issues. So please go ahead! Thanks!!

echopype/utils/log.py
```diff
@@ -0,0 +1,112 @@
...
+LOGFILE_HANDLE_NAME = "logfile_file_handler"
+
+
+class _ExcludeWarningsFilter(logging.Filter):
+    def filter(self, record):  # noqa
+        """Only lets through log messages with log level below ERROR ."""
+        return record.levelno < logging.WARNING
+
+
+def verbose(logfile: Optional[str] = None, override: Optional[bool] = None) -> None:
```
emiliom, line 18: As written, `override` has a default `None` value that will raise an error (line 33). Wouldn't it be more helpful to set the default to a boolean value? `False` seems the most logical to me: the user is calling a function whose name "verbose" suggests that it should make things verbose unless otherwise stated.

echopype/convert/utils/ek_raw_io.py
```diff
@@ -6,18 +6,18 @@
...
 import fsspec
 from fsspec.implementations.local import LocalFileSystem
 
+from ...utils.log import _init_logger
 from . import ek_raw_parsers as parsers
 
 __all__ = ["RawSimradFile"]
 
-log = logging.getLogger(__name__)
+log = _init_logger(__name__)
```
emiliom, line 20: Change the name of the assigned variable to `logger`, just for consistency with usage elsewhere.
```suggestion
logger = _init_logger(__name__)
```

emiliom:
A general comment about consistency (ie, style, not substance).

In about half of the modules that use `utils.log`, you use this package import pattern:
```python
from ..utils.log import _init_logger

logger = _init_logger(__name__)
```

In the other half, you use:
```python
from ..utils import log

logger = log._init_logger(__name__)
```

I suggest using a common pattern everywhere. Since it looks like the only thing you ever use from `log` is `_init_logger`, I'd vote for the first pattern.

echopype/utils/log.py
lsetiawan: It won't raise an error because I catch it by checking if `override is not None` then check if it's a boolean. On line 37 I have a check if `verbose` is `None` then by default set to `True`. The reason I do this logic is because I want the user to specifically call `override` if they'd like to override the actual verbose value. Otherwise `verbose` means that it will be verbose by default.

echopype/tests/visualize/test_plot.py
```diff
@@ -212,9 +214,10 @@ def test_plot_mvbs(
         (30.5, False),
     ],
 )
-def test_water_level_echodata(water_level, expect_warning):
+def test_water_level_echodata(water_level, expect_warning, caplog):
```
emiliom, line 217: I don't understand what `caplog` is and where it's assigned :confused:  .  Could you explain, briefly? (Same goes for `test_water_level_Sv_dataset`).

I've run the test suite and all tests are passing, so I'm not concerned that something is failing. I just can't figure out what's going on.

echopype/tests/visualize/test_plot.py
```diff
@@ -1,3 +1,5 @@
+import logging
```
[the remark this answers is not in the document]
lsetiawan: Not during test. I can delete that.

lsetiawan, line 1:
```suggestion
```

echopype/tests/visualize/test_plot.py
lsetiawan: `caplog` is a special pytest attribute to get at captured logs by pytest test. See documentation: https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture

emiliom: Ah!! Thanks for the explanation.

emiliom:
Alright, other than the narrow inline comments, I'd like to make a broader comment about the `override` argument and its intuitiveness. Briefly, I think its implementation may be counterintuitive and possibly the opposite of what one might assume

The default is no verbosity (I think that's what we discussed, and I agree with it). This is accomplished in `echopype/__init__.py`: `verbose(override=False)`. The use of `verbose()` results in verbosity. So far so good.

For a function call named "verbose", it's natural to assume that its default behavior is to be verbose. That's confirmed by the behavior of `verbose()`. I would then interpret `override=False` as meaning that verbose behavior remains, ie, same behavior as `verbose()`. The current docstring definition of the argument would seem to agree with that interpretation: "override: bool, optional. Optional boolean flag to override verbosity". But in fact, the behavior is the opposite. `verbose()` results in verbosity while `verbose(override=False)` results in no verbosity.

We could address this possible confusion by improving the docstring. But I think the common interpretation of an argument named "override" in a function named "verbose" is counter to the current implementation. So, in addition to docstring tweaks, I suggest flipping the "override" interpretation: `True` should results in no verbosity and `False` in verbosity.

Also, going back to our earlier inline exchange, you made this comment:

> The reason I do this logic is because I want the user to specifically call `override` if they'd like to override the verbosity. Otherwise `verbose` means that it will be verbose by default.

I find this logic and the use of `None` default for `override` overly complicated and probably unnecessary. It also reduces the usability of the `def` line:
```python
def verbose(logfile: Optional[str] = None, override: Optional[bool] = None) -> None:
```
From that line alone, a user won't know what the default behavior of `verbose()` is. It's more work for users. I suggest setting a default argument in the usual way, eg:
```python
def verbose(logfile: Optional[str] = None, override: bool = False) -> None:
```

lsetiawan:
Thanks for the comments @emiliom. I definitely see what you're saying and okay, that makes total sense. I do like that `override=True` means no verbosity, instead of `False`. I'll make the changes and push em up shortly. Thanks again for your suggestions! 😄

emiliom:
Regarding `logfile`:
- If the file path already exists, it concatenates the new logs. That could be confusing if a user inadvertently reuses a previous log file path. I suggest overwriting the file if it exists.
- If a `logfile` path is passed together with `override=False`, an empty file is created. I suggest that a file shouldn't be created at all.

I found myself wanting the use case of having verbosity but only in the log file, not inline. In other words, if I specify a `logfile` it's because I want stuff in it even if I choose not to display those log messages inline. Right now that's not possible. But I realize that may take more work/discussion, and may be best left to 0.6.3.

Otherwise, it's an awesome implementation and feature!! FYI, I ran multiple manual tests to get a clear idea of how  it behaved.

lsetiawan:
Your last suggestions in https://github.com/OSOceanAcoustics/echopype/pull/772#issuecomment-1212355025 regarding log files might have to go into an issue. The two behavior you describe are default implementations from the `logging` library, which I haven't fully delve into, which would require more research, and I think should be in future features. I think at the moment, it does the minimal of what we want which is the capability to print out logs when we want and turn it off when we want. 😛

emiliom:
> Your last suggestions in https://github.com/OSOceanAcoustics/echopype/pull/772#issuecomment-1212355025 regarding log files might have to go into an issue. The two behavior you describe are default implementations from the logging library, which I haven't fully delve into, which would require more research, and I think should be in future features. I think at the moment, it does the minimal of what we want which is the capability to print out logs when we want and turn it off when we want.

I agree. I'll try to create the issue later today.

emiliom (approved):
I haven't retested this manually after your changes, but they look good. My new inline comment is a minor point.

echopype/utils/log.py
```diff
@@ -15,26 +15,28 @@ def filter(self, record):  # noqa
...
+        Boolean flag to override verbosity,
+        which turns off verbosity if the value is `False`.
+        Default is `False`.
 
     Returns
     -------
     None
     """
-    if override is not None and not isinstance(override, bool):
+    if not isinstance(override, bool):
```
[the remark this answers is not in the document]
lsetiawan: No it doesn't have built-in error checking. It's just for hinting a type of an argument.

PR #786, opened 2022-08-14
Add attention block in docs to note more-itertools conda problem

leewujung:


leewujung:
I'll self-merge in the interest of time because v0.6.2 is in a half-hanging state only available via pip install, and the environment creating instructions are also failing.

Issue #742, opened 2022-06-28
Coordinate `filenames` of data variable `source_filenames` should be a coordinate


leewujung:
I am just working with the calibrated Sv dataset and trying to combine many of them, and I noticed that the `filenames` is a dimension but not a coordinate. This makes it not possible to automatically combining the `source_filenames` together along the `filenames` dimension.

It seems that this data variable is added here and we can make `filenames` a coordinate explicitly to make it easier to work with.
https://github.com/OSOceanAcoustics/echopype/blob/12693a72badcf54a18408e694b72e7b0e6d48023/echopype/calibrate/api.py#L95

@emiliom : tagging you since this is related to what we want to include in the data provenance. This seems like a one-line fix?

emiliom:
`source_files_vars` is a function that's reused in at least a couple of places, including here and in `set_provenance`. A general solution would be to change `source_files_vars` so that it returns dicts for both the variable and the coordinate, just like the [_beam_groups_vars](https://github.com/OSOceanAcoustics/echopype/blob/a3ddec1d9f9aa6bbd394aac82c0bdffcfde90477/echopype/convert/set_groups_base.py#L190) method. See this example of the use of that method:
https://github.com/OSOceanAcoustics/echopype/blob/a3ddec1d9f9aa6bbd394aac82c0bdffcfde90477/echopype/convert/set_groups_ek60.py#L198-L199

So, I'd change the line you mentioned into roughly this:
```python
source_filenames_var, filenames_coord = source_files_vars(source_file)
cal_ds = cal_ds.assign(**source_filenames_var).assign_coords(**filenames_coord)
```

Of course, we'll need to modify `source_files_vars`, too.

emiliom:
#803 already addressed this issue. I'm closing it.

Issue #677, opened 2022-05-12
`xml_path` is not contained in `source_filenames` for the AZFP sensor


b-reyes:
While looking into PR #674, it was found that currently we only store the `raw_file` path from `open_raw` as an attribute (after PR #674 this will become the variable `ed.provenance.source_filenames`). However, for AZFP we have both the `raw_file` and `xml_path` as inputs and the file names for these can be quite different. After discussing this with @emiliom, it seems like we should make it so that we store both the `raw_file` and `xml_path` in the variable `ed.provenance.source_filenames` or modify this variable in someway for the AZFP sensor.

emiliom:
I only see three options here, and none of them is great. For the case of applying `open_raw` on a single AZFP file:

1. Rather than than entering a single string for the data file (the file path, eg, "/my/path/17082117.01A"), enter a string array with both file paths (eg, `["17082117.01A", "17041823.XML"]`)
2. Introduce a new variable, say, `sourcemeta_filenames`, to hold the metadata/XML: "17041823.XML". Then `source_filenames` will remain as is, listing only the data file: "17082117.01A"
3. Ignore the metadata/XML file path

1 is simple to implement and doesn't introduce new variables. It passes the complexity down to the `combine_echodata` operation, which will need to figure out how to handles pairs of paths that represent a single data source. It could use string representations of string tuples as individual array elements, eg, `['("17082117.01A", "17041823.XML")', '("27082117.01A", "27041823.XML")', ...]`). But then that's not what the convention says, and it'll require special logic for echopype to be able to detect and handle both this form and the current, simple string array cases like EK60, eg, `["file1.raw", "file2.raw", ...]`.

2 is clean, but it introduces a new variable not in the convention.

3 is easy: do nothing! :wink:

@leewujung, thoughts?

leewujung:
I vote for 2. I think it is ok to have a new variable as long as it is clear what that is. I also like the part that it makes the core logic in `combine_echodata` intact. 

For `sourcemeta_filenames`: in `combine_echodata` I suggest we use the same logic as `source_filenames` to store the combined outcome (whatever form that takes as part of the `source_filename` clean-up).

b-reyes:
@emiliom thank you for looking into this. From the options you have proposed, option 2 does seem to be the best solution. 

[...]

PR #806, opened 2022-09-10
Turn filenames dimension into coordinate variable

emiliom:
Addresses #742 across the board, including in the converted dataset (provenance group) and not just for `Sv`.

### DIFF
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -119,7 +119,7 @@ class SetGroupsEK60(SetGroupsBase):
                     "old_ping_time": self.old_ping_time,
                     **source_files_var,
                 },
-                coords=source_files_coord
+                coords=source_files_coord,
             )
         else:
             ds = xr.Dataset(data_vars=source_files_var, coords=source_files_coord)
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -111,7 +111,6 @@ class SetGroupsEK60(SetGroupsBase):
         """Set the Provenance group."""
         prov_dict = echopype_prov_attrs(process_type="conversion")
         prov_dict["duplicate_ping_times"] = 1 if self.old_ping_time is not None else 0
-        # source_files = source_files_vars(self.input_file)
         source_files_var, source_files_coord = source_files_vars(self.input_file)
         if self.old_ping_time is not None:
             ds = xr.Dataset(
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -111,16 +111,18 @@ class SetGroupsEK60(SetGroupsBase):
         """Set the Provenance group."""
         prov_dict = echopype_prov_attrs(process_type="conversion")
         prov_dict["duplicate_ping_times"] = 1 if self.old_ping_time is not None else 0
-        source_files = source_files_vars(self.input_file)
+        # source_files = source_files_vars(self.input_file)
+        source_files_var, source_files_coord = source_files_vars(self.input_file)
         if self.old_ping_time is not None:
             ds = xr.Dataset(
                 data_vars={
                     "old_ping_time": self.old_ping_time,
-                    **source_files,
-                }
+                    **source_files_var,
+                },
+                coords=source_files_coord
             )
         else:
-            ds = xr.Dataset(data_vars=source_files)
+            ds = xr.Dataset(data_vars=source_files_var, coords=source_files_coord)
         ds = ds.assign_attrs(prov_dict)
         return ds
 
--- a/echopype/echodata/combine.py
+++ b/echopype/echodata/combine.py
@@ -107,11 +107,7 @@ def check_and_correct_reversed_time(
 def assemble_combined_provenance(input_paths):
     prov_dict = echopype_prov_attrs(process_type="combination")
     source_files_var, source_files_coord = source_files_vars(input_paths)
-    ds = xr.Dataset(
-        data_vars=source_files_var,
-        coords=source_files_coord,
-        attrs=prov_dict
-    )
+    ds = xr.Dataset(data_vars=source_files_var, coords=source_files_coord, attrs=prov_dict)
     return ds
 
 
--- a/echopype/echodata/combine.py
+++ b/echopype/echodata/combine.py
@@ -105,10 +105,14 @@ def check_and_correct_reversed_time(
 
 
 def assemble_combined_provenance(input_paths):
-    return xr.Dataset(
-        data_vars=source_files_vars(input_paths),
-        attrs=echopype_prov_attrs(process_type="conversion"),
+    prov_dict = echopype_prov_attrs(process_type="combination")
+    source_files_var, source_files_coord = source_files_vars(input_paths)
+    ds = xr.Dataset(
+        data_vars=source_files_var,
+        coords=source_files_coord,
+        attrs=prov_dict
     )
+    return ds
 
 
 def union_attrs(datasets: List[xr.Dataset]) -> Dict[str, Any]:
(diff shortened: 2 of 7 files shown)


echopype/utils/prov.py
```diff
@@ -59,4 +64,11 @@ def source_files_vars(source_paths: Union[str, List[Any]]) -> Dict[str, Tuple]:
         ),
     }
 
-    return source_files_var
+    source_files_coord = {
+        "filenames": (
+            "filenames",
+            list(range(len(source_files))),
+            {"long_name": "Source filenames index"},
```
b-reyes, line 71:
```suggestion
            {"long_name": "Index for data source file names"},
```
I made this small change because it looks like you are maybe referring to the variable `source_filenames`. I wanted to refrain from doing this because we plan on using this coordinate in issue #677 for another variable. Feel free to change it, if your think there is a better long name.

echopype/convert/set_groups_ek60.py
```diff
@@ -111,16 +111,18 @@ def set_provenance(self) -> xr.Dataset:
         """Set the Provenance group."""
         prov_dict = echopype_prov_attrs(process_type="conversion")
         prov_dict["duplicate_ping_times"] = 1 if self.old_ping_time is not None else 0
-        source_files = source_files_vars(self.input_file)
+        # source_files = source_files_vars(self.input_file)
```
b-reyes, line 114:
```suggestion
```

echopype/tests/visualize/test_plot.py
```diff
@@ -173,6 +173,10 @@ def test_plot_mvbs(
     azfp_xml_path,
     range_kwargs,
 ):
+
+    # pytest.xfail("Test momentarily (2022-9-9, EM) set to xfail due to 'Killed' failure")
```
b-reyes, line 177: @emiliom can you explain this comment and how it relates to this PR?

b-reyes:
@emiliom now that `source_files_vars` returns two dictionaries, [assemble_combined_provenance](https://github.com/emiliom/echopype/blob/7b0a6e8265d70acf42f15ee8318fe547b80613e5/echopype/echodata/combine.py#L107-L111) in `combine.py` should be updated. It currently assigns all output of `source_files_vars` to `data_vars`.

b-reyes:
@emiliom I ran the following code on an EK60 file 

```
Sv_ds = ep.calibrate.compute_Sv(ed)
MVBS_ds = ep.preprocess.compute_MVBS(
    Sv_ds,
    range_meter_bin=5,  # in meters
    ping_time_bin='20s'  # in seconds
)
```

I found that `filenames` is correctly displayed in `Sv_ds`. However, it appears in `MVBS_ds` as an unused coordinate without a dimension (please see picture below). I am not sure if this occurs in other downstream processes, but it appears that it is caused by changes in this PR.

Screen Shot 2022-09-13 at 5 03 06 PM

b-reyes (changes requested):
@emiliom I am done reviewing. The changes look good, but there are a couple of items that need to be addressed (please see my comments above).

echopype/tests/visualize/test_plot.py
emiliom: This was an oversight and should be removed. I was dealing with some test failures, but forgot to remove this line. I'll remove it here.

echopype/tests/visualize/test_plot.py
```diff
@@ -173,6 +173,10 @@ def test_plot_mvbs(
...
+
```
emiliom, line 178:
```suggestion
```

echopype/utils/prov.py
emiliom: Good point and good suggestion for a `long_name`. But I'll deal with it in the next PR, the one that actually implements #677. I think it'll be best to finalize that long_name string in that context.

emiliom:
> I found that filenames is correctly displayed in `Sv_ds`. However, it appears in `MVBS_ds` as an unused coordinate without a dimension (please see picture below). I am not sure if this occurs in other downstream processes, but it appears that it is caused by changes in this PR.

Great catch! I think you're right that it's cause by this PR, by `filenames` becoming a coordinate variable. I'll fix this.

emiliom:
> @emiliom now that `source_files_vars` returns two dictionaries, [assemble_combined_provenance](https://github.com/emiliom/echopype/blob/7b0a6e8265d70acf42f15ee8318fe547b80613e5/echopype/echodata/combine.py#L107-L111) in `combine.py` should be updated. It currently assigns all output of `source_files_vars` to `data_vars`.

Ah. That makes me realize that the `test_echodata_combine.py` failures I was sloppily attributing to the overall flux in combine functionality are actually a result of this bug I introduced. I'll fix it.

emiliom:
> @emiliom now that source_files_vars returns two dictionaries, [assemble_combined_provenance](https://github.com/emiliom/echopype/blob/7b0a6e8265d70acf42f15ee8318fe547b80613e5/echopype/echodata/combine.py#L107-L111) in combine.py should be updated. It currently assigns all output of source_files_vars to data_vars.

I've addressed this in the latest commit I just pushed.

echopype/echodata/combine.py
```diff
@@ -105,10 +105,10 @@ def check_and_correct_reversed_time(
 
 
 def assemble_combined_provenance(input_paths):
-    return xr.Dataset(
-        data_vars=source_files_vars(input_paths),
-        attrs=echopype_prov_attrs(process_type="conversion"),
-    )
+    prov_dict = echopype_prov_attrs(process_type="combination")
```
b-reyes, line 108: I see here that you changed `process_type` to `"combination"`. I agree that the name should be changed. However, maybe it is a better idea to create an issue for this (so that I can account for it in my combine routine) and keep `"conversion"` in this PR?

echopype/utils/prov.py
```diff
@@ -5,7 +5,7 @@
 from _echopype_version import version as ECHOPYPE_VERSION
 from typing_extensions import Literal
 
-ProcessType = Literal["conversion", "processing"]
+ProcessType = Literal["conversion", "combination", "processing"]
```
b-reyes, line 8: If we decide to not change the name in this PR (see my previous comment), then we should also remove this.

echopype/echodata/combine.py
emiliom: Yeah, I made that change because I happened to run into that `process_type` string while fixing the combine problem, and I had not noticed that usage before. It's not inherent to this PR, so it's not a problem undoing it for the time being.

All tests **are** now running fine, though (see the CI). So I'll leave the PR as is at this time. If you still think it's best not to include it in the PR, please go ahead and create the new issue and make the change to the two instances in this PR.

echopype/utils/prov.py
emiliom: See my comment above

emiliom:
> I found that filenames is correctly displayed in Sv_ds. However, it appears in MVBS_ds as an unused coordinate without a dimension (please see picture below). I am not sure if this occurs in other downstream processes, but it appears that it is caused by changes in this PR.

I've addressed this in my latest commit. While doing this, I also looked at the outcomes in `calibrate.compute_TS`, `preprocess.compute_MVBS_index_binning` and `preprocess.remove_noise`, for 4 different raw files. The unused `filenames` coordinate was only occurring with `preprocess.compute_MVBS`

emiliom:
@b-reyes I'm done. Once you've decided whether you'd like to revert the "conversion" > "combination" change (and applied it, if that's the case), please merge the PR.

echopype/echodata/combine.py
b-reyes: Ok, as long as all tests are passing this change is fine.

b-reyes (approved):

b-reyes:
@emiliom thank you for addressing all of my comments, everything looks good. I will go ahead and merge.

Issue #810, opened 2022-09-16
Account for `conversion` -> `combination` change in `Provenance` attribute


b-reyes:
In PR #806 `conversion` was changed to `combination` in `combine.py`. This causes issues when displaying a combined file because we specifically obtain the version from the `Provenance` attributes: 

https://github.com/OSOceanAcoustics/echopype/blob/47b827a927f73b7e00f3a27e48b6509037a1732a/echopype/echodata/echodata.py#L183-L186

I don't believe this will causes issues with other portions of the codebase.

emiliom:
Based on discussion in #811, I'm changing the Milestone tag to 0.6.4

leewujung:
This seems also an issue that we are in a position to address once and for all.

leewujung:
@emiliom and @leewujung discussed this: we will not remove these provenance attributes and just use a long `History` attribute to capture all the functions applied to the data. Some more thinking required to figure out what to do with the outcome of `combine_echodata`. We might add a set of equivalent attributes with name `combinatino_*`.

emiliom:
The `conversion_*` provenance attributes are actually specified by the convention and are "mandatory if applicable or available" (MA). If I tried to change them earlier (#806), that was a mistake! So, we will definitely not remove these attributes. 

Adding equivalent `combination_*` attributes when combining data (echodata objects) may be a good way to go. Then, the `version_info` method mentioned above can be modified so that it first checks if a `conversion_software_version` attribute exists; if it does, it'll use it to read the echodata (=echopype) version info, otherwise it falls back to the `conversion_software_version`.

But before settling on that scheme, I'll ask the SONAR-netCDF4 community about the interpretation of these "conversion" provenance attributes. Specifically, should a combine-data processing step be interpreted as another type of conversion step? In that case, we should just reuse the `conversion_*` attributes. The SONAR-netCDF4 documents provide too little information to make that determination; they just say that these attributes refer to the "software used to do the conversion". Is combination a type of conversion?

I'll hold off on working on this until PR #1042 is merged. But I'll plan to ask the SONAR-netCDF4 community soon.

leewujung:
I think reusing the “conversion” set of attributes to store the “combination” information is potentially problematic, because we will lose the info pertaining to each originally converted files. It will also make it much harder to keep track of what’s converted from the individual raw file and what’s combined afterwards (and therefore does not 1-1 corresponding raw files). I suggest that we just keep “conversion” and “combination” separate.

[...]

PR #812, opened 2022-09-16
Revert prov 'combination' process_type back to 'conversion'

emiliom:
A change I made in #806 had downstream ramifications that are best left to the next release, 0.6.4.

I am reverting the change the provenance `process_type` used in `combine_echodata`, from "combination" back to "conversion".

I'll self merge once the CI tests are complete.

See also discussions in #810 and #811


emiliom:
FYI it's possible that selective testing was applied in the CI, but I successfully ran the complete test suite locally.

PR #814, opened 2022-09-16
Handle xml_path in EchoData provenance on open_raw and combine_echodata

emiliom:
Initial implementation of #677, for discussion. For AZFP, adds new Provenance variable, `meta_source_filenames`, parallel to the existing `source_filenames` and reusing the `filenames` dimension. The capability is not hard-wired to AZFP, though, and can be easily reused in future instruments having an external "metadata" file.

Does not include downstream propagation beyond `open_raw`; specifically, it does not propagate it in `compute_Sv` and `combine_echodata`. I think there are potentially larger issues with doing that, especially for `compute_Sv`, which may require more discussion. We may decide to delay that for release 0.6.4.

I'll set this PR to draft mode. We may also decide not to include it in release 0.6.3.


leewujung:
> Does not include downstream propagation beyond `open_raw`; specifically, it does not propagate it in `compute_Sv` and `combine_echodata`.

The provenance for output of `compute_Sv` is pretty light right now and needs a thorough review and implementation, so I think we can (and likely should) delay to 0.6.4. 

I think we should include the propagation as part of `combine_echodata` since with our currently dump-everything-in-one-bucket type of attributes in the Provenance group, `source_filenames` is the only thing that is easy to use to figure out what happened to the combined data.

Please see #816 for my thoughts related to this but with a larger context.

emiliom:
> I think we should include the propagation as part of `combine_echodata` since with our currently dump-everything-in-one-bucket type of attributes in the Provenance group, `source_filenames` is the only thing that is easy to use to figure out what happened to the combined data.

@leewujung I'm not sure I follow what specifically you're recommending. Propagating `meta_source_filenames` into a combined dataset would invariably involve `combine_echodata`. It would not be hard at all for me to add this to this PR, following what's done with `source_filenames`. I hesitate because I think there's a more complicated question about provenance. `source_filenames` and `meta_source_filenames` provide the source of the raw data. But `combine_echodata` will typically operate on previously converted data, on converted files. Those files may have been generated from raw data seconds ago, months ago, or longer. So, the provenance question becomes more complicated: do we track the paths to those previously converted files, too? If so, how? Do we also propagate the echopype version used for conversion, in addition to the one for the combine operation?

For this release, we could:
- Not propagate `meta_source_filenames`. Don't invest effort now, and leave it to the next release when we can focus more attention w/o adding delays to this release.
- Propagate `meta_source_filenames` just as `source_filenames`. The new `combine_echodata` functionality being developed in this release may have to explicitly account for this new variable, for this release. And things may change later.

leewujung:
Below are for when we actually propagate `meta_source_filenames` down to the combined data set.


> So, the provenance question becomes more complicated: do we track the paths to those previously converted files, too?

I think we can just propagate what was in the original `source_filenames` before, so that there is no loss of information. It is useful to know where the original data source was located (like from a local drive or from s3, etc). This original path info is not important for most analyses, but could be useful in diagnosing issues like files missing from the combined data set, confirm the original source of data, etc.

> If so, how? Do we also propagate the echopype version used for conversion, in addition to the one for the combine operation?

We previously decided to NOT allow combining raw data converted using different version of echopype. So, how about we add another attribute to store the previous conversion software and version?

emiliom:
For the upcoming release (0.6.3), we've decided not to handle the downstream propagation of `meta_source_filenames` during combine and calibration. They'll be addressed in the following release.

This PR already implements the handling of `meta_source_filenames` in `open_raw`. The plan is to include that capability in 0.6.3. @leewujung this capability is ready for review. I'm removing the "draft" flag/status from the PR.

leewujung:
> we've decided not to handle the downstream propagation of `meta_source_filenames` during combine and calibration. They'll be addressed in the following release

I came away with a different understanding -- I think we should propagate `meta_source_filenames` to the combined `EchoData` but not calibration. That seems straightforward to include?

b-reyes:
> I came away with a different understanding -- I think we should propagate meta_source_filenames to the combined EchoData but not calibration. That seems straightforward to include?

Since this new variable will have `filenames` as a dimension, my code can easily handle this. I don't really see any issues with allowing it to propagate to `combine`.

emiliom:
@leewujung I was basing my statement about timing (whether to propagate `meta_source_filenames` in `combine_echodata` in 0.6.3) in part on your last comment here on this topic:

> I think this should be part of the Provenance work, and has to be done soon, but I really would like the combine_echodata new functionality to combine large number of files to be completed. Since this is not central to that issue, we can do it in v0.6.4.

As I [mentioned above](https://github.com/OSOceanAcoustics/echopype/pull/814#issuecomment-1251821928), propagating it using the same propagation scheme used with `source_filenames` will be easy. With @b-reyes commenting that he doesn't anticipate complications vis a vis his new combine code, I can go ahead and do it now. I'll submit a commit to this PR.

leewujung:
Sounds good. Thanks @emiliom @b-reyes !

echopype/echodata/combine.py
```diff
@@ -273,20 +281,23 @@ def in_memory_combine(
         if group_path is None:
             group_path = "Top-level"
 
-        for echodata in echodatas:
-            if echodata[group_path] is not None:
-                group_datasets.append(echodata[group_path])
+        for ed in echodatas:
```
emiliom, line 284: Note: I renamed the variable `echodata` to the shorter `ed`, here and below, for readability. IMHO.

emiliom:
I've pushed changes to support propagation of `meta_source_filenames` (when present) into the combined echodata object. I tested it on ek60 (ie, no `xml_path` entries) and azfp (has `xml_path` entries) files.

While creating this PR and #806, I've thought about tests. There are no tests (I think) that check for the correct creation of these file provenance variables. More broadly, I don't think we do much testing on provenance values, and that's probably a tricky area. Rather than trying to create very narrow tests in this PR, I think we should have a broader discussion. Maybe that could be folded into provenance work in 0.6.4.

echopype/echodata/combine.py
leewujung: I like this. I always have problems reading code using `x` and `xs` in loops... convenient but can easily be confusing.

echopype/utils/prov.py
```diff
@@ -5,6 +5,9 @@
 from _echopype_version import version as ECHOPYPE_VERSION
 from typing_extensions import Literal
 
+# TODO: It'd be cleaner to use PathHint, but it leads to a circular import error
+# from ..core import PathHint
```
leewujung, line 9: Do you mean that defining `ProcessType` and doing `from ..core import ProcessType` (instead of importing `PathHint`) here lead to a circular import error?

leewujung (approved):
@emiliom : thanks for the PR! I have a minor comment that does not require action, I was just wondering what you meant.

I have an overarching question on the function design:
`utils/prov::source_files_vars` returns 3 single-element dicts with keys almost identical to the names of the dicts, and values being tuples that can be passed in to construct xarray dataArrays. Why not just a single dict with those same key-value pairs? The downstream code don't seem like would be impacted much.

leewujung:
Hey @emiliom : @b-reyes and I just talked, and we decided that we'll merge this PR right now so that @b-reyes can run through these changes with his routine (which should _just_ work). I'll copy-paste the last section of my comment to the corresponding issue and point to this PR.

Issue #730, opened 2022-06-15
Add function to produce NASC


leewujung:
NASC (nautical area backscattering coefficients) is a high-level data product intimately related to biomass estimation. It can be viewed as a cleaned up form of the acoustic data (after removing noise, removing bottom, classifying regions of particular organisms, etc.) at regular spatial intervals along depth and geographical locations (say 10 m x 0.5 nmi "cells"). 

The hake survey uses PRC_NASC implementation in Echoview documented in this page for generating NASC: https://support.echoview.com/WebHelp/Reference/Algorithms/Analysis_variables/PRC_ABC_and_PRC_NASC.htm

leewujung:
This is addressed in #1005. The remaining issues are documented in #1014.

Issue #729, opened 2022-06-15
Add function to remove bottom


leewujung:
Given an existing bottom annotation generated from either manual editing or automatic detection, we need a function to remove everything below the bottom before computing NASC.

For this function it will be good to add an argument for users to put in a buffer zone, since in many processing people use 1 m or 3 m offset to avoid getting bottom echoes into biomass estimates.

This can probably use or be coordinated with functions in `echoregions` -- @valentina-s tagging you for input here.

leewujung:
We need to add some bottom detection algorithms at some point, but I think we can close this one now that we know the bottom mask creation given existing annotation will be handled by `echoregion` and the masking applied using `mask.apply_mask`.

Issue #726, opened 2022-06-15
Add function to regrid Sv data


leewujung:
It is a pretty common need to regrid Sv data, since different channels (frequencies) may be configured differently to collect data with different `sample_interval` even if the total range collected is the same. 

Proposal: add this to the `preprocess` subpackage.

leewujung:
@leewujung will add a link here to a ref implementation from IMR.

lsetiawan:
This should be optional... potentially use: https://xesmf.readthedocs.io/en/latest/index.html

leewujung:
I found my notes from before. Here are two resources, one from IMR and the other pyEcholab.

IMR:
- it seems that `_resampleWeight`, `_regrid`, and `regrid_sv` are the functions used by IMR [here](https://github.com/CRIMAC-WP4-Machine-learning/CRIMAC-preprocessing/blob/master/CRIMAC_preprocess.py#L355-L463).
- it also seems that the ditched the resampleWeight thing in [this line](https://github.com/CRIMAC-WP4-Machine-learning/CRIMAC-preprocessing/blob/master/CRIMAC_preprocess.py#L455) under `regrid_Sv`

pyEcholab:
- [grid.py](https://github.com/CI-CMG/pyEcholab/blob/RHT-EK80-Svf/echolab2/processing/grid.py) seems to be generating just the grid boundaries
- [integration.py](https://github.com/CI-CMG/pyEcholab/blob/RHT-EK80-Svf/echolab2/processing/integration.py) they basically do integration over pixels in linear domain that fall within the grid computed from grid.py, and then converting it to the log domain
- their grids here are much larger and since it is integration, it corresponds better with NASC, especially because there is "layer" and "interval" in the language, so not really the finer regridding we are discussing here.

lsetiawan:
@leewujung Could you point us to a test data for this?

leewujung:
I haven't generated any datasets for this. I think to generate outputs from the IMR functions and test against it would be the way to go. The input would be Sv with potentially irregular spacing in ping time and depth across channels, and output would be Sv gridded to the same ping time and depth grid across all channels.

anantmittal:
Explore `mock_Sv_dataset_irregular` and `mock_Sv_dataset_regular` fixture in `echopype/tests/commongrid/conftest.py`

anantmittal:
@leewujung How should we decide the correct ping time?

anantmittal:
[...]

PR #817, opened 2022-09-17
Prototype subpackages along processing levels

leewujung:
**NOT** ready for review/discussion yet!  I am still moving things around and will request for review when ready.

This PR prototypes subpackages that lie along data processing levels to bring data beyond the calibrated Sv. These include (from low to high processing levels, with some mixing in between):
- `convert` (existing)
- `calibrate` (existing)
- `filter`: remove noise or reducing variability in the data
- `unify`
- `mask`
- `metrics`: summary statistics and NASC


leewujung:
@emiliom : a question for you:
Currently we have `qc` as a subpackage that only contains checking and correcting for the monotonicity of the `ping_time` coordinate. But in this framing PR I put noise removal functions (e.g., thresholding based on estimated background noise, removing pings with spikes) in the `filter` subpacakge along with smoothing functions (e.g., median filter). Do you think all of these should just be in `filter`? Somehow it feels awkward to think of correcting time coordinates as a type of filtering, since it is more like rectifying things, but maybe it is ok?

emiliom:
@leewujung can you add links to references or web pages that describe the calculations?

Below I flesh out individual modules or functions as found in this PR, so we can see it in one place.

- `consolidate` (existing), not in @leewujung's list above but included among the modules edited in this PR
  - `add_location` (existing)
  - `add_splitbeam_angle`
- `convert` (existing)
- `calibrate` (existing)
- `filter`: remove noise or reducing variability in the data
  - `median`
  - `conv`
  - `remove_noise`
  - `noise`
    - `estimate_noise`
    - `mean_bkg`
    - `spike`
- `unify`
  - `compute_MVBS`
  - `compute_MVBS_index_binning`
  - `regrid_Sv`
- `mask`
  - `from_Sv`
    - `freq_diff`
  - `from_labels`
    - `boundary`
    - `region`
- `metrics`
  - summary statistics
  - `nasc`
    - `compute_NASC`

emiliom:
@leewujung can you add links to references or web pages that describe the calculations?

Below I flesh out individual modules or functions as found in this PR, so we can see it in one place.

- `consolidate` (existing), not in @leewujung's list above but included among the modules edited in this PR
  - `add_location` (existing)
  - `add_splitbeam_angle`
- `convert` (existing)
- `calibrate` (existing)
- `filter`: remove noise or reducing variability in the data
  - `median`
  - `conv`
  - `remove_noise`
  - `noise`
    - `estimate_noise`
    - `mean_bkg`
    - `spike`
- `unify`
  - `compute_MVBS`
  - `compute_MVBS_index_binning`
  - `regrid_Sv`
- `mask`
  - `from_Sv`
    - `freq_diff`
  - `from_labels`
    - `boundary`
    - `region`
- `metrics`
  - summary statistics. See initial implementation, [metrics/summary_statistics.py](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/metrics/summary_statistics.py)
  - `nasc`
    - `compute_NASC`

b-reyes:
Based on discussion with @leewujung and @emiliom we have agreed that the mask sub-package functions should only return a mask and not apply the mask. For this reason, we need to create a function that applies masks, say `apply_mask`. This function will take in a Dataset or DataArray (this is still up for discussion) and a list of masks to apply to the input Dataset or DataArray. Additionally, we need to decide where this function should live (perhaps the mask sub-package is the most natural location).

leewujung:
Thanks for the summary @b-reyes. My inclination is:

> This function will take in a Dataset or DataArray (this is still up for discussion) and a list of masks to apply to the input Dataset or DataArray.

My inclination is to restrict both to be a Dataset for the following reasons:
- consistency: all of our other processing functions take in/produce Dataset, so this is easy to remember
- flexibility: allow users to optionally store other metadata with the mask zarr file without interfering with the functionality
- xarray `to_zarr` converts DataArray to Dataset before storing into zarr
- xarray `open_dataset` reads zarr into a Dataset and we are familiar with its behavior

> Additionally, we need to decide where this function should live (perhaps the mask sub-package is the most natural location).

I agree that under the `mask` subpackage is the best place.

emiliom:
> Additionally, we need to decide where this function should live (perhaps the mask sub-package is the most natural location).

I, too, agree with `apply_mask` being in the `mask` subpackage.

> This function will take in a Dataset or DataArray (this is still up for discussion) and a list of masks to apply to the input Dataset or DataArray.

I'm on board with the function taking in an Sv Dataset. But I'd like to go further in the specification:
- In principle if more than one mask is passed to `apply_mask` , these may come in the form of more than one Dataset. The names of the mask variables will most likely not be the same across masks. So, to fully specify a mask for use in `apply_mask`, the variable's Dataset (or path) and name will be needed.
- The Sv source will be also specified via a Dataset and the name of the Sv variable name. Keep in mind that echopype can generate an Sv variable from either `compute_Sv`, in which case the variable name is `Sv`, or from `preprocess.remove_noise`, in which case the name is `Sv_corrected`.
- `apply_mask` output:
  - `apply_mask` will return a Dataset, right? Should this be a new Dataset, or is there a use case for also adding the new masked Sv to the source Sv dataset? 
  - provide an option to include the mask variable in the output Dataset; if multiple masks were passed, only the single, combined mask should be included.

It may be helpful to look over the `estimate_noise` and `remove_noise` functions while working on this. There are many parallels.

b-reyes:
> I, too, agree with apply_mask being in the mask subpackage.

Great, it sounds like we all agree on the location of the `apply_mask` function!

> In principle if more than one mask is passed to apply_mask , these may come in the form of more than one Dataset. The names of the mask variables will most likely not be the same across masks. So, to fully specify a mask for use in apply_mask, the variable's Dataset (or path) and name will be needed.

This is a good point that I had not thought of. Also supplying the name seems like a cumbersome task for the user. Perhaps this function should only take in a DataArray? That way we do not have to worry about this. 

> The Sv source will be also specified via a Dataset and the name of the Sv variable name. Keep in mind that echopype can generate an Sv variable from either compute_Sv, in which case the variable name is Sv, or from preprocess.remove_noise, in which case the name is Sv_corrected.

Another good point and something that is not accounted for in #901. I do not like the idea of the user needing to supply this name, but it seems like it might be needed. @leewujung pinging you as we should decide on this before full review of #901. 

> apply_mask will return a Dataset, right? Should this be a new Dataset, or is there a use case for also adding the new masked Sv to the source Sv dataset?

I would assume that it would output a new Dataset, so the original source could remain unmodified. However, we could always make this an input flag, say `inplace`. 

> provide an option to include the mask variable in the output Dataset; if multiple masks were passed, only the single, combined mask should be included.

I think including the mask in the output Dataset is not preferable. The masks are already an input of the function, so the user could already save them, if they want to. Additionally, attaching the mask would increase the data size. However, If we really want to make this an option, we could have a keyword, say `include_mask`. We could also include all masks individually that were passed and attach them as variables using their provided names (thinking about the size increase maybe the single combined mask is a better idea).

b-reyes:
Based on discussions between @leewujung and myself, we have arrived at the following for `apply_mask`:
- It will be in the `mask` sub-package and thus be accessed as `echopype.mask.apply_mask(...)`
- It will have the following function inputs:
    - `source_ds` - a Dataset or a path to a Dataset that contains the variable we want to apply a mask to. 
    - `var_name` - a string corresponding to the variable name in `source_ds` that the mask should be applied to
    - `mask` the actual mask to apply to the variable specified by `var_name`. This input can be a list or a single input that corresponds to a DataArray or a path to a DataArray
    - `fill_value` - specifies the value(s) at false indices. This variable can be a scalar, array, Variable, or DataArray (see [xr.where](https://docs.xarray.dev/en/stable/generated/xarray.where.html) for descriptions of these types).
- The output will be a Dataset with the same format of `source_ds` with the mask applied to `var_name`
    - We will not save this output for them to a zarr or netcdf
    - If `source_ds` is lazy, so will the output Dataset (likewise for in-memory)
- Other items discussed:
    - We will not store the final mask or any mask in the output Dataset
    - When implementing the method we will use [xr.where](https://docs.xarray.dev/en/stable/generated/xarray.where.html)  with `x=source_ds[var_name]` and `y=fill_value`

b-reyes:
In the above comment I mention that we should let `fill_value` be a [xr.Variable](https://docs.xarray.dev/en/stable/generated/xarray.Variable.html#xarray.Variable). After further investigation, I think we should not allow for this type. The reason for this is the documentation states: "However, manipulating data in the form of a Dataset or DataArray should almost always be preferred [in comparison to a xr.Variable], because they can use more complete metadata in context of coordinate labels."

leewujung:
I agree with omitting `xr.Variable`.

emiliom:
> `fill_value` - specifies the value(s) at false indices. This variable can be a scalar, array, ~~Variable~~, or DataArray (see [xr.where](https://docs.xarray.dev/en/stable/generated/xarray.where.html) for descriptions of these types).

Just a comment that in the context of xarray (and netcdf/zarr) a "fill value" is typically (or always?) a scalar. It's intended to be literally a single value. I see that the types you've listed come directly from `xr.where`, which makes sense in the more general context of that function. But in the context of applying a mask, we're talking about a scalar. Or, if you actually intended something different, I suggest you use a name other than `fill_value`.

b-reyes:
In addition to the inputs provided above for `apply_mask`, we will also need storage options for both the `source_ds` and `mask` inputs (as they can be paths). @leewujung and I have settled on `storage_options_ds: dict = {}` and `storage_options_mask: Union[dict, List[dict]] = {}` for `source_ds` and `mask`, respectively.

emiliom:
Current state of individual modules or functions, existing and planned, organized by subpackage. The order of subpackages roughly reflects processing order.

**TODO:**
- Add one-sentence descriptions of every subpackage category

(E) = Existing

- `convert` (E). *Includes here combine_echodata capability?*
- `qc`
  - `coerce_increasing_time` (E)
  - `exist_reversed_time` (E)
- `calibrate` (E)
- `consolidate` (E), not in @leewujung's list above but included among the modules edited in this PR
  - `add_location` (E)
  - `add_depth` (E)
  - `add_splitbeam_angle` (E)
  - `swap_dims_channel_frequency` (E). *This is a lower-level function that rearranges data a bit, doesn't create a new product or add new data per se. Move it to `qc`?*
- `filter`: remove noise or reducing variability in the data
  - `median`
  - `conv`
  - `remove_noise` (E)
  - `estimate_noise` (E)
  - `mean_bkg`
  - `spike`
- `commongrid`
  - `compute_MVBS` (E)
  - `compute_MVBS_index_binning` (E)
  - `regrid_Sv`. See #726
- `mask`
  - `apply_mask` (E)
  - `frequency_differencing` (E)
  - `from_labels`
    - `boundary`
    - `region`
    - For bottom removal discussion, see #729
- `metrics`
  - summary statistics. See initial (draft?) implementation at [metrics/summary_statistics.py](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/metrics/summary_statistics.py)
    - `delta_z` (E)
    - `convert_to_linear` (E)
    - `abundance` (E)
    - `center_of_mass` (E)
    - `dispersion` (E)
    - `evenness` (E)
    - `aggregation` (E)
  - `nasc`
    - `compute_NASC`. See #730

leewujung:
I'll close this PR now since it is very outdated. Some of the functions mentioned here in prototype form have been implemented, and others either won't be implemented in echopype (those related to region and line labels under `mask` -- they are now under `echoregions`) or are in active PRs (those related to noise removal under `clean`).

PR #820, opened 2022-09-22
🐛 Delete 'v' in version string for really old version to avoid error

lsetiawan:
## Overview

This PR is a small one to catch really old versions of echopype in Provenance. OOI used converted file that went back to 0.4.1 and the string for the version is `v0.4.1`. Rather than getting a nice not implemented error that @b-reyes has: `NotImplementedError: Conversion from echopype v0.4.1 to ...` we currently get `ValueError: invalid literal for int() with base 10: 'v0'`... so this PR fixes that bug.


leewujung:
This looks good.

PR #828, opened 2022-10-04
Change isel to sel to fix compute_Sv

lsetiawan:
This PR adress https://github.com/OSOceanAcoustics/echopype/pull/824#issuecomment-1267463453 where `idxmin` is actually xarray data array object not integers, so `isel` doesn't work.

### DIFF
--- a/echopype/calibrate/calibrate_ek.py
+++ b/echopype/calibrate/calibrate_ek.py
@@ -94,7 +94,7 @@ class CalibrateEK(CalibrateBase):
         )
 
         # Select corresponding index and clean up the original nan elements
-        da_param = da_param.isel(pulse_length_bin=idxmin, drop=True)
+        da_param = da_param.sel(pulse_length_bin=idxmin, drop=True)
         return da_param.where(~transmit_isnull, np.nan)  # set the nan elements back to nan
 
     def get_cal_params(self, cal_params, waveform_mode, encode_mode):


b-reyes:
@lsetiawan from my understanding, we should replace all `isel` operations on something that could potentially be a Dask array. Searching through just the `calibrate` folder, I found a couple of other occurrences of `isel`. Could you look into the following occurrences and determine if they also need to be changed? 
- In `calibrate_ek.py` Line 852: `return out.isel(beam=0).drop("beam")`
- In `env_params.py` Line 146: `extrap = extrap.isel(**extrap_unique_idx)`
- In `env_params.py` Line 152: `interp = interp.isel(**interp_unique_idx)`

echopype/calibrate/calibrate_ek.py
```diff
@@ -94,7 +94,7 @@ def _get_vend_cal_params_power(self, param, waveform_mode):
         )
 
         # Select corresponding index and clean up the original nan elements
-        da_param = da_param.isel(pulse_length_bin=idxmin, drop=True)
+        da_param = da_param.sel(pulse_length_bin=idxmin, drop=True)
```
b-reyes, line 97: I looked into this change and everything looks good.

lsetiawan:
> In env_params.py Line 146: extrap = extrap.isel(**extrap_unique_idx)
> In env_params.py Line 152: interp = interp.isel(**interp_unique_idx)

For these two, I'm not sure how to check for these, I'm not familiar enough with the `env_params` module. @leewujung I need your guidance on this. Thanks.

b-reyes:
> For this one, the value is clearly 0 so isel is appropriate, I'll take look at the other files.

Yes, I understand that `isel` would be appropriate here. However, from my understanding, if `out` is referring to a Dask array then `out.isel` will not work. Here, I think we are just trying to drop any lingering `beam` dimension. So, one could just find the first value in the `beam` dimension and use this value to replace `isel` with `sel`.

b-reyes (approved):
Everything looks good!

Issue #796, opened 2022-08-25
"WindowsPath" object is not iterable


IanTBlack:
I am trying to run through the [OOI EK60 Mooring Example](https://github.com/OSOceanAcoustics/echopype-examples/blob/main/notebooks/ms_OOI_EK60_mooringtimeseries.ipynb) and have received an error `"WindowsPath" object is not iterable` when running the line `ds_Sv = ep.calibrate.compute_Sv(ed)` in the "Convert Multiple Files" section.

I receive this error when running the example as a notebook in Pycharm and when running it line by line in my system console.

My work around for this is to loop through each file individually and append them to a list. ep.open_raw > ep.calibrate.compute_Sv > ep.preprocess.compute_MVBS. Once each file has been process individually, I use xarray.merge to merge all of the datasets together. 
I've had mixed success with this method.

Is there something I can change to avoid the `WindowsPath` error? Thanks!

leewujung:
Hi @IanTBlack : Could you let us know what echopype version you are using, whether you install echopype using conda or pip, and provide the full trace back error messages? These will help us with isolating the potential problem. Thanks!

IanTBlack:
Echopype Version: 0.6.2
Install Method: pip

Full Traceback:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [24], in <cell line: 1>()
----> 1 ds_Sv = ep.calibrate.compute_Sv(ed)

File ~\AppData\Roaming\Python\Python39\site-packages\echopype\calibrate\api.py:201, in compute_Sv(echodata, **kwargs)
    115 def compute_Sv(echodata: EchoData, **kwargs) -> xr.Dataset:
    116     """
    117     Compute volume backscattering strength (Sv) from raw data.
    118 
   (...)
    199     it must be set using `EchoData.update_platform()`.
    200     """
--> 201     return _compute_cal(cal_type="Sv", echodata=echodata, **kwargs)

File ~\AppData\Roaming\Python\Python39\site-packages\echopype\calibrate\api.py:103, in _compute_cal(cal_type, echodata, env_params, cal_params, waveform_mode, encode_mode)
    100 else:
    101     source_file = "SOURCE FILE NOT IDENTIFIED"
--> 103 cal_ds["source_filenames"] = source_files_vars(source_file)["source_filenames"]
    104 prov_dict = echopype_prov_attrs(process_type="processing")
[...]

PR #829, opened 2022-10-05
Fix WindowsPath error with compute_Sv when run on Windows

emiliom:
Fixes the error identified in #796. 

I tested it on a Windows computer by running the relevant parts of https://osoceanacoustics.github.io/echopype-examples/echopype_tour.html. 

All CI tests are passing on Ubuntu (ie, no negative impact on Linux).

### DIFF
--- a/echopype/utils/prov.py
+++ b/echopype/utils/prov.py
@@ -1,5 +1,5 @@
 from datetime import datetime as dt
-from pathlib import PosixPath
+from pathlib import Path
 from typing import Any, Dict, List, Tuple, Union
 
 from _echopype_version import version as ECHOPYPE_VERSION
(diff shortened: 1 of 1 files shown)


echopype/utils/prov.py
```diff
@@ -1,5 +1,5 @@
 from datetime import datetime as dt
-from pathlib import PosixPath
+from pathlib import Path
```
lsetiawan, line 2: Oh wow, didn't realize that this module was locked into `PosixPath`! Nice fix @emiliom! Do you know if this is the only place that `PosixPath` is found?? I hope so!

emiliom: > Do you know if this is the only place that PosixPath is found??

It looks that way.

> Oh wow, didn't realize that this module was locked into PosixPath!

My bad :disappointed: . I think I introduced that.

lsetiawan (approved):
LGTM. At this point since you've tested it locally I'll approve this. This seems like a critical fix. Thanks for fixing this! I'm glad that you found a windows environment!

I realized that we don't have a test for the `utils/prov.py`, maybe this will be great for our testing creation session in the future 😄

emiliom:
Thanks @lsetiawan ! I'll merge it.

I opened a new issue for developing prov tests, #831

Issue #834, opened 2022-10-06
`ValueError` during compute_MVBS


lsetiawan:
## Overview

`ValueError` was discovered while running the [OOI_EK60_mooringtimeseries](https://osoceanacoustics.github.io/echopype-examples/ms_OOI_EK60_mooringtimeseries.html) example notebook during a review of #824.

## Troubleshooting

While my initial thought of this issue was caused by the new `combine_echodata` method, further investigation suggest that this issue started appearing in version `0.6.1`!
- 0.6.0: Kernel died when trying to run through all cells up to `compute_MVBS`. Had to re-run from getting converted files, combining, then calculate Sv and MVBS. Other than that, everything ran through successfully.
- 0.6.1: Kernel died when trying to run through all cells up to `compute_MVBS`. Had to re-run from getting converted files, combining, then calculate Sv and MVBS. During the `compute_MVBS` I get 
    ```python
    ValueError: echo_range variable changes across pings in at least one of the frequency channels.
    ```
- 0.6.2: Everything ran smooth without any memory spikes up to `compute_Sv` with `offload_to_zarr=True`. When I got to `compute_Sv` it worked but was in borderline of my max memory capacity, running at about 98% with fully using all of my swap, but successfully get `ds_Sv`. During the `compute_MVBS` I get 
    ```python
    ValueError: echo_range variable changes across pings in at least one of the frequency channels.
    ```

Looking at git history, seems like the current code change to `compute_MVBS` happened in PR https://github.com/OSOceanAcoustics/echopype/pull/753. The full error points to the line with this change: https://github.com/OSOceanAcoustics/echopype/pull/753/files#diff-2848209b968526da06d5475e3df089d5a7138b944681119fbe9f64f9aad81603R118.

I'm not sure why this issue isn't caught during CI testing, but seems like there's a current issue to rework the test for `compute_MVBS` see https://github.com/OSOceanAcoustics/echopype/issues/756. 

## Full error

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [34], in <cell line: 1>()
----> 1 ds_MVBS = ep.preprocess.compute_MVBS(
      2     ds_Sv, 
      3     range_meter_bin=0.2,  # 0.2 meters
      4     ping_time_bin='10S'   # 10 seconds
      5 )

[...]

PR #837, opened 2022-10-08
Fix interpolating over NA issue in `EchoData._harmonize_env_param_time()`

leewujung:
In `EchoData._harmonize_env_param_time()` the interpolation from `Environment/time1` to `Sonar/Beam_group1/ping_time` was done without considering `NA` that may exist in `Environment/SOME_VAR`. This can result in producing additional `NA` at the locations adjacent to the `NA` entry. 

This interpolation was added in #755 to fix the problem from mismatched timestamp names in the `Environment` and `Sonar/Beam_groupX` groups.

This PR drops `NA`s first before interpolation. This should fix the issue in #834.

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -296,8 +296,7 @@ class EchoData:
                     return p.squeeze(dim="time1").drop("time1")
                 else:
                     if ping_time is not None:
-                        p_nona = p.dropna(dim="time1")
-                        return p_nona.interp(time1=ping_time)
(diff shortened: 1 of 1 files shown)


leewujung:
Oops @lsetiawan , it seems that this broke the plotting test... https://github.com/OSOceanAcoustics/echopype/actions/runs/3229962711/jobs/5287864284

lsetiawan:
Ooops! I should've tested on all the test suite! 😞

PR #844, opened 2022-10-11
Add requests and aiohttp to dependency

lsetiawan:
## Overview

When installing echopype from pip, I get error shown below when trying to do `open_raw` from an `http` source. This seem to be caused by `requests` and `aiohttp` libraries not being included in requirements by default. So this PR adds them in so that we don't have issues in the future and users don't need to do extra step to install it.

## Error screenshot

Screenshot from 2022-10-11 10-17-38

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,8 @@ xarray==2022.3.0
 zarr
 fsspec
 s3fs
+requests
+aiohttp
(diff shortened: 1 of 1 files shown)


emiliom (approved):
This seems straightforward, and the tests passed.

Thanks for identifying and addressing this issue.

Issue #847, opened 2022-10-12
Create issue template for "my files don't convert" problems


leewujung:
We see "my files don't convert" issues pretty frequently.

It will be nice if folks having this type of issues provide us a list of things that will help with debugging:
- echopype version and how it was installed (e.g., conda or pip)
- computing environment 
- minimum reproducible code snippet that generate the error
- error message printout
- link to an example file (as small as possible)

I added a "my_files_don't_convert" label so that we can identify this type of issues.

I think we can create an issue template for this so that it is clear the type of info required to facilitate resolving the issues, and the issues will be automatically tagged with the above label.

PR #850, opened 2022-10-13
Add "My files don't convert" issue template

leewujung:
This addresses #847.

### DIFF
--- a/.github/ISSUE_TEMPLATE/my-files-don-t-convert.md
+++ b/.github/ISSUE_TEMPLATE/my-files-don-t-convert.md
@@ -32,7 +32,8 @@ ERROR MESSAGES
 
 ### Provide an example file
 _Required: An example file that allow reproducing the problem._
-Here is a link to an example file: SOME_ONLINE_LOCATION
+- Link to an example raw file: SOME_ONLINE_LOCATION
+- Raw file sensor model: (FOR EXAMPLE: Simrad EK60)
 
 ### Troubleshooting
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/my-files-don-t-convert.md
```diff
@@ -0,0 +1,38 @@
...
+### Error message printouts
+_Required: Copy-paste the entire error messages you encounter here._
+Below is the error messages I received when running the above code:
+```python
+ERROR MESSAGES
+```
+
+### Provide an example file
+_Required: An example file that allow reproducing the problem._
+Here is a link to an example file: SOME_ONLINE_LOCATION
```
emiliom, line 35: Should we place limits on how big the file can be? Say, < 200 MB?

emiliom, line 35:
```suggestion
- Link to an example raw file: SOME_ONLINE_LOCATION
- Raw file sensor model: (FOR EXAMPLE: Simrad EK60)
```

emiliom:
I made a couple of small suggestions. Otherwise, it looks good, thanks!

.github/ISSUE_TEMPLATE/my-files-don-t-convert.md
leewujung: Yeah that will be good. I'll add that.

Issue #233, opened 2021-01-07
Consolidate Zarr metadata after conversion


ngkavin:
Consolidating metadata will greatly increase the speed of reading array metadata, as described here: https://zarr.readthedocs.io/en/stable/tutorial.html#consolidating-metadata.

It looks like the metadata can be consolidated with `zarr.consolidate_metadata(store)` after conversion.

lsetiawan:
If it's an xarray Dataset object, all you need is `Dataset.to_zarr(consolidated=True)`

ngkavin:
Do you know if this something that should be done for every group or just the last one?

lsetiawan:
Should be all the groups I think. @emiliom @leewujung , Do you guys have any thoughts on this?

emiliom:
Only groups that are chunked. Is any group besides `Beam` chunked?

ngkavin:
Groups with a time or range_bin dimension are chunked. For EK80 this also includes the platform group.

From testing, the '.zmetadata' KeyError is raised when setting `consolidated=True` in xarray `to_zarr()` if the .zmetadata does not exist and we are appending to an existing zarr file. .zmetadata needs to be created when calling `zarr.consolidate_metadata` after creating the zarr file in the top_level group because xarray is not being used to create zarr files, only appending to them.

ngkavin:
So it might be best to simply call `zarr.consolidate_metadata` at the very end of conversion. Otherwise we would need to call it at the beginning as well as set `consolidated=True` each time we save with xarray.

leewujung:
In the latest conversation we discussed:
- potentially adding useful info to the top-level group so that it is naturally combined into zmetadata
- create functions, maybe in the `print` statement for EchoData to pull only the consolidated metadata (the "lazy-lazy" load) before the files are actually lazy-opened

We still need to:
- decide on the use cases
- benchmark the loading speed wrt different file sizes in the remote

lsetiawan:
@b-reyes consolidated needs to be True for zar to consolidate. I think all you need is to make sure that every Dataset.to_zarr has the consolidate=True flag. Shouldn't be much to address for this upcoming release.

leewujung:
From discussion: we will set it up that all `.to_zarr/netcdf` will default to `cnosolidated=True`, but users will be able to pass in kwargs to optionally overwrite the default operation.

Issue #791, opened 2022-08-19
Create default artificial test objects/datasets


leewujung:
Increasingly we have the need to unit test functionalities that are higher on the processing chain and many of the current tests right now need to go through the convert-->calibrate process before the actual functionality can be tested. This has a few drawbacks:
- hard to keep track of which test data files are used in what tests
- "real" data files do not allow the flexibility of testing what may be missing or in peculiar forms in the parsed data, corresponding EchoData object, and calibrated Sv
- data files tend to be large, especially those containing complex samples from EK80

## Proposal
We can create artificial test objects and datasets the same way we have a default test path. A few thoughts on what we may want to have for these
- can create irregularity according to what we are aware of in the parsed data (`parser` objects): use these in tests for `set_group_X`
  - examples include and are not limited to: irregular `range_sample` across channels or across `ping_time`, missing `water_depth`, missing lat/lon info
- can create irregularity in `EchoData` object: use these in tests for the subpackages `calibrate` and `consolidate`
- can create irregularity in calibrated `Sv` datasets: use these in tests for the subpackages for higher levels

I also propose that we put together a couple hackday for this since our tests are pretty adhoc at the moment and would benefit from a group discussion with a top-down approach. Once that's setup, we can identify specific tasks, and then divide the tasks to each tackle a subset.

leewujung:
Focused on EK data for now:
- @lsetiawan : hackday chief, pytest fixture setup/modules
- @b-reyes : parsed output modules/functions
- @leewujung : irregularity in files

leewujung:
I'll close this now as it is very outdated and the content also superseded by #1349.

PR #855, opened 2022-10-14
Add default consolidated flag for to_zarr

lsetiawan:
## Overview

This PR adds a default of `consolidated=True` for `to_zarr` but also allows user to modify it, per conversation from #233.

### DIFF
--- a/echopype/echodata/zarr_combine.py
+++ b/echopype/echodata/zarr_combine.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Hashable, List, Set, Tuple
 
 import dask
 import dask.array
+import fsspec
 import numpy as np
 import pandas as pd
 import xarray as xr
@@ -600,7 +601,6 @@ class ZarrCombine:
         zarr_group: str,
         region: Dict[str, slice],
         storage_options: Dict[str, Any] = {},
-        consolidated: bool = True,
     ) -> None:
         """
         Constructs a delayed write of ``ds_in`` to the appropriate zarr
@@ -627,6 +627,7 @@ class ZarrCombine:
         """
 
         with Lock(lock_name):
+            # Write to region is not consolidated by default
             ds_in.to_zarr(
                 zarr_path,
                 group=zarr_group,
@@ -634,7 +635,6 @@ class ZarrCombine:
                 compute=True,
                 storage_options=storage_options,
                 synchronizer=zarr.ThreadSynchronizer(),
-                consolidated=consolidated,
             )
 
             # TODO: put a check to make sure that the chunk has been written
@@ -646,7 +646,6 @@ class ZarrCombine:
         zarr_group: str,
         ed_name: str,
         storage_options: Dict[str, Any] = {},
-        consolidated: bool = True,
     ) -> List[str]:
         """
         Creates a zarr store and then appends each Dataset
@@ -668,9 +667,6 @@ class ZarrCombine:
         storage_options: dict
             Any additional parameters for the storage
             backend (ignored for local paths)
-        consolidated : bool
-            Flag to consolidate zarr metadata.
-            Defaults to ``True``
 
         Returns
         -------
@@ -691,7 +687,7 @@ class ZarrCombine:
             encoding=encodings,
             storage_options=storage_options,
             synchronizer=zarr.ThreadSynchronizer(),
-            consolidated=consolidated,
+            consolidated=False,
         )
 
         # get all dimensions in ds that are append dimensions
@@ -740,7 +736,6 @@ class ZarrCombine:
                             zarr_group,
                             region,
                             storage_options,
-                            consolidated,
                         )
                     )
 
@@ -756,7 +751,6 @@ class ZarrCombine:
         zarr_path: str,
         zarr_group: str,
         storage_options: Dict[str, Any] = {},
-        consolidated: bool = True,
     ) -> None:
         """
         Appends all constant (i.e. not chunked) variables and dimensions to the
@@ -776,9 +770,6 @@ class ZarrCombine:
         storage_options: dict
             Any additional parameters for the storage
             backend (ignored for local paths)
-        consolidated : bool
-            Flag to consolidate zarr metadata.
-            Defaults to ``True``
 
         Notes
         -----
@@ -800,7 +791,7 @@ class ZarrCombine:
                 group=zarr_group,
                 mode="a",
                 storage_options=storage_options,
-                consolidated=consolidated,
+                consolidated=False,
             )
 
     def _write_append_dims(
@@ -809,7 +800,6 @@ class ZarrCombine:
         zarr_path: str,
         zarr_group: str,
         storage_options: Dict[str, Any] = {},
-        consolidated: bool = True,
     ) -> None:
         """
         Sequentially writes each Dataset's append dimension in ``ds_list`` to
@@ -827,9 +817,6 @@ class ZarrCombine:
         storage_options: dict
             Any additional parameters for the storage
             backend (ignored for local paths)
-        consolidated : bool
-            Flag to consolidate zarr metadata.
-            Defaults to ``True``
         """
 
         # get all dimensions in ds that are append dimensions
@@ -854,11 +841,11 @@ class ZarrCombine:
                     compute=True,
                     storage_options=storage_options,
                     synchronizer=zarr.ThreadSynchronizer(),
-                    consolidated=consolidated,
+                    consolidated=False,
                 )
 
     def _append_provenance_attr_vars(
-        self, zarr_path: str, storage_options: Dict[str, Any] = {}, consolidated: bool = True
+        self, zarr_path: str, storage_options: Dict[str, Any] = {}
     ) -> None:
         """
         Creates an xarray Dataset with variables set as the attributes
@@ -873,9 +860,6 @@ class ZarrCombine:
         storage_options: dict
             Any additional parameters for the storage
             backend (ignored for local paths)
-        consolidated : bool
-            Flag to consolidate zarr metadata.
-            Defaults to ``True``
         """
 
         xr_dict = dict()
@@ -918,7 +902,7 @@ class ZarrCombine:
             group="Provenance",
             mode="a",
             storage_options=storage_options,
-            consolidated=consolidated,
+            consolidated=False,
         )
 
     @staticmethod
@@ -1034,7 +1018,6 @@ class ZarrCombine:
                     zarr_group=grp_info["ep_group"],
                     ed_name=ed_group,
                     storage_options=storage_options,
-                    consolidated=consolidated,
                 )
 
                 self._append_const_to_zarr(
@@ -1043,20 +1026,28 @@ class ZarrCombine:
                     zarr_path,
                     grp_info["ep_group"],
                     storage_options,
-                    consolidated,
                 )
 
                 self._write_append_dims(
-                    ds_list, zarr_path, grp_info["ep_group"], storage_options, consolidated
+                    ds_list,
+                    zarr_path,
+                    grp_info["ep_group"],
+                    storage_options,
                 )
 
         # append all group attributes before combination to zarr store
         self._append_provenance_attr_vars(
-            zarr_path, storage_options=storage_options, consolidated=consolidated
+            zarr_path,
+            storage_options=storage_options,
         )
 
         # change filenames numbering to range(len(eds))
         self._modify_prov_filenames(zarr_path, len_eds=len(eds), storage_options=storage_options)
+
+        if consolidated:
+            # consolidate at the end if consolidated flag is True
+            store = fsspec.get_mapper(zarr_path, **storage_options)
+            zarr.consolidate_metadata(store=store)
 
         # open lazy loaded combined EchoData object
         ed_combined = open_converted(
--- a/echopype/echodata/zarr_combine.py
+++ b/echopype/echodata/zarr_combine.py
@@ -628,7 +628,6 @@ class ZarrCombine:
         """
 
         with Lock(lock_name):
-            # Write to region is not consolidated by default
             ds_in.to_zarr(
                 zarr_path,
                 group=zarr_group,
--- a/echopype/echodata/zarr_combine.py
+++ b/echopype/echodata/zarr_combine.py
@@ -636,6 +636,7 @@ class ZarrCombine:
                 compute=True,
                 storage_options=storage_options,
                 synchronizer=zarr.ThreadSynchronizer(),
+                consolidated=False,
             )
 
             # TODO: put a check to make sure that the chunk has been written
--- a/echopype/echodata/zarr_combine.py
+++ b/echopype/echodata/zarr_combine.py
@@ -600,6 +600,7 @@ class ZarrCombine:
         zarr_group: str,
         region: Dict[str, slice],
         storage_options: Dict[str, Any] = {},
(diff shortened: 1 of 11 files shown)


lsetiawan:
@b-reyes this is ready for review... I hope I covered all of the `to_zarr` appropriately. Thanks.

echopype/echodata/zarr_combine.py
```diff
@@ -600,6 +600,7 @@ def write_to_file(
         zarr_group: str,
         region: Dict[str, slice],
         storage_options: Dict[str, Any] = {},
+        consolidated: bool = True,
```
b-reyes, line 603: Please add a docstring for this input variable.

echopype/echodata/zarr_combine.py
```diff
@@ -633,6 +634,7 @@ def write_to_file(
                 compute=True,
                 storage_options=storage_options,
                 synchronizer=zarr.ThreadSynchronizer(),
+                consolidated=consolidated,
```
b-reyes, line 637: You should not add consolidated here. The metadata has already been created at this point and you are recomputing it.

echopype/echodata/zarr_combine.py
```diff
@@ -830,10 +854,11 @@ def _write_append_dims(
                     compute=True,
                     storage_options=storage_options,
                     synchronizer=zarr.ThreadSynchronizer(),
+                    consolidated=consolidated,
```
b-reyes, line 857: You should not add consolidated here. The metadata has already been created at this point and you are recomputing it.

echopype/echodata/zarr_combine.py
```diff
@@ -1022,6 +1064,7 @@ def combine(
             chunks={},
             synchronizer=zarr.ThreadSynchronizer(),
             storage_options=storage_options,
+            backend_kwargs={"consolidated": consolidated},
```
b-reyes, line 1067: Why is this necessary?

b-reyes (changes requested):
@lsetiawan I am done reviewing this PR. From what I can see, It looks like you hit all of the `to_zarr` statements. There are a couple of places where `consolidated` should be removed because the consolidation has already occurred.

With these changes, have you tried to run a large number of files in `combine_echodata`? If not, can you please do that to make sure nothing has changed? 

One thing that I noticed is that in functions such as `to_zarr` in `echodata.py` you provide docstrings for `kwargs`. This seems confusing to me. Is there a numpy doc way of specifying docstring parameters that are specifically for `kwargs`?

echopype/echodata/zarr_combine.py
lsetiawan, line 637:
```suggestion
```

echopype/echodata/zarr_combine.py
lsetiawan, line 603:
```suggestion
```

echopype/echodata/zarr_combine.py
lsetiawan: Aren't you writing the merged dataset into a new zarr file here?

echopype/echodata/zarr_combine.py
lsetiawan: To be explicit that we are opening a consolidated zarr file.

lsetiawan:
> One thing that I noticed is that in functions such as to_zarr in echodata.py you provide docstrings for kwargs. This seems confusing to me. Is there a numpy doc way of specifying docstring parameters that are specifically for kwargs?

I think this was from the implicity of `to_file` function... we should probably change it so that we do list out all the kwargs. 😛

echopype/echodata/zarr_combine.py
b-reyes: Within this function we are just filling in values to a zarr store that has already been consolidated. A quick way to make this determination is to see if `region` is an input of `to_zarr`. If `region` is an input, then the metadata should already be constructed.

echopype/echodata/zarr_combine.py
b-reyes: Where/how would we use/access  this information. I am not really familiar with `backend_kwards`.

b-reyes:
> I think this was from the implicity of to_file function... we should probably change it so that we do list out all the kwargs. 😛

Yeah, I think that is best. If you think this is best too, could you create an issue for it and we can deal with it after this release?

lsetiawan:
> With these changes, have you tried to run a large number of files in combine_echodata? If not, can you please do that to make sure nothing has changed?

I won't have time to fully test this 🙈 Maybe just push this off to the next release since it's not an urgent issue.

leewujung:
@lsetiawan @b-reyes : following up on this -- could you connect and make sure the remaining issues get resolved in this week? I think it would be good to add the required docstrings in this PR to not overcrowd the issue list.

echopype/echodata/zarr_combine.py
```diff
@@ -789,6 +809,7 @@ def _write_append_dims(
         zarr_path: str,
         zarr_group: str,
         storage_options: Dict[str, Any] = {},
+        consolidated: bool = True,
```
lsetiawan, line 812:
```suggestion
```

echopype/echodata/zarr_combine.py
```diff
@@ -806,6 +827,9 @@ def _write_append_dims(
         storage_options: dict
             Any additional parameters for the storage
             backend (ignored for local paths)
+        consolidated : bool
+            Flag to consolidate zarr metadata.
+            Defaults to ``True``
```
lsetiawan, line 832:
```suggestion
```

echopype/echodata/zarr_combine.py
lsetiawan, line 857:
```suggestion
```

echopype/echodata/zarr_combine.py
```diff
@@ -1005,23 +1021,41 @@ def combine(
...
+            storage_options=storage_options,
+        )
 
         # change filenames numbering to range(len(eds))
         self._modify_prov_filenames(zarr_path, len_eds=len(eds), storage_options=storage_options)
 
+        if consolidated:
+            # consolidate at the end if consolidated flag is True
+            store = fsspec.get_mapper(zarr_path, **storage_options)
+            zarr.consolidate_metadata(store=store)
```
lsetiawan, line 1050: @b-reyes I've modified the code so that `consolidated` flag will consolidate the metadata at the end when the file has been created fully.

1. This will ensure that we don't miss any metadata, say if only parts were consolidated.
2. This should make the process of actually writing faster and it doesn't try to consolidate everytime.

lsetiawan:
@b-reyes This PR is ready for another review. I've put in some tests in there to ensure that everything is consolidating properly. One of the biggest changes is that for your combine function, I put the consolidation at the very end so no metadata is missed, and default all the various private function calls to `to_zarr` to not consolidate.

echopype/tests/echodata/utils.py
```diff
@@ -0,0 +1,103 @@
...
+    tree_dict["Sonar"] = setgrouper.set_sonar()
+    tree_dict["Vendor_specific"] = setgrouper.set_vendor()
+
+    tree = DataTree.from_dict(tree_dict, name="root")
+    echodata = EchoData(
+        source_file=file_chk, xml_path=xml_chk, sonar_model=sonar_model
+    )
+    echodata._set_tree(tree)
+    echodata._load_tree()
+    return echodata
```
lsetiawan, line 103: Starting work with mock Echodata object referenced in https://github.com/OSOceanAcoustics/echopype/issues/791

lsetiawan: FYI @leewujung @emiliom

echopype/echodata/zarr_combine.py
```diff
@@ -627,6 +628,7 @@ def write_to_file(
         """
 
         with Lock(lock_name):
+            # Write to region is not consolidated by default
```
b-reyes, line 631: I am not seeing this in the docs anywhere. Perhaps you found this by looking at the actual code? Either way, I think it would be best to be explicit here. Could you please add `consolidated=False` in `ds_in.to_zarr`?

echopype/echodata/zarr_combine.py
b-reyes: Sounds good! Thank you for providing this explanation.

echopype/tests/conftest.py
```diff
@@ -28,6 +28,16 @@ def test_path():
     }
 
 
+@pytest.fixture(scope="session")
```
b-reyes, line 31: From my understanding, you moved `ek60_test_data` here because it was used within `test_echodata_combine` and `test_zarr_combine`. However, this fixture seems out of place. Also, in the future we may want to have  `ek60_test_data` be different for `test_echodata_combine` and `test_zarr_combine`. Would it be better to just copy this into both `test_echodata_combine` and `test_zarr_combine`?

echopype/tests/echodata/test_echodata.py
```diff
@@ -311,6 +319,49 @@ def test_get_dataset(self, converted_zarr):
         assert result is None
         assert isinstance(ed_result, xr.Dataset)
 
+    @pytest.mark.parametrize("consolidated", [True, False])
+    def test_to_zarr_consolidated(self, mock_echodata, consolidated):
+        """
+        Tests to_zarr method. Currently uses a mock echodata object
+        When consolidated=True, every to_zarr call in ea. group will use the flag
```
b-reyes, line 326:
```suggestion
        Tests to_zarr consolidation. Currently, this test uses a mock EchoData object that only
        has attributes. The consolidated flag provided will be used in every to_zarr call (which 
        is used to write each EchoData group to zarr_path). 
```

echopype/tests/echodata/test_echodata.py
```diff
@@ -311,6 +319,49 @@ def test_get_dataset(self, converted_zarr):
...
+
+        assert zmeta_path.exists() is check
+
+        if check is True:
+            # Check that every group is in
+            # the zmetadata if consolidated
+            expected_zgroups = [
+                ".zgroup",
+                "Environment/.zgroup",
+                "Platform/.zgroup",
```
b-reyes, line 342: I noticed that `"Platform/NMEA/.zgroup"` is not here. This should be added. I will also make a comment on the mock EchoData creation.

echopype/tests/echodata/test_echodata.py
```diff
@@ -311,6 +319,49 @@ def test_get_dataset(self, converted_zarr):
...
+                "Provenance/.zgroup",
+                "Sonar/.zgroup",
+                "Sonar/Beam_group1/.zgroup",
+                "Vendor_specific/.zgroup"
+            ]
+            import json
```
b-reyes, line 348: Is there a reason why this import needs to be here? If not, I would suggest moving it to the top of the file.

echopype/tests/echodata/test_echodata.py
```diff
@@ -311,6 +319,49 @@ def test_get_dataset(self, converted_zarr):
...
+            with open(zmeta_path) as f:
+                meta_json = json.load(f)
+
+            file_groups = [
```
b-reyes, line 352: Could you add a comment here? Something along the lines of "obtain each .zgroup path in zmeta_path".

echopype/tests/echodata/test_zarr_combine.py
```diff
@@ -300,3 +327,101 @@ def test_append_ds_list_to_zarr(append_ds_list_params):
...
+            echodata_filenames=echodata_filenames,
+            consolidated=consolidated,
+        )
+
+        check = True if consolidated else False
+        zmeta_path = Path(zarr_path) / ".zmetadata"
+
+        assert zmeta_path.exists() is check
+
+        if check is True:
```
b-reyes, line 399: All of my comments from `test_echodata.py` should apply here. From what I can see this is just a copy paste of items in `test_to_zarr_consolidated` could we make this into a function and put it into say `utils.py`?

echopype/tests/echodata/utils.py
```diff
@@ -0,0 +1,103 @@
...
+        return ds
+
+    def set_vendor(self) -> xr.Dataset:
+        # TODO: add mock data
+        ds = xr.Dataset(attrs={"created_by": "Mock test"})
+        return ds
+
+
+def get_mock_echodata(
+    output_path='test.zarr',
```
b-reyes, line 70: Out of curiosity, what was your vision for `output_path`? Would `get_mock_echodata` return None in this case?

lsetiawan, line 70:
```suggestion
```

echopype/tests/echodata/utils.py
```diff
@@ -0,0 +1,103 @@
...
+    xml_chk=None,
+):
+    # Setup tree dictionary
+    tree_dict = {}
+
+    setgrouper = SetGroupsTest(
+        parser_obj=None,
+        input_file=file_chk,
+        xml_path=xml_chk,
+        output_path=output_path,
```
lsetiawan, line 82:
```suggestion
        output_path=None,
```

echopype/tests/echodata/utils.py
lsetiawan: Thanks for catching that!

echopype/tests/echodata/test_echodata.py
lsetiawan: This is because we only use it for this test.

lsetiawan:
@b-reyes Okay. I've implemented your suggestions. Let me know if I miss anything. Thanks!

echopype/tests/echodata/utils.py
```diff
@@ -0,0 +1,142 @@
...
+    # the zmetadata if consolidated
+    expected_zgroups = [
+        os.path.join(p, '.zgroup') if p != 'Top-level' else '.zgroup'
+        for p in echodata.group_paths
+    ]
+
+    with open(zmeta_path) as f:
+        meta_json = json.load(f)
+
+    file_groups = [
```
b-reyes, line 135: Could you please put a comment here saying what is happening? I suggested one in the previous review.

echopype/tests/echodata/utils.py
```diff
@@ -0,0 +1,142 @@
...
+    tree = DataTree.from_dict(tree_dict, name="root")
+    echodata = EchoData(
+        source_file=file_chk, xml_path=xml_chk, sonar_model=sonar_model
+    )
+    echodata._set_tree(tree)
+    echodata._load_tree()
+    return echodata
+
+
+def check_consolidated(echodata, zmeta_path):
```
b-reyes, line 119: Could you put docstrings for the function inputs. I could see one being particularly useful for `zmeta_path`.

b-reyes:
@lsetiawan I am done reviewing. I only have minor remarks in regards to documentation/commenting.

echopype/tests/echodata/utils.py
lsetiawan: It's in the function docstring

b-reyes (approved):

b-reyes:
@lsetiawan thank you very much for making those changes. All of my concerns have been addressed. Please go ahead and merge these changes in.

I am slightly worried about the RTD failure, but I believe we can resolve that after a merge, if necessary.

lsetiawan:
> I am slightly worried about the RTD failure, but I believe we can resolve that after a merge, if necessary.

The RTD failure had nothing to do with this PR, no stress 😄 See it all works: https://readthedocs.org/projects/echopype/builds/18907478/

PR #874, opened 2022-10-27
Add `.nbytes` to get at echodata data size

lsetiawan:
## Overview

This PR allows for a `EchoData.nbytes` call, which sums all the individual `xr.Dataset.nbytes` to get a total size of the data.


emiliom (approved):
This looks good. Thanks

PR #896, opened 2022-12-07
Add echopype working directory

lsetiawan:
## Overview

This PR is an idea I had in regards to a hidden working directory for echopype. This directory can be used for any temp files/configs/helper files or sub directories, etc that we may need. It's hidden but can still be easily found in the user directory.

### DIFF
--- a/echopype/tests/test_core.py
+++ b/echopype/tests/test_core.py
@@ -1,10 +1,13 @@
 from typing import TYPE_CHECKING
+import tempfile
+from pathlib import Path
 
 import pytest
 
 if TYPE_CHECKING:
     from ..core import SonarModelsHint
-from ..core import SONAR_MODELS
+from ..core import SONAR_MODELS, init_ep_dir
+import echopype.core
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


echopype/tests/test_core.py
```diff
@@ -63,3 +66,21 @@ def test_file_extension_validation_should_fail(
         raise ValueError(
             f"\"{ext}\" should have been rejected for sonar model {sonar_model}"
         )
+
+
+def test_init_ep_dir(monkeypatch):
+    temp_zarr_dir = tempfile.TemporaryDirectory()
+    zarr_file_name = Path(temp_zarr_dir.name) / ".echopype"
```
emiliom, line 73: Why refer to zarr in these path names? I realize that Zarr writing will be one of the main usages for the user temp directory  you're proposing, but that's not central to this test.

emiliom:
I like it. Granted, this new user folder would be hidden, so users would need to be made aware of it, say, to clean it up occasionally. But if the official temp space (`tempfile.TemporaryDirectory()`) is not guaranteed to always clean up, at least this one is a little more visible.

leewujung:
Two things that should follow from this:
- move all the temp directories and temp zarr files into this hidden directory, instead of the current `temp_echopype_output` directory
- move all the testing framework and files into this hidden directory

Issue #907, opened 2022-12-20
Incorrect insertion of meta_source_filenames during conversion for EK80, ES70, ES80, EA640


emiliom:
PR #814 introduced a small bug where the `Provenance` variable `meta_source_filenames`  is being incorrectly inserted during file conversion (open_raw) for EK80, ES70, ES80 and EA640. Currently that attribute is needed only with the AZFP sensor. An empty `meta_source_filenames` variable is being inserted for these sensor models due to a bug in `prov.source_files_vars`. This bug does not impact EK60.

AFAIK this bug is innocuous, because that variable is never used outside of AZFP.

I'll submit a fix shortly.

emiliom:
Fixed in PR #908

Issue #895, opened 2022-12-06
Combining a combined echodata object with a converted echodata object leads to Zarr chunk size error


emiliom:
While addressing #678, I ran into a Zarr chunk size TypeError when using `combine_echodata` to combine an already combined echodata object with an echodata object generated from either `open_raw` (ie, in memory) or `open_converted`. @b-reyes and I discussed this error a bit.

I used 3 hake survey raw files found in our test data folder. The file paths are constructed like this:

```python
from echopype.testing import TEST_DATA_FOLDER, HERE
files = [
    ("ncei-wcsd", "Summer2017-D20170620-T011027.raw"),
    ("ncei-wcsd", "Summer2017-D20170620-T014302.raw"),
    ("ncei-wcsd", "Summer2017-D20170620-T021537.raw"),
]
rawfile_paths = [TEST_DATA_FOLDER / "ek60" / f[0] / f[1] for f in files]
```

`open_raw` is applied in a straightforward way:

```python
eds = [ep.open_raw(file, sonar_model) for file in rawfile_paths]
```

Running `combine_echodata` on 2 or all 3 of the converted echodata objects works w/o issues, as expected. But when I try to combine that combined echodata object with a simple converted echodata object, the error occurs. Here, `combined_ed` is the result of combining the first two converted echodata objects (`eds[0]` and `eds[1]`), and the statement below tries to combine `combined_ed` with the third converted echodata object, `eds[-1]`:

```python
combined_ed2 = ep.combine_echodata(
    [combined_ed, eds[-1]], 
    zarr_path='combined_echodatas2.zarr', 
    overwrite=True
)
```

I've pasted below a good chunk of the last error lines out of a long error stack:

```
File ~/miniconda3/envs/echopype_dev/lib/python3.9/site-packages/xarray/backends/zarr.py:581, in ZarrStore.set_variables(self, variables, check_encoding_set, writer, unlimited_dims)
    578     zarr_array = self.zarr_group[name]
    579 else:
    580     # new variable
--> 581     encoding = extract_zarr_variable_encoding(
    582         v, raise_on_invalid=check, name=vn, safe_chunks=self._safe_chunks
    583     )
    584     encoded_attrs = {}
    585     # the magic for storing the hidden dimension data

File ~/miniconda3/envs/echopype_dev/lib/python3.9/site-packages/xarray/backends/zarr.py:228, in extract_zarr_variable_encoding(variable, raise_on_invalid, name, safe_chunks)
[...]

Issue #678, opened 2022-05-12
Enable correct provenance `source_filenames` construction for "appending" a converted file to a combined file


b-reyes:
In `combine_echodata` we currently gather the source file names using `echodata.source_file` and then these variables are put into the new combined `echodata.provenance.source_filenames` (this variable will exist once PR #674 has been merged). The mechanism that allows for this are the following lines

https://github.com/OSOceanAcoustics/echopype/blob/e0b3cbff9459b57cc8320388cc1ac86da6063343/echopype/echodata/combine.py#L144-L151

If we perform the following actions 
```
ed_1_2 = ep.combine_echodata([ed1, ed2])

ed_1_2_3 = ep.combine_echodata([ed_1_2, ed3])
```

then it might be possible that the above lines produce a list of lists. If this is the case, then it may cause issues with how we are currently creating `ed_1_2_3.provenance.source_filenames`. I think this scenario should be investigated further. 

Additionally, in the lines above, we should consider changing `echodata.source_file` to `echodata.provenance.source_filenames`.

leewujung:
@emiliom : this has been addressed in the recent PRs, is this correct?

emiliom:
> @emiliom : this has been addressed in the recent PRs, is this correct?

It wasn't addressed, not quite. We decided this was an edge case (combining two or more datasets that are already the result of combine operations) that we were not going to address.

I suggest we close the issue, though.

leewujung:
@emiliom : Thanks. Could you create an issue to make sure that we don't forget about that, and close this issue?

emiliom:
We've clarified the primary use case where this issue would be important: when "appending" a new, converted file to a pre-existing combined file. For example, in a continuous monitoring deployment, where new files are generated continuously and a combined file is maintained and updated by appending each new file as it's generated.

We will address this issue in the next release, 0.6.4.

emiliom:
The original scope of this issue has been addressed in PR #908. 

#895 describes a larger problem uncovered in the process of working on this issue. That problem will be tracked there.

PR #908, opened 2022-12-20
Fix meta_source_filenames bug and enable (meta)source_filenames appending of path and list

emiliom:
Addresses #678. More specifically, it enables `source_files_vars` in `prov.py` to handle the case of an input `source_paths` or `meta_source_paths` that includes both scalar elements (string or pathlib Path) and a list of such elements. This is the case that would arise with `combine_echodata` to append (combine) a converted file to a previously combined echodata object. Note that beyond the enhancement to handle `source_filenames` and `meta_source_filenames`, this PR does not address the larger, blocking issue described in #895, which currently makes that use case not possible.

I've done offline manual tests to confirm that the underlying `prov._source_files` function addresses that use case (heterogeneous list elements).

Also fixes the small bug described in #907

### DIFF
--- a/echopype/utils/prov.py
+++ b/echopype/utils/prov.py
@@ -1,6 +1,6 @@
 from datetime import datetime as dt
 from pathlib import Path
-from typing import Any, Dict, List, Tuple, Union
+from typing import Dict, List, Tuple, Union
 
 import numpy as np
 from _echopype_version import version as ECHOPYPE_VERSION
@@ -8,10 +8,10 @@ from typing_extensions import Literal
 
 from .log import _init_logger
 
-# TODO: It'd be cleaner to use PathHint, but it leads to a circular import error
-# from ..core import PathHint
-
 ProcessType = Literal["conversion", "processing"]
+# Note that this PathHint is defined differently from the one in ..core
+PathHint = Union[str, Path]
+PathSequenceHint = Union[List[PathHint], Tuple[PathHint], np.ndarray[PathHint]]
 
 logger = _init_logger(__name__)
 
@@ -35,13 +35,13 @@ def echopype_prov_attrs(process_type: ProcessType) -> Dict[str, str]:
     return prov_dict
 
 
-def _source_files(paths: Union[str, Path, List[Any], Tuple[Any], np.ndarray]):
+def _sanitize_source_files(paths: Union[PathHint, PathSequenceHint]):
     """
     Create sanitized list of string paths from heterogeneous path inputs.
 
     Parameters
     ----------
-    paths : Union[str, Path, List[str, Path], Tuple[str, Path], np.ndarray]
+    paths : Union[PathHint, PathSequenceHint]
         File paths as either a single path string or pathlib Path,
         a sequence (tuple, list or np.ndarray) of strings or pathlib Paths,
         or a mixed sequence that may contain another sequence as an element.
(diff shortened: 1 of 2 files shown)


echopype/utils/prov.py
```diff
@@ -30,20 +35,63 @@ def echopype_prov_attrs(process_type: ProcessType) -> Dict[str, str]:
...
+        logger.warning(
+            f"Unrecognized file path element type, path element will not be written to (meta)source_file provenance attribute. {paths}"  # noqa
+        )
+        return []
+
+
 def source_files_vars(
-    source_paths: Union[str, List[Any]], meta_source_paths: Union[str, List[Any]] = None
+    source_paths: Union[str, Path, List[Any], Tuple[Any], np.ndarray],
+    meta_source_paths: Union[None, str, Path, List[Any], Tuple[Any], np.ndarray] = None,
```
leewujung, line 78: The `List[Any], Tuple[Any]` should actually be `List[str, Path], Tuple[str, Path]` as what you had in the docstring?

echopype/utils/prov.py
```diff
@@ -30,20 +35,63 @@ def echopype_prov_attrs(process_type: ProcessType) -> Dict[str, str]:
     return prov_dict
 
 
+def _source_files(paths: Union[str, Path, List[Any], Tuple[Any], np.ndarray]):
```
leewujung, line 38: How about renaming this to something like `_sanitize_source_list`? Thought it's good for "sanitize" to show up.

emiliom: I like that suggestion. I'll implement it.

echopype/utils/prov.py
emiliom: Heh, good eye. Initially I tried having `List[str, Path]` and `Tuple[str, Path]` in the argument signature. But it looks like type hinting doesn't allow that, b/c I got an error. So I went with `List[Any]` and `Tuple[Any]` but kept the more explicit version in the docstring in order to, well, be more explicit. I wonder if using `Union[str, Path]` instead of `str, Path` would do the trick. If it does, it'll make for an accurate but pretty long and hard-to-digest type specification. 

I also didn't try to see if `np.ndarray` can be followed by a sequence element type specification. I did some research just now, and it turns out that numpy added a typing library last year or so, [numpy.typing](https://numpy.org/devdocs/reference/typing.html). It includes a better replacement for `np.ndarray`: `numpy.typing.NDArray`. It also has `numpy.typing.ArrayLike`, which *may* be a more compact way to handle `List`, `Tuple` and `np.ndarray`. I'll look into it. It'll be great if that's the case!

emiliom:
@leewujung Thanks for the review! I like your suggestion of adding a unit test. I could do it by tomorrow, if that still works with the release timeframe.

echopype/utils/prov.py
emiliom: I've renamed the function to `_sanitize_source_files`. Though I just realized that your suggestion was `_sanitize_source_list`. How about we go with what I've done?

echopype/utils/prov.py
emiliom: After much R&D, I submitted a "perfect" (IMHO) implementation of the type hints using aliases, and it turns out it [fails with Python 3.8 ](https://github.com/OSOceanAcoustics/echopype/actions/runs/3744903528/jobs/6358775827#step:14:36) :sob: . I'll revisit it later.

emiliom:
I've completed the individual modifications you brought up. The remaining task is creating a unit test.

emiliom:
Alright, I've added unit tests. @leewujung please take a look when you get a chance.

(FYI in the process I also discovered and fixed a small bug I had introduced in my last commit involving the handling of input in the form an np.ndarray)

leewujung:
@emiliom : thanks, the test looks good! Please feel free to merge.

PR #909, opened 2022-12-21
Fix mkdir to not raise error and skip directory creation if exists

lsetiawan:
## Overview

To avoid parallel runs to fail, let's allow skipping of directory creation.


echopype/core.py
```diff
@@ -30,7 +30,7 @@
 def init_ep_dir():
     """Initialize hidden directory for echopype"""
     if not ECHOPYPE_DIR.exists():
-        ECHOPYPE_DIR.mkdir()
+        ECHOPYPE_DIR.mkdir(exist_ok=True)
```
b-reyes, line 33: I have looked into this and online it appears that this is the proper way to handle a race condition that occurs with multiple workers. This looks good!

b-reyes (approved):
This looks good @lsetiawan!

PR #915, opened 2022-12-22
Redo: Ignore docs/source/_build

leewujung:
This PR redos #910. See #910 for all conversations.

This is part of the fix to ensure that we have a clean squash-and-merge history on `dev` and eventually `main`.

### DIFF
--- a/.github/workflows/windows-utils.yaml
+++ b/.github/workflows/windows-utils.yaml
@@ -1,0 +1,57 @@
+name: windows-utils-test
+
+on:
+  push:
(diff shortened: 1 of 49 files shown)


emiliom:
Hmm, this PR includes several unrelated commits involving mask and prov work ... That shouldn't be there, right?

leewujung:
it shouldn't, but it will. This is done after the fact and won't have the same fine-grained control. The best I can do is to make sure all changes are committed.

Issue #831, opened 2022-10-05
Develop tests for prov module, specially for source file name functionality


emiliom:
- From @lsetiawan in PR #829: I realized that we don't have a test for the [utils/prov.py](https://github.com/OSOceanAcoustics/echopype/blob/dev/echopype/utils/prov.py), maybe this will be great for our testing creation session in the future :smile:
- From @emiliom in PR #814: While creating this PR and https://github.com/OSOceanAcoustics/echopype/pull/806, I've thought about tests. There are no tests (I think) that check for the correct creation of these file provenance variables. More broadly, I don't think we do much testing on provenance values, and that's probably a tricky area. Rather than trying to create very narrow tests in this PR, I think we should have a broader discussion. Maybe that could be folded into provenance work in 0.6.4.

leewujung:
This is likely related to #1043 and #1042.

emiliom:
See also the new issue #1050

Issue #769, opened 2022-08-04
Add more comprehensive tests to `consolidate.add_latlon`


leewujung:
This is a leftover to #749 , in which we added a new function to interpolate lat-lon locations from the `EchoData["Platform"]` group to calibrated Sv data. That PR includes a basic test to run `add_latlon` for an EK60 file and check for existence of the latitude and longitude variables.

We still need tests to cover the following cases:
- Data with empty latitude and longitude variables in the Platform group
- Data with a single point latitude and longitude point in the Platform group
- Checking the added (interpolated) lat/lon values between the `add_latlon` function and directly computed from `scipy.interpolate`

leewujung:
This will ensure #730 runs through without problem.

emiliom:
@leewujung in `add_location`, the `time1` coordinate variable is dropped at the very end:

https://github.com/OSOceanAcoustics/echopype/blob/f8082cac32bfbee2b21c62d266507491d221182d/echopype/consolidate/api.py#L188

I've looked closely at a couple of sample datasets. For two EK60 datasets, after `compute_Sv`, there are no variables with a `time1` dimension. The only other time dimension in use besides `ping_time` is `time3`, used with `water_level`. But for an AZFP dataset, after `compute_Sv`, `water_level` uses `time1`, not `time3`. Off the top of my head I don't remember which time dimension `water_level` is supposed to have or whether it'd be the same regardless of instrument. If `water_level` having a `time1` dimension is in fact valid, then it shouldn't be dropped for AZFP.

I also investigated the result of interpolation. For some reason that I haven't investigated, the first interpolated value appears to always be nan (in two EK60 example files). For now, I'm going to modify `add_location` and set the first value of interpolated `latitude` and `longitude` equal to the second value. We can discuss here or in the PR that I'll submit shortly.

leewujung:
For I looked at the [docs](https://echopype.readthedocs.io/en/stable/data-format-5to6.html) and couldn't completely remember what's up with our various `time*` dimensions, but I think within a single group the `time*` changes depending on the source of data (since they can be on different time base), which varies depending on the `sonar_model`.

I found this issue https://github.com/OSOceanAcoustics/echopype/issues/656 that documents what those `time*` are. I think we should replicate those in the docs.

You can see that `water_depth` is called out here to be aligned with `time3`.

AZFP should not have a `water_depth` variable stored by default, see code below -- which data file were you looking at?
[...]

PR #918, opened 2022-12-24
Add dataset and variable-level provenance to new mask and add_splitbeam_angle functions

emiliom:
**WARNING:** the PR apparently can't be automatically merged to `dev`! But it can be merged to `old_dev`. I haven't investigated this. I'll keep this PR as a draft until that's resolved. @leewujung you're the expert on the `dev` vs `old_dev` issue.

Add provenance attributes to the outputs of the mask functions `apply_mask` and `frequency_differencing`. Specifically:
- `mask_type` and `history` variable attributes added to the mask dataarray returned by `frequency_differencing`. `history` captures the actual operation performed, based on the arguments passed to the function.
- dataset-level standard provenance attributes (from `prov.echopype_prov_attrs`) and Sv `history` variable attribute. The latter is concatenated with the mask `history` string. In addition to these new attributes, the `long_name` and `actual_range` attributes are updated. Note that this initial implementation can only handle mask attribute extraction for a single mask dataarray; if more than one mask is passed, only the attributes from the first one will be used. Handling of attributes from >1 mask will be added in a future release.

Add variable-level `history` attributed to the variables (datarrays) added by the new function `add_splitbeam_angle` and the existing function `add_depth`, both of which are `Sv` consolidate functions.

A few small fixes were also applied.

No tests were added. Tests for attribute content is a larger issue that will be addressed in a future release. See #831 and #769


emiliom:
I could click on `Resolve conflicts` above and resolve the conflicts (I've looked at them). But I'm not sure if there's a larger issue of merging against the new, "squashed" dev vs the old dev :exploding_head:

leewujung:
@emiliom : I tested the resolve conflict and merge operations in my fork and everything worked out, so I resolved the conflicts again here (I think those are _after_ I updated the new `dev` branch) and will go ahead to **squash** and merge. Thanks for the provenance addition!

leewujung:
@emiliom : on the merge to `dev` you can see that there's a small assertion error that trips the test due to the added history and unit attributes. Could you please send another PR to patch these up? We'll have to address the larger issue of testing for attributes soon -- it has become cumbersome in recent developments to keep on having to look after this.

emiliom:
> on the merge to dev you can see that there's a small assertion error that trips the test due to the added history and unit attributes. Could you please send another PR to patch these up? 

Great, thanks! Could you point me to the assertion error? I can see that tests involving version conversion could fail, but the CI tests didn't fail and I also didn't see any errors when I ran the test suite locally just before submitting the PR.

> We'll have to address the larger issue of testing for attributes soon -- it has become cumbersome in recent developments to keep on having to look after this.

Yup

leewujung:
https://github.com/OSOceanAcoustics/echopype/actions/runs/3783076757/jobs/6431330051#step:11:1368

Not sure why this is not run at PR level.

PR #904, opened 2022-12-19
Refactor calibrate subpackage

leewujung:
## Overview

This PR refactors the `calibrate` subpackage and fixed 3 bugs to get everything ready for a series of PRs addressing issues tracked in #875.

### Refactoring
- factor out `env_params` and `cal_params` retrieval/computation to separate modules `calibration/env_params.py` and `calibration/cal_params.py` (and renamed the old `env_params.py` to `env_params_old.py`)
- factor out echo range calculation to `calibration.range`
- factor out all functions related to using EK80 complex data to `calibration/ek80_complex.py`
- set `waveform_mode` and `encode_mode` as class attributes for `CalibrateEK80` and simplified class methods
- take a top-down approach to reorganize the sequence of getting `env_params`, `cal_params`, and `echo_range` at init of cal objects.
- update existing tests

### Bug fixes
- fix bug re interpolating frequency-dependent gain at the center frequency of the transmit signal (before the interpolation was at the nominal frequency of a given channel)
  - this is related to #458 
- fix bug in computing wavelength for BB cal, to use the center frequency of transmit signal
  - before was wrongly using the nominal frequency of the channel, see [here](https://github.com/OSOceanAcoustics/echopype/blob/d4c73aac6b174bde5181b9849dc4940b47bff389/echopype/calibrate/calibrate_ek.py#L767-L770))
  - this is likely a regression bug during to the reorganization of data dimensions at v0.6.0
- fix bug in scaling equivalent sampling volume, use 20\*log10
  - before used 10\*log10 before, see [here](https://github.com/OSOceanAcoustics/echopype/blob/d4c73aac6b174bde5181b9849dc4940b47bff389/echopype/calibrate/calibrate_ek.py#L809-L812))

### Remaining work for later PR
- review and likely remove the input argument checking block in `calibrate/api.py::_compute_cal`
- clean up input argument ingestion in `CalibrateBase.__init__`

### Dependency
- Note this PR uses (and modifies) a couple functions from #894 and will have to pull in additional changes and resolve conflicts from that PR once it's merged.
  - **Update 2022/12/28:** The latest changes from `dev` were pulled and all conflicts resolved.

PR #926, opened 2022-12-30
Small fixes following calibration refactoring and mask history 

leewujung:
This PR includes fixes 2 small bugs:
- update use of `retrieve_correct_beam_group` in `add_splitbeam_angle` (from #904)
- fix bug in chanAB type in constructing freq-diff mask history (from #918)

### DIFF
--- a/echopype/consolidate/api.py
+++ b/echopype/consolidate/api.py
@@ -315,9 +315,7 @@ def add_splitbeam_angle(
 
     # check that the appropriate waveform and encode mode have been given
     # and obtain the echodata group path corresponding to encode_mode
-    encode_mode_ed_group = retrieve_correct_beam_group(
-        echodata, waveform_mode, encode_mode, pulse_compression
-    )
+    encode_mode_ed_group = retrieve_correct_beam_group(echodata, waveform_mode, encode_mode)
 
     # check that source_Sv at least has a channel dimension
     if "channel" not in source_Sv.variables:
--- a/echopype/mask/api.py
+++ b/echopype/mask/api.py
@@ -547,7 +547,7 @@ def frequency_differencing(
     history_attr = (
         f"{datetime.datetime.utcnow()} +00:00. "
         "Mask created by mask.frequency_differencing. "
-        f"Operation: Sv['{str(chanA.values)}'] - Sv['{str(chanB.values)}'] {operator} {diff}"
+        f"Operation: Sv['{chanA}'] - Sv['{chanB}'] {operator} {diff}"
     )
 
     da = da.assign_attrs({**mask_attrs, **{"history": history_attr}})


emiliom:
> fix bug in chanAB type in constructing freq-diff mask history (from #918)

@leewujung I've looked at this more closely. I think the code in my PR (#918) was actually correct and doing the right thing. But some tests in `mask/test_mask.py` may need to be updated to account for the new attributes being added.

With the change from this PR, here's what `ep.mask.frequency_differencing` produces for the `history` attribute:

image

That's basically the text repr for `chanA` and `chanB`. Which makes sense, given the code (`Sv['{chanA}'] - Sv['{chanB}']`). The test suite runs locally w/o errors, though.

When I revert the code back to what I had in my PR (`Sv['{str(chanA.values)}'] - Sv['{str(chanB.values)}']`), the `history` attribute is populated correctly (FYI I had tested this manually before submitting the PR):

"2023-01-03 02:30:59.938729 +00:00. Mask created by mask.frequency_differencing. Operation: Sv['GPT  38 kHz 009072058146 2-1 ES38B'] - Sv['GPT  18 kHz 009072058c8d 1-1 ES18-11'] > 5"

With this change, the test suite also runs locally w/o errors. This was puzzling, given the [CI error you pointed me to](https://github.com/OSOceanAcoustics/echopype/actions/runs/3783076757/jobs/6431330051#step:11:1368). To investigate this, I ran `test_mask.py` directly (`pytest -vv echopype/tests/mask/test_mask.py`). Now the errors show up!

It turns out `test_mask.py` was in fact not being executed when running the test suite using `run-test.py`. The reason is that the `mask` subpackage folder needs to be added to `run-test.py`, [here](https://github.com/OSOceanAcoustics/echopype/blob/dev/.ci_helpers/run-test.py#L26).

Now that I can reproduce the errors locally I'll work on updating `test_mask.py`. I'll also create a PR to fix `run-test.py`

leewujung:
Hmm.. thanks for finding out about and adding the `mask` to `run-test.py`!

For `ep.mask.frequency_differencing`, I was going by the [docstring](https://github.com/leewujung/echopype/blob/90830fbe58e344a8a77c8012425f78ec814a5924/echopype/mask/api.py#L406) that has `chanAB: Optional[List[str]] = None,`.

Looking at this again, it seems that the difference between what you saw and what I saw hinges on whether users use `freqAB` or `chanAB` in the inputs!
https://github.com/leewujung/echopype/blob/90830fbe58e344a8a77c8012425f78ec814a5924/echopype/mask/api.py#L521-L535
When `freqAB` is used, `chanA` and `chanB` becomes data arrays; when `chanAB` is used (list of strings), `chanA` and `chanB`  stay as strings.

@emiliom : How do you want to handle this?

emiliom:
I'll tackle this tomorrow morning.

PR #927, opened 2023-01-01
Update CITATION.cff

emiliom:
After the exchanges in https://github.com/OSOceanAcoustics/echopype/discussions/924, how about these updates?

- [ ] Update to the latest release (will have to update to the actual release date)
- [x] Add doi url for arxiv preprint, just to be more explicit and easy to use

Separately -- or as an addition to this PR? --, we could add a mention in the RTD docs to the citation link and preferred citation forms.


leewujung:
@emiliom : these look good -- the only thing we need to do is to actually update the manuscript to reflect the changes since then 😉  

> Separately -- or as an addition to this PR? --, we could add a mention in the RTD docs to the citation link and preferred citation forms.

Maybe to this PR? Could you put these in? Thanks!

emiliom:
> Maybe to this PR? Could you put these in? Thanks!

So, we'll add it to the docs on `dev`, not `main`, right?

CITATION.cff
```diff
@@ -23,8 +23,8 @@ title: "echopype"
 doi: 10.5281/zenodo.3906999
 # but pointing to a specific release/version will
 # require having to update this file with each release
-version: 0.5.4
-date-released: 2021-10-01
+version: 0.6.4
+date-released: 2023-01-03
 url: "https://github.com/OSOceanAcoustics/echopype"
```
emiliom, line 28: Add use of `repository-code` to be more specific. Saw it first in xarray CITATION.cff and double checked at https://github.com/citation-file-format/citation-file-format/blob/main/README.md. The `url` can then point to the RTD site.
```suggestion
url: "https://echopype.readthedocs.io"
repository-code: "https://github.com/OSOceanAcoustics/echopype"
```

leewujung:
> > Maybe to this PR? Could you put these in? Thanks!
> 
> So, we'll add it to the docs on `dev`, not `stable`, right?

I guess this can be to `stable`? since technically the preprint hasn't changed for either.

leewujung:
Hey @emiliom : just pinging to get this to the somewhere near top of your inbox. I think you could merge once you add a link to the RTD. I think either `dev` or `stable` would be fine, since hopefully we release 0.6.4 soon. Thanks!

emiliom:
@leewujung  Regarding this task:

> Update to the latest release (will have to update to the actual release date)

In this PR I used 0.6.4 and a place-holder date of 2023-01-03. If the 0.6.4 release will happen soon, I suggest we stick with 0.6.4 and update the date one the release is being prepared. Also, this PR is against `dev`, which makes sense for a change set to the upcoming release.

If I add changes to the docs to this PR, they'll have to go to `dev`, too (obviously).

CITATION.cff
```diff
@@ -23,9 +23,10 @@ title: "echopype"
 doi: 10.5281/zenodo.3906999
 # but pointing to a specific release/version will
 # require having to update this file with each release
-version: 0.5.4
-date-released: 2021-10-01
-url: "https://github.com/OSOceanAcoustics/echopype"
+version: 0.6.4
+date-released: 2023-01-03
```
emiliom, line 27:
```suggestion
date-released: 2023-03-10
```

emiliom:
@leewujung I've been thinking about this, about the CITATION.cff file, and what we want to see happen. In our https://github.com/OSOceanAcoustics/echopype/blob/main/CITATION.cff file I originally included two types of citations: one to the software (via zenodo), which should be kept up-to-date with each release; and one to the arxiv manuscript. I set the latter as the `preferred-citation`, and that's what users will get when they click on the GitHub citation widget. But from the perspective of software citation and reproducibility, citing the software directly, which refers to a specific version, is a better practice.

Should we remove the `preferred-citation`? Maybe the arxiv citation belongs more properly in the RTD documentation, for more academic citations rather than specific software version citation.

With respect to what to add to the RTD docs, I like what's done in https://docs.plasmapy.org/en/stable/about/citation.html and would recommend a much more modest version of it (for reference, that software's CITATION.cff is [here](https://github.com/PlasmaPy/PlasmaPy/blob/main/CITATION.cff)). That would allow us elaborate on what citation to use in what context. 

Finally, and maybe just for my own future reference, having the software citation in RTD docs as well as CITATION.cff would require us to remember to update the RTD doc with each release. That's brittle. I started looking into Sphinx extensions that might read, parse and render (format) the CITATION.cff content directly. I found a promising lead, https://github.com/citation-file-format/citation-file-format/issues/356 (see it in action at https://hereon-netcdf.readthedocs.io/en/latest/). But that's probably best left for a future enhancement.

emiliom:
- Added Brandon to the software citation part of CITATION.cff (not the arxiv citation)
- Added a short section to the main docs page, on citing echopype

PR #939, opened 2023-02-07
Add default chunk encoding for zarr output

lsetiawan:
## Overview

This PR adds a default chunk size encoding when performing `to_zarr` on `EchoData` object. By default, the max chunk size is `100MB`, which follows the dask chunking best practice. When the variable is already chunked via a dask array, this encoding default chunk computation will be ignored as xarray will already handle the creation of encoding based on the xarray data array chunks.

### DIFF
--- a/echopype/utils/coding.py
+++ b/echopype/utils/coding.py
@@ -1,8 +1,10 @@
 from re import search
+from typing import Tuple
 
 import numpy as np
 import xarray as xr
 import zarr
+from dask.array.core import auto_chunks
 from xarray import coding
 
 DEFAULT_TIME_ENCODING = {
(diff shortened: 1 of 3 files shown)


lsetiawan:
Current `windows-utils-test` error is not related to PR, but rather something with conda.

echopype/utils/coding.py
```diff
@@ -53,6 +55,30 @@ def _encode_dataarray(da, dtype):
...
+        The desired max chunk size for the array.
+        Default is 100MB
+
+    Returns
+    -------
+    tuple
+        The chunks
+    """
+    auto_tuple = tuple(["auto" for i in variable.shape])
+    chunks = auto_chunks(auto_tuple, variable.shape, chunk_size, variable.dtype)
```
b-reyes, line 78: @lsetiawan as we discussed today, I suggest specifying that chunks should occur along the time dimensions. This will be better for downstream methods.

lsetiawan: After conversation with @leewujung this is fine for the smaller files. The usecases for downstream combine echodata will be in a separate issue and PRs in the future.

PR #944, opened 2023-02-26
Improving pulse compressed Sv computation and echo range computation

leewujung:
## Overview

This PR improves pulse compressed Sv computation and addresses #308 and #458.

### Changes directly related to Sv computation include:
- change transmit chirp construction to follow [Anderson et al. 2021](https://arxiv.org/abs/2104.07248)
- fix regression bug introduced in #904 on decimated sampling frequency (`fs_deci_dict`) in constructing pulse compression replica
- add gain compensation for broadband mode related to angle offset and beamwidth
- add interpolation for getting angle offset and beamwidth along/athwartship for broadband transmission
  - previously only interpolating for the gain
- use transmit and receive impedance when exist in data file, otherwise default to fixed values
  - previously use only the default fixed values
- use receiver sampling frequency when exist in data file, otherwise default to values based on transceiver hardware (previously use only the default values for WBT)
- use `tau_effective` or `transmit_duration_nominal` for computing sampling volume depending on transceiver hardware (whether GPT or newer ones for Ex80)
- use `sa_correction` when computing Sv for CW mode complex data
- use sound speed and pH stored in data file in absorption calculation
  - previously: sound speed used the approximation in FG paper, pH used a default value (8.1)

### New data variables added to the `Vendor-specific` when exist
These are necessary to support the above changes (and are good to have anyways):
- transceiver type
- transmit impedance
- receive impedance
- receiver sampling frequency

### Others
- add error checking for the case where `waveform_mode="CW"` but all data are BB
- change range computation to depend on transceiver hardware (whether GPT or newer ones for Ex80)
  - previously this was based on `sonar_model`
- change `echo_range` outputs to not include TVG compensation component, referencing pyecholab implementation (i.e., the actual echo range and the range used for TVG compensation are different)
- `CalibrateEK.range_meter` now only contains the subset of channels that are either BB or CW depending on user input for calibration

### Tests
- new unit test for EK80 calibration details: file `test_calibrate_ek80.py`, testing for:
  - transmit replica (against pyecholab values)
  - broadband-interpolated calibration params (against pyecholab values)
  - pulse-compressed power values (against pyecholab and Echoview values)
  - range (against pyecholab values)

## Remaining tasks for later PRs
- test EK80 CW Sv params/outputs against pyecholab and Echoview values
- Sv integration test:
  - select test data files to cover different combinations of BB/CW transmission, complex/power samples, different EK80 software versions
- test TS outputs
- track down the difference between the first 8 samples for EK60 Sv data
- separate out tests for `env_params` and `cal_params` once routines of these params are revised
- #948
- #951


lsetiawan:
@leewujung could you redo this PR without the pre-commit changes? Thanks! Probably easiest to do that way.

leewujung:
@lsetiawan : what do you mean by "without the pre-commit changes" ?

leewujung:
Hmm that mamba problem persists somehow. Any clue what that might be from?

lsetiawan:
> Hmm that mamba problem persists somehow. Any clue what that might be from?

Weirdly enough, the dev branch is installing mamba < 1 and this PR is installing mamba == 1.3.1, which is the latest one.. sigh...

lsetiawan:
> Weirdly enough, the dev branch is installing mamba < 1 and this PR is installing mamba == 1.3.1, which is the latest one.. sigh...

I think I found the issue here. Trying out the solution in dev. Once that's merged. This PR should pull in the latest changes. Thanks.

leewujung:
Great, now this runs! There are some tests that pass locally so not sure why failing on CI end, but there's one test that I know will fail.

Regardless, I need to put new tests for this, so we'll see how things go once the new tests are in place.

Thanks @lsetiawan !

leewujung:
I'm going to self-merge this so that I can continue to work on the remaining tasks. All tests passed locally, but let's see if there are other hiccups at merge to `dev`.

leewujung:
All tests after merge passed except for the one with windows-utils-test. I'll get on to the other tasks.

PR #947, opened 2023-02-27
Fix windows github actions workflow

lsetiawan:
## Overview

This PR attempt to fix the windows CI github actions workflow that are currently failing.

## Related issues
- #956


lsetiawan:
The windows testing has now passed. The `pull_request_target` still uses the old action configs, so this can be ignored since it will be updated once this PR is merged.

Issue #958, opened 2023-03-08
Rename offload_to_zarr to use_swap to make it more user friendly


lsetiawan:
## Overview

`offload_to_zarr` should be renamed to `use_swap` as it is a more familiar term and users don't really need to know what data format we're using under the hood

There will NOT be a required destination path. `use_swap` will use the local drive and the "spilled" memory data files will be stored in echopype's temp folder in `temp_output`.

leewujung:
This is addressed in #962.

Issue #782, opened 2022-08-11
Add `'auto'` option to kwarg `offload_to_zarr` in `open_raw`


b-reyes:
In PR #774 we added the option to directly write variables with a large memory footprint to a temporary zarr store. Ideally, we want to include the option `offload_to_zarr='auto'`, which will trigger a routine that automatically determines if we should directly write these variables to a temporary zarr. In a first attempt, the choice of whether to offload to zarr was based on a threshold of the percentage of memory consumed, if the variables were to be expanded. However, as highlighted by @lsetiawan's [comment](https://github.com/OSOceanAcoustics/echopype/pull/774#issuecomment-1212328878), it may be difficult to make this robust for all computer architectures and take into account the memory increase downstream. 

One alternative to this approach would be to write those variables that when expanded exceed the user input value `max_zarr_mb`. The input `max_zarr_mb` is the maximum MB that each zarr chunk should hold, when offloading variables with a large memory footprint to a temporary zarr store.

leewujung:
@lsetiawan : this is related to the parsed2zarr refactoring we discussed today. I'm putting it in v0.7.1 as something we look into after March.

PR #962, opened 2023-03-08
Update the name of `offload_to_zarr` to `use_swap` and `max_zarr_mb` to `max_mb`

lsetiawan:
## Overview

This PR addresses the following:
- #958 
- Update the variable `max_zarr_mb` to `max_mb`
- Require `raw_file` and `sonar_model` for `open_raw` function, removing deprecation.

### DIFF
--- a/docs/source/convert.rst
+++ b/docs/source/convert.rst
@@ -86,7 +86,7 @@ all files from the same deployment.
     by allowing users to directly write variables that may consume a large amount of memory
     into a temporary zarr store (see `#774 <https://github.com/OSOceanAcoustics/echopype/pull/774>`_).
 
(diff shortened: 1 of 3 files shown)


leewujung (approved):
@lsetiawan : this looks good, feel free to merge. ~One thing I'd highlight here is that whatever `max_mb` controls is actually not working I think -- the argument was left in but the functionality was not implemented.~ I was wrong, actually it is the potential "auto" option for `offload_to_zarr` that has not been implemented (see #782). The use of `max_mb` for the max allowable chunk size is pretty hard-coded though, so something could be improved there. However, this is related to the needs to refactor parsed2zarr so let's not worry about that here.

PR #981, opened 2023-03-13
Unpin xarray version [all tests ci]

lsetiawan:
## Overview

This PR unpins `xarray` version, along with the unpinning, we now pin `pandas` to `<2` to be in sync with xarray (See https://github.com/pydata/xarray/pull/7650). Additionally with some unpinning, there are a few changes in the code below to ensure compatibility with the latest xarray version.


leewujung:
@lsetiawan : I looked into a couple of the failed tests, and it seems that it might be that xarray is:
- either being more strict on what counts as `.identical()`
- or that some internal cast of int values are now done differently. 

You can see that in the two cases below (these are the only cases I looked at):

`test_cal_params_intake_EK60`:
Screenshot 2023-03-15 at 2 29 33 PM


`test_get_vend_cal_params_power::in_no_nan`:
Screenshot 2023-03-15 at 2 35 30 PM

echopype/calibrate/cal_params.py
```diff
@@ -312,7 +312,19 @@ def get_vend_cal_params_power(beam: xr.Dataset, vend: xr.Dataset, param: str) ->
...
+    # Set the nan elements back to nan.
+    # Doing the `.where` will result in float64,
+    # which is fine since we're dealing with nan
+    da_param = da_param.where(~transmit_isnull, np.nan)
+
+    # Clean up for leftover plb variable
+    # if exists
+    plb_var = "pulse_length_bin"
+    if plb_var in da_param.coords:
+        da_param = da_param.drop(plb_var)
```
lsetiawan, line 325: This variable is still left after putting the `nan` back to `nan` type, so if it still exists, it will now be dropped.

echopype/calibrate/cal_params.py
```diff
@@ -312,7 +312,19 @@ def get_vend_cal_params_power(beam: xr.Dataset, vend: xr.Dataset, param: str) ->
 
     # Select corresponding index and clean up the original nan elements
     da_param = da_param.sel(pulse_length_bin=idxmin, drop=True)
-    return da_param.where(~transmit_isnull, np.nan)  # set the nan elements back to nan
+
+    # Set the nan elements back to nan.
+    # Doing the `.where` will result in float64,
+    # which is fine since we're dealing with nan
+    da_param = da_param.where(~transmit_isnull, np.nan)
```
lsetiawan, line 319: Calling `.where` in this line will always result in `nan`. An outside discussion suggest that there need to be a case for `range_sample` that should output int:
> yes I think it is reasonable to default to float, the only exception would be something like our range_sample  values, since it is sample count

I don't think this should be addressed in this PR, and probably further discussion is needed to figure out how to address it.

echopype/calibrate/env_params_old.py
```diff
@@ -105,8 +105,10 @@ def _apply(self, echodata) -> Dict[str, xr.DataArray]:
         if self.data_kind == "mobile":
             if np.isnan(echodata["Platform"]["time1"]).all():
                 raise ValueError("cannot perform mobile interpolation without time1")
+            # only grab needed variables for the interpolation
+            platform_data = echodata["Platform"][["latitude", "longitude"]]
             # compute_range needs indexing by ping_time
-            interp_plat = echodata["Platform"].interp(
+            interp_plat = platform_data.interp(
```
lsetiawan, line 111: Something is going on under the hood with xarray when calling `.interp` with all of the variables after version `2022.6.0`, which undergone the new `indexing`. By limiting the `.interp` to the data variables that are actually used, this works properly.

### All var
Screenshot 2023-04-17 at 4 50 37 PM

### Filtered var

Screenshot 2023-04-17 at 4 50 50 PM

echopype/utils/coding.py
```diff
@@ -35,6 +35,23 @@
...
+    Validates and fixes data type for expected variables
+    """
+
+    if isinstance(ds, xr.Dataset):
+        for name, var in ds.variables.items():
+            if name in EXPECTED_VAR_DTYPE:
+                expected_dtype = EXPECTED_VAR_DTYPE[name]
+                if not np.issubdtype(var.dtype, expected_dtype):
+                    ds[name] = var.astype(expected_dtype)
+    return ds
```
lsetiawan, line 52: This new `sanitize_dtypes` method will ensure that the data type of each variable is the expected one. Currently, the `channel` and `beam` data variable is `object` dtype and when trying to output to zarr it fails in `echopype.utils.coding.get_zarr_compression` since this dtype is not listed. So let's cast it to ensure that it is the correct data type.

leewujung:
Hey @lsetiawan : could you pull the latest `dev` into this? I saw something that have been fixed for v0.7.0, including the codespell exclusion of test_data. That would likely fix the windows tests also.

I'll continue to review other parts of this PR.

echopype/calibrate/cal_params.py
leewujung: I think this is fine, unless computationally this becomes super slow. 
Basically the `nan` here will end up being `nan` in the output Sv after calibrating the values in `backscatter_r`.

echopype/calibrate/cal_params.py
leewujung: Interesting how this `pulse_length_bin` variable get preserved in the new versions. I noticed some inconsistency before for dataarray sel operations in terms of what is preserved and what is dropped directly. Perhaps they made the behavior consistent now.

echopype/calibrate/env_params_old.py
leewujung: Geez, ok, thanks for catching this. There may be some other parts in echopype that will break because of this that are not caught by the current tests, but we'll fix them when we get to it.

echopype/utils/coding.py
leewujung: This is great -- and I didn't know there is `np.issubdtype`!

leewujung (approved):
@lsetiawan : I've reviewed this and it looks great! Feel free to merge once you get that psutil thing resolved or on a different PR if you prefer.

The dtype issues are elusive. The pydantic approach will be so much better than what we're doing now, but that's a conversation/task for later.

echopype/calibrate/cal_params.py
lsetiawan: Yea, I'll have to revisit this since this was already in the code, I didn't wanna change it much and break something else for now 😛

lsetiawan:
EVERYTHING PASSED! WOO HOO! 🎉 I'm self-merging since it's all approved and all checks passed! Thanks @leewujung for the reviews 👍🏽

Issue #965, opened 2023-03-09
Improve `env_params` intake and add test


leewujung:
This issue is to keep track of one of the remaining issues from #955:
- improve `env_params` intake in a form similar to what we do with `cal_params` (eg have an overarching dict defining all allowable params)
- add unit tests and integration tests that checked the intake of `env_params` into calibration objects. 

Note: #955 contains these tests for `cal_params` intake, under `test_cal_params.py` and `test_cal_params_integration.py`

leewujung:
This is addressed in #985.

PR #985, opened 2023-03-14
Ensure `env_params` intake for calibration

leewujung:
## Overview

~This PR fixes a regression bug in v0.6.4 and ensures `env_params` are taken up correctly for EK60 echosounder.~

It turned out that the bug was not omission, but that the `env_params` used for EK60 cal in v0.6.4 was not passed in from the correctly sanitized ones from the parent class method. This is now in a PR with minimal scope #987.

Therefore, I am putting in work in this PR to address #965 more comprehensively.

## Details
- Overhaul functions in `env_params.py` to address the core of #965
- Remove redundancy in passing in extra `ed_beam_group` in `_cal_power_samples` and `_cal_complex_samples`
- Create integration tests for the `get_env_params_*` functions in `env_params_integration.py`
  - These verify that user-provided params are taken in to the calibration objects and stored in the final Sv output correctly
- Create unit tests for the `get_env_params_*` functions in `env_params.py`


leewujung:
@emiliom : I think this is ready for a review when you are available. I will add that one integration test tomorrow morning.

leewujung:
@emiliom : This one is ready for a thorough review.

echopype/calibrate/calibrate_ek.py
```diff
@@ -130,22 +130,27 @@ class CalibrateEK60(CalibrateEK):
     def __init__(self, echodata, env_params, cal_params, **kwargs):
         super().__init__(echodata, env_params, cal_params)
 
-        # Set sonar_type
+        # Set sonar_type and waveform/encode mode
         self.sonar_type = "EK60"
```
emiliom, line 134: Small comment: we use `sonar_model` in the echodata class and elsewhere. If it's not too disruptive, I'd suggest calling it that here, too, to maintain consistency across echopype and reduce cognitive overload.

echopype/calibrate/calibrate_ek.py
```diff
@@ -130,22 +130,27 @@ class CalibrateEK60(CalibrateEK):
...
-        # Get env_params
-        self.env_params = get_env_params_EK60(echodata=echodata, user_env_dict=self.env_params)
         self.waveform_mode = "CW"
         self.encode_mode = "power"
 
-        # Compute range
-        self.compute_echo_range()
-
         # Get the right ed_group for CW power samples
         self.ed_group = retrieve_correct_beam_group(
```
emiliom, line 139: Consider renaming `ed_group` to, say, `ed_beam_group`? Just for clarity.

echopype/calibrate/env_params.py
```diff
@@ -59,199 +68,289 @@ def harmonize_env_param_time(
         return p
 
 
-def get_env_params_AZFP(echodata: EchoData, user_env_dict: Optional[dict] = None):
-    """Get env params using user inputs or values from data file.
+def sanitize_user_env_dict(
+    user_dict: Dict[str, Union[int, float, list, xr.DataArray]],
+    channel: Union[List, xr.DataArray],
```
emiliom, line 73: Any reason to use `list` in one case and `typing` `List` in the other?

echopype/calibrate/env_params.py
```diff
@@ -59,199 +68,289 @@ def harmonize_env_param_time(
...
     Parameters
     ----------
-    echodata : EchoData
-        an echodata object containing the env params to be pulled from
-    user_env_dict : dict
-        user input dict containing env params
+    user_dict : dict
+        A dictionary containing user input calibration parameters
+        as {parameter name: parameter value}.
+        Parameter value has to be a scalar (int or float) or an ``xr.DataArray``.
```
emiliom, line 87:
```suggestion
        Parameter value has to be a scalar (int or float), a list or an ``xr.DataArray``.
```

echopype/calibrate/env_params.py
emiliom: Actually, given how frequently you're using the combination of a list and a dataarray, I suggest you define a global variable for that union, then reuse that. It'll make for more compact and consistent code. eg:

```python
sequence = Union[List, xr.DataArray]

# ...

user_dict: Dict[str, Union[int, float, sequence]],
channel: sequence,
```

echopype/calibrate/env_params.py
```diff
@@ -59,199 +68,289 @@ def harmonize_env_param_time(
...
+    #  -- transform params in list to xr.DataArray
+    #  -- directly pass through params that are scalar or str
+    #  -- check channel coordinate if params are xr.DataArray and pass it through
+    out_dict = dict.fromkeys(ENV_PARAMS)
+    for p_name, p_val in user_dict.items():
+        if p_name in out_dict:
+            # Param "sound_absorption" has to be an xr.DataArray or a list because it is freq-dep
+            if p_name == "sound_absorption" and not (p_val, (xr.DataArray, list)):
+                raise ValueError(
+                    "The 'sound_absorption' parameter has to be an xr.DataArray, "
```
emiliom, line 124: So, does it have to be an xr.DataArray or either it or a list, per the comment above?

echopype/calibrate/env_params.py
```diff
@@ -59,199 +68,289 @@ def harmonize_env_param_time(
...
+
+    # Screen parameters: only retain those defined in ENV_PARAMS
+    #  -- transform params in list to xr.DataArray
+    #  -- directly pass through params that are scalar or str
+    #  -- check channel coordinate if params are xr.DataArray and pass it through
+    out_dict = dict.fromkeys(ENV_PARAMS)
+    for p_name, p_val in user_dict.items():
+        if p_name in out_dict:
+            # Param "sound_absorption" has to be an xr.DataArray or a list because it is freq-dep
+            if p_name == "sound_absorption" and not (p_val, (xr.DataArray, list)):
```
emiliom, line 122: Did you mean this?
```suggestion
            if p_name == "sound_absorption" and not isinstance(p_val, (xr.DataArray, list)):
```

echopype/calibrate/env_params.py
```diff
@@ -59,199 +68,289 @@ def harmonize_env_param_time(
...
+    # Check absorption and sound speed formula
+    if out_dict["formula_absorption"] is not None and not out_dict["formula_absorption"] in [
+        "AM",
+        "FG",
+    ]:
+        raise ValueError("'formula_absorption' has to be None, 'FG' or 'AM' for EK echosounders.")
+    if (
+        out_dict["formula_sound_speed"] is not None
+        and out_dict["formula_sound_speed"] != "Mackenzie"
+    ):
```
emiliom, line 286: I _think_ this works with `None`, too
```suggestion
    if out_dict["formula_sound_speed"] not in (None,  "Mackenzie"):
```

echopype/calibrate/env_params.py
```diff
@@ -59,199 +68,289 @@ def harmonize_env_param_time(
...
+    # Sound speed
+    if out_dict["sound_speed"] is None:
+        if not tspa_all_exist:
+            # sounds speed always exist in EK60 and EK80 data
+            out_dict["sound_speed"] = env["sound_speed_indicative"]
+            out_dict.pop("formula_sound_speed")
+        else:
+            # default to Mackenzie sound speed formula if not in user dict
+            if out_dict["formula_sound_speed"] is None:
+                out_dict["formula_sound_speed"] = "Mackenzie"
```
emiliom, line 313: see my suggestion and comment above
```suggestion
            if out_dict["formula_sound_speed"] in (None, "Mackenzie"):
```

emiliom (commented):
I wasn't thorough, but I think it looks good. I made a bunch of comments, most of them stylistic. 

There's just one where I think you're missing something important, an `inistance()`. Take a look.

echopype/calibrate/calibrate_ek.py
leewujung: Hmm, I used `sonar_type` because here I use "EK60" to stand for the older style of data, and "EK80" to stand for the newer style of data. Perhaps I can use "Ex60" (though this does not include ES70) and "Ex80" if we use `sonar_model`. Thoughts?

echopype/calibrate/calibrate_ek.py
leewujung: Sure, this is something inherited, but I agree to change it for clarify.

echopype/calibrate/env_params.py
leewujung: But if we do this, the type hint users get would also be `sequence`, which may not be clear to users?

echopype/calibrate/env_params.py
leewujung: Oh geez, yes! I don't know how I missed that! And I wonder why the test passes through!

echopype/calibrate/env_params.py
leewujung: Apparently, I have not finished the test cases I planned to do!! 🤯 Good that you caught this!

echopype/calibrate/env_params.py
leewujung: Apparently, I have not finished the test cases I planned to do!! 🤯 Good that you caught this!

echopype/calibrate/env_params.py
leewujung: Yes, I forgot to change this.
```suggestion
                    "The 'sound_absorption' parameter has to be a list or an xr.DataArray, "
```

echopype/calibrate/env_params.py
leewujung: Oh you are right, it does work with `None`! Great, this makes it so much more succint.

echopype/calibrate/env_params.py
```diff
@@ -59,199 +68,289 @@ def harmonize_env_param_time(
...
+
+    # Use sanitized user dict as blueprint
+    # out_dict contains only and all of the allowable cal params
+    out_dict = sanitize_user_env_dict(user_dict=user_dict, channel=beam["channel"])
+
+    # Check absorption and sound speed formula
+    if out_dict["formula_absorption"] is not None and not out_dict["formula_absorption"] in [
+        "AM",
+        "FG",
+    ]:
```
leewujung, line 281:
```suggestion
    if out_dict["formula_absorption"] not in [None, "AM", "FG"]:
```

echopype/calibrate/env_params.py
leewujung: I made the same change to the `formula_sound_absorption` if statement above too

echopype/calibrate/env_params.py
leewujung: I added this but this was actually to set the `formula_sound_speed` to Mackenzie when it does not exist, not an if statement like anove. I've reverted it back.

echopype/calibrate/env_params.py
leewujung: I've added this separately.

echopype/calibrate/calibrate_ek.py
leewujung: Done! In the process I discovered a redundancy in input argument for `_cal_complex_samples` and `_cal_power_samples` and removed them as well.

leewujung:
@emiliom : Thanks for reviewing this! I've made the following changes:
- committed all but one of the suggested changes (that one was incorrect even though the pattern looks similar, so I reverted)
- added missed test cases for `test_sanitize_user_env_dict`
- remove redundancy in passing in `ed_beam_group` in `_cal_power_samples` and `_cal_complex_samples`

I had questions about two of your comments so left them open.

echopype/calibrate/env_params.py
emiliom: That's a very good point. The strategy I suggested may not be a net benefit for user-facing functions, for the reason you bring up. But it may be a net benefit for internal functions, where code consistency is a higher priority.

So, scratch that. My only remaining comment here is the original one, which is very minor:

> Any reason to use `list` in one case and `typing` `List` in the other?

leewujung: Oh i forgot the `list` part. I'll change that to `List` in the function typing, for consistency.

echopype/calibrate/calibrate_ek.py
emiliom: Hmm, I don't know what "Ex60" and "Ex80" are, so I can't comment on that.

But this may be related to the broader relationship between Simrad-Kongsberg `sonar_model` entries (eg, "EK60", "ES70") and the broader sonar "type" ("EK60" vs "EK80") that determines the parser, set_group_, and many other processing decisions. I've created #995 just now to describe this relationship and a need for describing it centrally in an easily reusable way.

But back to the question at hand, *here.* No need to belabor my minor naming point. If you think `sonar_type` is more appropriate, stick with it. Or if you think this really is related to #995, stick with `sonar_type` and let's discuss the broader I raised in #995 issue after 0.7.0.

leewujung: "Ex60" and "Ex80" are what you called "class" in #995. I think we can keep it as is here, and address those code consolidation changes later in a more systematic way.

emiliom: sounds good

emiliom (approved):

Issue #737, opened 2022-06-21
Review/change handling of negative `echo_range` in calibration outputs


leewujung:
Right now we set all negative `echo_range` to 0 at the end of `compute_range`, which causes problems when trying to combine datasets (#578)
https://github.com/OSOceanAcoustics/echopype/blob/12693a72badcf54a18408e694b72e7b0e6d48023/echopype/echodata/echodata.py#L473

This issue is a reminder to review this and change/fix related operations in calibration.

leewujung:
See https://github.com/OSOceanAcoustics/echopype/pull/968#issuecomment-1464282197 for the divide by zero warnings on ek80_bb_complex and ek80_cw_power range tests.

emiliom:
@leewujung I've isolated the source of these log10 warnings. They're all in `calibrate_ek.py`.

- For power data, `tvg_mod_range` can be <= 0 in https://github.com/OSOceanAcoustics/echopype/blob/90e4377e7cb46953bf15d8b57f05ad0ab968d4de/echopype/calibrate/calibrate_ek.py#L68
- For complex data,  `tvg_mod_range` can be <= 0 in https://github.com/OSOceanAcoustics/echopype/blob/90e4377e7cb46953bf15d8b57f05ad0ab968d4de/echopype/calibrate/calibrate_ek.py#L378 and `prx` can be <= 0 in https://github.com/OSOceanAcoustics/echopype/blob/90e4377e7cb46953bf15d8b57f05ad0ab968d4de/echopype/calibrate/calibrate_ek.py#L406

Strictly speaking, something like this for both `tvg_mod_range` and `prx` would fix the warning:
```python
x = x.where(x > 0, eps)
```
where `eps` is a very small positive value (eg, 0.00001). But, I don't know what impact that would have on the calculations. On the other hand, those invalid `log10` applications must be leading to `nan` or `-inf` values right now ...

What do you suggest?

emiliom:
Can we close this now? I'm not 100% sure that the exact scope of the issue was addressed by #986

PR #986, opened 2023-03-15
Handling of non-positive values causing log10 warnings on EK calibration

emiliom:
Handle values <= 0 (0 and negative values) in calibration compute to eliminate log10 warnings. For the dataarrays causing these warnings, sets non-negative values to a slightly positive non-zero value.

Addresses #737

Note: when I ran the test suite locally there were a bunch of errors unrelated to the warnings addressed here. I'm going to assume there's some other mess in my test setup or some temporary glitch, and let the CI tests take a crack. All the log10 warnings in `test_range_integration.py` are gone.

### DIFF
--- a/echopype/calibrate/calibrate_ek.py
+++ b/echopype/calibrate/calibrate_ek.py
@@ -17,6 +17,7 @@ logger = _init_logger(__name__)
 # Small positive number, to use in place of zero
 EPS = 1e-5
 
+
 class CalibrateEK(CalibrateBase):
     def __init__(self, echodata: EchoData, env_params, cal_params):
         super().__init__(echodata, env_params, cal_params)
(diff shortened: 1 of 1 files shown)


emiliom:
There was just [one failure in the CI](https://github.com/OSOceanAcoustics/echopype/actions/runs/4422969355/jobs/7755260515#step:14:142), and it appears to be directly related to the change in this PR.

leewujung:
@emiliom : Thanks! I will take a look tomorrow morning!

leewujung:
I checked the CI failure and found the following:
- for the parts involving `np.log10(tvg_mod_range)`: the warning will appear because the first small section of `tvg_mod_range` is negative, and the value will be NaN
- for the parts involving `10 * np.log10(prx)`: the warning will appear because the last element is 0, so the resulting value will be Inf

Let's change both so that the values are replaced with NaN before the log10 operation, since these the ranges that NaN at output does not impact practical functionality.

emiliom:
Done. I've pushed a commit that replaces values with `np.nan` instead of 0. CI tests are churning.

PR #987, opened 2023-03-15
Fix bug to pass in `env_params` from parent class method for EK60

leewujung:
This PR fixes the bug that user-provided env params are not correctly taken in in `CalibrateEK60`.


emiliom (approved):
Looks good (after talking to you)!

Issue #212, opened 2020-12-16
Expose transducer serial number on the converted data from the raw unconverted data.


lsetiawan:
## Overview

For OOI data, there are serial numbers somewhere within the raw data files. It would be great if this can be shown in the converted nc/zarr file.

leewujung:
Currently we have the serial numbers for the transducer and for the transceiver in the EK80 nc/zarr files (but probably not in the most obvious place!). I need to check where things are in the OOI EK60 files.  I know they are parsed, but not sure if it made it's way into the nc/zarr... (again the decisions related to correspondence/discrepancy with the convention #210 )

emiliom:
> I need to check where things are in the OOI EK60 files. I know they are parsed, but not sure if it made it's way into the nc/zarr... 

At least for EK60, it's not being written into the `sonar_serial_number` attribute; see [here](https://github.com/OSOceanAcoustics/echopype/blob/class-redesign/echopype/convert/set_groups_ek60.py#L88). It looks to me like it'd be straightforward to fix this, if the information can be parsed easily from the EK60 datagrams.

leewujung:
> if the information can be parsed easily from the EK60 datagrams

Let's add this to **TODO**, I don't know immediately where in the parsed config datagram that is stored.

emiliom:
It would be nice to address this long-standing issue in the near future. Maybe @imranmaj could tackle it?

leewujung:
@imranmaj @lsetiawan : I looked into 3 EK60 test files we have, and the only place that I think has info resembling the serial number type is in field `"channel_id"` in `parses_obj` (if you set break point after [here]). 

It doesn't seem to me that the transducer serial number was recorded in these files.

@gavinmacaulay :
- Do you know what the middle string ("009072034d45", "0090720346a8", etc) are?
- To confirm: the "2-1", "6-1", "6-2" are: "transceiver channel-transducer number in that channel", is this correct?

---
### File from OOI-CE02SHBP
path: ./echopype/test_data/ek60/ooi/CE02/SHBP-MJ01C-07-ZPLSCB101_OOI-D20191201-T000000.raw

![Screen Shot 2022-02-27 at 9 13 01 AM](https://user-images.githubusercontent.com/15334215/155892318-38011a21-17ce-46aa-883b-2ed770f60585.png)

[...]

PR #992, opened 2023-03-16
Standardize use of sonar metadata as `Sonar` global attributes vs (for EK80 only) variables

emiliom:
Addresses all tasks in #681 involving sonar metadata in the `Sonar` group as implemented in echopype vs in the SONAR-netCDF convention.

Nearly all the changes are to EK80, where the main issue was the presence of some sonar metadata as variables rather than global attributes. Specifically:
- Changed `sonar_software_name` and `sonar_software_version` to be global attributes rather than variables
- Renamed the custom (non-convention) variable `serial_number` to `transducer_serial_number` (@leewujung please confirm if this is accurate; see https://github.com/OSOceanAcoustics/echopype/issues/681#issuecomment-1472502461)
- Added a `long_name` attribute to variables `sonar_serial_number`, `transducer_serial_number` and `transducer_name`

For AZFP, Set `sonar_software_name` to "AZFP" instead of "Based on AZFP Matlab Toolbox".

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -124,7 +124,7 @@ class SetGroupsAZFP(SetGroupsBase):
             "sonar_manufacturer": "ASL Environmental Sciences",
             "sonar_model": self.sonar_model,
             "sonar_serial_number": int(self.parser_obj.unpacked_data["serial_number"]),
-            "sonar_software_name": "Based on AZFP Matlab Toolbox",
+            "sonar_software_name": "AZFP",
             "sonar_software_version": "1.4",
             "sonar_type": "echosounder",
         }
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -194,6 +194,7 @@ class SetGroupsEK80(SetGroupsBase):
             "transducer_frequency",
             "serial_number",
             "transducer_name",
+            "transducer_serial_number",
             "application_name",
             "application_version",
             "channel_id_short",
@@ -236,18 +237,11 @@ class SetGroupsEK80(SetGroupsBase):
                     "standard_name": "sound_frequency",
                 },
             ),
-            "sonar_serial_number": (
-                ["channel"],
-                var["channel_id_short"],
-                {
-                    "long_name": "Sonar serial number",
-                },
-            ),
-            "transducer_serial_number": (
+            "transceiver_serial_number": (
                 ["channel"],
                 var["serial_number"],
                 {
-                    "long_name": "Transducer serial number",
+                    "long_name": "Transceiver serial number",
                 },
             ),
             "transducer_name": (
@@ -255,6 +249,13 @@ class SetGroupsEK80(SetGroupsBase):
                 var["transducer_name"],
                 {
                     "long_name": "Transducer name",
+                },
+            ),
+            "transducer_serial_number": (
+                ["channel"],
+                var["transducer_serial_number"],
+                {
+                    "long_name": "Transducer serial number",
                 },
             ),
         }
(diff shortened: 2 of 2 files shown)


echopype/convert/set_groups_azfp.py
```diff
@@ -124,7 +124,7 @@ def set_sonar(self) -> xr.Dataset:
             "sonar_manufacturer": "ASL Environmental Sciences",
             "sonar_model": self.sonar_model,
             "sonar_serial_number": int(self.parser_obj.unpacked_data["serial_number"]),
-            "sonar_software_name": "Based on AZFP Matlab Toolbox",
+            "sonar_software_name": "AZFP",
```
leewujung, line 127: I agree with this change as you proposed in https://github.com/OSOceanAcoustics/echopype/issues/681#issuecomment-1472502461

echopype/convert/set_groups_ek80.py
```diff
@@ -236,17 +236,27 @@ def set_sonar(self, beam_group_type: list = ["power", None]) -> xr.Dataset:
...
+            ),
+            "transducer_name": (
                 ["channel"],
-                var["application_version"],
-            ),  # identical for all channels
+                var["transducer_name"],
+                {
+                    "long_name": "Transducer name",
+                },
+            ),
```
leewujung, line 259: I think something is not quite right with these changes. 

- `channel_id_short` are the transducer model plus serial number
- `transducer_name` contains the transducer model. 
- `serial_number` contains the transceiver serial number (transceivers are things like GPT and WBT).

See https://github.com/OSOceanAcoustics/echopype/issues/212#issuecomment-1053626927 for example printouts, I think you could see how these numbers are combined to become what's in `channel_id`

leewujung (changes requested):
@emiliom : thanks for looking into this.I think there's something needing changes in the EK80 block. Please see my comments below.

echopype/convert/set_groups_azfp.py
emiliom: Great, thanks.

echopype/convert/set_groups_ek80.py
emiliom: To clarify, my changes in the block you're commenting on amount to just two things:
- Rename `serial_number` to `transducer_serial_number`
- Add a `long_name` attribute with values that are just the variable name spelled out.

I'm not sure what changes you're proposing. But since `transducer_name` and `serial_number` are variable names from echopype, not the convention, we can rename them as we see fit. Are you proposing we rename them to `transducer_model` and `transceiver_serial_number`, respectively?

As for `sonar_serial_number`, that's defined as a global attribute in the convention and as a variable by echopype, for EK80 only b/c it varies with channel. In your examples in #212, it's "0" in one case (which I interpret as not available / known) and "202" in another (a suspiciously short value for a serial number, IMHO). Should we use that parameter instead of `channel_id_short`? In https://github.com/OSOceanAcoustics/echopype/issues/212#issuecomment-1099298069 you had equated transducer serial number with `channel_id_short`.

leewujung: My comment above is to clarify what each of the variables are coming out from the parser, and that transceiver and transducer are different things.

I think `sonar_serial_number` is ill-defined since it can mean too many things. 

Below is what I propose:
- store the parsed `transducer_name` as `transducer_name` in echopype (ie. no change), so that it is easy for folks familiar with EK80 raw data know what they are
- store the parsed `serial_number` as `transceiver_serial_number` in echopype, so that it is specific that it is not the transducer serial number
- store the parsed `transducer_serial_number` as `transducer_serial_number` in echopype (ie. no change), for clarify
- do not store `sonar_serial_number` for EK80: This is marked as recommended and not mandatory in the convention. I think it is confusing what it is in this context. And, not populating it makes it consistent with EK60. 😬

`channel_id_short` seems a combination of `transducer_name` and `transducer_serial_number`, so I am thinking that we can safely ignore it.

emiliom: I've pushed a commit that implements this:
- `sonar_serial_number` is an empty global attribute, no longer a variable
- variables:
  - `transducer_name`, based on parser `transducer_name` parameter
  - `transducer_serial_number`, based on parser `transducer_serial_number` parameter
  - `transceiver_serial_number`, based on parser `serial_number` parameter

That follows what you recommended, but please review.

leewujung:
Looks great, please feel free to merge. Thanks for putting in comments about the `sonar_serial_number`. I checked the counterpart in EK60 and was happy to find that it is in there too! :)

emiliom:
Thanks! Will merge.

Issue #889, opened 2022-11-28
Create and move functions to `commongrid` and `clean` subpackages


leewujung:
**UPDATE, 3/20/2023:** We've settled on `commongrid` instead of `unify`. Also, expanded this issue to include moving the noise functions to the new `clean` subpackage (which we initially had named `filter`).

To manage functions that all could be called as "processing" or "preprocessing" type, in #817 we laid out some subpackage organizations. This issue addresses one of them, to create a new subpackage called `unify` and move a subset of the functions currently under `preprocess` under it:
- `unify`
  - `compute_MVBS` (existing) -- the efficiency was enhanced in #878
  - `compute_MVBS_index_binning` (existing) -- we know more work may be needed for this, but we won't change the implementation at the moment.
  - `regrid_Sv` (to be implemented) -- See https://github.com/OSOceanAcoustics/echopype/issues/726

leewujung:
I don't think "unify" and "consolidate" are exactly synonyms, but if there are other names that are better we can use those. 

What I think the PR addressing this issue should achieve is to move those functions into a separate subpackage that is not called "preprocess." I think we have until v0.6.4 is released to finalize what subpackage name we want. 😛

emiliom:
Currently there is a third user function under `preprocess`: `remove_noise`. Our ultimate goal is to move (reorganize) it to a new subpackage, `filter` (per https://github.com/OSOceanAcoustics/echopype/pull/817#issuecomment-1294332886). I think we should try to move it at the same time as we move `compute_MVBS` and `compute_MVBS_index_binning`, so we can retire `preprocess` all at once.

BTW, as we move these functions, we should look into maintaining working stubs under `preprocess` that simply point to the new location of the functions and emit a warning that function calls using `preprocess` will be retired in 0.7.1. eg, a request for `preprocess.compute_MVBS` will result in a request to `unify.compute_MVBS` plus a warning.

leewujung:
> I think we should try to move it at the same time as we move `compute_MVBS` and `compute_MVBS_index_binning`, so we can retire `preprocess` all at once.

I like this idea, this way we won't be moving things around that much afterwards.

Also agreed with deprecation warning and remove everything in 0.7.1.

emiliom:
We've settled on `commongrid` instead of `unify`. PR #993 addresses the reorganization of functions to both `commongrid` and `filter`. BUT, we now have an open question about `filter`, because it's a Python reserved word :weary:

[...]

PR #993, opened 2023-03-16
Move `preprocess` functions and tests new `clean` and `commongrid` subpackages

emiliom:
~~Moved noise functions (public API: `estimate_noise` and `remove_noise`) from `preprocess` to `filter`. Split off noise tests in `preprocess/test_preprocess.py` to `filter/test_noise.py`~~

Fully addresses #889,  moving functions from `preprocess` to `filter` or `commongrid`. We agreed to call the new subpackage `commongrid` rather than `unify` (I'll make a note of it in #889). Updated all tests.

See https://github.com/OSOceanAcoustics/echopype/pull/817#issuecomment-1294332886, but note that I did not carry out the `filter.noise` hierarchy proposed there. Looking through all subpackages, there are only a couple of cases of another subpackage (folder) under a subpackage, and those are not publicly exposed.

**NOTE: We'll likely need to revisit the subpackage name `filter`, because `filter` is a Python reserved word!**

~~Partly addresses #889. This PR could be treated as stand-alone, with a follow up PR to address the rest of #889. Or we could add those other changes on top of this PR, when ready.~~


leewujung:
> Partly addresses #889. This PR could be treated as stand-alone, with a follow up PR to address the rest of #889. Or we could add those other changes on top of this PR, when ready.

Hey @emiliom : thanks for submitting this! I think we could keep this PR open so that you could add the rest of #889 on top of this? I can review this PR again at that time. I did look through the changes in this one quickly and your proposed changes to where `estimate_noise` sits is fine by me.

echopype/filter/api.py
```diff
@@ -0,0 +1,72 @@
+"""
+Functions for enhancing the spatial and temporal coherence of data.
```
leewujung, line 2: I think this is the description to use for the `unify` subpackage. 
I would suggest replacing this with "Functions for reducing variabilities in backscatter data."

emiliom: Thanks for catching that and offering a replacement. It was an oversight. I'll push a commit with the change.

emiliom: Done.

emiliom:
Alright, it took me much more work than I anticipated to complete the rest of this PR, but here it is.

- All preprocess functions have now been distributed to either `filter` or `commongrid`
- All tests were updated and renamed as needed
- `.ci_helpers/run-test.py` is updated
- Added stub, pass-through functions to `preprocess` with a deprecation warning saying that they'll be removed in 0.7.1. I didn't add tests for these stub functions, since they seem unnecessary. I did test them locally, manually.
- Updated the provenance attribute statements to replace "preprocess" with "filter" or "commongrid". Note that I didn't make other changes to that code, since we're having a parallel conversation about the future of those attributes.

All tests ran locally.

echopype/commongrid/api.py
```diff
@@ -0,0 +1,217 @@
...
+        coords={
+            "ping_time": ping_interval,
+            "channel": ds_Sv.channel,
+            "echo_range": range_interval[:-1],
+        },
+    )
+
+    # Added this check to support the test in test_process.py::test_compute_MVBS
+    if "filenames" in ds_MVBS.variables:
+        ds_MVBS = ds_MVBS.drop_vars("filenames")
```
leewujung, line 84: Do you remember what this is? It's ok if not. I think the MVBS code needs to be looked at further as well anyways.

leewujung (approved):
@emiliom : Thanks for this PR! I think this is ready to be merged once you address the `estimate_noise` docstring change (which I know didn't come from you...). My other comments were just questions which may or may not lead to further changes.

echopype/__init__.py
```diff
@@ -18,10 +18,12 @@
     "open_raw",
     "open_converted",
     "combine_echodata",
+    "verbose",
     "calibrate",
+    "clean",
+    "commongrid",
```
leewujung, line 24: I have similar question about the order of these too like in https://github.com/OSOceanAcoustics/echopype/pull/993/files#r1142620782

.ci_helpers/run-test.py
```diff
@@ -27,13 +27,15 @@
     "root": {},  # This is to test the root folder.
     "convert": {},
     "calibrate": {},
+    "consolidate": {},
     "echodata": {},
+    "clean": {},
+    "mask": {},
+    "metrics": {},
+    "commongrid": {},
```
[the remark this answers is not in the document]
emiliom: Yeah. I think I started rearranging it alphabetically but then hesitated and stopped, to minimize changes unrelated to the goals of this PR. But the end result is mixed! 

I think we should go ahead and order it alphabetically, except for "root".

echopype/clean/api.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+
+from ..utils.prov import echopype_prov_attrs
+from .noise_est import NoiseEst
+
+
+def estimate_noise(ds_Sv, ping_num, range_sample_num, noise_max=None):
+    """
+    Remove noise by using estimates of background noise
+    from mean calibrated power of a collection of pings.
```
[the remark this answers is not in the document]
emiliom: Nice catch! I didn't pay any attention to that docstring. I'll merge your suggestion.

echopype/commongrid/api.py
emiliom: I don't remember.

echopype/__init__.py
emiliom: In this case, I followed the pattern that seemed to be in place: functions first, then subpackages. I moved `verbose` to follow that pattern more strictly (but `combine_echodata` doesn't follow that order ...). But we could just follow a strict alphabetical order. Any preference? The order has no user-visible impact, so it's more about code readability.

.ci_helpers/run-test.py
```diff
@@ -27,13 +27,15 @@
...
     "preprocess": {},
     "utils": {},
     "visualize": {},
```
emiliom, line 38: Here's the list in alphabetical order, leaving "root" as the first entry. I suggest we go with this order.

```suggestion
    "calibrate": {},
    "clean": {},
    "commongrid": {},
    "consolidate": {},
    "convert": {},
    "echodata": {},
    "mask": {},
    "metrics": {},
    "preprocess": {},
    "utils": {},
    "visualize": {},
```

.ci_helpers/run-test.py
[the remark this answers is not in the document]
emiliom: I've gone ahead and re-sorted alphabetically, except for root.

emiliom:
I think the only open issue is whether to reorder the `__all__` strings in `echopype/__init__.py`. Any preference?

echopype/commongrid/api.py
leewujung:
```suggestion
    # TODO: look into why 'filenames' exist here as a variable
    # Added this check to support the test in test_process.py::test_compute_MVBS
    if "filenames" in ds_MVBS.variables:
        ds_MVBS = ds_MVBS.drop_vars("filenames")
```
Let's add a TODO here.

echopype/__init__.py
leewujung: Let's do alphabetical order then? That way it's easier to find and update things if needed in the future.

leewujung:
> whether to reorder the `__all__` strings in `echopype/__init__.py`. Any preference?

I think it will be good to reorder that as well! :)

echopype/__init__.py
emiliom: Done.

PR #1000, opened 2023-03-20
Add more add_location tests and weak add_location

emiliom:
Addresses all but one of the tests specified in #769.

`test_consolidate_integration.py::test_add_location` now includes two additional sample raw files, including an AZFP file. The test files now encompass:
- EK60 with Platform lat-lon data (the file previously tested)
- EK60 with no lat-lon data (from OOI)
- AZFP with no lat-lon data, where the test uses `update_location` to first add a fixed-location single point

The tests in this PR currently include:
- Previously existing, but now run also on an AZFP file:
  - Presence of `latitude` and `longitude` variables
  - Absence of `time` variables
  - Running the above set of tests on `nmea_sentence="GGA"` for the EK60 dataset with Platform lat-lon data
- New tests:
  - When source echodata object has no valid lat-lon data, test for expected `ValueError`
  - No `longitude` or `latitude` value is null
  - `longitude` and `latitude` variables have a single dimension, `ping_time`
  - (Added in latest commit) Interpolated lat/lon values added by `add_latlon` function match up with values interpolated in the test
  - (Added in latest commit) For single-point case, lat-lon coordinates each contains a single unique value that matches the value passed to `update_platform`

~~Remaining test: Checking the added (interpolated) lat/lon values between the `add_latlon` function and directly computed from `scipy.interpolate`~~

Also, see https://github.com/OSOceanAcoustics/echopype/issues/769#issuecomment-1477089359 for findings about `time1` vs `time3` dimensions associated with `water_level`, and for the outcome of the first interpolated `latitude` and `longitude` value being `nan`. In this PR I modified `add_location` to ~~set the first value of interpolated `latitude` and `longitude` equal to the second value~~ use the [`scipy.interpolate.interp1d`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d) `fill_value="extrapolate"` scheme. Let's discuss this.


emiliom:
Ok, done. I decided to use xarray interpolate in the test rather than `scipy.interpolate`, for expedience. I don't think we gain much by using `scipy.interpolate`; or at least, I don't think it's worth the extra effort right now.

echopype/consolidate/api.py
```diff
@@ -168,7 +168,8 @@ def sel_interp(var):
             )
         else:
             # Interpolation. time1 is always associated with location data
-            return coord_var.interp(time1=ds["ping_time"])
+            # The first value ends up being always nan. Use "extrapolate"
+            return coord_var.interp(time1=ds["ping_time"], kwargs={"fill_value": "extrapolate"})
```
leewujung, line 172: I think this may be that the first `ping_time` precedes the first NMEA sentence time, so there's nothing to interpolate against. How about we keep it as NaN? That is technically more correct (since we do not know the exact ping location), and also likely avoid problems later on for computing distances between pings (for NASC computation).

For most of the ship survey extrapolating the first ping is fine since the ping rate is high, but if that's not the case the extrapolation error may be non-negligible.

leewujung: BTW, how about changing `coord_var` to something like `da_var`? When I was reading the code, I spent a few confused minutes trying to figure out why we can interpolate the coordinates of an xr data array. 😅

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -223,6 +205,119 @@ def _create_array_list_from_echoview_mats(paths_to_echoview_mat: List[pathlib.Pa
...
+    ed = ep.open_raw(raw_path, xml_path=xml_path, sonar_model=sonar_model)
+    env_params = None
+    if location_type == "fixed-location":
+        point_ds = xr.Dataset(
+            {
+                "latitude": (["time"], np.array([float(extras['latitude'])])),
+                "longitude": (["time"], np.array([float(extras['longitude'])])),
+            },
+            coords={
+                "time": (["time"], np.array([ed['Sonar/Beam_group1'].ping_time.values.min()]))
```
leewujung, line 258:
```suggestion
                "time": (["time"], np.array([ed["Sonar/Beam_group1"]["ping_time"].values.min()]))
```
I found it easier to read the code to always use ["string"] form for variables and coordinates, so that it's not confused with `.values`. Is there an actual difference? (something I didn't look up and never knew!)

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -223,6 +205,119 @@ def _create_array_list_from_echoview_mats(paths_to_echoview_mat: List[pathlib.Pa
...
+            },
+        )
+        ed.update_platform(point_ds)
+
+        # AZFP require external salinity and pressure
+        if sonar_model == "AZFP":
+            env_params = {
+                'temperature': ed["Environment"]['temperature'].values.mean(),
+                'salinity': extras['salinity'],
+                'pressure': extras['pressure'],
```
leewujung, line 268: nitpicking here: could you change all your `'` to `"`?

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -223,6 +205,119 @@ def _create_array_list_from_echoview_mats(paths_to_echoview_mat: List[pathlib.Pa
...
+                "longitude": (["time"], np.array([float(extras['longitude'])])),
+            },
+            coords={
+                "time": (["time"], np.array([ed['Sonar/Beam_group1'].ping_time.values.min()]))
+            },
+        )
+        ed.update_platform(point_ds)
+
+        # AZFP require external salinity and pressure
+        if sonar_model == "AZFP":
```
leewujung, line 264: This if block is indented. It does not matter because the only case testing for fixed location is under AZFP, but it seems that this may be an accidental indentation?

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -223,6 +205,119 @@ def _create_array_list_from_echoview_mats(paths_to_echoview_mat: List[pathlib.Pa
...
+            # coordinate existence
+            assert "latitude" in ds_test
+            assert "longitude" in ds_test
+            assert "time1" not in ds_test
+
+            # no coordinate value is nan
+            assert not ds_test["longitude"].isnull().any()
+            assert not ds_test["latitude"].isnull().any()
+
+            # coordinate has single dimension, ping_time
```
leewujung, line 290:
```suggestion
            # coordinate has a single dimension: 'ping_time'
```

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -223,6 +205,119 @@ def _create_array_list_from_echoview_mats(paths_to_echoview_mat: List[pathlib.Pa
...
+    else:
+        def _tests(ds_test, location_type, nmea_sentence=None):
+            # coordinate existence
+            assert "latitude" in ds_test
+            assert "longitude" in ds_test
+            assert "time1" not in ds_test
+
+            # no coordinate value is nan
+            assert not ds_test["longitude"].isnull().any()
+            assert not ds_test["latitude"].isnull().any()
```
leewujung, line 288: These will have to change if we do not extrapolate the first position.

leewujung (changes requested):
@emiliom : I am done reviewing this. My comments are mostly small and some completely textual. The only exception is the one involving extrapolation. Let me know what you think on that.

echopype/tests/consolidate/test_consolidate_integration.py
emiliom: I'm pretty sure there's no difference in terms of computation or interpretation. But I'm happy to go with your suggestion, which is probably more consistent anyways. I'll merge it.

echopype/tests/consolidate/test_consolidate_integration.py
emiliom: Will do. And yikes, I was pretty messy there, with mixed `'` and `"`!

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -223,6 +205,119 @@ def _create_array_list_from_echoview_mats(paths_to_echoview_mat: List[pathlib.Pa
...
+            },
+        )
+        ed.update_platform(point_ds)
+
+        # AZFP require external salinity and pressure
+        if sonar_model == "AZFP":
+            env_params = {
+                'temperature': ed["Environment"]['temperature'].values.mean(),
+                'salinity': extras['salinity'],
+                'pressure': extras['pressure'],
```
emiliom, line 268:
```suggestion
                "temperature": ed["Environment"]["temperature"].values.mean(),
                "salinity": extras["salinity"],
                "pressure": extras["pressure"],
```

echopype/tests/consolidate/test_consolidate_integration.py
emiliom: Hmm. The indentation was deliberate, but as you point out it doesn't matter here. Maybe it is better not to indent it, as that would potentially allow for additional AZFP files not tied to a "fixed-location" case (though those may be very rare).

For expedience, let's leave it as is for now, though.

echopype/consolidate/api.py
emiliom: Regarding `coord_var` vs `position_var`, I see your point. I'll make the change in my local clone and push it to the PR.

echopype/tests/consolidate/test_consolidate_integration.py
emiliom: It's fixed.

emiliom:
Thanks! I'll remove the extrapolation and update the tests to reflect that change.

echopype/tests/consolidate/test_consolidate_integration.py
emiliom: Right! Thanks. I've made the change locally and will push commits shortly.

echopype/consolidate/api.py
emiliom: In the latest commit I've removed extrapolation and renamed "coord" variables to "position" in both `consolidate/api.py` and the test.

echopype/tests/consolidate/test_consolidate_integration.py
emiliom: Done.

echopype/tests/consolidate/test_consolidate_integration.py
emiliom: I've removed these tests. Now that `nan` lon & lat values may be valid, I can't think of a general, useful test involving nan's. But I think the other tests already cover all ground: lon & lat contain the `ping_time` dimension (and no other dim), and the values match against interpolated values (that may include nan's) or fixed-point broadcasting. BTW, I modified the comparison of interpolated values to account for nan's.

leewujung (approved):
Looks good. Thanks @emiliom. Please feel free to merge this!

Issue #980, opened 2023-03-13
Add global attribute for data processing level


emiliom:
Add a single global attribute that specifies the data processing/product level for the dataset. 

The [ACDD](http://wiki.esipfed.org/index.php/Attribute_Convention_for_Data_Discovery) convention defines the *Recommended* global attribute `processing_level`, "A textual description of the processing (or quality control) level of the data". Usage of this attribute is also recommended (possibly as part of its ACDD implementation?) in the NASA Earth Science Data and Information System Standards Office document, "Data Product Development Guide (DPDG) for Data Producers version1.1", https://doi.org/10.5067/DOC/ESO/RFC-041VERSION1

Another relevant ACDD global attribute is `product_version`, "Version identifier of the data file or product as assigned by the data creator. For example, a new algorithm or methodology could result in a new product_version." ACDD classifies it as *Suggested*.

Related to #817

leewujung:
A minimal example of decorator to add attribute:
```python
In [29]: def test_add_attrs(func):
    ...:     def wrapper():
    ...:         da = func()
    ...:         da = da.assign_attrs({"test_attr": "some attribute"})
    ...:         return da
    ...:     return wrapper
    ...: 

In [30]: @test_add_attrs
    ...: def build_da():
    ...:     return xr.DataArray([1, 2, 3], coords={"channel": ["chA","chB","chC"]})
    ...: 

In [31]: da = build_da()

In [32]: da
Out[32]: 
<xarray.DataArray (channel: 3)>
array([1, 2, 3])
Coordinates:
  * channel  (channel) <U3 'chA' 'chB' 'chC'
Attributes:
    test_attr:  some attribute
```

leewujung:
Out of curiosity, I tried another one with parameters:
```python
In [50]: def test_add_attrs(some_str):
    ...:     def inner(func):
    ...:        def wrapper(*args, **kwargs):
    ...:           da = func()
    ...:           da = da.assign_attrs({"input_str": some_str})
    ...:           da = da.assign_attrs({"test_attr": "some attribute"})
    ...:           return da
    ...:        return wrapper
    ...:     return inner
    ...: 

In [51]: @test_add_attrs(some_str="add this string!")
    ...: def build_da():
    ...:     return xr.DataArray([1, 2, 3], coords={"channel": ["chA","chB","chC"]})
    ...: 

In [52]: da = build_da()

In [53]: da
Out[53]: 
<xarray.DataArray (channel: 3)>
[...]

PR #1001, opened 2023-03-22
Implementation of data-processing-level attributes

emiliom:
Draft, rough, partial implementation of data-processing-level attribute insertion, addressing #980.  I've got a "representative" core working now, that handles:
- Assign L1A on the Top-level group of the EchoData object returned by  `open_raw`, but only if Platform lat-lon data are present.
- Add L2A on `consolidate.add_location`.
- Add L3A or L3B on `commongrid.compute_MVBS`. It looks for an existing `processing_level` attribute in the input `Sv` dataset. If it's not present, nothing is added. Otherwise, L3A or L3B are assigned based on the input `Sv` being L2A or L2B, respectively.

Two attributes are inserted: `processing_level` and `processing_level_url`. The `processing_level` value currently looks like this: "Level 1A". For EchoData object (Level 1A), the attributes are inserted into the Top-level group dataset.

The functionality is implemented as a decorator function, `add_processing_level`, with two optional arguments in addition to the required processing level code. There is also a complementary function `insert_input_processing_level` that must be called in the decorated function to enable sublevel "propagation" -- eg, L2A > L3A, L2B > L3B.

See https://github.com/uw-echospace/data-processing-levels/blob/main/data_processing_levels.md for the draft description of data processing levels (and discussions about it [here](https://github.com/uw-echospace/data-processing-levels/issues/4)), and https://github.com/OSOceanAcoustics/echopype/pull/817#issuecomment-1474564449 for a listing of data processing functions that will ultimately receive this feature.

The code is full of comments for myself at this point, though mostly outside actual functions. I'll set it as draft. As is, it should give you a good idea of how it works and what it covers.

No tests yet, though I've run it locally, manually, on a couple of specific cases. Tomorrow.

### DIFF
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -6,7 +6,7 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 
-from ..utils.prov import echopype_prov_attrs
+from ..utils.prov import echopype_prov_attrs, add_processing_level, insert_input_processing_level
 from .mvbs import get_MVBS_along_channels
 
 
@@ -35,6 +35,7 @@ def _set_MVBS_attrs(ds):
     }
 
 
+@add_processing_level("L3", inherit_sublevel=True)
 def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
     """
     Compute Mean Volume Backscattering Strength (MVBS)
@@ -138,6 +139,8 @@ def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
     prov_dict["processing_function"] = "commongrid.compute_MVBS"
     ds_MVBS = ds_MVBS.assign_attrs(prov_dict)
     ds_MVBS["frequency_nominal"] = ds_Sv["frequency_nominal"]  # re-attach frequency_nominal
+
+    ds_MVBS = insert_input_processing_level(ds_MVBS, input_ds=ds_Sv)
 
     return ds_MVBS
 
--- a/echopype/consolidate/api.py
+++ b/echopype/consolidate/api.py
@@ -9,6 +9,7 @@ from ..calibrate.ek80_complex import get_filter_coeff
 from ..echodata import EchoData
 from ..echodata.simrad import retrieve_correct_beam_group
 from ..utils.io import validate_source_ds_da
+from ..utils.prov import add_processing_level
 from .split_beam_angle import add_angle_to_ds, get_angle_complex_samples, get_angle_power_samples
 
 
@@ -128,6 +129,7 @@ def add_depth(
     return ds
 
 
+@add_processing_level("L2A")
 def add_location(ds: xr.Dataset, echodata: EchoData = None, nmea_sentence: Optional[str] = None):
     """
     Add geographical location (latitude/longitude) to the Sv dataset.
--- a/echopype/convert/api.py
+++ b/echopype/convert/api.py
@@ -16,6 +16,7 @@ from ..echodata.echodata import XARRAY_ENGINE_MAP, EchoData
 from ..utils import io
 from ..utils.coding import COMPRESSION_SETTINGS
 from ..utils.log import _init_logger
+from ..utils.prov import add_processing_level
 
 BEAM_SUBGROUP_DEFAULT = "Beam_group1"
 
@@ -306,6 +307,7 @@ def _check_file(
     return str(raw_file), str(xml)
 
 
+@add_processing_level("L1A", is_echodata=True)
 def open_raw(
     raw_file: "PathHint",
     sonar_model: "SonarModelsHint",
--- a/echopype/utils/prov.py
+++ b/echopype/utils/prov.py
@@ -1,11 +1,14 @@
 from datetime import datetime as dt
+import functools
 from pathlib import Path
+
 from typing import Dict, List, Tuple, Union
+from typing_extensions import Literal
 
 import numpy as np
-from _echopype_version import version as ECHOPYPE_VERSION
 from numpy.typing import NDArray
-from typing_extensions import Literal
+import xarray as xr
+from _echopype_version import version as ECHOPYPE_VERSION
 
 from .log import _init_logger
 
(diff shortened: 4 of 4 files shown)


emiliom:
I got the decorator working on the `echodata.update_platform` class method, too!

leewujung:
Thanks @emiliom, I did a quick scan through the code -- it's exciting to see these functionalities implemented!

My only comment at this moment is having to call `insert_input_processing_level` separately within the decorated function. Since you have access to and manipulate the dataset within the decorated function already in the decorator, it seems that you could have this as an optional argument that is a flag for whether or not to insert the input processing level attribute? The function does not have to change, but it would then be called under the decorator and not the decorated function.

emiliom:
> Thanks @emiliom, I did a quick scan through the code -- it's exciting to see these functionalities implemented!

Thanks! Yeah, isn't it? I was thrilled to also be able to reuse the same decorator function with class methods.

> My only comment at this moment is having to call `insert_input_processing_level` separately within the decorated function. Since you have access to and manipulate the dataset within the decorated function already in the decorator, it seems that you could have this as an optional argument that is a flag for whether or not to insert the input processing level attribute? The function does not have to change, but it would then be called under the decorator and not the decorated function.

Definitely, having to explicitly call that function *within* the decorated function is not ideal (though not awful, either). The challenge with folding this into the decorator function is having access to the "input" xr.Dataset. The decorator function does have access to all inputs, but it doesn't know which function argument is which. I suppose that if there's a common pattern for the position of the input dataset (or the use of a keyword argument), the decorator function could be made to work. For example, in `compute_MVBS` the input xr.Dataset is the first argument, so `args[0]` could be used. I'd have to examine all target functions to see if that's always the case; if it is, that'll be wonderful! I'll take a look, but I suspect things won't be so simple.

Either way, these implementation decisions are not user-facing, so in principle we could go with one scheme for 0.7.0 and change it later if a better option is identified.

emiliom:
In the latest commits:
- Added the decorator to `clean.remove_noise`, `mask.apply_mask` and `commongrid.compute_MVBS_index_binning`.
- Introduced a decorator `processing_level_code` "wildcard" pattern like "L*A", "L2*", etc (and removed the decorator argument `inherit_sublevel`, as it's no longer necessary). 
  - Sublevel wildcard (eg, L2*): The decorated function's "input" dataset's sublevel (A or B) will be propagated to the function's output dataset. Used with L2 and L3. A function that takes L2A and L2B and produces corresponding L3 products will result in L2A -> L3A and L2B -> L3B.
  - Level wildcard (eg, L*B): The decorated function's "input" dataset's level (2 or 3) will be propagated to the function's output dataset. Used with L2 and L3. A function that takes L2A and L3A and produces corresponding B-sublevel products will result in L2A -> L2B and L3A -> L3B.

I've cleaned up the  code in `prov.py`. I think this PR is ready for a final review and possible merge. I'm removing the draft label.

Regarding the folding of the `insert_input_processing_level` function into a call from the decorator function: I tried the strategy I mentioned in my previous comments. Using `arg[0]` does work, but there's a twist. When the corresponding first argument in the decorated functions can be called with an argument name, `arg[0]` is `None` and we'd have to rely on a `kwarg` whose name varies across functions. Let's leave this discussion for 0.7.1, when we can take a fresh look. I can open a new issue about it.

emiliom:
I just remembered that I haven't added any tests :disappointed:. I'll start working on that.

emiliom:
Added tests (in a new, single module) that includes two test files, a hake survey EK60 raw file and an AZFP raw file. The tests check for the correction addition of processing level attributes (or they expected absence) on multiple processing steps along a sequence:
- open_raw
- ed.update_platform, if needed
- compute_Sv
- consolidate.add_location
- clean.remove_noise
- mask.apply_mask on a frequency-difference mask based on: a) Sv with add_location; b) Sv with add_location and remove_noise
- compute_MVBS

@leewujung  This PR is ready for review! The only final change I expect to make, other than those based on your input, is on the final url for `processing_level_url`.

echopype/tests/utils/test_processinglevels_integration.py
```diff
@@ -0,0 +1,121 @@
...
+            "temperature": ed["Environment"]["temperature"].values.mean(),
+            "salinity": extras["salinity"],
+            "pressure": extras["pressure"],
+        }
+
+    ds = ep.calibrate.compute_Sv(echodata=ed, env_params=env_params)
+    _absence_test(ds)
+
+    sv_ds = ep.consolidate.add_location(ds=ds, echodata=ed)
+    assert "longitude" in sv_ds.data_vars and "latitude" in sv_ds.data_vars
```
leewujung, line 86: I suggest changing `sv_ds` to either `Sv_ds` or `ds_Sv` because Sv and sv unfortunately mean different things in the common notations...

echopype/tests/utils/test_processinglevels_integration.py
```diff
@@ -0,0 +1,121 @@
...
+    # On Sv w/o noise removal
+    ds = _freqdiff_applymask(sv_ds)
+    _presence_test(ds, "Level 3A")
+
+    # On denoised Sv
+    ds = _freqdiff_applymask(denoised_ds)
+    _presence_test(ds, "Level 3B")
+
+    # ---- Compute MVBS
+    # compute_MVBS expects a variable named "Sv"
```
leewujung, line 118:
```suggestion
    # compute_MVBS expects a variable named "Sv"
    # No product level is assigned because at present compute_MVBS drops the lat/lon data associated with the input Sv dataset
```
Just adding a clarification, since I was confused for a sec.

leewujung (approved):
@emiliom : Thanks for the PR -- My comments are small and IMO this is ready to go! 

BTW, I like the test scheme with `_presence_test` and `_absense_test` :)

Are you thinking the addition of `processing_level_url` will be another PR or in this one?

emiliom:
> Are you thinking the addition of processing_level_url will be another PR or in this one?

I think that'll be the cleanest. That way we can get this PR into dev ASAP.

In the meantime, I'll go through your comments now.

echopype/tests/utils/test_processinglevels_integration.py
emiliom: Done

emiliom:
Alright, the `sv_ds` to `Sv_ds` change is in. I'll merge the PR once the CI tests are complete.

Then I'll submit another PR updating `processing_level_url`, within the hour.

emiliom:
BTW, I have no idea what the windows build test failure is. I've never looked at that GH action.

leewujung:
Yeah, no idea why that particular windows test failed. The other windows test passed...!

PR #1005, opened 2023-03-24
Add `compute_NASC`

leewujung:
This PR adds a brute force version of `commongrid.compute_NASC`. 

The main operations are identical to those needed in `commongrid.compute_MVBS` and `clean.estimate_noise`, and since these core blocks need to be refactored and optimized, I decided to not to work on those here.

Note:
- Here we do not require the sampling interval in the along track (along `ping_time`) -- this is partly why the computation takes longer.
- The current implement assumes the Sv dataset having uniform depth step size. This restriction is easy to generalize but will require more computation, so we'll leave it as part the optimization task in the near future.


leewujung:
Oh haha, forgot to change the test file path. Will do that now.

echopype/commongrid/api.py
```diff
@@ -214,5 +221,108 @@ def compute_MVBS_index_binning(ds_Sv, range_sample_num=100, ping_num=100):
...
+            sv_mean_dist_depth.append(sv_mean_depth)
+
+        sv_mean.append(sv_mean_dist_depth)
+
+    # Compute mean height
+    # For data with uniform depth step size, mean height = vertical size of cell
+    height_mean = cell_depth
+    # TODO: generalize to variable depth step size
+
+    ds_NASC = xr.DataArray(
```
emiliom, line 312: `ds_NASC` is a DataArray. But the variable name and the function signature (`-> xr.Dataset`) and docstring point to an expectation of an xr.Dataset.

echopype/commongrid/api.py
```diff
@@ -214,5 +221,108 @@ def compute_MVBS_index_binning(ds_Sv, range_sample_num=100, ping_num=100):
...
+        np.array(sv_mean) * height_mean,
+        dims=["channel", "distance", "depth"],
+        coords={
+            "channel": ds_Sv["channel"].values,
+            "distance": np.arange(bin_num_dist) * cell_dist,
+            "depth": np.arange(bin_num_depth) * cell_depth,
+        },
+    )
+
+    # TODO: Attach attributes
```
emiliom, line 322: Variable (DataArray) attributes, right?

echopype/commongrid/api.py
leewujung: Oh right, I forgot about this. I was initially thinking we might need to add other info (in the form of data variables) for NASC. 

Looking at what we have for `compute_MVBS`, I think we probably need to add `frequency_nominal` (coordinate `channel`). Seems something we should keep around, and there's not actual difference once it's saved since things are always saved as a dataset. What do you think?

echopype/commongrid/api.py
leewujung: Yes, I was thinking something along the line of what we have for `compute_MVBS`.

echopype/commongrid/api.py
```diff
@@ -214,5 +221,108 @@ def compute_MVBS_index_binning(ds_Sv, range_sample_num=100, ping_num=100):
...
+    cell_depth: Union[int, float],  # TODO: allow xr.DataArray
+) -> xr.Dataset:
+    """
+    Compute Nautical Areal Scattering Coefficient (NASC) from an Sv dataset.
+
+    Parameters
+    ----------
+    ds_Sv : xr.Dataset
+        A dataset containing Sv data.
+        The Sv dataset must contain 'latitude' and 'longitude' as data variables.
```
emiliom, line 236: And `depth` too, right?

echopype/commongrid/api.py
emiliom: Ah. Thanks for pointing to `compute_MVBS`; that'll make it easier!

echopype/commongrid/api.py
emiliom: Definitely, let's add `frequency_nominal`!

echopype/commongrid/api.py
leewujung: Oh gosh, yes!

echopype/commongrid/api.py
leewujung: Ok, I will add that now along with the docstring addition of requiring `depth`.

emiliom:
On a general observation, it looks like the computation of NASC results in the complete disappearance of timestamp information. Assuming I go that right, I think the resulting product will be lacking an important context. 

We can fix this easily by adding timestamp start and end attributes, ideally using the ACDD attributes  [`time_coverage_start`](https://wiki.esipfed.org/Attribute_Convention_for_Data_Discovery_1-3#time_coverage_start) and `time_coverage_end`. A next step up (in *addition* to those attributes) would be to create a mean time variable with `distance` dimension. **IF** we think there's time today to do that, I'd also throw in mean longitude and latitude variables, also with `distance` dimension. Regardless, at a minimum we can also easily add ACDD attributes [`geospatial_lat_min`](https://wiki.esipfed.org/Attribute_Convention_for_Data_Discovery_1-3#geospatial_lat_min) and `geospatial_lat_max`, ditto for `lon`.

echopype/commongrid/api.py
emiliom: Great. I see that this is taken care of. I'll pull this into my local clone and work on the attributes.

emiliom:
Addresses #730

echopype/commongrid/api.py
leewujung: Added.

echopype/commongrid/api.py
emiliom: The conversion to Dataset fails b/c the DataArray requires a name. I've fixed this locally so I can work on adding the attributes.

Also, the test needs to be updated now that a dataset is returned. I've fixed that locally, too.

emiliom:
@leewujung see the PR I submitted against your fork, https://github.com/leewujung/echopype/pull/35

emiliom (approved):
I'm done reviewing, plus my changes via the PR to your fork.

Good to go, IMHO -- short of adding the mean time, lon and lat variables (with `distance` dim) that I mentioned, if we wanted to do it now.

leewujung:
> short of adding the mean time, lon and lat variables (with `distance` dim) that I mentioned, if we wanted to do it now.

I have mixed feeling about this, mainly two things:
- right now the `distance` and `depth` dimensions uses the starting bin edges and not the mean or mid point of the bins (see next point) as the values -- we should probably stick with one scheme
- for the means: should it be true means (as in the mean of the values), or should it just be the middle point of the bins?

Thoughts?

emiliom:
> I have mixed feeling about this, mainly two things:
>
> - right now the distance and depth dimensions uses the starting bin edges and not the mean or mid point of the bins (see next point) as the values -- we should probably stick with one scheme
> - for the means: should it be true means (as in the mean of the values), or should it just be the middle point of the bins?
> Thoughts?

Hmm. I agree that this makes things trickier, and not just a matter of coding it correctly, quickly.

I suggest we leave this for 0.7.1. We can discuss these things then. At least now the global attributes do provide the general temporal and spatial context.

leewujung:
I agree -- let's revisit this in v.7.1. I'll merge this once the tests run through.

Issue #963, opened 2023-03-08
Update Windows CI so it can also work to run the full test suite


lsetiawan:
## Overview

Currently the Windows CI only runs the `utils` module and NOT the full test suite for echopype (https://github.com/OSOceanAcoustics/echopype/actions/workflows/windows-utils.yaml). This needs to be updated so that the whole test suite will run for Windows (https://github.com/OSOceanAcoustics/echopype/actions/workflows/windows.yaml). The backbone is there, but it's currently not activated due to Docker limitations in working with Windows.

To make this work, we can potentially do the following:
- Manually setup minio straight into the CI server: https://github.com/minio/minio#microsoft-windows
- Manually setup apache http server into the CI server: https://httpd.apache.org/docs/2.4/platform/windows.html
- Perform rclone at the beginning of the CI run with the `gdrive-rclone` action that we have: https://github.com/OSOceanAcoustics/echopype/tree/main/.github/actions/gdrive-rclone

LOCEANlloydizard:
Hi all,

As I'm developing on Windows, I encountered errors locally when running the tests (same as #883). I wanted to compare with Windows runners, but as mentioned in this issue there's Docker limitations in working with Windows (from what I've read, it will stay like this).

@leewujung mentioned GitHub assets as an alternative, so I explored that and managed to get a Windows runner working on my fork, pulling data from a GitHub release instead of Docker. For the integration, as I don’t have write access to the main echopype repo, a maintener needs to create and load the test data in the release (I've written what worked for me in my fork below).

After running the test, and after a quick look at the test results, I’m seeing the same errors as my local ones and as well as #883, mostly related to how temporary files are handled on Windows. I’ll dig into that soon, I think it's just a matter of how the test close the files?

<img width="1897" height="865" alt="Image" src="https://github.com/user-attachments/assets/8c6e94f8-2b0d-434f-b468-783c5bcacaf3" />



################ Workflow to import data in github assets to have Windows CI runner for echopype

**1. Extract the test data from docker** (if you don't have it locally)

We first pull the latest Docker images and deply test data services locally using the helper script:

    `python .ci_helpers/docker/setup-services.py --deploy`

This copies a large set of test data folders into the local echopype/test_data directory.

**2. Zip each test data folder separately**

The _**separately**_ is important to overcome the GitHub Release asset import size limit (~2GB), so we zip each subfolder individually instead of the entire test data at once.

Example PowerShell code to zip each folder:

```
[...]

PR #1007, opened 2023-03-24
Update windows-utils.yaml

emiliom:
Remove `test_raw_to_mvbs` in `test_processinglevels_integration.py` from the `windows-utils.yaml` CI test. Unlike other `utils` tests, this one uses a data file and that was causing a failure, like [this one](https://github.com/OSOceanAcoustics/echopype/actions/runs/4515367995/jobs/7952579646)

We'll revisit this in the near future.

### DIFF
--- a/.github/workflows/windows-utils.yaml
+++ b/.github/workflows/windows-utils.yaml
@@ -66,4 +66,4 @@ jobs:
             python -m pip install -e .
         - name: Running all Tests
           run: |
-            pytest -k 'not test_raw_to_mvbs' -vvv -rx --cov=echopype --cov-report=xml --log-cli-level=WARNING --disable-warnings echopype/tests/utils |& tee ci_${{ matrix.python-version }}_test_log.log
+            pytest -k "not test_raw_to_mvbs" -vvv -rx --cov=echopype --cov-report=xml --log-cli-level=WARNING --disable-warnings echopype/tests/utils |& tee ci_${{ matrix.python-version }}_test_log.log
--- a/.github/workflows/windows-utils.yaml
+++ b/.github/workflows/windows-utils.yaml
@@ -66,4 +66,4 @@ jobs:
             python -m pip install -e .
         - name: Running all Tests
           run: |
-            pytest -vvv -rx --cov=echopype --cov-report=xml --log-cli-level=WARNING --disable-warnings echopype/tests/utils |& tee ci_${{ matrix.python-version }}_test_log.log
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-utils.yaml
```diff
@@ -66,4 +66,4 @@ jobs:
             python -m pip install -e .
         - name: Running all Tests
           run: |
-            pytest -vvv -rx --cov=echopype --cov-report=xml --log-cli-level=WARNING --disable-warnings echopype/tests/utils |& tee ci_${{ matrix.python-version }}_test_log.log
+            pytest -k 'not test_raw_to_mvbs' -vvv -rx --cov=echopype --cov-report=xml --log-cli-level=WARNING --disable-warnings echopype/tests/utils |& tee ci_${{ matrix.python-version }}_test_log.log
```
emiliom, line 69: Maybe Windows doesn't like single quotes, so changing to double quotes.
```suggestion
            pytest -k "not test_raw_to_mvbs" -vvv -rx --cov=echopype --cov-report=xml --log-cli-level=WARNING --disable-warnings echopype/tests/utils |& tee ci_${{ matrix.python-version }}_test_log.log
```

emiliom:
Well, this isn't fully working. 2 out of 3 of the windows-utils testes do work as expected, and one doesn't. It's as if the change I made is being ignored, or the CI is using a cache and not the yaml file itself.

In the [job report](https://github.com/OSOceanAcoustics/echopype/actions/runs/4516288851/jobs/7954443940?pr=1007#step:8:1), the `pytest` statement doesn't have the `-k "not test_raw_to_mvbs"` option I introduced!

In the other successful tests, the option is present.

@leewujung can we ignore this one failure? Does it impact anything with the build?

leewujung:
Hmm I am not sure what tests are run in conda feedstock. What about if we move `test_processinglevels_integration.py` outside of `tests/utils`? then it won't be run in windows-utils-test...

emiliom:
> What about if we move `test_processinglevels_integration.py` outside of `tests/utils`? then it won't be run in windows-utils-test...

What I recommend for today/tomorrow, for 0.7.0, is that we simply xfail this test. We know the tests work, locally and on the CI. Once the release is issued, we remove the xfail. Then we can address the overall issue more carefully.

Moving it elsewhere should work too, but I think it's a bigger change to things and one we'd also have to revisit after 0.7.0.

leewujung:
Sure let's give that a try (xfail). In one of the previous runs I saw other tests not working in the same form as not being able to find files so I was not sure what exactly is going on.

leewujung:
I'll close this now since the issue was resolved by #1012. The need for windows testing is tracked in #963.

PR #1013, opened 2023-03-25
Move test to to prevent windows-utils action from running it

emiliom:
Move test to directory other than utils to prevent windows-utils action from running it.

Previous attempts, none with full success: PRs #1007, #1011, #1012. The `pull_request_target` checks keep on ignoring whatever setting there is


emiliom:
The tests all pass, but it may just be what you mentioned elsewhere: Merging PR #1012 may have already fixed things (which probably means that if we had merged PR #1007, the problem may have been fixed, too)

I won't merge this.

leewujung:
I'll close this now since #1012 seems to have worked.

Issue #1018, opened 2023-03-29
impedance int > float


rolandproud:
Hi, impedance is not always an integer. I suggest changing to a float e.g. line 854 in ek_raw_parsers.py. I have some EK80 files where impedance varies and is a float.

leewujung:
@rolandproud : Great, thanks for pointing this out! We can put this in no problem!

leewujung:
This is addressed in #1019.

PR #1019, opened 2023-04-02
Impedance type change, name change, bug fix

leewujung:
This PR contains the following changes:
- change the data type of impedance to `float` (addresses #1018)
- finish renaming `impedance_transmit` to `impedance_transducer` and `impedance_receive` to `impedance_transceiver` (leftover that resulted into bug from v0.7.0)
- rename transmit chirp replica construction methods: 
   - correct typo `Anderson` to `Andersen`
   - ~change `Macaulay` to `Matlab` since original code came from Andersen~
   - remove the previous `Macaulay` chirp replica implementation and only keep the one in reference to Andersen

### DIFF
--- a/echopype/calibrate/calibrate_ek.py
+++ b/echopype/calibrate/calibrate_ek.py
@@ -192,8 +192,8 @@ class CalibrateEK60(CalibrateEK):
 class CalibrateEK80(CalibrateEK):
     # Default EK80 params: these parameters are only recorded in later versions of EK80 software
     EK80_params = {}
-    EK80_params["z_et"] = 75  # transmit impedance
-    EK80_params["z_er"] = 1000  # receive impedance
+    EK80_params["z_et"] = 75  # transducer impedance
+    EK80_params["z_er"] = 1000  # transceiver impedance
     EK80_params["fs"] = {  # default full sampling frequency [Hz]
         "default": 1500000,
         "GPT": 500000,
@@ -407,6 +407,10 @@ class CalibrateEK80(CalibrateEK):
             EchoData["Sonar/Beam_group1"] with selected channel subset
         chirp : dict
             a dictionary containing transmit chirp for BB channels
+        z_et : float
+            impedance of transducer [ohm]
+        z_er : float
+            impedance of transceiver [ohm]
 
         Returns
         -------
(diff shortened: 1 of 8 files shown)


emiliom:
> change `Macaulay` to `Matlab` since original code came from Andersen

I don't know anything about this, but an implementation name of "Matlab" seems not specific enough, unlike Andersen and Macaulay. The Andersen implementation code has a helpful 2-line comment that provides background. Is the "Matlab" implementation from a specific m-script file, from the general Matlab code base, or something else?

echopype/calibrate/calibrate_ek.py
```diff
@@ -407,6 +407,10 @@ def _get_power_from_complex(
             EchoData["Sonar/Beam_group1"] with selected channel subset
         chirp : dict
             a dictionary containing transmit chirp for BB channels
+        z_et : float
```
emiliom, line 410: How about adding `float` type hints to `z_et` and `z_er`, while you're at it? Especially given that the other function arguments already have type hints.

leewujung: Done!

leewujung:
@emiliom : I've added type hint to `z_et` and `z_er`. Also added a source reference to the gain correction based on angle offset and beamwidth that was undocumented anywhere else -- I asked Lars about it and added the explanation in the docstring.

Let me know what you think about the "Matlab" implementation of chirp.

emiliom:
> The "Matlab" implementation came from Lars Andersen in matlab Echolab code that was passed through multiple people (I got it from Chu).

:+1: 

> In fact, I think the best way to do this is to remove this implementation completely. I spoked with Lars about this and the specific chirp implementation is due to the implementation in hardware, so we should stick to it.

Well, the issue goes away if you remove the "Matlab" implementation :wink:. So, no additional comment here. It's up to you whether to keep it or not. If you do keep it, my suggestion would be to simply change the label to something like "MatlabEcholab" and add a comment like the sentence above describing its origin.

leewujung:
I think I'll opt for removing that section entirely, since it does not have the simplicity of just using a general chirp function, not does it conform with what Simrad actually uses. I'll push up a commit for the removal.

Issue #198, opened 2020-09-28
Convert AZFP from glider with separate ancillary data files


nlbeaird:
Hi,
Apologies if this is the wrong forum for this question, I haven't used github issues before. This is not really an issue, just a question.

I'm working with a student who is hoping to use echopype to process AZFP data collected by an autonomous underwater glider. We've tried to run the Convert / Process pipeline, but are striking out. I believe the issue is that the AZFP onboard the glider is not reporting ancillary measurements like temperature, pressure, pitch and roll to it's raw datafiles. These datastreams exist as part of the glider engineering data, however are not written to the '.01A' files. Thus when we run `.raw2nc()` we get the following error, I assume because the `unpacked_data` doesn't contain any temperature field (or because the temperature coefficients are all `0` in the XML file):

<img width="1040" alt="Screen Shot 2020-09-28 at 7 37 33 PM" src="https://user-images.githubusercontent.com/8039481/94496931-c2992900-01c3-11eb-905d-ced1f6d18b61.png">

I'm not very familiar with the AZFP, but your echopype effort seems wonderful! I'm wondering if you have a recommendation, or warning, about how hard it would be to try to use the temp, pressure, pitch, roll data collected separately by the vehicle in the echopype processing?  

Thanks much,
Nick

leewujung:
@nlbeaird : This is great! We don't actually have a specific forum for this type of discussion, so thanks for raising an issue!

Could you point us to an example file (+ the XML)? This type of diversity in raw data files are exactly what we haven't had for AZFP. A few people have reported similar problems for EK60 and EK80 data also and we added those into the package (to either no save those parameters at all or pad with NaN, depending on the circumstance).

A related note: @emiliom and I were just talking today about adding a plug-in type of thing for adding GPS data into AZFP files or EK files that do not already have them. There is definitely downstream data processing consequences, since to the minimum when compensating for the sound attenuation the environmental parameters come into play. It'll be cool if you're interested in brainstorming this together!

nlbeaird:
@leewujung : Thanks for responding so quickly! Glad to hear this is of interest. 

I've attached a small example raw file with the XML. I hope this is a helpful example. 

Yes! we (the student and I ) would be happy to help think about this in any way. 

> A few people have reported similar problems for EK60 and EK80 data also and we added those into the package (to either no save those parameters at all or pad with NaN, depending on the circumstance).

[...]

PR #1020, opened 2023-04-03
AZFP parser error catch to handle case of raw file with no temperature

emiliom:
Addresses remaining issue in #198, where an AZFP raw file without temperature data leads to a failure during `open_raw`. As diagnosed there, the failure comes about in `parser_azfp.compute_temp` due to invalid `R` values passed to `math.log`. A test has been added that will catch cases of the `R` denominator being <= 0 and return `np.nan` instead.

This has been tested with an AZFP file (and corresponding XML) provided by @dsmossman. 

Remaining tasks:
- [ ] Add AZFP test file to our test data
- [ ] Add a test

### DIFF
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -112,9 +112,14 @@ class ParseAZFP(ParseBase):
             and the counts from ancillary
             """
             v_in = 2.5 * (counts / 65535)
+            # Occurs when temperature sensor is not active (eg, glider deployments)
+            if self.parameters["kc"] - v_in <= 0:
+                return np.nan
+
             R = (self.parameters["ka"] + self.parameters["kb"] * v_in) / (
                 self.parameters["kc"] - v_in
             )
+
             # fmt: off
             T = 1 / (
                 self.parameters["A"]


echopype/tests/convert/test_convert_azfp.py
```diff
@@ -161,3 +161,17 @@ def test_convert_azfp_01a_different_ranges(azfp_path):
 
     # check convention-required variables in the Platform group
     check_platform_required_vars(echodata)
+
+
+def test_convert_azfp_01a_notemperature(azfp_path):
+    """Test converting file with no temperature data."""
+    azfp_01a_path = str(azfp_path.joinpath('rutgers_glider_notemperature', '22052500.01A'))
+    azfp_xml_path = str(azfp_path.joinpath('rutgers_glider_notemperature', '22052501.XML'))
```
leewujung, line 169: question: wouldn't the code work without the use of `str`?

echopype/convert/parse_azfp.py
```diff
@@ -112,9 +112,14 @@ def compute_temp(counts):
             and the counts from ancillary
             """
             v_in = 2.5 * (counts / 65535)
+            # Occurs when temperature sensor is not active (eg, glider deployments)
+            if self.parameters["kc"] - v_in <= 0:
+                return np.nan
+
```
leewujung, line 118: `self.parameters["kc"] - v_in <= 0` is fairly arbitrary since v_in is essentially trying to get some sensor voltage values.

Since in the case of the glider, all temperature-related parameters in `self.parameters` are 0, it is likely safer to test for these cases?

Or, alternatively, test for the case when `R<=0` so that we don't go into log(negative values).

leewujung (changes requested):
@emiliom : I have a suggestion re the statement you used to handle the temperature conversion error.

Also, while you are on it, I think we should change the tilt calculations too. All tilt-related parameters in `self.parameters` are also all 0 in the case of the glider, likely due to the same reason that the glider is recording the attitude separately just like the temperature. I suggest that in `compute_tilt` we check for the parameter values, and if they are all 0, return NaN (instead of 0 as it currently returns), so that users do not think that they are actually getting tilt=0 data.

echopype/convert/parse_azfp.py
emiliom: Are we certain that the relevant temperature parameters (`ka`, `kb`, `kc`) will always be 0 in circumstances like the Rutgers glider where the temperature sensor is turned off (or not deployed?). Do you know if it's documented that that will always be the case? Assuming that is true, can we reduce the 0 tests to just one of the parameters, to minimize the footprint of the test? 

BTW, `v_in` values are small (and apparently constant) but non-zero. See https://github.com/OSOceanAcoustics/echopype/issues/198#issuecomment-791560031

I'll comment on `compute_tilt` below your comment.

emiliom:
> Also, while you are on it, I think we should change the tilt calculations too. All tilt-related parameters in `self.parameters` are also all 0 in the case of the glider, likely due to the same reason that the glider is recording the attitude separately just like the temperature. I suggest that in `compute_tilt` we check for the parameter values, and if they are all 0, return NaN (instead of 0 as it currently returns), so that users do not think that they are actually getting tilt=0 data.

Makes sense. I'll take a closer look. I think it'll be more efficient, computationally, to do these tests outside the `compute_temp` and `compute_tilt` functions. I'll submit a commit with an implementation that you can review.

echopype/tests/convert/test_convert_azfp.py
emiliom: I started the new test code by copying and pasting one of the existing tests in that module. Yes, `str` is unnecessary, but its use is also present in that context in every test in the module. I'd rather deal with that in the context of a broader clean up.

But feel free to push a commit that cleans this up :wink:

leewujung:
As an overall comment: I am about to write AZFP an email so I'll include the questions about whether these values are always zero in it!

echopype/tests/convert/test_convert_azfp.py
leewujung: haha, I am going to push back and say: how about clean up just this test file? It's unclear to me when a broader clean up will happen, so might as well do small bits first.

emiliom:
New commits:
- Refactored the `compute_temp` test for invalid (= 0) temperature parameters.
- Added `compute_tilt` test for invalid (= 0) tilt parameters.

For both tests, I set the testing of actual parameter values outside the `compute_*` functions, so they happen only once. The functions are then passed a simple boolean.

Also, in the test module I replaced the use of `joinpath` with pathlib path concatenation.

echopype/tests/convert/test_convert_azfp.py
emiliom: Addressed in a new commit.

echopype/convert/parse_azfp.py
emiliom: Addressed in a new commit, for both `compute_temp` and `compute_tilt`.

leewujung:
Great, thanks @emiliom! I'll comment on this once we hear from ASL and if the 0s are indeed the default behavior, we can merge this.

Also, in looking at the code, I had the urge to move the `compute_*` functions to be private functions outside of `parse_raw` to make things more readable, and some location of the constants. Do you prefer for those to be in a separate PR?

emiliom:
Pushed new commit to add test checks for `tilt_x` and `tilt_y` similar to the one for `temperature`.

emiliom:
> Also, in looking at the code, I had the urge to move the `compute_*` functions to be private functions outside of `parse_raw` to make things more readable, and some location of the constants. Do you prefer for those to be in a separate PR?

You can submit a PR against my PR. I'll take a quick look before merging.

leewujung:
@emiliom : the movements end up to be a much broader job than I originally thought (it's too fun to refactor!), so I think I'll submit a separate PR for that...

leewujung:
I haven't heard from AZFP. I think we can merge this tomorrow if I don't hear back by then, so that this and #1024 can both get merged. We have a few small fixes in v0.7.1 and it'll be good to release by end of this week so that those changes get to actual package.

emiliom:
Per email from ASL Environmental to @leewujung, it's been confirmed that the temperature (and tilt?) parameters in the XML file are set to zero when the sensor is deactivated.

So, we can merge this PR as is, right? Then, PR #1024 can be merged, after updating it to reflect this PR.

leewujung:
Yes, the values for temperature and tilt sensor coefficients will be 0 in glider-specific AZFP deployments.

Feel free to merge this, and I'll update #1024 and merge that too!

PR #1024, opened 2023-04-05
Refactor AZFP parser code

leewujung:
This PR refactors `parse_raw`:
- move `compute_*` subfunctions to separate private functions under `ParseAZFP` as `._compute_*`
- refactor input arguments for `ParseAZFP._compute_*` to keep `parse_raw` clean
- move XML and header constants and field specification to be class attributes for `ParseAZFP`
- change use of `math` to just use `np`

@emiliom : I'll merge `dev` once #1020 is merged so the history here looks nice, but it should be pretty easy for you to see what I have done here.

### DIFF
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -190,7 +190,8 @@ class ParseAZFP(ParseBase):
         if not is_valid:
             return np.nan
         else:
-            N = self.unpacked_data["ancillary"][ping_num][0]
+            idx = 0 if xy == "X" else 1
+            N = self.unpacked_data["ancillary"][ping_num][idx]
             a = self.parameters[f"{xy}_a"]
             b = self.parameters[f"{xy}_b"]
             c = self.parameters[f"{xy}_c"]
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -258,7 +258,7 @@ class ParseAZFP(ParseBase):
                         )
                         # compute x tilt from unpacked_data[ii]['ancillary][0]
                         self.unpacked_data["tilt_x"].append(
-                            self._compute_tilt(ping_num, "Y", tilt_x_is_valid)
+                            self._compute_tilt(ping_num, "X", tilt_x_is_valid)
                         )
                         # Compute y tilt from unpacked_data[ii]['ancillary][1]
                         self.unpacked_data["tilt_y"].append(
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -1,4 +1,3 @@
-import math
 import os
 import xml.dom.minidom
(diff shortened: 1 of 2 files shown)


echopype/convert/parse_azfp.py
```diff
@@ -40,112 +131,119 @@ def get_value_by_tag_name(tag_name, element=0):
...
+            ping number
+        xy
+            either "X" or "Y"
+        is_valid
+            whether the associated parameters have valid values
+        """
+        if not is_valid:
+            return np.nan
+        else:
+            N = self.unpacked_data["ancillary"][ping_num][0]
```
emiliom, line 193: Modifying the list index to be 0 or 1, per the original code.
```suggestion
            idx = 0 if xy == "X" else 1
            N = self.unpacked_data["ancillary"][ping_num][idx]
```

echopype/convert/parse_azfp.py
```diff
@@ -156,48 +254,36 @@ def compute_battery(N):
...
                         # compute x tilt from unpacked_data[ii]['ancillary][0]
                         self.unpacked_data["tilt_x"].append(
-                            compute_tilt(
-                                self.unpacked_data["ancillary"][ping_num][0],
-                                self.parameters["X_a"],
-                                self.parameters["X_b"],
-                                self.parameters["X_c"],
-                                self.parameters["X_d"],
-                            )
+                            self._compute_tilt(ping_num, "Y", tilt_x_is_valid)
```
emiliom, line 261:
```suggestion
                            self._compute_tilt(ping_num, "X", tilt_x_is_valid)
```

emiliom:
> change use of `math` to just use `np`

Any particular reason for this? It seems fine either way.

emiliom (changes requested):
I like what you've done and don't see any down sides.

But see my bug fixes.

echopype/convert/parse_azfp.py
leewujung: Ah you're right, thanks for catching this!

echopype/convert/parse_azfp.py
leewujung: many bugs... I have done this too quickly!! and too bad that tests do not catch it, because the tests only checks the backscatter data.

leewujung:
Thanks @emiliom for catching those pretty stupid bugs! 

For the change from `math` to `np`, it just seems simpler if we use numpy for all the common functions, and especially in this case I don't think there was a real reason behind the original choice of `math`.

echopype/convert/parse_azfp.py
leewujung: I thought I added it to the batch but apparently not. I'll commit it now.

emiliom:
I've merged PR #1020.

leewujung:
I've merged changes from #1020 and will merge this PR now.

PR #1029, opened 2023-04-18
Update windows github actions for testing

lsetiawan:
## Overview

This PR updates the windows github workflow to be able to run the full testing suite in windows.


lsetiawan:
After some trial/error. Looks like using docker on windows just doesn't work! Let's find another solution. Closing this.

PR #1032, opened 2023-04-18
Update windows testing

lsetiawan:
## Overview

This PR works on moving testing away from docker usage for compatibility across platforms, starting with windows.

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -6,10 +6,12 @@ on:
       - dev
       - main
     paths-ignore: [".ci_helpers/docker/**", "**/docker.yaml"]
+  pull_request:
   workflow_dispatch:
 
 env:
   CONDA_ENV: echopype
+  NUM_WORKERS: 2
 
 jobs:
   windows-test:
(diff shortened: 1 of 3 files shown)


lsetiawan:
## Progress summary

This PR is still in progress as of 4/19/2023. Will be put on hold for now due to other pressing issues. I thought I was really close, but this will need some more thinking and experimentation in regards on how the test data should be distributed.

## Ideas

1. Replace docker build workflow with package test data workflow where the data is downloaded to google drive, zipped, and hosted somewhere
2. The CI workflows would then download the zip file, extract, and use that as usual
3. Update the tests to utilize built in python http server and moto package so that it can run cross platforms

## TODOs

There are still a few todos based on the ideas above. 

- After playing around creating the test data package workflow, one thing occured to me as how I can get the google drive file id and pass that into each individual CI test run so it can get the correct file.
- All of the test reference to `minio_bucket` need to change, as well as the http url should not have `/data`.

A partial solution that can be implemented quickly ASAP maybe a hybrid one, where the windows CI testing only run during push to `dev` or `main` and it can directly access the google drive avoiding the need to package the data and such. Then the other CI testing with linux can be done by still using docker as a way of getting the test data rather than the whole file hosting.

Issue #976, opened 2023-03-11
Overhaul `combine_echodata` code


leewujung:
The `combine_echodata` routine is working but buggy in the cases when the combination is across a previously combined `EchoData` object and one ore more `EchoData` objects generate from `open_raw`. There are also questions about whether/how the zarr files associated with the `EchoData` objects to be combined are chunked, since `auto_chunks` is only used when `echodata.to_zarr` is explicitly called (from #939), which is not the case for the current `combine_echodata` code.

This issue tracks all the associated issues for reviewing, debugging, and refactoring `combine_echodata` code.
- [x] #975 
- [x] #961 
- [x] #827 
- [x] #882 
- ~#798 -- I seem to remember this issue being no longer relevant, but am not sure. Will need investigation.~ This is something that still needs to be dealt with, but it does not belong to the `combine_echodata` context.

PR #1042, opened 2023-05-15
Overhaul `combine_echodata` method

lsetiawan:
## Overview

This PR aims to overhaul the `combine_echodata` functionality to be in favor of using xarray's functions to concat the datasets. This relates to issue https://github.com/OSOceanAcoustics/echopype/issues/976. The changes that happens in this PR are as follows:

- Remove the usage of ZarrCombine object
- Remove spinning up dask client under the hood during combine
- Remove the need for zarr path as `combine_echodata` input
- Utilizes `xr.concat` to combine the datasets under the hood so it doesn't require any monotonic values, keeping the data lossless
- Updates `group_paths` attribute to be a `tuple` rather than `set` to keep order of paths.

**NOTE: Currently this works for EK60, but there are still some issues with attributes combining within the Vendor_specific group for EK80 since some of the attribute values are arrays.**

In another PR, I will address the issue as stated above.

### DIFF
--- a/echopype/echodata/combine.py
+++ b/echopype/echodata/combine.py
@@ -1,7 +1,6 @@
 import datetime
 import itertools
 import re
-from collections import defaultdict
 from functools import reduce
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -497,114 +496,6 @@ def _check_ascending_ds_times(ds_list: List[xr.Dataset], ed_group: str) -> None:
                 raise RuntimeError(
                     f"The coordinate {time} is not in ascending order for "
                     f"group {ed_group}, combine cannot be used!"
-                )
-
-
-def _get_ds_info(ds_list: List[xr.Dataset], ed_group: str) -> Dict[str, list]:
-    """
-    Constructs useful dictionaries that contain information
-    about the dimensions of the Dataset. Additionally, collects
-    the attributes from each Dataset in ``ds_list`` and saves
-    this group specific information to the class variable
-    ``group_attrs``.
-
-    Parameters
-    ----------
-    ds_list: list of xr.Dataset
-        The Datasets that will be combined
-    ed_group: str
-        The name of the EchoData group corresponding to the
-        Datasets in ``ds_list``
-
-    Notes
-    -----
-    If attribute values are numpy arrays, then they will not be included
-    in the ``self.group_attrs``. Instead, these values will only appear
-    in the attributes of the combined ``EchoData`` object.
-    """
-
-    group_attrs = defaultdict(list)
-
-    # format ed_name appropriately
-    ed_group = ed_group.replace("-", "_").replace("/", "_").lower()
-
-    if ed_group == "top_level":
-        # Let's skip the attribute comparison for top level
-        numpy_keys = []
-    else:
-        if len(ds_list) == 1:
-            # get numpy keys if we only have one Dataset
-            numpy_keys = _compare_attrs(ds_list[0].attrs, ds_list[0].attrs)
-        else:
-            # compare attributes and get numpy keys, if they exist
-            for ind in range(len(ds_list) - 1):
-                numpy_keys = _compare_attrs(ds_list[ind].attrs, ds_list[ind + 1].attrs)
-
-    # collect Dataset attributes
-    for count, ds in enumerate(ds_list):
-        # get reduced attributes that do not include numpy keys
-        red_attrs = {key: val for key, val in ds.attrs.items() if key not in numpy_keys}
-
-        if count == 0:
-            group_attrs[ed_group + "_attr_key"].extend(red_attrs.keys())
-
-        group_attrs[ed_group + "_attrs"].append(list(red_attrs.values()))
-    return group_attrs
-
-
-def _compare_attrs(attr1: dict, attr2: dict) -> List[str]:
-    """
-    Compares two attribute dictionaries to ensure that they
-    are acceptably identical.
-
-    Parameters
-    ----------
-    attr1: dict
-        Attributes from Dataset 1
-    attr2: dict
-        Attributes from Dataset 2
-
-    Returns
-    -------
-    numpy_keys: list
-        All keys that have numpy arrays as values
-
-    Raises
-    ------
-    RuntimeError
-        - If the keys are not the same
-        - If the values are not identical
-        - If the keys ``date_created`` or ``conversion_time``
-        do not have the same types
-
-    Notes
-    -----
-    For the keys ``date_created``, ``conversion_time`` the values
-    are not required to be identical, rather their type must be identical.
-    """
-
-    # make sure all keys are identical (this should never be triggered)
-    if attr1.keys() != attr2.keys():
-        raise RuntimeError(
-            "The attribute keys amongst the ds lists are not the same, combine cannot be used!"
-        )
-
-    # make sure that all values are identical
-    numpy_keys = []
-    for key in attr1.keys():
-        if isinstance(attr1[key], np.ndarray):
-            numpy_keys.append(key)
-
-            if not np.allclose(attr1[key], attr2[key], rtol=1e-12, atol=1e-12, equal_nan=True):
-                raise RuntimeError(
-                    f"The attribute {key}'s value amongst the ds lists are not the "
-                    f"same, combine cannot be used!"
-                )
-        elif key in ["date_created", "conversion_time"]:
-            if not isinstance(attr1[key], type(attr2[key])):
-                raise RuntimeError(
-                    f"The attribute {key}'s type amongst the ds lists "
-                    f"are not the same, combine cannot be used!"
                 )
 
 
--- a/echopype/echodata/combine.py
+++ b/echopype/echodata/combine.py
@@ -572,8 +572,13 @@ def _merge_attributes(attributes: List[Dict[str, str]]) -> Dict[str, str]:
     for attribute in attributes:
         for key, value in attribute.items():
             if value == "" and key not in merged_dict:
+                # checks if current attr value is empty,
+                # and doesn't exist in merged attribute
                 merged_dict[key] = value
             elif value != "":
+                # if current attr value is not empty,
+                # then overwrite the merged attribute,
+                # keeping attribute from latest value
                 merged_dict[key] = value
     return merged_dict
 
--- a/echopype/echodata/combine.py
+++ b/echopype/echodata/combine.py
@@ -621,10 +621,7 @@ def _capture_prov_attrs(
     )
     prov_ds = df_merged.to_xarray()
     # Set these provenance as string
-    if sonar_model.lower() != "ek80":
-        # Skip for EK80
-        # TODO: This is right now problematic with array attribute for filter coeff
-        prov_ds = prov_ds.fillna("").astype(str)
+    prov_ds = prov_ds.fillna("").astype(str)
     prov_ds[ED_GROUP] = prov_ds[ED_GROUP].astype(str)
     prov_ds[ED_FILENAME] = prov_ds[ED_FILENAME].astype(str)
     return prov_ds
--- a/echopype/echodata/combine.py
+++ b/echopype/echodata/combine.py
@@ -746,8 +746,6 @@ def combine_echodata(
 ) -> EchoData:
     """
     Combines multiple ``EchoData`` objects into a single ``EchoData`` object.
-    This is accomplished by writing each element of ``echodata_list`` in parallel
-    (using Dask) to the zarr store specified by ``zarr_path``.
 
     Parameters
     ----------
--- a/echopype/echodata/combine.py
+++ b/echopype/echodata/combine.py
@@ -153,22 +153,22 @@ def _check_channel_selection_form(
             raise TypeError("Each value of channel_selection must be a list of strings!")
 
 
-def check_echodatas_input(echodatas: List[EchoData]) -> Tuple[str, List[str]]:
+def check_eds(echodata_list: List[EchoData]) -> Tuple[str, List[str]]:
     """
     Ensures that the input list of ``EchoData`` objects for ``combine_echodata``
     is in the correct form and all necessary items exist.
 
     Parameters
     ----------
-    echodatas: list of EchoData object
+    echodata_list: list of EchoData object
         The list of `EchoData` objects to be combined.
 
     Returns
     -------
     sonar_model : str
-        The sonar model used for all values in ``echodatas``
+        The sonar model used for all values in ``echodata_list``
     echodata_filenames : list of str
-        The source files names for all values in ``echodatas``
+        The source files names for all values in ``echodata_list``
 
     Raises
     ------
@@ -183,17 +183,19 @@ def check_echodatas_input(echodatas: List[EchoData]) -> Tuple[str, List[str]]:
     """
 
     # make sure that the input is a list of EchoData objects
-    if not isinstance(echodatas, list) and all([isinstance(ed, EchoData) for ed in echodatas]):
+    if not isinstance(echodata_list, list) and all(
+        [isinstance(ed, EchoData) for ed in echodata_list]
+    ):
         raise TypeError("The input, eds, must be a list of EchoData objects!")
 
     # get the sonar model for the combined object
-    if echodatas[0].sonar_model is None:
+    if echodata_list[0].sonar_model is None:
         raise ValueError("all EchoData objects must have non-None sonar_model values")
     else:
-        sonar_model = echodatas[0].sonar_model
+        sonar_model = echodata_list[0].sonar_model
 
     echodata_filenames = []
-    for ed in echodatas:
+    for ed in echodata_list:
         # check sonar model
         if ed.sonar_model is None:
             raise ValueError("all EchoData objects must have non-None sonar_model values")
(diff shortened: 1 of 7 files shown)


lsetiawan:
## TODO

Benchmarking is needed before merging this PR into `dev`!

echopype/echodata/combine.py
```diff
@@ -435,14 +447,357 @@ def _check_echodata_channels(
...
+                    f"The attribute {key}'s value amongst the ds lists are not the "
+                    f"same, combine cannot be used!"
+                )
+        elif key in ["date_created", "conversion_time"]:
+            if not isinstance(attr1[key], type(attr2[key])):
+                raise RuntimeError(
+                    f"The attribute {key}'s type amongst the ds lists "
+                    f"are not the same, combine cannot be used!"
+                )
+
```
aniketfadia96, line 609: I think we need to add `return numpy_keys`

echopype/echodata/combine.py
```diff
@@ -435,14 +447,357 @@ def _check_echodata_channels(
...
+        The provenance dataset for all the attribute values
+
+    """
+    index_keys = [ED_GROUP, ED_FILENAME]
+    df_list = []
+    for group, attributes in attrs_dict.items():
+        df = pd.DataFrame.from_records(attributes)
+        df.loc[:, ED_FILENAME] = echodata_filenames
+        df[ED_GROUP] = group
+        df.loc[:, ED_FILENAME] = echodata_filenames
```
lsetiawan, line 665:
```suggestion
```

echopype/echodata/combine.py
```diff
@@ -435,14 +446,248 @@ def _check_echodata_channels(
...
+    echodata_filenames: List[str] = [],
+    ed_group_chan_sel: Dict[str, Optional[List[str]]] = {},
+) -> Dict[str, xr.Dataset]:
+    """
+    Combines the echodata objects and export to a dictionary tree.
+
+    Parameters
+    ----------
+    sonar_model : str
+        The sonar model used for all elements in ``eds`
```
aniketfadia96, line 585: NIT: ``eds``

aniketfadia96, line 585:
```suggestion
        The sonar model used for all elements in ``eds``
```

Code suggestion for the above NIT.

aniketfadia96 (approved):
LGTM

lsetiawan:
@leewujung This is currently failing because the `Vendor specific` identical check fails for `AZFP`. Should AZFP have same filter params for all files to be valid for merging?

emiliom:
> @leewujung This is currently failing because the Vendor specific identical check fails for AZFP. Should AZFP have same filter params for all files to be valid for merging?

I'll chime in, in case it helps. I believe the content of the Vendor group will be quite different for AZFP.

lsetiawan:
## Dask Reports

Below are links for the resulting benchmarking with dask. I ran this test with 131GB of OOI Data locally on my MacBook Pro (Apple M1 Pro, 32GB RAM).

combine_echodata: https://gistcdn.githack.com/lsetiawan/93ac199a107c76739c22851af4233ace/raw/953238380197a772e2e0d44e9cc8594bcd3ac90f/dask-report-combine.html

to_zarr on combined data: https://gistcdn.githack.com/lsetiawan/fc00688b26310c17b1cd7b31d02f34c2/raw/5abe6dfe2c7c5272637e408e771b1a23cd6644de/dask-report-combine-to-zarr.html

## Summary

Jupyter notebook: https://nbviewer.org/gist/lsetiawan/3ad22ddb41e12fdad2e4e368ae3dab0f

The actual combining of lazy loaded converted zarr files took about 7 minutes. This consist of the actual concatenations of dimensions and lazy variables. After this step, we get a lazy loaded echodata object that one can explore. The actual writing of the file into a combined zarr file took about 1 minute, so in total, the new operation took about 8-9 minutes.

Looking at previous benchmarking [here](https://nbviewer.org/gist/lsetiawan/ebb3faed65e53a3188518d62dbe0968a). The old way would take about 15 minutes to combine data, with already written path. However this way, there was no way to possible setup lazy computation before actually writing the file. It also created a copy of all the data.

echopype/tests/echodata/test_echodata_combine.py
```diff
@@ -211,13 +212,76 @@ def test_combine_echodata(raw_datasets):
...
+
+@pytest.mark.parametrize("test_param", [
+        "single",
+        "multi"
+    ]
+)
+def test_combine_echodata_combined_and_others(ek60_multi_test_data, test_param, sonar_model="EK60"):
+        """
+        Integration test for combine_echodata with
+        a single combined ed and a single echodata
```
leewujung, line 229:
```suggestion
        Integration test for combine_echodata with the following cases:
        - a single combined echodata object and a single echodata object
        - a single combined echodata object and 2 single echodata objects
```
I also wonder if there should be a test for:
- a single combined echodata object and another combined echodata object

something like:
```
combined_ed = echopype.combine_echodata(eds[:2])
combined_ed_other = echopype.combine_echodata(eds[2:])
combined_ed2 = echopype.combine_echodata([combined_ed, combined_ed_other])
```

echopype/tests/echodata/test_echodata_combine.py
```diff
@@ -211,13 +212,76 @@ def test_combine_echodata(raw_datasets):
...
+            if (combined_group is not None) and (test_ds is not None):
+                assert test_ds.identical(combined_group.drop_dims(grp_drop_dims))
+
+
+@pytest.mark.parametrize("test_param", [
+        "single",
+        "multi"
+    ]
+)
+def test_combine_echodata_combined_and_others(ek60_multi_test_data, test_param, sonar_model="EK60"):
```
leewujung, line 226: Thinking maybe change the test function name to make it more obvious that this is testing for the "appending" cases:
```suggestion
def test_combine_echodata_combined_append(ek60_multi_test_data, test_param, sonar_model="EK60"):
```

echopype/utils/coding.py
```diff
@@ -142,16 +144,19 @@ def set_zarr_encodings(ds: xr.Dataset, compression_settings: dict) -> dict:
...
-        # currently defaults to 100MB
-        if not ds.chunks and len(val.shape) > 0:
+        # Always optimize chunk if not specified already
+        # user can specify desired chunk in encoding
+        existing_chunks = encoding[name].get("chunks", None)
+        if len(val.shape) > 0 and existing_chunks is not None:
             chunks = _get_auto_chunk(val)
-            val_encoding.update({"chunks": chunks})
-        encoding[name] = val_encoding
+            encoding[name]["chunks"] = chunks
```
leewujung, line 155: Should the second of the if statement be `existing_chunks is None` ?
The way it is if `existing_chunks` is given, chunks from `_get_auto_chunk` will be used, which seems the opposite of what should be happening?

Should we have a simple test for this function?

echopype/utils/coding.py
```diff
@@ -142,16 +144,19 @@ def set_zarr_encodings(ds: xr.Dataset, compression_settings: dict) -> dict:
...
+
+        if PREFERRED_CHUNKS in encoding[name]:
+            # Remove 'preferred_chunks', use chunks only instead
+            encoding[name].pop(PREFERRED_CHUNKS)
```
leewujung, line 159: Is this block due to the current `zarr_combine` implementation to be dropped later? I couldn't figure out what this is but a full codebase search showed `preferred_chunks` under `zarr_combine.py`

echopype/utils/io.py
```diff
@@ -60,6 +60,9 @@ def save_file(ds, path, mode, engine, group=None, compression_settings=None, **k
     if engine == "netcdf4":
         ds.to_netcdf(path=path, mode=mode, group=group, encoding=encoding, **kwargs)
     elif engine == "zarr":
+        # Ensure that encoding and chunks match
+        for var, enc in encoding.items():
+            ds[var] = ds[var].chunk(enc.get("chunks", {}))
```
leewujung, line 65: For my curiosity: is this for when `chunks` has not been specified for a given `var`?

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+    """
+    Check for identical filter params for all inputs in Vendor specific group
+
+    Parameters
+    ----------
+    ds_list: list of xr.Dataset
+        List of Datasets to be combined
+    ed_group: "Vendor_specific"
+        The name of the ``EchoData`` group being combined,
+        this is only works for "Vendor_specific" group.
```
leewujung, line 514: minor correction
```suggestion
        this only works for "Vendor_specific" group.
```

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+        if not is_identical:
+            raise RuntimeError(
+                f"Non identical filter parameters in {ed_group} group. " "Objects cannot be merged!"
+            )
+        dataset = next_dataset
+
+
+def _merge_attributes(attributes: List[Dict[str, str]]) -> Dict[str, str]:
+    """
+    Merges a list of attributes dictionary
```
leewujung, line 552: super minor: just to keep consistency with other docstring
```suggestion
    Merge a list of attributes dictionary
```

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+    dict
+        The merged attribute dictionary
+    """
+    merged_dict = {}
+    for attribute in attributes:
+        for key, value in attribute.items():
+            if value == "" and key not in merged_dict:
+                merged_dict[key] = value
+            elif value != "":
+                merged_dict[key] = value
```
leewujung, line 571: This logic looks like if `key` already exists in `merged_dict` from a previous `attribute` in the loop, it will be overwritten with the latest `value`, but I think we want to do the reverse -- ie to keep only the first `value` entry?

Probably would be a good idea to add some comments to the if-else block to make it obvious what the expected handling is for the new key-value pair

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+        E.g. {'Group': [{'attr1': 'val1'}, {'attr2': 'val2'}, ...]}
+    echodata_filenames : list of str
+        The filenames of the echodata objects
+    sonar_model : str
+        The sonar model
+
+    Returns
+    -------
+    xr.Dataset
+        The provenance dataset for all the attribute values
```
leewujung, line 595: small clarification in docstring:
```suggestion
        The provenance dataset for all attribute values from the list of echodata objects that are combined.
```

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+            else:
+                combined_ds = xr.Dataset()
+                for dim in ds_append_dims:
+                    drop_dims = [c_dim for c_dim in ds_append_dims if c_dim != dim]
+                    sub_ds = xr.concat(
+                        [ds.drop_dims(drop_dims) for ds in ds_list],
+                        dim=dim,
+                        coords="minimal",
+                        data_vars="minimal",
+                    )
```
leewujung, line 698: it's awesome how clean this is with `minimal`!

echopype/echodata/combine.py
```diff
@@ -474,24 +756,17 @@ def combine_echodata(
...
         will have their selected channels chosen automatically.
-    consolidated: bool
-        Flag to consolidate zarr metadata.
-        Defaults to ``True``
 
     Returns
     -------
     EchoData
-        A lazy loaded ``EchoData`` object obtained from ``zarr_path``,
+        A lazy loaded ``EchoData`` object,
```
leewujung, line 763: Not about this line, but I couldn't highlight it above: could you remove the sentence about `zarr_path` in the first section of the docstring? Thanks!

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+    prov_ds = tree_dict["Provenance"].assign(prov_ds)
+    # Update filenames to iter integers
+    prov_ds[FILENAMES] = prov_ds[FILENAMES].copy(data=np.arange(*prov_ds[FILENAMES].shape))
+    tree_dict["Provenance"] = prov_ds
+
+    return tree_dict
+
+
 def combine_echodata(
     echodatas: List[EchoData] = None,
```
leewujung, line 737: How about changing `echodatas` to `echodata_list`? I couldn't stand adding s to data, and it is better to be explicit about it needs a list.
```suggestion
    echodata_list: List[EchoData] = None,
```

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+        df_list.append(df)
+
+    df_merged = reduce(
+        lambda left, right: pd.merge(left, right, on=index_keys, how="outer"), df_list
+    )
+    prov_ds = df_merged.to_xarray()
+    # Set these provenance as string
+    if sonar_model.lower() != "ek80":
+        # Skip for EK80
+        # TODO: This is right now problematic with array attribute for filter coeff
```
leewujung, line 614: Now that the filter coeffs are data variables in the Vendor-specific group, this would pass?

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+            is_descending = (np.diff(first_times) < np.timedelta64(0, "ns")).any()
+
+            if is_descending:
+                raise RuntimeError(
+                    f"The coordinate {time} is not in ascending order for "
+                    f"group {ed_group}, combine cannot be used!"
+                )
+
+
+def _check_filter_params(
```
leewujung, line 502: Maybe change this to `_check_no_append_vendor_params` or something similar since the concept of filter does not exist for sonar model other than EK80, but there are still vendor-specific params that are not varying across the appending dimensions (e.g., `ping_time` in AZFP)

echopype/echodata/combine.py
```diff
@@ -435,14 +446,296 @@ def _check_echodata_channels(
...
+        The provenance dataset for all the attribute values
+
+    """
+    index_keys = [ED_GROUP, ED_FILENAME]
+    df_list = []
+    for group, attributes in attrs_dict.items():
+        df = pd.DataFrame.from_records(attributes)
+        df.loc[:, ED_FILENAME] = echodata_filenames
+        df[ED_GROUP] = group
+        df = df.set_index(index_keys)
```
leewujung, line 604: By setting index for both `ED_FILENAME` and `ED_GROUP` the resulting xarray data variables have both of these as dimensions, like below:
Screenshot 2023-05-30 at 10 44 25 AM
Screenshot 2023-05-30 at 10 44 35 AM

I found this to be a little confusing? It seems that we can omit the `ED_GROUP` dimension and still append all the attributes?

I assume this is for preserving the "origin" of each attribute. If that is the case, is there another mechanism to store this info, such as in the attributes of the constructed data variables?

leewujung (commented):
Hey @lsetiawan : Thanks for this great PR! I've gone through the code and also went through a few test cases. I think there might be a potential bug in `_merge_attributes` and `set_zarr_encodings `, and something to discuss in `_capture_prov_attrs` regarding whether `ED_GROUP` should be a dimension there. All my other comments are minor. It'll be awesome once we get this merged!!

leewujung (commented):
Hey @lsetiawan : Thanks for this great PR! I've gone through the code and also went through a few test cases. I think there might be a potential bug in `_merge_attributes` and `set_zarr_encodings `, and something to discuss in `_capture_prov_attrs` regarding whether `ED_GROUP` should be a dimension there. All my other comments are minor. It'll be awesome once we get this merged!!

echopype/echodata/combine.py
lsetiawan: I didn't even think about storing the group name in each of the data variables! I think this is a really great idea. I'll try that. 😄 Thank you!!!

echopype/echodata/combine.py
lsetiawan: I think it would be better to keep the latest because if a user wants to append a new data, rather than messing with the already combined data, it might be better to update the attribute in the new data, which will overwrite the current attributes in the first data to be appended. This was the reasoning behind that. If I remember correctly, I had a discussion about this with @emiliom and he thought that was good. Could you chime in @emiliom  and let me know what you think? Thanks 😄

leewujung: I think one of the attribute is the first ping time, so using the latest one doesn't make sense...

echopype/utils/coding.py
lsetiawan: Seems like this encoding comes from xarray directly, I was thinking this was problematic, but then I just tried it and seem like it doesn't affect anything, so I'll remove this check. See: https://github.com/pydata/xarray/blob/960f15ce35826b85745adf5a7b5d8f7decbfa16d/xarray/backends/zarr.py#L493

echopype/echodata/combine.py
lsetiawan: Which one is that for?

leewujung: i think it's in the AZFP vendor group -- in a sense it is not needed, since it is already in the dataset. 

In a sense using neither the first nor the last makes sense. Let's discuss in the slack channel...

echopype/utils/io.py
lsetiawan: This is for ensuring that both dask chunks and encoding chunks match otherwise, you'd get an error around here: https://github.com/pydata/xarray/blob/960f15ce35826b85745adf5a7b5d8f7decbfa16d/xarray/backends/zarr.py#L176

echopype/utils/coding.py
lsetiawan: Just updated this to be a little smarter, where it checks the existing chunk as to whether it's optimal or not! If not optimal then run the auto chunk. I wanna make sure it's always optimal especially as data get's larger. At some point we'll think about letting user choose the chunk size, but right now, we're "optimizing" the chunk size for simplicity sake.

echopype/echodata/combine.py
emiliom: In the Slack channel you asked that I comment here. I don't think I follow the code logic, since both parts of the if statement result in `merged_dict[key] = value`. So, I'll comment on the discussion. I'm not sure what @leewujung means by "I think one of the attributes is the first ping time, so using the latest one doesn't make sense...", but in my discussion with @lsetiawan we talked about storing either the first or last values. Based on that discussion alone, there weren't clear arguments for picking one over the other (the first or the last value). So, if you have good arguments for one, that's fine with me.

lsetiawan:
Thank you both @emiliom and @leewujung for your helpful comments. I've integrated the changes needed from the comments. Please give it another look and let me know what you think! Thanks again 😄

leewujung (approved):
Hey @lsetiawan : Thanks for putting in the new tests and the revised chunking schemes. I like the flexibility it provides, and we can see about allowing users rechunking options later since that would involve an additional input argument -- though I think we can probably leave the rechunking to the Sv stage since it probably makes sense to have that along with the regridding to bring everything onto the same timebase.

I just have a question about the pre-optimized chunk size of the combined ds -- how is that related to the chunk size of the individual ds (those that get combined), but I'll ask that on Slack.

PR #1044, opened 2023-05-18
Move filter coefficients and decimation factor to variables

lsetiawan:
## Overview

This PR moves the filter coefficient and decimation factors within the `Vendor specific` group from being in the `Attributes` to data variables for `EK80` sonar model as seen in the screenshot below.

Screenshot 2023-05-18 at 4 44 45 PM

### DIFF
--- a/echopype/convert/set_groups_ek80.py
+++ b/echopype/convert/set_groups_ek80.py
@@ -1317,29 +1317,30 @@ class SetGroupsEK80(SetGroupsBase):
         if "impedance" in ds_cal:
             ds_cal = ds_cal.rename_vars({"impedance": "impedance_transducer"})
 
-        #  Save decimation factors and filter coefficients
+        # Save decimation factors and filter coefficients
+        # Param map values
+        # 1: wide band transceiver (WBT)
+        # 2: pulse compression (PC)
+        param_map = {1: WIDE_BAND_TRANS, 2: PULSE_COMPRESS}
         coeffs_and_decimation = {
-            t: {FILTER_IMAG: [], FILTER_REAL: [], DECIMATION: []}
-            for t in [WIDE_BAND_TRANS, PULSE_COMPRESS]
+            t: {FILTER_IMAG: [], FILTER_REAL: [], DECIMATION: []} for t in list(param_map.values())
         }
 
         for ch in self.sorted_channel["all"]:
-            # filter coeffs and decimation factor for wide band transceiver (WBT)
-            if self.parser_obj.fil_coeffs and (ch in self.parser_obj.fil_coeffs.keys()):
-                coeffs_vals = self.parser_obj.fil_coeffs[ch][1]
-                coeffs_and_decimation[WIDE_BAND_TRANS][FILTER_IMAG].append(np.imag(coeffs_vals))
-                coeffs_and_decimation[WIDE_BAND_TRANS][FILTER_REAL].append(np.real(coeffs_vals))
-                coeffs_and_decimation[WIDE_BAND_TRANS][DECIMATION].append(
-                    self.parser_obj.fil_df[ch][1]
-                )
-            # filter coeffs and decimation factor for pulse compression (PC)
-            if self.parser_obj.fil_df and (ch in self.parser_obj.fil_coeffs.keys()):
-                coeffs_vals = self.parser_obj.fil_coeffs[ch][2]
-                coeffs_and_decimation[PULSE_COMPRESS][FILTER_IMAG].append(np.imag(coeffs_vals))
-                coeffs_and_decimation[PULSE_COMPRESS][FILTER_REAL].append(np.real(coeffs_vals))
-                coeffs_and_decimation[PULSE_COMPRESS][DECIMATION].append(
-                    self.parser_obj.fil_df[ch][2]
-                )
+            fil_coeffs = self.parser_obj.fil_coeffs.get(ch, None)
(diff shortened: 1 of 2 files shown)


echopype/convert/set_groups_ek80.py
```diff
@@ -1311,31 +1318,88 @@ def set_vendor(self) -> xr.Dataset:
...
+        # Add the coeffs and decimation
+        ds = ds.pipe(self._add_cd, coeffs_and_decimation)
 
         # Save the entire config XML in vendor group in case of info loss
         ds.attrs["config_xml"] = self.parser_obj.config_datagram["xml"]
 
         return ds
+
+    @staticmethod
+    def _add_cd(
```
leewujung, line 1356:
```suggestion
    def _add_filter_params(
```
Trying to make it more readable for our future selves...

echopype/convert/set_groups_ek80.py
```diff
@@ -1311,31 +1318,88 @@ def set_vendor(self) -> xr.Dataset:
...
+                        ]
+                    )
+                    dims = ["channel", f"{cd_type}_filter_n"]
+                else:
+                    attrs = {
+                        "long_name": f"{attribute_values[cd_type]} {attribute_values[DECIMATION]}"
+                    }
+                    dims = ["channel"]
+                # Set the xarray data dictionary
+                coeffs_xr_data[f"{cd_type} {key}"] = (dims, data, attrs)
```
leewujung, line 1403: From the print out you showed, there's a space instead of an underscore between PC or WBT and filter_*. Seems better to replace it?
```suggestion
                coeffs_xr_data[f"{cd_type}_{key}"] = (dims, data, attrs)
```

leewujung (commented):
Hey @lsetiawan : the changes look great and I only made small suggestions. I certainly like the cleanness of the outcome! 

The test failure is on `test_convert_ek80_no_fil_coeff` so should just take a check for existence. What do you think is better between the below options?
- save these data variables as NaN arrays with a channel dimension
- do not save these data variables

aniketfadia96:
LGTM

lsetiawan:
> What do you think is better between the below options?

I think for simplicity and cleanliness, if data variable doesn't exist then they shouldn't be there. When combining, it xarray should be smart enough to fill that with NaN's I think.

echopype/convert/set_groups_ek80.py
```diff
@@ -1311,31 +1318,91 @@ def set_vendor(self) -> xr.Dataset:
...
-        coeffs = dict()
-        decimation_factors = dict()
+        coeffs_and_decimation = {
+            t: {FILTER_IMAG: [], FILTER_REAL: [], DECIMATION: []}
+            for t in [WIDE_BAND_TRANS, PULSE_COMPRESS]
+        }
+
         for ch in self.sorted_channel["all"]:
             # filter coeffs and decimation factor for wide band transceiver (WBT)
             if self.parser_obj.fil_coeffs and (ch in self.parser_obj.fil_coeffs.keys()):
```
lsetiawan, line 1328: @leewujung do you know why this one checks for only `fil_coeffs` for WBT, even though you also appear to need the decimation value? Same thing below with PC, you only check for `fil_df` even though you also need `fil_coeffs`?

leewujung: I think for best practice both should be checked. These parameters are usually all or none, so if one does not exist it means the other also does not exist.

leewujung:
> I think for simplicity and cleanliness, if data variable doesn't exist then they shouldn't be there. When combining, it xarray should be smart enough to fill that with NaN's I think.

Sounds good to just not save them. I think we should not allow combining data with variations of these WBT and PC params.

echopype/convert/set_groups_ek80.py
lsetiawan: > These parameters are usually all or none, so if one does not exist it means the other also does not exist.

To clarify, is this between the WBT and PC or between filter coefficient and decimation? WBT can't exist without PC? or filter coefficient can't exist without decimation?

leewujung: I think it is safer to check them separately:
- WBT filter coeff (and if it does not exist, WBT decimation factor likely won't exist)
- PC filter coeff (and if it does not exist, PC decimation factor likely won't exist)

echopype/convert/set_groups_ek80.py
```diff
@@ -1310,32 +1317,93 @@ def set_vendor(self) -> xr.Dataset:
...
+                # get filter coefficient values
+                for type_num, values in fil_coeffs.items():
+                    param = param_map[type_num]
+                    coeffs_and_decimation[param][FILTER_IMAG].append(np.imag(values))
+                    coeffs_and_decimation[param][FILTER_REAL].append(np.real(values))
+
+                # get decimation factor values
+                for type_num, value in fil_df.items():
+                    param = param_map[type_num]
+                    coeffs_and_decimation[param][DECIMATION].append(value)
```
lsetiawan, line 1343: Cleaned this up, let me know what you think @leewujung. Thanks!

leewujung: `param = param_map[type_num]`
I like this! :D

lsetiawan:
@leewujung if all looks good, please approve PR. @aniketfadia96 could you go through the changes one more time, there has been more commits since your last approval. Thank you both for your help on this 😄

leewujung (approved):

aniketfadia96 (approved):
LGTM

PR #1046, opened 2023-05-19
Fix calibrate to grab variables not attributes

lsetiawan:
## Overview

This PR fixes calibrate to now grab the filter coefficient and decimation values from data variables rather than attributes.


leewujung (approved):
This looks great. Thank you!

emiliom:
FYI I edited the PR title so the scope is clearer, for future reference.

Issue #643, opened 2022-04-22
"Processed" raw power data for EK60/80


leewujung:
Currently a minimum processing is applied in the EK parser with the power data to convert the counts to power.
https://github.com/OSOceanAcoustics/echopype/blob/bc8afa190fa2ca4fab5944bac83cd4b20f7abcf6/echopype/convert/parse_base.py#L122-L140

This means that the power data we stored as `backscatter_r` are technically "processed" data, which conflicts a bit with our intention to store the most "raw" form of data in at level 0.

@emiliom: If we change this and actually save just the counts in `backscatter_r` and move the scaling to part of `compute_Sv`, does it still work with the convention specification **"Real part or amplitude or power of backscatter measurements."** ?

This would be a breaking change since it changes the content of data in `open_raw` and have downstream impact in `compute_Sv`. We can mitigate the latter by including this scaling in the set of changes for v0.5.x --> v0.6.0 conversion (#606). The breaking aspect though does mean that we need to include this in v0.6.0.

emiliom:
Circling back to this. It's a good question / observation. I've pasted below, for reference, the complete SONAR-netCDF4 v1  information about `backscatter_r`.

In general, I agree with your point about counts being more appropriate as rawer, less processed data. But looking at the code, the conversion to power is based on a simple, fixed constant (`INDEX2POWER`) that's not dependent on any variable. So, in this case the distinction between "power" and "counts" seems very small, from a processing perspective.

I think there are other factors to consider, as we make a decision about this:
- It looks like echopype retains counts for AZFP, for `backscatter_r`. I don't see a conversion factor analogous to `INDEX2POWER`. But the variable `long_name` is the same as with EK, "Backscatter power".
- The convention itself is open ended about what exactly `backscatter_r` can hold. I interpret the descriptions as encompassing either counts or power. Do you agree?
- Is there an argument to be made for echopype returning consistent `backscatter_r` (eg, always counts) regardless of the instrument? Or are instruments too heterogeneous in practice for this to be a realistic aim?

> When we consider this revisit putting in detail of what this is in variable attribute.

Agreed. It could go into a `comment` attribute. BTW, the convention specifies the `long_name` "Raw backscatter measurements (real part)". I realize "Backscatter power" is more specific and user friendly, but we should probably revisit this.

Description | Obligation | Comment
--- | --- | ---
[...]

Issue #966, opened 2023-03-09
Refactor and clean up parsed2zarr implementation


leewujung:
@lsetiawan and I talked today about refactoring and cleaning up the current parsed2zarr code. This issue is serve as a reminder for us to do this around v0.7.1 timeframe, and also as a place to link related issues/PRs/discussions.

Related issues/PRs/discussions:
- [x] #795 
- [x] #782
- [x] #777
- [x] refactor/remove `_get_complex_dataarrays` and `_get_angle_dataarrays` that are currently arbitrarily placed in `set_groups_base.py`
- [x] #1053
- [x] Remove `use_swap` arg to `open_raw`
- [x] Add `destination_path` with default `None` with an optional argument of `"swap"` which essetially make `use_swap=True`.
- [x] Check destination path so that it requires `storage_options` when it's remote file... probably already have this feature somewhere!

leewujung:
This section of the code exposes problems in the logic where p2z object is created and then deleted and then created again, which can be changed by refactoring the class methods.
https://github.com/OSOceanAcoustics/echopype/blob/3e16b928d2bc102be3154870d6301a97359b41fe/echopype/convert/api.py#L430-L462

PR #1047, opened 2023-05-22
Standardize backscatter_r/i long_name, and correct units

emiliom:
Addresses #643.

- Standard `long_name` attribute in `backscatter_r` and `backscatter_i` variables for all instruments (except ad2cp) to use the SONAR-netCDF v1 strings. Now being set in single place, in the `long_name` attribute strings in the `echodata/convention/1.0.yml` file.
- Set AZFP `backscatter_r` `units` string to "count", per recent discussion with @leewujung and in #643
- For EK80, correct wrong unit `backscatter_r` / `backscatter_i` assignment being made in some cases, where the unit was set to "V" rather than "dB". @leewujung  please double check this!

Note: I used `# noqa` for a couple of long lines, but black apparently ignored it. I guess we should remove the `# noqa` flags?

-------

Side note: in my local test runs, I had two failures: 
```
FAILED echopype/tests/calibrate/test_cal_params.py::test_get_cal_params_EK80_BB[in_da_freq_dep_with_scaling] (line 549)
  - AssertionError: assert False

FAILED echopype/tests/utils/test_utils_log.py::test_init_logger (line 25)
  - AssertionError: assert 4 == 2
```
I've had these failures happen sporadically since March! I don't know why.

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -265,7 +265,7 @@ class SetGroupsAZFP(SetGroupsBase):
                         "long_name": self._varattrs["beam_var_default"]["backscatter_r"][
                             "long_name"
                         ],  # noqa
-                        "units": "dB",
+                        "units": "count",
                     },
                 ),
                 "equivalent_beam_angle": (
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -261,7 +261,12 @@ class SetGroupsAZFP(SetGroupsBase):
                 "backscatter_r": (
                     ["channel", "ping_time", "range_sample"],
                     N,
-                    {"long_name": "Backscatter power", "units": "dB"},
+                    {
+                        "long_name": self._varattrs["beam_var_default"]["backscatter_r"][
+                            "long_name"
+                        ],  # noqa
+                        "units": "dB",
+                    },
                 ),
                 "equivalent_beam_angle": (
                     ["channel"],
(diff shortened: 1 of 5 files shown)


leewujung (approved):
Thanks @emiliom : the changes look good. The "noqa" stuff is weird. I wonder if it does not work after an "," since all other working cases are not preceded by a comma.

One thing I saw while going through this is the messe organization of parsed2zarr code sprinkled in `set_groups_base.py`. This is something hopefully #966 will address.

leewujung (approved):
Thanks @emiliom : the changes look good. The "noqa" stuff is weird. I wonder if it does not work after an "," since all other working cases are not preceded by a comma.

One thing I saw while going through this is the messe organization of parsed2zarr code sprinkled in `set_groups_base.py`. This is something hopefully #966 will address (I've added this to the issue description TODOs).

emiliom:
> Thanks @emiliom : the changes look good. 

Great!

> The "noqa" stuff is weird. I wonder if it does not work after an "," since all other working cases are not preceded by a comma.

Good suggestion. I may give it a try.

> One thing I saw while going through this is the messe organization of parsed2zarr code sprinkled in set_groups_base.py. This is something hopefully https://github.com/OSOceanAcoustics/echopype/issues/966 will address (I've added this to the issue description TODOs).

Yeah, there was effectively a duplication of code. I didn't try to clean it up. As you said, hopefully some it will be addressed in #966. Thanks for flagging it in that issue's TODOs.

You approved this PR, so it sounds like I can merge it.

leewujung:
> it sounds like I can merge it
Yes, if you could either remove the #noqa or fix it, that'll be good ;)

emiliom:
I've removed the `# noqa` flags that are being ignored. BTW, while doing that I saw other `# noqa` flags in one of these modules that follow lines that end in a comma. So, there must be something else going on. Oh well.

Issue #978, opened 2023-03-13
beam dimension in EK80 calibrated data


emiliom:
This is a a subset of #684 that would be very helpful to address in milestone 0.7.0.

@leewujung : can you flesh out the task, edit the issue title, etc?

leewujung:
@emiliom : I've updated the description above. Let me know if you have questions.

emiliom:
> EK80 has >1 length in the `beam` dimension, but these beams are actually within the same transducer (channel).

> Note that these stored values in the Beam_groupX or Vendor-specific group (if exist, some may not) are transducer-specific and therefore do not vary with `beam` or `ping_time`.

@leewujung I'm trying to see if I understand. You're saying there are multiple beams within each transducer (channel), but the values of the variables you've listed are constant across beam (and ping_time), *within* each channel, right?

Do other EK80 variables that currently have a `beam` dimension vary across the `beam` dimension? The main such variables are `backscatter_r` and `backscatter_i`, but these also occur:

```
beam_direction_x
beam_direction_y
beam_direction_z
angle_alongship
angle_athwartship
frequency_start
frequency_end
```

emiliom:
For reference: the addition of the `beam` dimension was originally discussed in #520

leewujung:
> @leewujung I'm trying to see if I understand. You're saying there are multiple beams within each transducer (channel), but the values of the variables you've listed are constant across beam (and ping_time), _within_ each channel, right?

Yes, they are constant across beam and ping_time, _within_ each channel.


> Do other EK80 variables that currently have a `beam` dimension vary across the `beam` dimension? The main such variables are `backscatter_r` and `backscatter_i`, but these also occur:
> 
> ```
> beam_direction_x
> beam_direction_y
> beam_direction_z
> angle_alongship
> angle_athwartship
> frequency_start
> frequency_end
> ```

For EK80 complex samples (both `backscatter_r` and `backscatter_i` exist):
- the `beam` dimension is necessary and valid -- they are the transducer sectors/quadrants or the "subbeam" dimension in ver.2 definition.
- `angle_alongship` and `angle_athwartship` do not exist in the raw data -- they need to be computed using `add_splitbeam_angle` (right now we have that to add to Sv datasets)
- `frequency_start` and `frequency_end` can vary across ping_time, but there is no beam dimension
- `beam_direction_x/y/z` does not have beam dimension and do not vary across ping_time (for the rigid installation for most echosounders anyway)

[...]

PR #1056, opened 2023-06-05
Drop beam dimension

emiliom:
I'll leave this PR as a draft. For now, it's meant only for discussion, especially for @leewujung to see the minimal changes I made that fully remove the "unnecessary" use of the `beam` dimension for EK60 and AZFP data. ~~only, for now; I'll add AZFP later, where we're also dropping the `beam` dimension completely.~~ For EK80, the removal will not be complete, as the `beam` dimension will remain for a few variables.

- [x] EK60 code updates
- [x] EK60 test updates
- [x] AZFP code updates
- [x] AZFP test updates
- [x] EK80 code updates
- [x] EK80 test updates

Addresses #978

As this will be a breaking change, our intent is to release it with version 0.8.0.


echopype/convert/set_groups_ek60.py
```diff
@@ -45,6 +47,9 @@ class SetGroupsEK60(SetGroupsBase):
         "gain_correction",
     }
 
+    # Variables that need beam and ping_time dimensions added to them.
+    beam_ping_time_names = set()
```
leewujung, line 51: This can be removed?

leewujung (commented):
@emiliom : I checked the changes and they look fine. I think it makes more sense to review this once all changes for different sonar_models are in, since there are some redundant code in `set_groups_base.py` that needs changing too (I just found out). 

Also, a quick note that please make sure for EK80 CW mode data the `beam` dimension is removed too, since it is identical to the EK60 setup. Thanks!

leewujung:
Related to this, I just did #1057 to capture the other half of what #978 exposed but not specifically called out. The artificial expansion (repetition of the same value) along the `ping_time` dimension is more costly than the length=1 `beam` dimension when data is loaded into memory, and I think we should consider removing it at v0.8.0.

echopype/convert/set_groups_ek60.py
emiliom: For now, it needs to be there. The variable is passed to `set_groups_base.beam_groups_to_convention`, which expects a `set` type.

emiliom:
- Removed the `beam` dimension from AZFP
- Updated all EK60 and AZFP tests to account for the removal of the beam dimension

emiliom:
Note: It's quite possible that in calibration modules there are some checks for the presence of the `beam` dimension that are no longer necessary, and I didn't catch that. For example, this block:

https://github.com/OSOceanAcoustics/echopype/blob/c5afb6521a30abce1d74d0a53edc807f0fae6567/echopype/calibrate/calibrate_ek.py#L579-L584

echopype/calibrate/cal_params.py
```diff
@@ -497,9 +497,14 @@ def _get_fs():
...
-                            beam["frequency_nominal"] / freq_center
-                        )
+                        if "beam" in beam[p].dims:
+                            out_dict[p] = beam[p].isel(beam=0).drop("beam") + 20 * np.log10(
+                                beam["frequency_nominal"] / freq_center
+                            )
+                        else:
+                            out_dict[p] = beam[p] + 20 * np.log10(
+                                beam["frequency_nominal"] / freq_center
+                            )
```
leewujung, line 507: remove the added if-else as it should not be there

echopype/calibrate/calibrate_ek.py
```diff
@@ -124,9 +124,12 @@ def _cal_power_samples(self, cal_type: str) -> xr.Dataset:
...
 
-        # Squeeze out the beam dim
-        # doing it here because both out and self.cal_params["equivalent_beam_angle"] has beam dim
-        return out.squeeze("beam", drop=True)
+        # Squeeze out the beam dim, if present.
+        # Doing it here because both out and self.cal_params["equivalent_beam_angle"] have` beam dim
+        if "beam" in out.dims:
+            return out.squeeze("beam", drop=True)
+        else:
+            return out
```
leewujung, line 132: once that the above added if-else is gone, this can be removed.

echopype/convert/set_groups_azfp.py
```diff
@@ -19,13 +19,18 @@ class SetGroupsAZFP(SetGroupsBase):
...
-    beam_only_names = {"backscatter_r"}
+    beam_only_names = set()
 
     # Variables that need only the ping_time dimension added to them.
-    ping_time_only_names = {"sample_interval", "transmit_duration_nominal"}
+    ping_time_only_names = {
+        "sample_interval",
+        "transmit_duration_nominal",
+        "equivalent_beam_angle",
+        "gain_correction",
```
leewujung, line 29: Why are these added? These should be constant throughout `ping_time`.

leewujung: Ok I see they are added since you remove them from the block below.

echopype/tests/echodata/test_echodata.py
```diff
@@ -459,12 +459,16 @@ def test_nan_range_entries(range_check_files):
             echodata, env_params=None, cal_params=None, ecs_file=None, waveform_mode="BB", encode_mode="complex",
         )
         range_output = cal_obj.range_meter
-        nan_locs_backscatter_r = ~echodata["Sonar/Beam_group1"].backscatter_r.isel(beam=0).drop("beam").isnull()
+        if "beam" in echodata["Sonar/Beam_group1"].dims:
+            nan_locs_backscatter_r = ~echodata["Sonar/Beam_group1"].backscatter_r.isel(beam=0).drop(
+                "beam").isnull()
+        else:
+            nan_locs_backscatter_r = ~echodata["Sonar/Beam_group1"].backscatter_r.isnull()
```
leewujung, line 466: this is not needed as the file is EK80 broadband complex data -- the `beam` dimension always needs to be dropped.

leewujung: Done

leewujung:
I've caught the cases in calibration code where the `beam` dim already does not exist and hence does not need to be dropped anymore. Also reverted the if-else handling related to `equivalent_beam_angle`, which should not have a `beam` dimension to begin with.

There are few more cases where the `beam` dimension handling needs to be removed:
- `cal_parasm.py::_get_interp_da`
- `test_cal_params.py::beam_AZFP`
- `test_cal_params.py::test_get_interp_da` test function inputs

leewujung:
I've added the remaining changes. Will close this now to move forward to #1057.

emiliom:
Great, thanks!!

Issue #1051, opened 2023-05-26
Add/handle vertical (z) variables in Platform group, on `echodata.update_platform`


emiliom:
Ensure that variables related to the vertical (z) coordinate are properly handled and populated in the EchoData `update_platform` method:
https://github.com/OSOceanAcoustics/echopype/blob/d08f1f6e460732497fa689d49b4da4d946137cd3/echopype/echodata/echodata.py#L265-L272

This may involve adding an optional argument for the depth relative to the water level, or related variables. Reworking this function will also entail addressing the time dimension name problems mentioned in #740 and #1003.

The two depth-related variables in the `Platform` group, from the convention, are `vertical_offset` and `water_level`.

Related open issues that we will act on now are:
- #740 (see also the [sample notebook](https://gist.github.com/emiliom/9cddce2417713208f40800dc5ba0abb1) I refer to, which I created)
- #1003
- #1062
- #790 

Here are some relevant background discussions (issues) we've had previously. See also the associated PR's. These discussions include specific references to the SONAR-netCDF4 convention.
- #259 (and follow ups, #578 and #579)
- #516
- #592 
- #517

emiliom:
For completeness, there are three other position related Platform variables that have a z component: `position_offset_z`, `MRU_offset_z` and `transducer_offset_z`. They are all z offsets relative to the "Platform Coordinate System" (PCS) origin, which itself is arbitrary.

For a platform with a pressure sensor like a glider or a mooring, the depth (pressure) is actually equivalent to `water_level` + `vertical_offset`. The strategy here should probably be to set `water_level` to the pressure-sensor based depth and set `vertical_offset` to 0.

emiliom:
The [preview](https://echopype-examples--31.org.readthedocs.build/en/31/glider-azfp-and-externaldata.html) to the draft, glider-based echopype examples notebook provides a concrete example of depth and pressure data from a glider external file.

emiliom:
**(TODO)** Another set of relevant variables is `beam_direction_x/y/z`, from `Beam_groupX`. They are mandatory (M), but AZFP has not been creating them. Their addition to AZFP is discussed in #1100 and will be in PR #1102.

`beam_direction_x/y/z` together define a *unit* vector "in the beam direction for each beam and ping, as per the sonar beam coordinate system."

[...]

Issue #1003, opened 2023-03-23
`echodata.update_platform` adds a `water_level` with incorrectly assigned `time1` dimension


emiliom:
This came up in #769. I'm pasting a comment from there, with slight edits. See #656, where we discussed and documented the usage of the various `time*` dimensions. `water_level` should be assigned a `time3` dimension (I think -- we'll review this).

I looked at the AZFP file that I've used in `tests/consolidate/test_consolidate_integration.py`: "azfp/17082117.01A" with XML "azfp/17041823.XML". This file is used in other tests, too. A `water_level` variable is present but it contains all nan's. Looking more closely, it appears that it's an outcome of `update_platform` (with external data), so it's possible that the presence of a `time1` dimension in the `water_level` variable (which is empty, anyways) may be an artifact or bug in `echodata.update_platform`.

emiliom:
This issue was actually already addressed by #1058 and #1060! So, I'm closing it.

PR #1058, opened 2023-06-07
Bring more consistency in the Platform group across sensors on conversion

emiliom:
I went through the behavior of each sensor type (EK60, EK80, AZFP) during conversion (`open_raw`) for the `Platform` group. There were a number of inconsistencies, some of which have been mentioned before in #740 and #1003. This PR also broadly addresses #1051, or creates a more consistent starting point for `update_platform`.

I've addressed most of the inconsistencies:

- AZFP now produces largely the same set of variables as the other two sensors, though most of them are empty (nan). The exceptions are `sentence_type`, a non-convention variable found in EK60 and EK80; and `tilt_x` and `tilt_y`, non-convention variables only used with AZFP.
- Modified the dimensionality of several variables in AZFP to match that in EK80.
- Fixed some incorrect or missing `long_name` attributes, either through edits to `convention/1.0.yml` or via code changes.
- Fixed an issue with EK60 where the global `platform_*` attributes were being dropped after initial inclusion.

I also made some cosmetic edits to all three sensor types to use similar code patterns and order of variables and coordinates, to improve clarity and consistency across sensors.

I ran manual tests but didn't run the test suite locally. If the CI exposes some new errors resulting from these changes, I'll update the PR.

Some open areas for improvement (but possibly not for 0.7.2):
- In EK80, several variables are being populated with 0.0 where I suspect they should be Nan. I think this is happening in the parser code, but I only focused on the `set_group_*` modules.
- EK60 adds the `channel` dimension to a number of variables for which EK80 doesn't, and EK80 adheres more closely to the convention with respect to variable dimensionality. I think the presence of the extra channel dimension is a mistake, but wanted to first double check with @leewujung before removing that dimension.
- The convention specifies that `water_level` is a scalar and does not have a time dimension; time variability is found in `vertical_offset` instead, which in a ship corresponds to heave. But we've been assigning a `time3` dimension to `water_level`.
- More broadly, it looks like the convention doesn't *require* the creation of *empty* (all Nan) variables when their "obligation" flag is not "Mandatory" (M). See this [protracted discussion](https://github.com/ices-publications/SONAR-netCDF4/issues/26) and [my own recent question at the end](https://github.com/ices-publications/SONAR-netCDF4/issues/26#issuecomment-1555397003). It'll be up to us to decide whether to adopt a general practice for these cases. For reference, *none* of the variables in the Platform group are "Mandatory"!

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -142,6 +142,7 @@ class SetGroupsAZFP(SetGroupsBase):
         platform_dict = {"platform_name": "", "platform_type": "", "platform_code_ICES": ""}
         unpacked_data = self.parser_obj.unpacked_data
         time2 = self.parser_obj.ping_time
+        time1 = [time2[0]]
 
         # If tilt_x and/or tilt_y are all nan, create single-value timme2 dimension
         # and single-value (np.nan) tilt_x and tilt_y
@@ -187,7 +188,7 @@ class SetGroupsAZFP(SetGroupsBase):
                     tilt_x,
                     {
                         "long_name": "Tilt X",
-                        "units": "degree",
+                        "units": "arc_degree",
                     },
                 ),
                 "tilt_y": (
@@ -195,7 +196,7 @@ class SetGroupsAZFP(SetGroupsBase):
                     tilt_y,
                     {
                         "long_name": "Tilt Y",
-                        "units": "degree",
+                        "units": "arc_degree",
                     },
                 ),
                 **{
@@ -244,7 +245,7 @@ class SetGroupsAZFP(SetGroupsBase):
                 "time1": (
                     ["time1"],
                     # xarray and probably CF don't accept time coordinate variable with Nan values
-                    [time2[0]],
+                    time1,
                     {
                         **self._varattrs["platform_coord_default"]["time1"],
                         "comment": "Time coordinate corresponding to NMEA position data.",
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -149,6 +149,36 @@ class SetGroupsAZFP(SetGroupsBase):
 
         ds = xr.Dataset(
             {
+                "latitude": (
+                    ["time1"],
+                    [np.nan],
+                    self._varattrs["platform_var_default"]["latitude"],
+                ),
+                "longitude": (
+                    ["time1"],
+                    [np.nan],
+                    self._varattrs["platform_var_default"]["longitude"],
+                ),
+                "pitch": (
+                    ["time2"],
+                    [np.nan] * len(time2),
+                    self._varattrs["platform_var_default"]["pitch"],
+                ),
+                "roll": (
+                    ["time2"],
+                    [np.nan] * len(time2),
+                    self._varattrs["platform_var_default"]["roll"],
+                ),
+                "vertical_offset": (
+                    ["time2"],
+                    [np.nan] * len(time2),
+                    self._varattrs["platform_var_default"]["vertical_offset"],
+                ),
+                "water_level": (
+                    ["time3"],
+                    [np.nan],
+                    self._varattrs["platform_var_default"]["water_level"],
+                ),
                 "tilt_x": (
                     ["time2"],
                     unpacked_data["tilt_x"],
@@ -166,6 +196,18 @@ class SetGroupsAZFP(SetGroupsBase):
                     },
                 ),
                 **{
+                    var: (
+                        ["channel"],
+                        [np.nan] * len(self.channel_ids_sorted),
+                        self._varattrs["platform_var_default"][var],
+                    )
+                    for var in [
+                        "transducer_offset_x",
+                        "transducer_offset_y",
+                        "transducer_offset_z",
+                    ]
+                },
+                **{
                     var: ([], np.nan, self._varattrs["platform_var_default"][var])
                     for var in [
                         "MRU_offset_x",
@@ -177,15 +219,34 @@ class SetGroupsAZFP(SetGroupsBase):
                         "position_offset_x",
                         "position_offset_y",
                         "position_offset_z",
-                        "transducer_offset_x",
-                        "transducer_offset_y",
-                        "transducer_offset_z",
-                        "vertical_offset",
-                        "water_level",
                     ]
                 },
+                "frequency_nominal": (
+                    ["channel"],
+                    self.freq_sorted,
+                    {
+                        "units": "Hz",
+                        "long_name": "Transducer frequency",
+                        "valid_min": 0.0,
+                        "standard_name": "sound_frequency",
+                    },
+                ),
             },
             coords={
+                "channel": (
+                    ["channel"],
+                    self.channel_ids_sorted,
+                    self._varattrs["beam_coord_default"]["channel"],
+                ),
+                "time1": (
+                    ["time1"],
+                    # xarray and probably CF don't accept time coordinate variable with Nan values
+                    [time2[0]],
+                    {
+                        **self._varattrs["platform_coord_default"]["time1"],
+                        "comment": "Time coordinate corresponding to NMEA position data.",
+                    },
+                ),
                 "time2": (
                     ["time2"],
                     time2,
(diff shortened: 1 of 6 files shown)


emiliom:
Alright, the tests are passing now! The only failure was with AZFP. `check_platform_required_vars` was actually checking for a set of variables that were all scalar; I removed the variables that are no longer scalar.

leewujung:
I happened to see this following block while reviewing the EK60 changes. What do you think we just remove this TODO, and also remove the said EK80 capability? In the echopype examples notebook we showed how to modify it directly on the EchoData object, so maybe we should just remove the capability to put it in as part of conversion all together?
https://github.com/OSOceanAcoustics/echopype/blob/8aa4d9f77d7f09c269c3441d22920b3985b5a0de/echopype/convert/set_groups_ek60.py#L279-L285

echopype/convert/set_groups_ek60.py
```diff
@@ -241,6 +241,11 @@ def set_platform(self) -> xr.Dataset:
         time1, msg_type, lat, lon = self._extract_NMEA_latlon()
 
         # NMEA dataset: variables filled with nan if do not exist
+        platform_dict = {
+            "platform_name": self.ui_param["platform_name"],
+            "platform_type": self.ui_param["platform_type"],
+            "platform_code_ICES": self.ui_param["platform_code_ICES"],
+        }
```
leewujung, line 248: For this and line 279-285: What do you think we just remove this TODO, and also remove the said EK80 capability? In the echopype examples notebook we showed how to modify it directly on the EchoData object, so maybe we should just remove the capability to put it in as part of conversion all together?
https://github.com/OSOceanAcoustics/echopype/blob/8aa4d9f77d7f09c269c3441d22920b3985b5a0de/echopype/convert/set_groups_ek60.py#L279-L285

emiliom:
Additional changes I'll add to this PR:
- Remove `time3` dimension from `water_level`
- Remove inappropriate / incorrect `channel` dimension from several variables in EK60

echopype/convert/set_groups_azfp.py
```diff
@@ -147,24 +147,73 @@ def set_platform(self) -> xr.Dataset:
...
+                "roll": (
+                    ["time2"],
+                    [np.nan] * len(time2),
+                    self._varattrs["platform_var_default"]["roll"],
+                ),
+                "vertical_offset": (
+                    ["time2"],
+                    [np.nan] * len(time2),
+                    self._varattrs["platform_var_default"]["vertical_offset"],
+                ),
```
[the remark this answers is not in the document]
emiliom: > Are the variables separated out to be either with the time1 or the time2 dimension because we want to keep the consistency across sensors? 

Yes, that's the goal here.

> Also I wonder if it'll be useful to have time1=[time2[0]] somewhere before the dataset construction block, just for readability.

I like that. Maybe right below the statement `time2 = self.parser_obj.ping_time` (line 148), so that the data for all time dimensions are specified in one block?

echopype/convert/set_groups_ek60.py
emiliom: > For this and line 279-285: What do you think we just remove this TODO, and also remove the said EK80 capability? In the echopype examples notebook we showed how to modify it directly on the EchoData object, so maybe we should just remove the capability to put it in as part of conversion all together?

I confess I'd never bothered to see what `ui_param` is! If I get that right (from a quick code scan), it allows passing certain parameter values during the conversion process, on `open_raw`, and some of those parameters are used for setting those global `param_*` attributes. If that's correct, I wasn't aware that was there! Let's talk about this today to make sure I understand. 

I'm all for removing that long TODO comment, and likewise removing the EK80 water_level updating capability that's based on `self.ui_param['water_level']`.

echopype/convert/set_groups_azfp.py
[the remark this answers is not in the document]
emiliom: Done. Set `time1=[time2[0]]`.

emiliom:
> Remove time3 dimension from water_level

Done. Removed `time3` completely from the Platform group, for all 3 sensor types. In addition to `water_level`, 3 additional, non-convention EK80 variables became scalar.

> Remove inappropriate / incorrect channel dimension from several variables in EK60

Done.

emiliom:
I suspect some tests will now fail, but we'll see. I again ran into errors building the local test environment, so I couldn't run the test suite locally.

echopype/convert/set_groups_ek60.py
emiliom: I've removed that ek60 comment block, and also removed the `self.ui_param` statements that referred to the Platform global `param_*` attributes in all 3 sensors, while I was at it.

I'll create a new issue about removing the `ui_param` property.

emiliom: Opened new issue #1059

emiliom:
Only ek80 convert tests were failing. I've fixed them by addressing two changes:
- Removal of `time3` dimension from Platform group
- A change I made in `set_group_ek80.py` originally for code readability actually fixed an unrecognized bug. For `drop_keel_offset`, a scheme involving `hasattr` was incorrectly being used on a dictionary, resulting in `np.nan` always being assigned even when there was valid data present. This was fixed by using the dictionary `get` method, and the fix required a change to a test.

emiliom:
Woo-hoo, all tests are passing! This PR is ready for a final review, and hopefully merging soon after.

echopype/tests/convert/test_convert_ek80.py
```diff
@@ -58,23 +58,23 @@ def check_env_xml(echodata):
...
     assert np.array_equal(echodata["Environment"]["sound_velocity_profile_depth"], [1, 1000])
 
-    # check plat vars
+    # check a subset of plat vars
     plat_vars = {
-        "drop_keel_offset": [np.nan],
+        "drop_keel_offset": [np.nan, 0, 7.5],
         "drop_keel_offset_is_manual": [0, 1],
         "water_level": [0],
         "water_level_draft_is_manual": [0, 1]
```
leewujung, line 66: I am puzzled by this block -- aren't these variables all supposed to just a single value? Also in line 73 below, it does not test for actual values, so the fact that the dimensions do not match does not show up.

leewujung (commented):
@emiliom : thanks for the changes! Everything looks fine. I just have a question about the test in test_convert_ek80, since the dimension do not match what we intended for the data variables, and the assert does not check the actual values. I know that this is an inherited test, but to avoid confusion later, I think we should fix it. (Or clarify what's going on if I am reading the code wrong -- I have been really slow this morning.)

echopype/tests/convert/test_convert_ek80.py
emiliom: I was very puzzled too, for the same reason as you! What's going on is that `plat_vars` specifies a list of possible *expected* values for each of those variables. The next block, starting in line 68 steps through each of those variables and tests (last line in the block) if the value in the data file matches one of the expected values.

I'll add a comment about this to clarify.

echopype/tests/convert/test_convert_ek80.py
```diff
@@ -58,23 +58,23 @@ def check_env_xml(echodata):
     assert "sound_velocity_profile_depth"
     assert np.array_equal(echodata["Environment"]["sound_velocity_profile_depth"], [1, 1000])
 
-    # check plat vars
+    # check a subset of plat vars
```
emiliom, line 61:
```suggestion
    # check a subset of platform variables. plat_vars specifies a list of possible, expected scalar values 
    # for each variable. The variables from the EchoData object are tested against this dictionary
    # to verify their presence and their scalar values
```

echopype/tests/convert/test_convert_ek80.py
emiliom: I've added the comment. It's worth noting also that the tests for `water_level` and `drop_keel_offset` are not good ones, in general, b/c they assume a limited set of possible values, which in theory is incorrect. They work for the ek80 raw files currently being tested. That's in contrast to the two `*_is_manual` variables, which, based on what I've seen, I assume are boolean and therefore only 0 or 1, so the tests are correct.

emiliom:
NOTE: The build CI failed. Looking into it, I realized that I forgot that the PR CI by default runs in the selective mode so that only "convert" tests were run; plus b/c I was having problems with running tests locally, I never ran the full suite.

So, there are apparently more fixes to make resulting from the changes in this PR. From https://github.com/OSOceanAcoustics/echopype/actions/runs/5226125815/jobs/9436435332, the failures are in `tests/calibrate/test_calibrate.py::test_env_params`, `tests/utils/test_processinglevels_integration.py::test_raw_to_mvbs[AZFP-AZFP-raw_and_xml_paths1-extras1]` and `tests/visualize/test_plot.py::test_water_level_echodata[True-False]`

leewujung:
Ah that's too bad! Hopefully these are easy to fix.

emiliom:
After a quick look, two out of three seem easy. The 3rd one, I'm not sure yet.

PR #1060, opened 2023-06-09
update_platform overhaul [all tests ci]

emiliom:
Addresses #1003 and #1051. Overhauls `update_platform` functionality while updating to ensure it works with the updates to the Platform group introduced in #1058.

Relevant changes in PR #1058: `time3` is no longer used in the Platform group. `water_level` now follows the convention in being a scalar value, while `vertical_offset` retains the `time2` dimension

The primary overhaul to `update_platform` was to remove the existing mapping of a pre-defined set of Platform variables to a pre-defined set of target variable names on the external data source. This previous mapping also resulted in all those variables being dropped from the Platform group regardless of user preference or the presence of valid data in the Platform variables. This PR introduces a new argument `variable_mappings` that specifies the mapping between a target Platform variable name and the matching variable name in the external data. Only those variables will be updated explicitly (but see exceptions below).

Data conversion (`open_raw`) results in only three variables assigned a `time1` dimension: `latitude`, `longitude` and for ek60 & ek80, `sentence_type`. All other variables are either scalar or assigned a `time2` dimension. However, in many cases such as deployments from AUVs, all data may be on a single time dimension, including lat & lon, pitch, roll, and pressure (depth). This necessitates a decision: assign a `time1` dim for lat & lon and create a duplicate `time2` dim for all other variables, to maintain consistency of time dimension assignments in the Platform group. Or break the consistency and either use a single time dimension, say `time1` or a `time1` for lat & lon and a new, unused dimension, such as `time4` for other time-dependent variables.

- If creating a duplicate `time2` dim, existing `time2`-dependent variables having valid data will be dropped because the `time2` dimension must first be dropped. That's clearly an undesirable effect. eg, for AZFP `tilt_x` and `tilt_y` are on `time2` may contain valid data.
- If using a single `time1` dimension, we'll break the consistent expectation for variables that are assigned a `time2` dimension during conversion.

The current implementation in this PR uses a single `time1` dimension.


leewujung:
@emiliom : I skimmed through the code changes and wanted to propose a different solution: to increment the time coordinate (`time3`, `time4`, etc) for variables added or substituted via `update_platform`, and in the attributes of these variables specify that the variable comes from `update_platform`. 

The reason is that even though it is conceptually nice to always have lat/lon with dimension `time1` and heave(vertical_offset)/roll/pitch with dimension `time2`, it is way too messy to determine a rule of what to be dropped/substituted, which could result in unintentional outcome. Checking the number of existing time coordinates and determine what the next one should be is relatively straightforward. 

We can enforce some simple rules that makes intuitive sense
- the final variables `lat` `lon` should have identical time dimension
- the final variables  `vertical_offset` `roll` and `pitch` I think we can allow the flexibility for them to be different, especially since in the case of underwater vehicles, the pressure value (which will give values for `vertical_offset`) is possibly from a different sensor than `roll` and `pitch`.

Concrete examples of my proposal below:
#### Case 1 (substitute 2 variables with `time3`)
- setup:
  - the original `Platform` group contains `lat` and `lon` with coordinate `time1`
  - the external dataset contains only `ext_lat` and `ext_lon` with coordinate `ext_timeA`
  - the input mapping dict is: `{"lat": "ext_lat", "lon": "ext_lon"}`
- behavior:
   - the existing `lat` and `lon` are dropped
   - the external variables `ext_lat` and `ext_lon` are added to the `Platform` group with names `lat` `lon`, with coordinate `time3`

#### Case 2 (substitute 2 variables with `time3` and add 3 variables with `time4`)
- setup:
  - the original `Platform` group contains `lat` and `lon` with coordinate `time1`
  - the external dataset contains `ext_lat` and `ext_lon` with coordinate `ext_timeA`
  - the external dataset also contains `heave`, `roll`, `pitch` with coordinate `ext_timeB`
  - input mapping dict: `{"lat": "ext_lat", "lon": "ext_lon", "heave": "vertical_offset", "roll": "roll", "pitch": "pitch"}`
- behavior:
   - the existing `lat` and `lon` are dropped
   - the external variables `ext_lat` and `ext_lon` are added to the `Platform` group with names `lat` `lon`, with coordinate `time3`
   - the external variables `heave` `roll` and `pitch` are added to the `Platform` group with names `vertical_offset` `roll` and `pitch`, with coordinate `time4`

#### Case 3 (add 3 variables with `time3`)
- setup:
  - the original `Platform` group contains `lat` and `lon` with coordinate `time1`
  - the external dataset contains `ext_lat` and `ext_lon` with coordinate `ext_timeA`
  - the external dataset also contains `heave`, `roll`, `pitch` with coordinate `ext_timeB`
  - input mapping dict: `{"heave": "vertical_offset", "roll": "roll", "pitch": "pitch"}`
- behavior:
   - the external variables `heave` `roll` and `pitch` are added to the `Platform` group with names `vertical_offset` `roll` and `pitch`, with coordinate `time3`

emiliom:
I fully agree with your proposal, and I don't expect it to be hard to implement. Here I'll add a 4th case that's a variation of case 2:

#### Case 4 (substitute 5 variables with `time3`)
- setup:
  - the original `Platform` group contains `lat` and `lon` with coordinate `time1`
  - the external dataset contains `ext_lat` and `ext_lon` with coordinate `ext_timeA`
  - the external dataset also contains `heave`, `roll`, `pitch`, also with coordinate `ext_timeA`
  - input mapping dict: `{"lat": "ext_lat", "lon": "ext_lon", "vertical_offset": "heave", "roll": "roll", "pitch": "pitch"}`
- behavior:
   - the existing `lat`, `lon`, `vertical_offset`, `roll` and `pitch` are dropped
   - the external variables `ext_lat`, `ext_lon`, `heave`, `roll` and `pitch`  are added to the `Platform` group with names `lat` `lon`, `vertical_offset` `roll` and `pitch` with coordinate `time3`

emiliom:
I've implemented everything (I think!) and fixed all downstream code and tests that broke. That includes updates to `consolidate.add_location` and corresponding tests.

I believe all the use cases listed in this issue can now be handled. *But*, I did not add new tests specifically for the case where more than one time dimension (case 2) will be added. That will be added later on, but not on this PR unless we drag it on into next week.

I also fixed the `add_processing_level` decorator function so that when longitude and latitude are not passed to `update_platform` (via `variable_mappings`), the processing level code "L1A" will not be added.

Note: when running the local test suite, I again rain into the segmentation fault with the netcdf4 export test (see comment in #1061). Everything ran fine when I deselected the netcdf4 "engine" from the test. Hopefully, as in #1061, the CI will not run into that problem.

echopype/echodata/echodata.py
```diff
@@ -315,114 +314,167 @@ def update_platform(
...
+            # external_ds is extra_platform_data[vars-list-on-time_dim_name]
+            sorted_external_time = external_ds[ext_time_dim_name].data
+            sorted_external_time.sort()
+            # fmt: off
+            min_index = max(
+                np.searchsorted(
+                    sorted_external_time, self["Sonar/Beam_group1"]["ping_time"].min(), side="left"
+                ) - 1,
+                0,
+            )
```
leewujung, line 362: Just wanted to make it the same format as in max_index below
```suggestion
            min_index = max(
                np.searchsorted(
                    sorted_external_time,
                    self["Sonar/Beam_group1"]["ping_time"].min(),
                    side="left"
                ) - 1,
                0,
            )
```

echopype/echodata/echodata.py
```diff
@@ -315,114 +314,167 @@ def update_platform(
...
-        dropped_vars_target = platform_vars_sourcenames.keys()
-        dropped_vars = []
-        for var in dropped_vars_target:
-            if var in platform and (~platform[var].isnull()).all():
-                dropped_vars.append(var)
-        if len(dropped_vars) > 0:
+        platform = self["Platform"]
+
+        # Drop variable_mappings items where either the Platform group or
+        # extra_platform_data don't contain the corresponding variables or contain only nan data
```
leewujung, line 389: Suggesting changing the docstring since the operation is subsetting and not dropping
```suggestion
        # Retain only variable_mappings items where 
        # either the Platform group or extra_platform_data
        # contain the corresponding variables or contain valid (not all nan) data
```

echopype/echodata/echodata.py
```diff
@@ -315,114 +314,167 @@ def update_platform(
...
+
+            # Create new time coordinate and dimension
+            platform = platform.assign_coords(**{time_dim: ext_ds[ext_time_dim].values})
+            # TODO: Update long_name and comment to something more appropriate
+            time_attrs = {
+                "axis": "T",
+                "standard_name": "time",
+                "long_name": "Timestamps for platform motion and orientation data",
+                "comment": "Time coordinate corresponding to platform motion and "
+                "orientation data.",
```
leewujung, line 441:
```suggestion
                "comment": "Time coordinate corresponding to platform motion and "
                "orientation data originated from a dataset external to the sonar data files.",
```

echopype/echodata/echodata.py
```diff
@@ -315,114 +314,167 @@ def update_platform(
...
+                platform_validvalues = not platform[platform_var].isnull().all()
+                ext_present, ext_validvalues, ext_time_dim_name = _extvar_properties(
+                    extra_platform_data, external_var
+                )
+                if ext_present and ext_validvalues:
+                    mappings_expanded[platform_var] = (
+                        external_var,
+                        ext_time_dim_name,
+                        platform_validvalues,
+                    )
```
leewujung, line 402: I found myself flipping back multiple times to this block to check what index 0, 1, 2 means. How about changing this to a dictionary of dictionaries? Then the code below can explicitly use "external_var", "ext_time_dim_name", and "platform_validvalues" in the places where you have `v[0]`, `v[1]`, `v[-1]`.

echopype/echodata/echodata.py
```diff
@@ -315,114 +314,167 @@ def update_platform(
...
-            if var in platform and (~platform[var].isnull()).all():
-                dropped_vars.append(var)
-        if len(dropped_vars) > 0:
+        platform = self["Platform"]
+
+        # Drop variable_mappings items where either the Platform group or
+        # extra_platform_data don't contain the corresponding variables or contain only nan data
+        mappings_expanded = {}
+        for platform_var, external_var in variable_mappings.items():
+            if platform_var in platform:
```
leewujung, line 392: Here we rely on whatever existing variables in the Platform group to say whether a particular `platform_var` is allowed. I think this can be vulnerable if the Platform group is not well-formed (though in theory they all are). How about we declare a tuple `VALID_PLATFORM_VARIABLES` up front in this module and we check explicitly against it?

echopype/echodata/echodata.py
```diff
@@ -268,43 +268,42 @@ def __setattr__(self, __name: str, __value: Any) -> None:
...
-            - `"vertical_offset"`
-            - `"latitude"`
-            - `"longitude"`
-            - `"water_level"`
-        The data inserted into the Platform group will be indexed by a dimension named
-        `"time1"`.
+        `extra_platform_data` must be an xarray Dataset. Data is extracted from
+        `extra_platform_data` by variable name. Only data assigned to a pre-existing
+        (but possibly all-nan) variable name in Platform will be processed. These
+        Platform variables include latitude, longitude, pitch, roll, vertical_offset, etc.
```
leewujung, line 280: How about putting a bullet point list here for allowable variables (those in `VALID_PLATFORM_VARS` -- see [my other comment](https://github.com/OSOceanAcoustics/echopype/pull/1060/files#r1232423132)) so people know what these are?

echopype/echodata/echodata.py
```diff
@@ -268,43 +268,42 @@ def __setattr__(self, __name: str, __value: Any) -> None:
...
+        Different external variables may be dependent on different time dimensions, but
+        latitude and longitude (if specified) must share the same time dimension. New
```
leewujung, line 282: I don't see a check for this in the code. How about adding a simple defensive check to make sure this is true? Seems like once you have `mappings_expanded` this would be straightforward to do.

echopype/echodata/echodata.py
emiliom: I like that. Thanks. I'll merge the commit.

echopype/echodata/echodata.py
```diff
@@ -315,114 +314,167 @@ def update_platform(
...
+                len(sorted_external_time) - 1,
+            )
+            return external_ds.sel(
+                {
+                    ext_time_dim_name: np.logical_and(
+                        sorted_external_time[min_index] <= external_ds[ext_time_dim_name],
+                        external_ds[ext_time_dim_name] <= sorted_external_time[max_index],
+                    )
+                }
+            )
```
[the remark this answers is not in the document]
emiliom: I don't remember specifically, but the case it's accounting for is what you describe. The whole scheme of determining `min_index` and `max_index` plus this part seems clearly more inefficient than a simple `sel`. I suggest we stick with it for now and revise it later if need be.

echopype/echodata/echodata.py
emiliom: Thank you!! That looks great

echopype/echodata/echodata.py
emiliom: Yeah ... I don't disagree. I ran out of steam to replace it with a dict. Let me see if I can do it now.

echopype/echodata/echodata.py
emiliom: > Here we rely on whatever existing variables in the Platform group to say whether a particular platform_var is allowed. I think this can be vulnerable if the Platform group is not well-formed (though in theory they all are).

That's true, but as you say, in theory they'll be well formed going forward after the recent Platform overhaul. 

> How about we declare a tuple VALID_PLATFORM_VARS up front in this module and we check explicitly against it?

Declaring a tuple `VALID_PLATFORM_VARS` does ensure a more explicit and uniform checking, but it introduces a redundant specification of the list of variables, which is a problem of a different sort. I'm reluctant to do that. In theory that tuple could be generated dynamically from a source of truth, like `echodata/convention/1.0.yml`; but that's not there yet as a source of complete truth.

echopype/echodata/echodata.py
emiliom: Right! I'll add that. I thought I had done it, but I guess not.

echopype/echodata/echodata.py
emiliom: Hmm, I'd make the same comment as the one on your other comment. The user could simply inspect the set of variables in their echodata object's Platform group. After all, to apply `update_platform`, the user needs to already have a Platform object. How about adding a statement to that effect in this docstring? Here's a suggestion:
```suggestion
        Platform variables include latitude, longitude, pitch, roll, vertical_offset, etc.
        See the variables present in the EchoData object's Platform group to obtain a
        complete list of possible variables.
```

echopype/echodata/echodata.py
emiliom: I don't know what I did that ended up dropping your suggestion for `long_name`. Sorry. I've tried to recreate and will submit it in my next push to this PR.

echopype/echodata/echodata.py
leewujung: Yeah, I agree ultimately that should come from somewhere else, but since we're not going to get there anytime soon, I feel that it would help if those are explicitly declared up top. It could come with a comment that ultimately that should be changed to something dynamically generated from a source of truth.

echopype/echodata/echodata.py
emiliom: Ok, I've pushed these changes just now. The code is much more readable now!

echopype/echodata/echodata.py
emiliom: Done, in my latest commits.

emiliom:
From https://github.com/OSOceanAcoustics/echopype/pull/1060#pullrequestreview-1483801706

> Yeah, I agree ultimately that should come from somewhere else, but since we're not going to get there anytime soon, I feel that it would help if those are explicitly declared up top. It could come with a comment that ultimately that should be changed to something dynamically generated from a source of truth.

I'll have to pass on doing anything further on this at this time. I've run out of time and won't be able to get back to echopype for a couple of days. 

Thanks for your input! In addition to your edits to docstrings and comments that I committed directly within the PR, my new commits address the latitude-longitude same-time-dimension check, the change of `mappings_expanded` to a much more readable dict of dicts, and the new test for the case of external variables on two time dimensions plus a scalar variable.

All CI tests passed. I can merge this PR tomorrow morning. Or I can leave it to you to merge later if you prefer to make additional tweaks.

echopype/echodata/echodata.py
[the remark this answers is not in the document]
leewujung: I added a TODO item to come back to this later at the workflow stage.

leewujung:
Alright, I added a few commits on top of your changes:
- add tests for lat/lon time dimension checking:
  - when lat and lon do not align on the same time dimension
  - when only one of them exists
- add warning when no variables are updated when `mappings_expanded` is empty
- add test for catching the warning when no variables are updated

I looked into adding `VALID_PLATFORM_VARS` here, but discovered that some Platform variables exist only for specific sonar_models:
- AZFP: tilt_x and tilt_y
- EK80: drop_keel_offset, drop_keel_offset_is_manual, water_level_draft_is_manual
I think these are caught in the format checker you are working on.
I still think it is important to let users know for what dataset they can add what variables, so once the CDL and checker are in place, let's come back to this.
I added a TODO item to the code.

I'll merge this once the tests run through!

echopype/echodata/echodata.py
leewujung: I added a TODO item in the code to come back to this once the CDL and checker are in place.

PR #1061, opened 2023-06-11
Additional test and related fixes stemming from Platform group PR #1058 [all tests ci]

emiliom:
Adapting what I wrote in https://github.com/OSOceanAcoustics/echopype/pull/1058#issuecomment-1585755148. The previously missed test failures were:
- `tests/calibrate/test_calibrate.py::test_env_params`: I removed the test. It was testing the obsolete `EnvParam` class in `calibrate/env_params_old.py`, which was replaced, though not 1:1, by `env_params.py::ENVPARAM` and a new accompanying test. Per @leewujung : `env_params.py` does not have the interpolation capability but it is more complete in terms of being able to take user input dictionary to `compute_Sv`. @leewujung will follow up with a PR to remove `env_params_old.py`
- `tests/utils/test_processinglevels_integration.py::test_raw_to_mvbs[AZFP-AZFP-raw_and_xml_paths1-extras1]`: Upgraded it to account for the case (especially with AZFP) where Platform `latitude` and `longitude` variables are present but are all nan. That was not the case for AZFP prior to PR #1058
-  `tests/visualize/test_plot.py::test_water_level_echodata[True-False]`: Made substantial changes to the test and to `visualize/api.py` module to change the use of `water_level` to `vertical_offset` (and similar references to water level), and to remove in the test the handling of the former, inappropriate `channel` dimension in ek60 `water_level` and `vertical_offset`. Made a few minor, additional improvements along the way, such as for type hints.

Note: I successfully ran the complete test suite locally *except* I temporarily removed the "netcdf4" export engine option from `test_convert_source_target_locs.py::export_engine`, to sidestep the new, unexpected (and unrelated) problem with the `Vendor_specific` group and netcdf4 files. On the first run of the CI, all tests ran successfully and the netcdf4 segmentation fault did not take place. Hmm.


leewujung (approved):
@emiliom : thanks, the changes look good. I am still puzzled by the segmentation problem, but it can be investigated later as long as we don't normally run into that problem. It is also weird to me that I ran into that in the `Sonar/Beam_group1` rather than the `Vendor_specific` group like you had.

For the sake of clarity, could you add the definition of `water_level` and `vertical_offset` with a simple diagram of which one means which for an underwater vehicle and a ship? This should actually be in the docs, but if we have something parked here it'll be easy to plug them in afterwards.

Other than that, please feel free to merge!

emiliom:
Great, thanks! Merging.

I created a new issue, #1062, about the documentation to be added (great idea, BTW).

emiliom:
This PR is related to #1051

Issue #798, opened 2022-08-26
Improve and add documentation for `_clean_ping_time`


b-reyes:
In PR #297 we created the function [_clean_ping_time](https://github.com/OSOceanAcoustics/echopype/blob/828587df6f49c6b7e49cef6d4af9246fc2e0be19/echopype/qc/api.py#L4), but did not include documentation. This function involves a recursive call, so it is not apparent what is happening. After discussing this with @leewujung, we decided that it is important to include some documentation for this function. 

Additionally, @leewujung do you think it is possible to remove `local_win_len` as an input and determine it within `_clean_ping_time`? I think that would make using this function less error-prone.

b-reyes:
@leewujung I was thinking about the `_clean_ping_time` function and I have constructed an alternative approach that avoids recursion, improves readability, and removes the need for a window. 

The goal of this code is to create ascending values and preserve the step size (wherever possible). Below is my code/example to do this. If you think it is a good candidate to replace `_clean_ping_time`, I can create a PR that fully implements it. Note: I quickly implemented it, so if you see any issues, please let me know. 

```
import numpy as np

arr = np.array([0,1,2,3,4,2,3,4,6,8,10,11,12,13,15,17,19,13,15,17,21])
arr_diff = np.diff(arr)

print(f"arr = {arr}")
print(f"arr_diff = {arr_diff} \n")

# get indices of arr_diff with negative values                                                        
neg_ind = np.argwhere(arr_diff < 0).flatten()

# TODO: account for neg_ind + 1 = len(arr_diff)                                                       
# TODO: do we need to account for neg diff next to eachother?                                         
# replace negative difference with difference one index after                                         
arr_diff[neg_ind] = arr_diff[neg_ind + 1]

# perform cummulative sum of differences after 1st neg index                                          
c_diff = np.cumsum(arr_diff[neg_ind[0]:], axis=0)

# create new array that preserves differences, but enforces increasing vals                           
new_arr = arr[:]
new_arr[neg_ind[0]+1:] = new_arr[neg_ind[0]] + c_diff
print(f"new_arr = {new_arr}")
print(f"np.diff(new_arr) = {np.diff(new_arr)}")
```

The output of the above example: 

```
arr = [ 0  1  2  3  4  2  3  4  6  8 10 11 12 13 15 17 19 13 15 17 21]
arr_diff = [ 1  1  1  1 -2  1  1  2  2  2  1  1  1  2  2  2 -6  2  2  4] 

[...]

PR #1065, opened 2023-06-17
Revise qc function to clean up reversed time

leewujung:
This PR addresses #798:
- Changed from the original recursive function to the new implementation based on cumulative summation of the differences
- Rename `_clean_ping_time` to `_clean_timestamps`
- Updated `coerce_increasing_time`
- Add tests for `_clean_timestamps` and `coerce_increasing_time`

Note:
- No change was needed for `exist_reversed_time`
- Need to decide if to keep `create_old_time_array` and `orchestrate_reverse_time_check`, and then to revise or deprecate: pinging @emiliom for thoughts


echopype/qc/api.py
```diff
@@ -9,32 +9,34 @@
...
-            )
-        )
-        return _clean_ping_time(ping_time_new, local_win_len=local_win_len)
-    else:
-        return ping_time_old  # no negative diff
+def _clean_reversed(time_old: np.ndarray, win_len: float):
+    time_old_diff = np.diff(time_old)
+
+    # get indices of arr_diff with negative values
+    neg_idx = np.argwhere(time_old_diff < np.timedelta64(0, "ns")).flatten()
```
emiliom, line 16: Isn't `time_old_diff` 1D, and therefore `np.argwhere` returns a 1D array? If so, why is flatten() needed?

echopype/qc/api.py
```diff
@@ -9,32 +9,34 @@
...
-            (
-                ping_time_old[: ni + 1],
-                ping_time_old[ni]
-                + np.cumsum(np.hstack((local_pt_median, ping_time_old_diff[(ni + 1) :]))),
-            )
-        )
-        return _clean_ping_time(ping_time_new, local_win_len=local_win_len)
-    else:
-        return ping_time_old  # no negative diff
+def _clean_reversed(time_old: np.ndarray, win_len: float):
```
emiliom, line 12:
```suggestion
def _clean_reversed(time_old: np.ndarray, win_len: int):
```

echopype/qc/api.py
```diff
@@ -9,32 +9,34 @@
...
+    # create new array that preserves differences, but enforces increasing vals
+    new_time = time_old.copy()
+    new_time[neg_idx[0] + 1 :] = new_time[neg_idx[0]] + c_diff
+
+    return new_time
 
 
 def coerce_increasing_time(
-    ds: xr.Dataset, time_name: str = "ping_time", local_win_len: int = 100
+    ds: xr.Dataset, time_name: str = "ping_time", win_len: float = 100
```
emiliom, line 39: Same type change as in `_clean_reversed()`
```suggestion
    ds: xr.Dataset, time_name: str = "ping_time", win_len: int = 100
```

echopype/tests/qc/test_qc.py
```diff
@@ -0,0 +1,82 @@
...
+            np.array([0,2,4,6,8,10,12,14,18,22,26,28,30,32,36,40,44,47,51,55,63]).astype("datetime64[ns]"),
+        ),
+    ],
+    ids=[
+        "win_len2",
+        "win_len6"
+    ]
+)
+def test__clean_reversed(local_win_len, input_arr, expected_arr):
+    arr_fixed = _clean_reversed(input_arr, local_win_len)
```
emiliom, line 60: Cosmetic change, just to use an argument name consistently with its new name in `_clean_reversed`:
```suggestion
def test__clean_reversed(win_len, input_arr, expected_arr):
    arr_fixed = _clean_reversed(input_arr, win_len)
```

emiliom:
> Need to decide if to keep `create_old_time_array` and `orchestrate_reverse_time_check`, and then to revise or deprecate: pinging @emiliom for thoughts

It looks like they're not currently being used anywhere, including in tests. But the use case for `orchestrate_reverse_time_check` still seems relevant, so it seems like a shame to get rid of it. I didn't go through the two functions, though, to see if they need to be updated.

emiliom (commented):
I've gone through it. I follow the logic in `_clean_reversed`, and it looks fine. The new CI tests passed. My remaining recommendations vis a vis the actual code changes are another type-hint change and a small tweak.

leewujung:
> > Need to decide if to keep `create_old_time_array` and `orchestrate_reverse_time_check`, and then to revise or deprecate: pinging @emiliom for thoughts
> 
> It looks like they're not currently being used anywhere, including in tests. But the use case for `orchestrate_reverse_time_check` still seems relevant, so it seems like a shame to get rid of it. I didn't go through the two functions, though, to see if they need to be updated.

These functions are not being used at the moment because `combine_echodata` right now does not do the reversed timestamp check automatically -- this way the data is preserved as from the raw data.

I pinged on these because the function `create_old_time_array` is to create a variable in the Provenance group to preserve the change made to the data once `orchestrate_reverse_time_check` is applied to a combined echodata object. I have not tried to work on these functions, since if we don't want to keep time there's no point of spending time on them. `orchestrate_reverse_time_check` will require a check and likely boost for efficiency when the combined echodata zarr stores are big. `create_old_time_array` is in the provenance domain -- what should be kept and what should be added to the Provenance group to note the change.

emiliom:
I've looked at `create_old_time_array` and `orchestrate_reverse_time_check`. I haven't tried to run them yet, but I could do that -- especially if we can identify test data that include at least one reversed-time case.

The functions seem to do what's expected. The main thing I'm puzzled about is why `orchestrate_reverse_time_check` is writing the results to both an in-memory `EchoData` object and a zarr store. I don't see a reason for it. Could it be a remnant of the old combine data scheme that used the old `ZarrCombine` object to keep memory use in check? I also may modify how the parameter `possible_time_dims` is handled, to both do more error checking and allow for processing all time dimensions (they can be identified based on the variable attribute `standard_name="time"`).

On a smaller note, in `create_old_time_array` the creation of a dataarray copy seems unnecessary: `old_time = old_time_in.copy()`. I don't see anything that might result in changing the source dataarray.

I also have some suggestions for improving readability and intent (changes to variable names, docstrings, etc), but they're all cosmetic.

More broadly, do these functions provide a valuable functionality? To the extent that we want to enable the retention of source and provenance information, I'd say yes. They could be optional features, based on arguments to `combine_echodata`, say: `correct_reversed_time=False, store_time_reversal_provenance=False`. But that would be best addressed in a separate PR. If we keep the functions, we should verify that they work and fix them if not.

leewujung:
I like the idea of adding `correct_reversed_time=False, store_time_reversal_provenance=False` options! Also agree with you on many of the issues and likely improvements of the current code. One thing we know is that with the current files generated by `combine_echodata` (which does not require time checking), any subsequent operation requires using the `ping_time` dimension will fail if the original combined data contains reversed time time.

I agree with putting these into a separate issue. It would be a good idea to add these soon, so that if folks do use `combine_echodata` _and_ there is reversed timestamp, subsequent processing would run through.

I will resolve the conflict and merge this now, and create an issue for adding `correct_reversed_time=False, store_time_reversal_provenance=False` options.

PR #1067, opened 2023-06-24
Handling problematic NMEA messages when setting lat/lon in Platform group

leewujung:
We just encountered a few files that contain problematic NMEA strings in the GPS lat/lon messages. This PR adds error handling and set those entries to NaN. The ill-formed NMEA messages look like below:

Screenshot 2023-06-23 at 6 47 40 PM

### DIFF
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -1,6 +1,6 @@
 import abc
-from typing import List, Set, Tuple
 import warnings
+from typing import List, Set, Tuple
 
 import dask.array
 import numpy as np
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -1,5 +1,6 @@
 import abc
 from typing import List, Set, Tuple
+import warnings
 
 import dask.array
 import numpy as np
@@ -179,11 +180,19 @@ class SetGroupsBase(abc.ABC):
         for x in nmea_msg:
             try:
                 lat.append(x.latitude if hasattr(x, "latitude") else np.nan)
-            except:
+            except ValueError as ve:
+                warnings.warn(
+                    "At least one latitude entry is problematic and "
+                    f"are assigned None in the converted data: {str(ve)}"
+                )
                 lat.append(np.nan)
             try:
                 lon.append(x.longitude if hasattr(x, "longitude") else np.nan)
-            except:
+            except ValueError as ve:
+                warnings.warn(
+                    f"At least one longitude entry is problematic and "
+                    f"are assigned None in the converted data: {str(ve)}"
+                )
                 lon.append(np.nan)
         msg_type = (
             np.array([x.sentence_type if hasattr(x, "sentence_type") else np.nan for x in nmea_msg])
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -176,24 +176,27 @@ class SetGroupsBase(abc.ABC):
                 pynmea2.ParseError,
             ):
                 nmea_msg.append(None)
-        lat, lon = [], []
-        for x in nmea_msg:
-            try:
-                lat.append(x.latitude if hasattr(x, "latitude") else np.nan)
-            except ValueError as ve:
-                warnings.warn(
-                    "At least one latitude entry is problematic and "
-                    f"are assigned None in the converted data: {str(ve)}"
-                )
-                lat.append(np.nan)
-            try:
-                lon.append(x.longitude if hasattr(x, "longitude") else np.nan)
-            except ValueError as ve:
-                warnings.warn(
-                    f"At least one longitude entry is problematic and "
-                    f"are assigned None in the converted data: {str(ve)}"
-                )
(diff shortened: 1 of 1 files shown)


leewujung:
@emiliom: this is something we're running into on the ship. I couldn't put a file up now since it is quite large (100MB) and the internet is slow.

leewujung:
Oops - it's weird that the dimensions are messed up, will try to fix it and ping again. Thanks!

leewujung:
@emiliom : This is ready for you to review! :)

emiliom:
I have a couple of comments / questions, regarding the pattern used for both lat & lon. I'm pasting one of them here for reference, and when I refer to `lat` the same comment will apply to `lon`.

```python
try:
   lat.append(x.latitude if hasattr(x, "latitude") else np.nan)
except:
   lat.append(np.nan)
```

In the original code the `lat` list is converted to a numpy array, but in the new code it isn't. I don't know if there are consequences downstream. But based on the sample output you included, the bad messages (eg, "768$$.$$$") result in strings. The current code would result in `lat` being a numpy array of strings (`object` type?), whereas the new one may result in an inhomogeneous list.

I can't tell how the above code would trigger the `except` clause when a bad message ("768$$.$$$") occurs. It looks like the `lat.append` statement would execute w/o any issues.

I think the new test that needs to be added is for the value of `x.latitude` to be a float; if it isn't, the `lat` list element should be set to `np.nan`. Is that it? If so, here's a code snippet that I think would do that:

```python
def is_number(s):
    try:
        float(s)
        return True
    except ValueError:
        return False

lat.append(x.latitude if hasattr(x, "latitude") and is_number(x.latitude) else np.nan)
```

The last statement would replace the try-except block I pasted at the start of this comment.

Hopefully I'm not off track!

leewujung:
> I can't tell how the above code would trigger the `except` clause when a bad message ("768$$.$$$") occurs. It looks like the `lat.append` statement would execute w/o any issues.

So `hasattr(x, "latitude")` would error out with strings like "768$$.$$$", so it wouldn't make it to `lat.append` in the current code. This is why I made them into the try-catch block. Here `x` is an object from `pynmea2`.

emiliom:
> So `hasattr(x, "latitude")` would error out with strings like "768$$.$$$", so it wouldn't make it to `lat.append` in the current code. This is why I made them into the try-catch block. Here `x` is an object from `pynmea2`.

Hmm. Why would `hasattr(x, "latitude")` *ever* error out? If the object `x` doesn't have a "latitude" attribute, it should just return `False`. Even `None` returns `False`: `hasattr(None, "latitude")`. So, that behavior is puzzling. But, if the error catching step you added does the trick, then that's that.

leewujung:
I looked into the `pynmea2` code. The reason why `hasattr(x, "longitude")` errors out is because both `latitude` and `longitude` are properties and the conversion to number is only run when the property is accessed. So I think the error catching step is probably going to be the easiest for us. I am open to suggestions though!

emiliom:
> I  looked into the pynmea2 code. The reason why hasattr(x, "longitude") errors out is because both latitude and longitude are properties and the conversion to number is only run when the property is accessed. So I think the error catching step is probably going to be the easiest for us. I am open to suggestions though!

Ah. That makes sense now. Yeah, the error catching approach seems like the way to go.

Longer term, we should consider being good open source citizens and submitting a PR to `pynmea2` so it doesn't fail when it encounters this case (and returns None when the string is not convertible to a number?). I might volunteer to do that.

> And I added some warning messages so that users will be aware of this happening.

I think the warning will be emitted every single time a bad lat or lon value is encountered. If so, I'd change the wording to, say:
```python
"The latitude entry is problematic and "
f"is assigned None in the converted data: {str(ve)}"
```
and replace None with numpy.nan?

Feel free to self merge once you finalize the warning message. Thanks!

Oh, one last thing. Earlier I made this comment: "In the original code the lat list is converted to a numpy array, but in the new code it isn't. I don't know if there are consequences downstream." It looks like you've been running your fix and I assume it runs fine, so that must mean that the original conversion to a numpy array wasn't necessary.

leewujung:
> Longer term, we should consider being good open source citizens and submitting a PR to `pynmea2` so it doesn't fail when it encounters this case (and returns None when the string is not convertible to a number?). I might volunteer to do that.

I completely agree. And in fact I was wondering fetching lat/lon at run time may not necessarily be the most ideal operation scenario, but there may be other reason the code is set up that way.


> I think the warning will be emitted every single time a bad lat or lon value is encountered.

For the default case it would raise the warning only at the first instance, so I decided to keep it simple (with the "at least one" phrasing).


> Oh, one last thing. Earlier I made this comment: "In the original code the lat list is converted to a numpy array, but in the new code it isn't. I don't know if there are consequences downstream." It looks like you've been running your fix and I assume it runs fine, so that must mean that the original conversion to a numpy array wasn't necessary.

Yes I guess so! And I just removed that conversion to numpy too so that it stays as a list, and seems like it runs through without problem too (locally). I'll push that now and if things all run through I'll merge this PR.

Thanks for the thorough review!

PR #1068, opened 2023-06-25
Fix call to register EK500 colormap

leewujung:
This PR fixes the call to register the EK500 colormap per the recent deprecation warning:
> The register_cmap function was deprecated in Matplotlib 3.7 and will be removed two minor releases later. Use ``matplotlib.colormaps.register(name)`` instead.

### DIFF
--- a/echopype/visualize/cm.py
+++ b/echopype/visualize/cm.py
@@ -68,7 +68,7 @@ for cmapname in cmapnames:
 
     if 'over' in colors_d:
         reg_map.set_over(colors_d['over'])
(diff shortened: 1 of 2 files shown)


lsetiawan (approved):

lsetiawan:
This is great! Thanks! Could you pin the matplotlib to be version `>=3.7`?

leewujung:
Pinned. Let me know if we actually need space around the >=...

leewujung:
Ok i've tested `pip install -e ".[plot]"` and it does work. I'll merge this now.

Issue #1053, opened 2023-05-31
use_swap= True,  TypeError: Object of type int32 is not JSON serializable


rolandproud:
I have tried use_swap = True on a number of different files (EK60/EK80/ES70) and everytime I get the same error:

ek = ep.open_raw(raw_file_name, sonar_model= self.raw_file_type,use_swap = True)
  File "C:\Users\proud\AppData\Roaming\Python\Python39\site-packages\echopype\utils\prov.py", line 215, in inner
    dataobj = func(*args, **kwargs)
  File "C:\Users\proud\AppData\Roaming\Python\Python39\site-packages\echopype\convert\api.py", line 442, in open_raw
    p2z.datagram_to_zarr(max_mb=max_mb)
  File "C:\Users\proud\AppData\Roaming\Python\Python39\site-packages\echopype\convert\parsed_to_zarr_ek60.py", line 289, in datagram_to_zarr
    self._write_power(df=self.datagram_df, max_mb=max_mb)
  File "C:\Users\proud\AppData\Roaming\Python\Python39\site-packages\echopype\convert\parsed_to_zarr_ek60.py", line 89, in _write_power
    self.write_df_column(
  File "C:\Users\proud\AppData\Roaming\Python\Python39\site-packages\echopype\convert\parsed_to_zarr.py", line 388, in write_df_column
    self.write_chunks(pd_series, zarr_grp, is_array, chunks, chunk_shape)
  File "C:\Users\proud\AppData\Roaming\Python\Python39\site-packages\echopype\convert\parsed_to_zarr.py", line 326, in write_chunks
    full_array = zarr_grp.array(
  ...

TypeError: Object of type int32 is not JSON serializable

Are you experiencing the same problem?

emiliom:
I don't think we've run into this. Can you report the version of echopype you are using and how you installed echopype (pip or conda)? Also, can you share a small file that produces this error?

leewujung:
Since the entire mechanism of `use_swap` is being refactored and we've not heard back on this, we'll close it for now. @rolandproud : if you run into the same issue later after the next release (v0.8.2) feel free to reopen this.

Issue #795, opened 2022-08-25
Account for RAW4 datagrams in parsed to zarr method


b-reyes:
In PR #794, it was found that RAW4 datagrams are not being correctly handled in the parsed to zarr (P2Z) method introduced in PR #774. 

The issue was discovered after running the EK80 file `raw4-D20220514-T172704.raw`. The issue with this file is the RAW4 datagram produces receive ping data and the P2Z method did not adequately address this. To fix this issue, we just need to rectangularize this data and the method will work. However, after reviewing the data, it appears that this data can be quite large and it is probably best to directly write it to a zarr too.

Once the RAW4 datagrams are correctly accounted for in the P2Z method, we should then add `raw4-D20220514-T172704.raw` to the integration tests.

Issue #777, opened 2022-08-10
Add unit tests for methods that directly write variables to zarr


b-reyes:
In PR #774 we added the functionality that allows users to directly write variables with a large memory footprint directly to a temporary zarr store. Currently, we have only applied this new functionality to a couple of files. Although these files produce the expected output, a suite of unit tests need to be completed to ensure this addition is behaving as expected.

leewujung:
I'm pushing this to v0.7.1 since it's related to refactoring parsed2zarr.

PR #1070, opened 2023-06-28
refactor(convert): refactor and cleanup parsed2zarr

lsetiawan:
## Overview

This PR refactors and cleans up some pieces for parsed2zarr functionalities, which tries to address the tasks within #966 

Pasting the checklist here:
- [x] #795 
- [x] #782
- [ ] #777
- [x] refactor/remove `_get_complex_dataarrays` and `_get_angle_dataarrays` that are currently arbitrarily placed in `set_groups_base.py`
- [x] ~#1053~ **Note: need more details**
- [x] Remove `use_swap` arg to `open_raw`
- [x] Add `destination_path` with default `None` with an optional argument of `"swap"` which essetially make `use_swap=True`.
- [x] Check destination path so that it requires `storage_options` when it's remote file... probably already have this feature somewhere!


echopype/convert/api.py
```diff
@@ -342,11 +345,21 @@ def open_raw(
...
     use_swap: bool
+        **DEPRECATED: This flag is ignored**
         If True, variables with a large memory footprint will be
         written to a temporary zarr store at ``~/.echopype/temp_output/parsed2zarr_temp_files``
+    destination_path: str
+        The path to a swap directory in a case of a large memory footprint.
+        This path can be a remote path like s3://bucket/swap_dir.
+        By default, it will create a temporary zarr store at
+        ``~/.echopype/temp_output/parsed2zarr_temp_files`` if needed,
+        when set to "auto".
```
leewujung, line 359: Suggest the following wording changes. Also, from the description `destination_path="auto"` is the default, but in the code the default is `None`. Which one is it?

```suggestion
        The path to a swap directory when the memory footprint is large.
        This path can be a remote path like s3://bucket/swap_dir.
        When ``destination_path="auto"`` (default),
        a temporary zarr store at ``~/.echopype/temp_output/parsed2zarr_temp_files``
        will be created when necessary.
```

echopype/convert/api.py
```diff
@@ -369,10 +382,23 @@ def open_raw(
...
     -----
-    ``use_swap=True`` is only available for the following
+    In a case of a large memory footprint, the program will determine if using
+    a temporary swap space is needed. If so, it will use that space
+    during conversion to prevent out of memory errors.
+    Users can override this behaviour by either passing ``"swap"`` or ``"no_swap"``
+    into the ``destination_path`` argument.
+    This feature is only available for the following
     echosounders: EK60, ES70, EK80, ES80, EA640. Additionally, this feature
     is currently in beta.
```
leewujung, line 392:
```suggestion
    When the required memory footprint is large, the program will determine if using
    a temporary swap space is needed during conversion to prevent out of memory errors.
    Users can override this behavior by setting ``destination_path="swap"``
     or ``destination_path="no_swap"``.
    This feature is only available for the following
    echosounders: EK60, ES70, EK80, ES80, EA640. 
    This is a beta feature.
```

echopype/convert/api.py
```diff
@@ -369,10 +382,23 @@ def open_raw(
...
     """
+
+    # Initially set use_swap False
+    use_swap = False
```
leewujung, line 396: since `use_swap` is set explicitly in the if-else below, I think we should remove this to avoid this implicit default creeping in and causing bugs.

echopype/convert/api.py
```diff
@@ -423,29 +440,49 @@ def open_raw(
...
+            "swap": True,
+            "no_swap": False,
+        }
+        if destination_path == "auto":
+            # Overwrite use_swap if it's True below
+            # Use local swap directory
+            use_swap = should_use_swap(parser.zarr_datagrams, dgram_zarr_vars, mem_mult=0.4)
+        elif destination_path in swap_map:
+            use_swap = swap_map[destination_path]
+        else:
```
leewujung, line 464: Suggest adding a line of comment that this is the case when a local or cloud path is given for readability.

leewujung:
Notes for myself before finishing review all changes: my comments to the previous version: https://github.com/OSOceanAcoustics/echopype/pull/853#pullrequestreview-1143144157. I went as far as implementing it but that did not get merged.

leewujung:
The few lines below currently under `rectangularize_transmit_ping_data` IMO should be absorbed into `rectangularize_data` to make the code more readable. I don't see obvious reason why this should be a separate function.
https://github.com/OSOceanAcoustics/echopype/blob/a4fa9f0089086b6bb1a24cb381e99c941ecb9813/echopype/convert/parse_base.py#L114-L119

leewujung:
Hey @lsetiawan thanks for looking into this. I am not convinced that keeping the whole parsed2zarr class is the right approach. It almost seem unnecessary and the whole thing can be replaced by writing part of the lists sequentially to the same zarr store, without directly using the zarr library

The reason why i think the current complicated approach is unnecessary is because there are 2 stages of padding that is happening:
1. the "padding" for each channel to become rectangular (so to pad the shorter pings so that all pings within a channel have the same length) is in `pad_shorter_ping`. That is the first NaN-padding expansion.
2. The second NaN-padding expansion is when some channels lack a certain type of data, so NaN is padded for the entire "slice" of that channel to match the size of the other channels.

Expansion 1 happens for all data without xarray/zarr/dask (which can potentially be sped up or make more efficient?). After that, if we take note of the shape of variables in `self.ping_data_dict["angle"][ch_id]` for each `ch_id` , we would know exactly whether or not expansion 2 is needed. There, we can decide what is the strategy to use to write to the zarr store -- can we append sequentially along segments of `ping_time` since we know all the pings are same length? and for the channels that lack any data at all, creating a NaN slice for each sections (along `ping_time` segments) would likely let us hover under the memory limit.

lsetiawan:
This currently works, merging... there will be a new issue regarding the comment above in a full clean up of the code.

Issue #1059, opened 2023-06-09
Remove `ui_param` EchoData property


emiliom:
The `ui_param` EchoData property is a remnant from an old code pattern that's no longer used. Remove it.

See https://github.com/OSOceanAcoustics/echopype/pull/1058#discussion_r1223433158

leewujung:
Closing as this is addressed in #1071.

PR #1071, opened 2023-07-05
Remove `ui_param`

leewujung:
This PR addresses #1059 by removing `ui_param` from the code base.

This is pretty minor since we really do not use it anywhere anymore.


emiliom (approved):
Looks good. You've captured all the occurrences of ui_param. Thanks!

PR #1072, opened 2023-07-09
Fix object dtype and cast to string

lsetiawan:
## Overview

This PR fixes issues with some variable being an `object` data type, which caused issues during file writing. This PR defaults the `object` data type to be casted to string.

### DIFF
--- a/echopype/echodata/echodata.py
+++ b/echopype/echodata/echodata.py
@@ -14,7 +14,7 @@ from zarr.errors import GroupNotFoundError, PathNotFoundError
 if TYPE_CHECKING:
     from ..core import EngineHint, FileFormatHint, PathHint, SonarModelsHint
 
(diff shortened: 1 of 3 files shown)


leewujung (approved):
@lsetiawan : thanks for the quick turn-around! I pushed a commit to add `cal_channel_id` into `EXPECTED_VAR_DTYPE` since this var will always be of the same type as `channel`. I think it is good that you added the block to catch other unexpected ones -- I wonder what the xarray rules are to determine when strings are stored as objects or strings? 

Also it is great that `sanitize_dtypes` is now in `echodata.__get_dataset`!

Feel free to merge once the tests run through!

leewujung:
I am going to merge this so that this change gets into `dev`, and we can further debug if other issues come up.

Issue #1055, opened 2023-06-05
Change most AZFP Vendor-specific attributes to data variables


leewujung:
Many of the current Vendor-specific group attributes for AZFP can (and should) be changed to data variables. This is something discovered while @lsetiawan was working on #1042. The code changes in #1042 should handle the new data variables without problem, so it is a matter of putting these changes in place.

This is a leftover from #1042.

lsetiawan:
@aniketfadia96 I've assigned you to this, figured this would get you some experience with xarray. Please consult @leewujung for direction. Thanks!

aniketfadia96:
@lsetiawan Okay I'll look into this. Thank you!

lsetiawan:
To let you know about some vocabulary:

AZFP is the instrument type of echosounder and Vendor-specific is a group within the set of datasets. This group is essentially an [xarray dataset object](https://docs.xarray.dev/en/stable/user-guide/data-structures.html#dataset).

These attributes are assigned in https://github.com/OSOceanAcoustics/echopype/blob/main/echopype/convert/set_groups_azfp.py#L329. Currently there are actual data within the attributes of this dataset, but we'd like to make it into `data_vars` or [DataArray objects](https://docs.xarray.dev/en/stable/user-guide/data-structures.html#dataarray) instead.

leewujung:
I'll pick up this issue because we need this for the upcoming release and for generating a CDL for the [sonarnetcdf4-echopype-checker](https://github.com/OSOceanAcoustics/sonarnetcdf4-echopype-checker).

aniketfadia96:
Hey @leewujung @lsetiawan, sorry I couldn't work on this issue. I was occupied with my primary project.

leewujung:
Addressed in #1075 .

Issue #1045, opened 2023-05-19
Add option to save configuration XML separately during `open_raw`


leewujung:
Currently when doing `open_raw`, we save a copy of the Simrad echosounder configuration XML into an attribute in the `Vendor-specific` group. This is pretty messy, especially for `combine_echodata`. 

Since most, if not all (need some careful checks), of the info in the XML files are already captured in the converted file, I propose we do the following:
- default to not saving the XML when doing `open_raw`
- add an option in `open_raw` to save an external xml with the same filename as the raw file

leewujung:
@emiliom and I discussed this further and decided that still keeping the configuration XML in the Vendor-specific group is a good idea, as there are precedences. One of the main reasons is that when one file becomes two in the processing steps, one can easily get dropped.

The only change (made in #1075) is that the config XML is now saved as a data variable instead of an attribute so that it does not interfere with lazy-loading zarr files.

Issue #640, opened 2022-04-21
Revisit AZFP unprocessed variables currently in Vendor group


emiliom:
AZFP has several variables that are unprocessed (eg, sensor signals) or have no analog in SONAR-netCDF4 v1. In echopype v0.6.0 we're moving many of these variables from the `Beam_group1` group (formerly `Beam`) into the `Vendor` group. See #642 and https://github.com/OSOceanAcoustics/echopype/issues/520#issuecomment-1104626403520#issuecomment-1104626403.

This issue is a reminder to revisit these variables *eventually*, especially to explore the possibility of processing currently unprocessed raw variables such as `temperature_counts`.

leewujung:
@emiliom : Now that #1075 is merged. Could you do this along with https://github.com/OSOceanAcoustics/sonarnetcdf4-echopype-checker/issues/5#event-9801811799 ? I'll get the AZFP vendor group nc files to you shortly. Thanks!

emiliom:
> especially to explore the possibility of processing currently unprocessed raw variables such as `temperature_counts`.

@leewujung I don't remember what this was about, specifically. Do you? We're already computing `temperature` per the AZFP equations, which start with counts.

I'm going over other variables in the `Vendor_specific` group.

emiliom:
### temperature_counts

So, neither of us remembers the context for the `temperature_counts` comment. I don't remember if it came from me alone or a conversation between us! But I think we can ignore it. temperature is already calculated in `_compute_temperature` based on what looks like the index-4 "ancillary" parameter, which is a count:
https://github.com/OSOceanAcoustics/echopype/blob/6fc5a1deae69b23d6feb3b46abb06ca6816c7403/echopype/convert/parse_azfp.py#L164
And index-4 is also the ancillary parameter written out to `Vendor_specific.temperature_counts`:
https://github.com/OSOceanAcoustics/echopype/blob/6fc5a1deae69b23d6feb3b46abb06ca6816c7403/echopype/convert/set_groups_azfp.py#L478

So, case closed, I think.

### Other variables

As for other variables in `Vendor_specific` that we could promote to another group, I looked at the list of variables and nothing strikes me as matching `Beam_group1` or `Platform` M or MA variables we're not currently populating. But I'm not in a good position to make that determination, especially when none of the `Vendor_specific` variables have attributes, and checking against the manual would be pretty tedious.

[...]

PR #1075, opened 2023-07-12
Move some Vendor-specific group attributes to data variables

leewujung:
This PR moves some Vendor-specific group attributes to data variables to make it more efficient to lazy-load and combine echodata objects, and potentially on the writing speed to cloud. This addresses #1055 and potentially #640 (AZFP) and #1045 (though the goal changed, EK80).

The changes include:
- EK80: move `xml_config` which is the entire XML string from an attribute to a data variable
- AZFP: move all Vendor-specific group attributes to data variables
  - some of these variables come from the XML associated with the 01A file
  - some of these variables come from the unpacked ping by ping data
  - a common feature of these variables is that they are *not* time-varying (which is probably why we saved them as attributes in the first place)

I've tested `combine_echodata` with the new format and these are combined correctly -- that all these variables remain data variables without a dimension (constants). The combine would error out if any of these variables change across the echodata objects to be combined, which is the desired behavior since those echodata objects should not be combined.

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -473,18 +473,26 @@ class SetGroupsAZFP(SetGroupsBase):
                 # parameters with channel dimension from XML file
                 "XML_transmit_duration_nominal": (["channel"], tdn),  # tdn comes from parameters
                 "XML_gain_correction": (["channel"], parameters["gain"][self.freq_ind_sorted]),
-                "XML_digitization_rate": (["channel"], parameters["dig_rate"][self.freq_ind_sorted]),  # noqa
-                "XML_lockout_index": (["channel"], parameters["lockout_index"][self.freq_ind_sorted]),  # noqa
+                "XML_digitization_rate": (
+                    ["channel"],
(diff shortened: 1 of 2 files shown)


emiliom (commented):
Looks good! Things are so nice and simple when you don't have to worry about having variable attributes :laughing: . I like how you organized the variable assignments into blocks with a helpful but brief comment line.

Just one question: I assume that these changes have no downstream effect on calibration functions, since there are no changes to that code?

leewujung:
> Just one question: I assume that these changes have no downstream effect on calibration functions, since there are no changes to that code?

Oh right, this completely slipped my mind. The ones that I can think of on top of my head are the conversion for temperature and tilt_x/tilt_y, from the raw count combining with the coefficients in the variables (`_compute_tilt`, `_compute_temperature`). I looked them up and luckily it actually uses the parsed data dict instead of the echodata vendor-specific group entries for that, since the conversion is in the parser. So I guess the changes are not used anywhere. I'll take another look on the ones I moved, just to be sure.

leewujung:
Ok, confirmed -- none of the attributes that I moved to be data variables are used in later calculations. :)

Issue #1076, opened 2023-07-17
Remove deprecated `preprocess` subpackage


emiliom:
When we deprecated the `preprocess` subpackage by moving its functionality into the new `commongrid` and `clean` subpackages (see #889), the plan was to remove `preprocess` with version 0.7.1. Our plan now is to remove it in [0.8.0, which will be released soon](#1063).

Also, we inadvertently left "preprocess" in the [run-test.py](https://github.com/OSOceanAcoustics/echopype/blob/dev/.ci_helpers/run-test.py#L36), which led to running many tests twice! I'll submit a PR shortly to address this.

- [x] Remove "preprocess" from `run-test.py`. Addressed in #1077
- [x] Remove `preprocess` subpackage

leewujung:
This can now be closed!

PR #1077, opened 2023-07-17
Remove `preprocess` from subpackages to test in run-test [all tests ci]

emiliom:
See #1076.

### DIFF
--- a/.ci_helpers/run-test.py
+++ b/.ci_helpers/run-test.py
@@ -33,7 +33,6 @@ MODULES_TO_TEST = {
     "echodata": {},
     "mask": {},
     "metrics": {},
-    "preprocess": {},
     "utils": {},
     "visualize": {},
 }


lsetiawan:
This is great! I'm glad it's pretty straightforward. I am wondering however, has this submodule been removed? The main CI do run all the tests using `pytest` directly rather than using the `run-test.py` script.

emiliom:
The preprocess subpackage is not removed. All the preprocess tests were removed back when preprocess was deprecated.

`run-test.py` is still being used in `pr.yml`:
https://github.com/OSOceanAcoustics/echopype/blob/e11c665b8098d1badf213cb1a326afaa92d12f67/.github/workflows/pr.yaml#L89-L93

Am I missing something?

emiliom:
BTW, I forgot to use "[all tests ci]" in the PR title, so it looks like most of the tests didn't run here. I added it after the fact  here, but it looks like a title edit doesn't spawn a re-run of tests.

lsetiawan:
> BTW, I forgot to use "[all tests ci]" in the PR title, so it looks like most of the tests didn't run here. I added it after the fact here, but it looks like a title edit doesn't spawn a re-run of tests.

Yea, I haven't found a way to trigger actions when changing the title, but closing and opening seem to always, work and that's what I did. Seems like now the proper action step is running.

emiliom:
> Yea, I haven't found a way to trigger actions when changing the title, but closing and opening seem to always, work and that's what I did. Seems like now the proper action step is running.

Oooh, great trick, thanks! I'll remember that.

lsetiawan (approved):

Issue #1057, opened 2023-06-06
Remove redundant `ping_time` dimension


leewujung:
Similar to #978, the convention requires the `beam` dimension for many variables, but for the echosounder models we support many of these variables do not change over `ping_time` during a deployment or survey, since they are properties of the transducers. This dimension was artificially created in the current code.

Since `ping_time` is a major dimension that data is being concatenated on, this repetition is costly when data is loaded into memory (though it compresses well). I think we should consider removing this redundant dimension.

@emiliom : tagging you so that it goes with work in #1056.

leewujung:
Addressed in #1083.

PR #1083, opened 2023-07-22
Drop artificially created `ping_time` dimension

leewujung:
This PR addresses #1057 by removing `ping_time` dimension artificially added to some variables in `set_groups_*`.

The PR also adds comments on sections of test functions where explicit `ping_time` dim drop is required.

### DIFF
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -29,23 +29,14 @@ class SetGroupsEK60(SetGroupsBase):
     # the SONAR-netCDF4 v1 convention. For the time being, we are retaining the
     # infrastructure that adds this dimension, but updating the variables lists.
 
+    # 2023-07-21 note: We are also removing the ping_time dimension for parameters
+    # that do not change across ping.
+
     # Variables that need only the beam dimension added to them.
     beam_only_names = set()
 
     # Variables that need only the ping_time dimension added to them.
-    ping_time_only_names = {
-        "beam_direction_x",
-        "beam_direction_y",
-        "beam_direction_z",
-        "beamwidth_twoway_alongship",
-        "beamwidth_twoway_athwartship",
-        "angle_offset_alongship",
-        "angle_offset_athwartship",
-        "angle_sensitivity_alongship",
-        "angle_sensitivity_athwartship",
-        "equivalent_beam_angle",
-        "gain_correction",
-    }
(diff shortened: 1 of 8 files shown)


echopype/tests/calibrate/test_cal_params_integration.py
```diff
@@ -79,6 +79,8 @@ def test_cal_params_intake_EK60(ek60_path):
     cal_obj = ep.calibrate.calibrate_ek.CalibrateEK60(echodata=ed, env_params=None, cal_params=None, ecs_file=None)
 
     # Check cal params ingested from both ways
+    # Need to drop ping_time for cal_obj.cal_params since get_vend_cal_params_power are related to
```
emiliom, line 82: Not so much being picky as confirming that something isn't missing in this comment.
```suggestion
    # Need to drop ping_time for cal_obj.cal_params since get_vend_cal_params_power is related to
```

echopype/convert/set_groups_ek60.py
```diff
@@ -29,23 +29,14 @@ class SetGroupsEK60(SetGroupsBase):
     # the SONAR-netCDF4 v1 convention. For the time being, we are retaining the
     # infrastructure that adds this dimension, but updating the variables lists.
 
+    # 2023-07-21 note: We are also removing the ping_time dimension for parameters
```
emiliom, line 32: This comment seems unnecessary. Otherwise an analogous comment should be added to the EK80 and AZFP modules, and for the removal of the beam dimension.

emiliom (commented):
Looks good! My comments are only about some of the comments you added.

echopype/convert/set_groups_ek60.py
leewujung: I added this because you added something like that just above this line. I was thinking this should actually be in set_groups_base. I'll move both comments to the leftover mechanism (which is entirely bypassed at this point).

leewujung:
Sound good. I think I've addressed your comments and will merge after the tests run through (though I did not make any code changes in the new commits).

echopype/convert/set_groups_ek60.py
emiliom: Did I? Sigh. It may have been a temporary note to self that I forgot to remove. I agree that it makes more sense to have something like that in `set_groups_base` or elsewhere. Thanks.

Issue #1081, opened 2023-07-20
Rename frequency_start/end to transmit_frequency_start/stop


emiliom:
Beam_groupX `frequency_start` and `frequency_end` are only used by EK80. Rename them to the equivalent v1 convention names, `transmit_frequency_start` and `transmit_frequency_stop`

emiliom:
I've examined the current usage of `frequency_start` and `frequency_end` and the definitions of the corresponding `transmit_frequency_start` and `transmit_frequency_stop` in the convention. I have some observations and questions.

In the convention (both v1 and v2), the two variables are defined as Mandatory. In echopype, they are only used (present) with EK80, and within EK80 only for BB data and apparently assigned valid values only for a subset of channels. For example, in one EK80 file I've been examining (`ek80/ek80_bb_with_calibration/2018115-D20181213-T094600.raw`), the `frequency_start/end` variables are assigned valid values only in 2 of the 6 channels, while the other 4 are assigned `nan`. 

My interpretation is that these variables are simply unnecessary (redundant) for data other than BB data (though I don't understand the case I described above). In other words, for EK60, AZFP and EK80 power data, one could set `transmit_frequency_start` = `transmit_frequency_stop` = `frequency_nominal`. 

If that's correct, I recommend always creating these variables. That would be more consistent with the convention, since the variables are mandatory (their description is: "Frequency at the start/end of the transmit pulse."). If we do this, there'll be an additional change needed. Currently the presence of `frequency_start` and/or `frequency_end` in `Beam_group1` is used as a test and decision point for further processing steps; in some cases, the test is for `nan` values instead. For example:
https://github.com/OSOceanAcoustics/echopype/blob/ed79cd82332f177e063d1b1341bb9d3f1bb5f734/echopype/calibrate/calibrate_ek.py#L312-L315

If the variables are set to always be present, these checks would  need to be updated. If we assign `nan` values in cases where the variables are not applicable or used, it should be fairly straightforward to modified these checks. Assigning valid values in all cases (`frequency_nominal` instead of `nan`) seems more consistent, but the checks would then have to be done either on a different variable or based on `transmit_frequency_start != transmit_frequency_stop`.

One last observation: the convention defines these variables as `float` type, but echopype sets them as `int` type. Both are defined in Hz units.

leewujung:
### Mixed BB/CW setup

[...]

Issue #951, opened 2023-03-02
Add some variables not currently stored from ping-by-ping data


leewujung:
Currently there are some ping data variables not stored in the converted EK files. These are not used at the moment but could be useful (see below). This issue is a reminder to add them after we have the calibration code is in good shape.

For EK80 each ping contains the following variables:
```python
# data in parser.ping_data_dict
['type', 'low_date', 'high_date', 'channel_id', 'data_type', 'spare', 'offset', 'count', 'timestamp', 'bytes_read', 'complex_dtype', 'n_complex', 'channel_mode', 'pulse_form', 'frequency', 'pulse_duration', 'sample_interval', 'transmit_power', 'slope', 'frequency_start', 'frequency_end']
```

We want to add:
```python
['data_type', 'offset', 'count', 'channel_mode', 'pulse_form']
```
Note:
- `offset` is used for echo range calculation, though so far I've only seen files with offset=0 so currently there is no impact in not saving it
- `pulse_form`: right now in the code we use the presence of `frequency_start/end` to check which ping/channel is transmitting narrowband or broadband signals; probably more convenient to use `pulse_form`, but that could cause backward compatibility issues.

Already stored or captured in other forms:
```python
['type', 'low_date', 'high_date', 'channel_id', 'timestamp', 'frequency', 'pulse_duration', 'sample_interval', 'transmit_power', 'slope', 'frequency_start', 'frequency_end']
```

Those we do not plan to store because they are already captured in other forms include:
```python
['type', 'low_date', 'high_date', 'timestamp', 'bytes_read', 'complex_dtype', 'n_complex']
```

Not sure:
```python
['spare']
```

Tagging @emiliom here for decisions on what to add and where. Note most of the already stored variables are in the `Sonar/Beam_group1` group even though those are not defined in the convention. We need to decide if to change the current implementation, or continue this approach to save those new variables in the `Sonar/Beam_group1` group also.

leewujung:
Related to #1043.

emiliom:
> We want to add: ['data_type', 'offset', 'count', 'channel_mode', 'pulse_form']

- Is there a brief description of  `data_type`, `count` and `channel_mode` somewhere, such as the EK80 manual? That's to create a `long_name` attribute. You've already described the other two variables.
- Since they have no matching convention variable, should we retain the variable names? `data_type`, `offset` and `count` seem overly generic.
[...]

PR #1091, opened 2023-07-26
Standardize use of `transmit_frequency_start/stop` and use `pulse_form` for BB checks

emiliom:
Addresses #1081 (and has linkage to #951). Specifically:

- For EK80
  - [x] Rename `frequency_start/end` to `transmit_frequency_start/stop`
  - [x] Set attributes to convention specification
  - [x] Change data type from `int` to `float`
  - [x] Overhaul checks for narrowband vs broadband (BB) channels to use `transmit_type` / `pulse_form`.
  - [x] For channels that are not `BB`, set values to `frequency_nominal` rather than `nan`
  - [x] Update checks for narrowband vs BB channels once `transmit_frequency_start/stop` are set to always exist and always have valid, non-`nan` values.
- [x] For EK60, add `transmit_frequency_start/stop` with only a `channel` dimension and set their values to `frequency_nominal`
- [x] For AZFP, add `transmit_frequency_start/stop` with only a `channel` dimension and set their values to `frequency_nominal`


emiliom:
Based on https://github.com/OSOceanAcoustics/echopype/issues/951#issuecomment-1652198520, the new checks for broadband channels should use `pulse_form > 0` (0 = "CW"). I couldn't find a definition for "FMD" in the EK80 manual; but I assume it's also BB, just like "FM".

leewujung:
I also don't know what "FMD" is, but in the files we have access to it is either 0 or 1.

emiliom:
I'll add the `pulse_form > 0` BB check to this PR once PR #1094 is merged.

emiliom:
Renamed EK80 `pulse_form` to `transmit_type` and changed to string, type per convention.

Not really part of the original scope of this PR, but this change will need to be in place before the new check for BB based on pulse_form (transmit_type) can be implemented.

leewujung:
I went back to the function _get_chan_dict  to check again and think that:
1. like what we discussed, there is a bug if within the channel there was a switch from BB to CW or from CW to BB
2. `np.unique` is pretty expensive, and the way the code was written, it first uses the existence of `frequency_start/end` to determine if there are BB channels (if `frequency_start/end` both exist then there are BB channels in this `Beam_group1`) (more precise, the operation actually checks for if there exists at least 1 BB ping out of all pings across all channels), and then if there is BB channels, drop channels that have all-NaN `freq_center` (this is the case when the BB to CW or CW to BB switch would not be processed correctly). The code is below for reference:
```python
# Use center frequency for each ping to select BB or CW channels
# when all samples are encoded as complex samples
if "frequency_start" in beam and "frequency_end" in beam:
    # At least some channels are BB
    # frequency_start and frequency_end are NaN for CW channels
    freq_center = (beam["frequency_start"] + beam["frequency_end"]) / 2

    return {
        # For BB: drop channels containing CW samples (nan in freq start/end)
        "BB": freq_center.dropna(dim="channel").channel,
        # For CW: drop channels containing BB samples (not nan in freq start/end)
        "CW": freq_center.where(np.isnan(freq_center), drop=True).channel,
    }
```

For 1: I think we should probably document it but not making any changes for the time being. Most of the time things do not get fiddled in the middle of a file, especially because complex data is pretty huge so the number of pings within a typical file, say 100 MB, is pretty small.

For 2.: What do you think about the following:
```python
if 1 in beam["pulse_form"]:
    # At least 1 BB ping exists -- this is analogous to what we had from before
    # Before: when at least 1 BB ping exists, "frequency_start" and "frequency_end" will exist
    freq_center = (beam["frequency_start"] + beam["frequency_end"]) / 2

    # This part below doesn't require changes (so still buggy)
    return {
        # For BB: drop channels containing CW samples (nan in freq start/end)
        "BB": freq_center.dropna(dim="channel").channel,
        # For CW: drop channels containing BB samples (not nan in freq start/end)
        "CW": freq_center.where(np.isnan(freq_center), drop=True).channel,
    }

```

emiliom:
> 1. like what we discussed, there is a bug if within the channel there was a switch from BB to CW or from CW to BB

Quick note: by coincidence, I discovered that this raw file may be one such case: `TEST_DATA_FOLDER / "ek80/2019118 group2survey-D20191214-T081342.raw"`

emiliom:
> For 2.: What do you think about the following:
```python
if 1 in beam["pulse_form"]:
   # At least 1 BB ping exists -- this is analogous to what we had from before
   # Before: when at least 1 BB ping exists, "frequency_start" and "frequency_end" will exist
```

That'll work. Come to think of it, my use of `np.unique` was unnecessary, and not just in this particular check! 

For this check, though, I know you've said that you haven't encountered the "FMD" (`pulse_form` = 5) type, but are you saying we should omit checking for it? I realize adding the `5 in beam["pulse_form"]` to the `if` condition adds to the computation, but it seems to me that if it's in the Simrad EK80 manual, we should be comprehensive and account for it. No?

Now, back to `np.unique`. There are 5 places, in 4 modules, where a test for BB vs CW occurs, and these checks come in two types: 

**1. The one above from `calibrate_ek.py` (below) and two occurrences, which test for the occurrence of a BB ping:**

https://github.com/OSOceanAcoustics/echopype/blob/eb92185e83577b3e30bdcaf00e3e43ec0ada547a/echopype/calibrate/calibrate_ek.py#L314-L316

https://github.com/OSOceanAcoustics/echopype/blob/eb92185e83577b3e30bdcaf00e3e43ec0ada547a/echopype/echodata/simrad.py#L141-L145

https://github.com/OSOceanAcoustics/echopype/blob/eb92185e83577b3e30bdcaf00e3e43ec0ada547a/echopype/convert/set_groups_ek80.py#L647-L653

**2. A test for the *absence* of BB pings (ie, all CW), with two occurrences:**

https://github.com/OSOceanAcoustics/echopype/blob/eb92185e83577b3e30bdcaf00e3e43ec0ada547a/echopype/echodata/simrad.py#L121-L126

https://github.com/OSOceanAcoustics/echopype/blob/eb92185e83577b3e30bdcaf00e3e43ec0ada547a/echopype/calibrate/ek80_complex.py#L214-L218

I can now see that this test can be handled with the `not in` operator: `1 not in beam[`pulse_form`] and 5 not in beam[`pulse_form`]`.

Ok, I think the path is clear now. Thanks!

leewujung:
> > For 2.: What do you think about the following:
> 
> ```python
> if 1 in beam["pulse_form"]:
>    # At least 1 BB ping exists -- this is analogous to what we had from before
>    # Before: when at least 1 BB ping exists, "frequency_start" and "frequency_end" will exist
> ```
> 
> That'll work. Come to think of it, my use of `np.unique` was unnecessary, and not just in this particular check!
> 
> For this check, though, I know you've said that you haven't encountered the "FMD" (`pulse_form` = 5) type, but are you saying we should omit checking for it? I realize adding the `5 in beam["pulse_form"]` to the `if` condition adds to the computation, but it seems to me that if it's in the Simrad EK80 manual, we should be comprehensive and account for it. No?

Good point. I guess then it'll be the reverse of the previous code, making it like below:
```
if not np.all(beam["pulse_form"] == 0):
```

leewujung:
> > 1. like what we discussed, there is a bug if within the channel there was a switch from BB to CW or from CW to BB
> 
> Quick note: by coincidence, I discovered that this raw file may be one such case: `TEST_DATA_FOLDER / "ek80/2019118 group2survey-D20191214-T081342.raw"`

Oh! ok I'll take a look. Seems that we've not used this file for cal tests, but I suspect that the operations will go through, but the results will be wrong. I can create a new issue for this.

emiliom:
> Good point. I guess then it'll be the reverse of the previous code, making it like below:
> if not np.all(beam["pulse_form"] == 0):

Nicer and cleaner. But wouldn't it be more expensive computationally than this?
```python
1 in beam["pulse_form"] or 5 in beam["pulse_form"]
# or, more compactly:
np.isin(beam["pulse_form"], [1,5]).any()
```

emiliom:
>> Quick note: by coincidence, I discovered that this raw file may be one such case: TEST_DATA_FOLDER / "ek80/2019118 group2survey-D20191214-T081342.raw"
>
> Oh! ok I'll take a look. Seems that we've not used this file for cal tests, but I suspect that the operations will go through, but the results will be wrong. I can create a new issue for this.

That's correct: the operations went through. I was using that file (don't remember why I picked it) in my echopype checker, when I noticed the results.

emiliom:
I've implemented the checks for BB based on `transmit_type` / `pulse_form`. 

The last remaining task is: For channels that are not `BB`, set values to `frequency_nominal` rather than `nan`. Then, the checks for BB may need to receive small updates.

Note that I haven't yet merged into this PR the changes from PR's #1099 and #1096.

emiliom:
Ok, this PR is now complete and ready for review (once the CI completes sucessfully)!

emiliom:
A note: the comments above about ek80 `pulse_form` all use the integer codes (0,1,5). Now, with the changeover to the convention-based `transmit_type`, these are strings: "CW", "FM", "FMD"

echopype/calibrate/ek80_complex.py
```diff
@@ -12,22 +12,19 @@ def tapered_chirp(
     fs,
     transmit_duration_nominal,
     slope,
-    frequency_nominal=None,
-    frequency_start=None,
-    frequency_end=None,
+    transmit_type,
```
leewujung, line 15: seems like `transmit_type` is unused and can be removed

echopype/calibrate/ek80_complex.py
```diff
@@ -229,27 +226,21 @@ def get_transmit_signal(
...
         raise TypeError("File does not contain BB mode complex samples!")
 
     # Generate all transmit replica
     y_all = {}
     y_time_all = {}
+    # TODO: expand to deal with the case with varying tx param across ping_time
+    tx_param_names = [
+        "transmit_duration_nominal",
+        "slope",
+        "transmit_type",
```
leewujung, line 239: In theory this can be removed since it is not used in `tapered_chirp`, but I kinda like having the uniqueness check on `transmit_type`. How about adding the below comment?
```suggestion
        "transmit_type",  # not needed for tapered_chirp but good to check uniqueness
```

echopype/convert/set_groups_ek80.py
```diff
@@ -620,54 +638,44 @@ def _add_freq_start_end_ds(self, ds_tmp: xr.Dataset, ch: str) -> xr.Dataset:
...
-                    ),
-                },
-            )
+        if not np.all(np.array(self.parser_obj.ping_data_dict["pulse_form"][ch]) == 0):
+            freq_start = np.array(self.parser_obj.ping_data_dict["frequency_start"][ch])
+            freq_stop = np.array(self.parser_obj.ping_data_dict["frequency_end"][ch])
+        elif not self.sorted_channel["power"]:
+            freq = self.parser_obj.config_datagram["configuration"][ch]["transducer_frequency"]
+            # freq_start = np.full(self.parser_obj.ping_data_dict["power"][ch], freq)
+            freq_start = np.full(len(self.parser_obj.ping_time[ch]), freq)
```
leewujung, line 649: I like your change. Maybe we can remove the commented out line?

echopype/convert/set_groups_ek80.py
```diff
@@ -909,21 +923,25 @@ def _assemble_ds_common(self, ch, range_sample_size):
...
                     {
-                        "long_name": "Pulse type",
-                        "flag_values": [0, 1, 5],
-                        "flag_meanings": ["CW", "FM", "FMD"],
+                        "long_name": "Type of transmitted pulse",
+                        "flag_values": ["CW", "FM", "FMD"],
+                        "flag_meanings": [
+                            "Continuous Wave",
+                            "Frequency Modulated",
+                            "Frequency Modulated D",
```
leewujung, line 938: Should we change this to "Frequency Modulated user-Defined" or do you think this is extrapolating too far from the documentation about ADCP?

echopype/echodata/simrad.py
```diff
@@ -118,11 +120,12 @@ def _retrieve_correct_beam_group_EK80(
     power_ed_group = None
     complex_ed_group = None
 
+    transmit_type_has_BB = not np.all(echodata["Sonar/Beam_group1"]["transmit_type"] == "CW")
     if waveform_mode == "BB":
         # check BB waveform_mode, BB must always have complex data, can have 2 beam groups
-        # when echodata contains CW power and BB complex samples, and frequency_start
+        # when echodata contains CW power and BB complex samples, and transmit_frequency_start
         # variable in Beam_group1
```
leewujung, line 127:
```suggestion
        # when echodata contains CW power and BB complex samples.
```
I think we can safely remove the rest of the sentence since now `transmit_frequency_start` would always exist.

echopype/echodata/simrad.py
```diff
@@ -139,16 +142,9 @@ def _retrieve_correct_beam_group_EK80(
...
-            if (
-                echodata["Sonar/Beam_group1"]["channel"].size  # total number of channels
-                == echodata["Sonar/Beam_group1"]["frequency_start"]
-                .dropna(dim="channel")["channel"]
-                .size  # number of BB channel
-            ):  # if all channels are BB
+        if encode_mode == "complex" and transmit_type_has_BB:
+            # complex + BB data (all channels are BB)
+            if np.all(echodata["Sonar/Beam_group1"]["transmit_type"] != "CW"):
                 raise ValueError("waveform_mode='CW', but all data are broadband (BB)!")
```
leewujung, line 148: I think there is a logic error in the first if here. `transmit_type_has_BB` says whether or not there is one ping that is not CW, but does not tell us whether there is no CW ping at all (which is the case we want to catch, and is in the second if. I think this can be:
```python
        # Raise error if waveform_mode="CW" but CW data does not exist (not a single ping is CW)
        if encode_mode == "complex" and np.all(echodata["Sonar/Beam_group1"]["transmit_type"] != "CW"):
            raise ValueError("waveform_mode='CW', but all data are broadband (BB)!")
```

echopype/echodata/simrad.py
```diff
@@ -118,11 +120,12 @@ def _retrieve_correct_beam_group_EK80(
     power_ed_group = None
     complex_ed_group = None
 
+    transmit_type_has_BB = not np.all(echodata["Sonar/Beam_group1"]["transmit_type"] == "CW")
```
leewujung, line 123: This is probably an expensive operation across all `ping_time`. Since right now we have this bug that we can only correctly process all-CW channels or all-BB channels (but not CW-BB mixed _within_ the same channel), do you think it's a bad idea to just test for `transmit_type` of the first ping?

echopype/calibrate/calibrate_ek.py
```diff
@@ -309,18 +309,18 @@ def _get_chan_dict(beam: xr.Dataset) -> Dict:
...
             return {
-                # For BB: drop channels containing CW samples (nan in freq start/end)
-                "BB": freq_center.dropna(dim="channel").channel,
-                # For CW: drop channels containing BB samples (not nan in freq start/end)
-                "CW": freq_center.where(np.isnan(freq_center), drop=True).channel,
+                # For BB: Keep only non-CW channels (FM or FMD) based on transmit_type
+                "BB": freq_center.where(beam["transmit_type"] != "CW", drop=True).channel,
+                # For CW: Keep only CW channels based on transmit_type
+                "CW": freq_center.where(beam["transmit_type"] == "CW", drop=True).channel,
             }
```
leewujung, line 323: We do not need `freq_center` here anymore now that `transmit_type` is available.

The `BB: ...` and `CW: ...` selections do not work if there are within-channel changes of CW and FM. In those cases, either only the single elements fits the criteria are kept, or only the single elements fits the criteria are set to nan. For example, if you have such within-channel changes:
Screenshot 2023-08-02 at 5 36 33 PM
Then the following happens:
first selection:
Screenshot 2023-08-02 at 5 37 54 PM
second selection:
Screenshot 2023-08-02 at 5 36 48 PM

I think we can do the following
```python
first_ping_transmit_type = beam["transmit_type"].isel(ping_time=0)  # assume transmit_type identical for all pings
return {
    # For BB: Keep only non-CW channels (FM or FMD) based on transmit_type
    "BB": first_ping_transmit_type.where(first_ping_transmit_type != "CW", drop=True).channel,
    # For CW: Keep only CW channels based on transmit_type
    "CW": first_ping_transmit_type.where(first_ping_transmit_type == "CW", drop=True).channel,
}
```

leewujung (changes requested):
@emiliom : Thanks for the PR! I think there are a couple places that have bugs. Could you look through my comments and change them? 

I think we will have to add a disclaimer in the documentation that the code does not currently handle within-channel FM-CW switches if they are both encoded in complex samples. (I still need to write an issue to keep track of this.)

echopype/calibrate/ek80_complex.py
emiliom: Right, thanks! I'll include this in the next batch of commits.

echopype/calibrate/ek80_complex.py
emiliom: Right! I added it for an initial implementation that used `transmit_type` in `tapered_chirp`, then removed that usage (simplified the code) and forgot to remove the unused `transmit_type`.

Sure, I'll add that comment in the next set of commits.

echopype/convert/set_groups_ek80.py
emiliom: Sigh, thanks. Will do.

echopype/echodata/simrad.py
emiliom: Right, thanks. Will do.

echopype/convert/set_groups_ek80.py
emiliom: I don't know. As I mentioned in https://github.com/OSOceanAcoustics/echopype/pull/1101#issuecomment-1657451416, even the description next to LFMD seems scrambled and unclear. Having the acronyms spelled out here, as far as we can, seemed like progress, but I'm hesitant to go further. Still, these are only attribute values that would have no downstream implications, so they can be easily tweaked in the future.

echopype/echodata/simrad.py
emiliom: I think it's not so much that the logic is in error (or led to errors), but rather that having `transmit_type_has_BB` in the first if is unnecessary and doesn't add anything useful. I now see that I translated the original logic too literally, but didn't take the further step you did. I agree with your suggestion and will implement it. Thanks! FYI, I'll also remove the variable `transmit_type_has_BB` since it will no longer serve the purpose of reusability.

echopype/echodata/simrad.py
emiliom: Surprisingly (at least to me), it's no more expensive than the `"CW" in dataarray` scheme we adopted elsewhere. See the tests I reported above, https://github.com/OSOceanAcoustics/echopype/pull/1091#issuecomment-1656841321

> Since right now we have this bug that we can only correctly process all-CW channels or all-BB channels

That's true, but the test is done across *all* channels, not for an individual channel. So, while the first ping tells us about the channel it corresponds to, it doesn't tell us anything about other channels.

emiliom: On second thought, maybe I wasn't seeing the multi-dimensional aspect of the data .... The "first" ping is the first ping in *each* channel. In that case, I guess you're right.

echopype/calibrate/calibrate_ek.py
emiliom: I follow the logic and it seems like it'd work. But I ran into multiple errors when testing it. I'll have to go into debug mode on it tomorrow morning, to look at it more carefully.

echopype/calibrate/ek80_complex.py
emiliom: I take this back. `transmit_type` can't remain in `tx_params` b/c the `tx_params` dict is passed to `tapered_chirp`, but `transmit_type` is no longer an argument in that function.

echopype/calibrate/calibrate_ek.py
emiliom: Alright, I found and fixed the problem. The first statement in your suggestion retained the `ping_time` variable, which needs to be removed. So, the fixed statement is:
```python
first_ping_transmit_type = beam["transmit_type"].isel(ping_time=0).drop_vars("ping_time")
```

echopype/echodata/simrad.py
emiliom: I've implemented a solution based on the first ping_time.

echopype/echodata/simrad.py
emiliom: This change is in the latest commits.

echopype/convert/set_groups_ek80.py
emiliom: Let's leave it as is until we have better information.

emiliom:
I've addressed all your comments. Thanks so much! Hopefully the CI will complete w/o issues.

I'm not sure if the proper Github step here is to "Dismiss" your previous review, or if you should do that after reviewing my changes. I requested another review, but maybe that was unnecessary. Oh well.

emiliom:
Woo-hoo, the CI tests all passed!

echopype/echodata/simrad.py
```diff
@@ -120,32 +120,28 @@ def _retrieve_correct_beam_group_EK80(
...
         # 3) power samples are in Sonar/Beam_group2 if two beam groups exist
 
-        # Raise error if waveform_mode="CW" but CW data does not exist
-        if encode_mode == "complex" and transmit_type_has_BB:
-            # complex + BB data (all channels are BB)
-            if np.all(echodata["Sonar/Beam_group1"]["transmit_type"] != "CW"):
-                raise ValueError("waveform_mode='CW', but all data are broadband (BB)!")
+        # Raise error if waveform_mode="CW" but CW data does not exist (not a single ping is CW)
+        if encode_mode == "complex" and np.all(transmit_type != "CW"):
+            raise ValueError("waveform_mode='CW', but all data are broadband (BB)!")
```
leewujung, line 144: Since we only test for `first_ping_transmit_type` in the BB case, to make things more clear when we pick up the issue of within-channel CW-BB switches, maybe we change the check below to just the first ping too?
```suggestion
        # Raise error if waveform_mode="CW" but CW data does not exist (not a single ping is CW)
        # TODO: this needs to be changed when fixing within-channel CW-BB switch
        if encode_mode == "complex" and np.all(first_ping_transmit_type != "CW"):
            raise ValueError("waveform_mode='CW', but all data are broadband (BB)!")
```

echopype/echodata/simrad.py
```diff
@@ -120,32 +120,28 @@ def _retrieve_correct_beam_group_EK80(
     power_ed_group = None
     complex_ed_group = None
 
-    transmit_type_has_BB = not np.all(echodata["Sonar/Beam_group1"]["transmit_type"] == "CW")
+    transmit_type = echodata["Sonar/Beam_group1"]["transmit_type"]
+    # assume transmit_type identical for all pings in a channel
```
leewujung, line 124: Just adding a TOOD:
```suggestion
    # assume transmit_type identical for all pings in a channel
    # TODO: change when allowing within-channel CW-BB switch
```

leewujung (approved):
LGTM, thanks! I suggested a small tweak to just to make things parallel when we get around to fix the within-channel CW-BB switch.

echopype/echodata/simrad.py
emiliom: Looks good. I'll accept the suggestion here rather than via a push from me.

emiliom:
Thanks! I committed your two inline suggestions here. As soon as the CI is done, I can merge.

emiliom:
I've pushed the changes we discussed:
- FM -> LFM
- More fleshed out `flag_meanings`, largely from the convention descriptions

Assuming everything checks out, we can then merge this PR

leewujung:
Looks good. I'll merge now!

PR #1094, opened 2023-07-26
Add missing EK80 Beam_group variables

emiliom:
Addresses #951

Note: `flag_values` and `flag_meanings` are [CF attributes used with variables whose values are codes ("flags")](https://cfconventions.org/cf-conventions/cf-conventions.html#flags) that stand for specific meanings. We could stuff all that information in a `comment` attribute instead, but since CF already accommodates this specific situation, let's use the flag attributes.

Also added a `long_name` attribute to the `slope` variable, based on https://github.com/OSOceanAcoustics/echopype/issues/951#issuecomment-1652398704


leewujung:
> Note: `flag_values` and `flag_meanings` are [CF attributes used with variables whose values are codes ("flags")](https://cfconventions.org/cf-conventions/cf-conventions.html#flags) that stand for specific meanings. We could stuff all that information in a `comment` attribute instead, but since CF already accommodates this specific situation, let's use the flag attributes.

Great to learn about these!

leewujung (approved):
Looks good. Thanks for sending this PR!

PR #1099, opened 2023-07-28
Update a couple of "extra" EK60 Beam_group1 variables

emiliom:
Follows from https://github.com/OSOceanAcoustics/echopype/issues/951#issuecomment-1654246070, but for EK60. Complements PR #1094.

In `Beam_group1`:
- Drop `count` variable
- Rename `offset` to `range_sample_offset`
- Cast some variables to integer types: `range_sample_offset` to `dtype=int`, `transmit_mode` to `dtype=np.byte` (analogous to `channel_mode` in EK80) and `data_type` to `dtype=np.byte`. 
- Improve attributes for `range_sample_offset`, `transmit_mode`, and `data_type`; partly for consistency with EK80

**NOTES:**
- ~~The casting to integer types is currently not working! The EchoData DataArrays end up being float types. I'm looking into it.~~
- `transmit_mode` seems to be identical to `channel_mode` in EK80. @leewujung can you confirm? If they are, let's use the same variable name (`channel_mode`? Or change EK80 to `transmit_mode`?) and attribute values.


emiliom:
- Renamed `transmit_mode` to `channel_mode` and harmonized its attributes with the attributes from ek80 `channel_mode`
- Finalized the specific `int` types used
- Ensured that `int` casting works as intended by adding the following to `EXPECTED_VAR_DTYPE` in `utils/coding.py`:
    ```python
        "channel_mode": np.byte,
        "range_sample_offset": np.int32,
    ```

leewujung (approved):
This looks great. Thank you!

emiliom:
Great, thank you!

Issue #1100, opened 2023-07-28
Add missing mandatory `Beam_group1` variables to AZFP


emiliom:
Following up from https://github.com/OSOceanAcoustics/echopype/issues/640#issuecomment-1654596279

The following `Beam_group1` variables are mandatory in SONAR-netCDF4 v1 but are not currently created in `set_groups_azfp.py`:

- [x] `beam_direction_x/y/z`
- [x] `beam_type`
- [x] `transmit_type`

This is in addition to the 3 mandatory variables addressed in #1097.

Add them. Set to `nan` if no relevant data can be assigned.

emiliom:
@leewujung AZFP is single-beam, right? So, `beam_type` will be 0 (1 is split aperture).

leewujung:
Yes, AZFP is single beam!

emiliom:
For `transmit_type`, I looked through the AZFP manual and did some searching, but couldn't confirm that it the AZFP can be described as "CW". I assume it is, but can you confirm?

It turns out that EK60 is also missing `transmit_type`. It looks like that will be "CW", right?

For both instruments, this variable will be set as a scalar. I think I'll submit a PR for just this variable, for both instruments.

leewujung:
Yes, both AZFP and EK60 are "CW".

emiliom:
Thanks!

emiliom:
Based on the AZFP manual (pages 64-65, including Fig. 33) and input from ASL staff (Steve P. and Julek), I've gleaned the following about `tilt_x` and `tilt_y`:

- They are 0 when the transducer tube points up or down.
- They are calibrated to operate properly in a range of +/- 45 degrees. They're not reliable at tilts beyond that range.
- `tilt_x` and `tilt_y` correspond to rotation around the "x" and "y" axis, respectively. These axes are defined as the usual platform or ship, right-handed coordinate convention x and y axes. But the orientation of the axes (say, the x axis as the reference) is arbitrary. The y axis points in the direction of the "reference edge" in the transducer tube (see Fig. 33).
- Given these correspondences to x and y axes, `tilt_x` and `tilt_y` correspond to `roll` and `pitch`, respectively. But, this is with reference to the transducer tube's orientation, not the overall orientation of the "platform" (mooring, etc). This distinction is probably something that needs more discussion.

[...]

Issue #1097, opened 2023-07-28
Add missing convention mandatory `Beam_groupX` variables to all instruments


emiliom:
These 3 variables that are specified as mandatory in SONAR-netCDF4 v1 are currently missing from `Beam_groupX`, for all instruments (EK80, EK60, AZFP): `beam_stabilisation`, `non_quantitative_processing`, `sample_time_offset`. We'll add them.

@leewujung  could you chime in with information about how we would populate the variables?

For `beam_stabilisation`, the EK80 manual describes the optional "Teledyne TSS1 datagram format" for "for heave, roll and pitch compensation", and the EK60 manual also refers to it ("Simrad Sounder/TSS1 protocol"). But it doesn't look like echopype handles it. Here's an extended discussion about this from the v1 convention manual: "The coordinate system for beam direction uses the same convention as the platform coordinate system, but with the origin at the transducer. Sonar beams can be compensated or uncompensated for platform motion. If uncompensated, the sonar beam co- ordinate system follows the yaw, pitch, and roll of the platform. If beams are compensated, the sonar beam coordinate system is translated and rotated per ping so that the y-axis is always horizontal and the z-axis is always vertical. The x-axis is always parallel to the platform's x-axis. The beam stabilisation variable indicates whether the beams are compensated or uncompensated."

`non_quantitative_processing` accommodates flags. Here's some description from the v1 convention: "Settings of any processing that is applied prior to recording backscatter data that may prevent the calculation of calibrated backscatter. A value of 0 always indicates no such processing.... [for the `flag_values` attribute] List of unique values (e.g. 0, 1, 3, 4) that indicate different non-quantitative processing settings that could be present in the sonar data. Must have the same number of values as settings given in the `flag_meanings` attribute." I'm guessing that we can just set it to 0 by default, unless you think there is relevant information present in the datagrams?

`sample_time_offset`: From the v1 convention: "Time offset applied to sample time-stamps and intended for applying a range correction (e.g. as caused by signal processing delays). Positive values reduce the calculated range to a sample."

leewujung:
For `beam_stabilisation`: Yeah I don't think echopype handles that right now, so I think this would be `not_stabilised = 0` (a side question: I remember trying to implement the enum type but got lost in it, are we just using True/False here or 1/0?)

For `non_quantitative_processing`: I agree with you that for echopype it'll just be 0

[...]

PR #1101, opened 2023-07-30
EK60, add mandatory `transmit_type` as a scalar

emiliom:
See https://github.com/OSOceanAcoustics/echopype/issues/1100#issuecomment-1657239544

### DIFF
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -585,6 +585,17 @@ class SetGroupsEK60(SetGroupsBase):
                     ["channel"],
                     beam_params["gpt_software_version"],
                 ),
+                "transmit_type": (
+                    [],
+                    "CW",
+                    {
+                        "long_name": "Type of transmitted pulse",
+                        "flag_values": ["CW"],
+                        "flag_meanings": [
+                            "Continuous Wave",
+                        ],
+                    },
+                ),
             },
             coords={
                 "channel": (


leewujung (approved):
Good to add this! 

Question: The "Continuous Wave" flag_meanings seems to come from the EK80 reference manual? I wonder if we should say add "(gated single-frequency sine waves)" for added explanation. 

Also, for FMD, the "D" seems to be user-defined?

emiliom:
> Question: The "Continuous Wave" flag_meanings seems to come from the EK80 reference manual? I wonder if we should say add "(gated single-frequency sine waves)" for added explanation.

Right. But I also found a reference to CW in the EK60 manual, though it was indirect.

Also, today I realized that the convention uses a special `enum` type for `transmit_type`, not a `string` per se. Here's what the enum description says:

> `byte enum transmit_t {CW = 0, LFM = 1, HFM = 2}`
>
> Type of transmit pulse. CW = continuous wave – a pulse nominally of one frequency, LFM = linear frequency modulation – a pulse which varies from transmit_frequency_start to transmit_frequency_stop in a linear manner over the nominal pulse duration (transmit_duration_nominal), HFM = hyperbolic frequency modulation – a pulse which varies from transmit_frequency_start to transmit_frequency_stop in a hyperbolic manner over the nominal pulse duration (transmit_duration_nominal).
```

> Also, for FMD, the "D" seems to be user-defined?

In the EK80 manual under `PulseForm` (p. 82), FMD is not defined. But I just noticed that "FMD" does appear in one other place (p. 87), under ADCP `PulseForm`, as "LFMD". There, the acronyms are spelled out, but the LFMD part is kind of jumbled:

![image](https://github.com/OSOceanAcoustics/echopype/assets/742403/44abd129-9115-45ac-8dde-9db198a45b02)

It's hard to be sure, but it does look like "D" might stand for user-defined.

leewujung:
> but the LFMD part is kind of jumbled:
> [...]
> It's hard to be sure, but it does look like "D" might stand for user-defined.

Yes... in fact that was where I found the "user-defined" because I was just doing a full text search.

> Tweaking the attributes is easy to do at any time, and has no real downstream impact. But I wonder if we need to revisit `transmit_type` one last time in light of the convention enum type .... Sigh.

... I thought you said enum type is not supported in netCDF (I forgot in which issue/PR)?

emiliom:
> ... I thought you said enum type is not supported in netCDF (I forgot in which issue/PR)?

The thread you're referring to is probably this one: https://github.com/OSOceanAcoustics/echopype/issues/1097#issuecomment-1656489638. I'll paste it here: "let's not try to bring some kind of enum type, which I don't think actually exists in the netcdf/CF/zarr world. I don't think there's a boolean type either, so I'll use np.byte."

After some digging, I can see I misspoke, partly. The `enum` type [does exist in netcdf4](https://docs.unidata.ucar.edu/nug/current/md_types.html#data_type) (which should've been obvious to me, given that it's used in the convention!). However, it doesn't appear to be widely implemented. As far as I can tell it's not supported by `CF` or `xarray`. I see no statements saying "enum types are not supported", but a search through their documentation doesn't say they do; CF only discusses "enumerations" in the context of the "flag" attributes -- their way of representing enums via attributes rather than via hard coding a custom data type. The `xarray` has only one hit for the string "enum", and I think it refers to something else. Also, I don't think I've ever run into a netcdf enum type in actual use.

In summary: NetCDF4 (at least the C library -- I haven't checked the Python library) does support enum types, but to  me, for all practical purposes it seems impractical, maybe even unwise, to try to use them.

emiliom:
Since you approved this PR (thanks!), I'll merge it and we can continue the `transmit_type` and `enum` discussion here or in a new issue.

leewujung:
> In summary: NetCDF4 (at least the C library -- I haven't checked the Python library) does support enum types, but to me, for all practical purposes it seems impractical, maybe even unwise, to try to use them.

Let's not worry enum type then in echopype, which makes things easier. :)

PR #1102, opened 2023-07-30
AZFP: add missing mandatory Beam_group1 variables

emiliom:
Addresses #1100

- [x] beam_type
- [x] transmit_type
- [x] beam_direction_x/y/z

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -340,6 +340,19 @@ class SetGroupsAZFP(SetGroupsBase):
                         "standard_name": "sound_frequency",
                     },
                 ),
+                # TODO: Add channel dim, to match ek60 and ek80?
+                "beam_type": (
+                    [],
+                    0,
+                    {
+                        "long_name": "Beam type",
+                        "flag_values": [0, 1],
+                        "flag_meanings": [
+                            "Single beam",
+                            "Split aperture beam",
+                        ],
+                    },
+                ),
                 "backscatter_r": (
                     ["channel", "ping_time", "range_sample"],
                     N,
@@ -380,6 +393,17 @@ class SetGroupsAZFP(SetGroupsBase):
                         "long_name": "Nominal bandwidth of transmitted pulse",
                         "units": "s",
                         "valid_min": 0.0,
+                    },
+                ),
+                "transmit_type": (
+                    [],
+                    "CW",
+                    {
+                        "long_name": "Type of transmitted pulse",
+                        "flag_values": ["CW"],
+                        "flag_meanings": [
+                            "Continuous Wave",
+                        ],
                     },
                 ),
             },
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -352,6 +352,20 @@ class SetGroupsAZFP(SetGroupsBase):
                         ],
                     },
                 ),
+                **{
+                    f"beam_direction_{var}": (
+                        ["channel"],
+                        [np.nan] * len(self.channel_ids_sorted),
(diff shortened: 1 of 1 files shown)


echopype/convert/set_groups_azfp.py
```diff
@@ -340,6 +340,19 @@ def set_beam(self) -> List[xr.Dataset]:
                         "standard_name": "sound_frequency",
                     },
                 ),
+                # TODO: Add channel dim, to match ek60 and ek80?
+                "beam_type": (
+                    [],
```
emiliom, line 345: `beam_type` has a ` channel` dimension in EK60 and EK80. Strictly speaking, it's unnecessary for EK60, right? It's unnecessary for AZFP, for the same reason. I'm hesitant to suggest changes to EK60, in case it has impacts downstream. But then the dimensionality would be inconsistent from the similar case for AZFP, where here I'm adding it as a scalar.

Thoughts.

leewujung: `beam_type` can actually be 1 (split-beam) for EK60, and we do use it for processing in `consolidate.split_beam_angle::get_angle_power_samples`

For AZFP it is unnecessary. Since it has a small footprint (only the `channel` dimension), how about adding that for AZFP and set them all to 0 (single-beam)?

emiliom: I was also  leaning towards adding the `channel` dim in AZFP. Great. I'll change it.

> beam_type can actually be 1 (split-beam) for EK60, and we do use it for processing in consolidate.split_beam_angle::get_angle_power_samples

What I meant in comparing AZFP with EK60 was that in EK60, `beam_type` is the same (1) across channels; or at least that's my understanding!

leewujung: > > beam_type can actually be 1 (split-beam) for EK60, and we do use it for processing in consolidate.split_beam_angle::get_angle_power_samples
> 
> What I meant in comparing AZFP with EK60 was that in EK60, `beam_type` is the same (1) across channels; or at least that's my understanding!

So `beam_type` could change across channel in EK60, depending on the type of transducer installed.

emiliom: Ah, thanks. Good to know.

emiliom:
Note to self: When addressing `beam_direction_x/y/z`, it may be helpful to refer back to the AZFP `Platform` group overhaul PR's merged in the last couple of months, #1058 and #1061

emiliom:
Added `beam_direction_x/y/z` variables, set to `nan`. That is, nothing is being read from AZFP data. The variables are created because they're mandatory (M) in the convention.

echopype/convert/set_groups_azfp.py
```diff
@@ -382,6 +408,17 @@ def set_beam(self) -> List[xr.Dataset]:
...
                     },
                 ),
+                "transmit_type": (
+                    [],
+                    "CW",
+                    {
+                        "long_name": "Type of transmitted pulse",
+                        "flag_values": ["CW"],
+                        "flag_meanings": [
+                            "Continuous Wave",
```
leewujung, line 418: I've been mulling over about this. It is true that CW is Continuous Wave, but it is an approximation. The actual transmit signal is gated sine wave. I wonder if this would be misleading. This is something defined in the convention, but do you think we can add something to make it "Continuous Wave (gated sine wave)"?

leewujung (approved):
Looks good! I only have a small comment on one of the `flag_meanings`.

emiliom:
We've discussed the `flag_meanings` comment. We decided to stick with "CW", but a new PR will have a more fleshed out `flag_meanings` entry that better addresses the concern.

PR #1103, opened 2023-07-31
Add missing, mandatory `Beam_groupX` variables

emiliom:
Addresses #1097. Adds mandatory variables

- [x] `beam_stabilisation`, as a scalar
- [x] `non_quantitative_processing`, as a scalar
- [x] `sample_time_offset`: Will replace `range_sample_offset`, at least for EK60 & EK80. See https://github.com/OSOceanAcoustics/echopype/issues/1097#issuecomment-1656418738, and note that this change revisits issue #951

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -382,6 +382,24 @@ class SetGroupsAZFP(SetGroupsBase):
                         "valid_min": 0.0,
                     },
                 ),
+                "beam_stabilisation": (
+                    [],
+                    np.array(0, np.byte),
+                    {
+                        "long_name": "Beam stabilisation applied (or not)",
+                        "flag_values": [0, 1],
+                        "flag_meanings": ["not stabilised", "stabilised"],
+                    },
+                ),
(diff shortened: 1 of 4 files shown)


leewujung (approved):
@emiliom : Thanks, looks good! I only have one question about the coding:
```
"beam_stabilisation": np.byte,
"non_quantitative_processing": np.int16,
```
Since they are both scalar and take 0 or 1 values, why is one `np.byte` and `np.int16`?

emiliom:
Thanks!

> beam_stabilisation": np.byte,
> "non_quantitative_processing": np.int16,
>
> Since they are both scalar and take 0 or 1 values, why is one np.byte and np.int16?

`beam_stabilisation` is defined to be only 0 or 1. But the definition for `non_quantitative_processing` is more open ended. I  was being perhaps overly open to future possibilities in setting it to int16! But we could change it to byte, especially since in echopype we'll be assigning it a value of 0 in echopype in all cases.

emiliom:
Thanks! Merging, based on our side conversation.

PR #1104, opened 2023-07-31
Rename `fs_receiver` in Vendor-specific group

leewujung:
This PR rename the data variable `fs_receiver` in the Vendor-specific group to `receiver_sampling_frequency`. This should have been changed around the time of #1019, but somehow didn't get fixed.


emiliom (approved):
Nice! And BTW I can see that `receiver_sampling_frequency` occurs in many places in the code, whereas `fs_receiver` only occurred here.

PR #1105, opened 2023-07-31
Small fixes for BB and splitbeam angle handling

leewujung:
This PR addresses a couple bugs and makes one refactoring change:
- [x] Remove the handling of BB filter coefficients that are 1-element long. The handling was needed before when the filter coefficients were stored as attributes (instead of data variables) in the Vendor-specific group. This was changed in #1044 as part of the `combine_echodata` overhaul (#1042), and the handling actually became a bug! It'd be nice to add a test to check this case, so I'll keep this as a draft PR for now. **[edit 08/18/2023: I've now added a test!]**
- [x] Factor out a normalization factor from `ek80_complex.compress_pulse` so that the function only does pulse compression. This way we can reuse this for getting the spectrum in addition to the current band-average processing.
- [x] Add error checking for `source_Sv` containing specified angle parameters, in `consolidate.api.add_splitbeam_angle`:
   https://github.com/OSOceanAcoustics/echopype/blob/6b734eb86958c036919216719a8b3c6e1b41c0a1/echopype/consolidate/api.py#L316-L325

### DIFF
--- a/.ci_helpers/py3.11.yaml
+++ b/.ci_helpers/py3.11.yaml
@@ -1,0 +1,8 @@
+name: echopype
+channels:
+  - conda-forge
+dependencies:
+  - python=3.11
+  - pip
+  - pip:
+    - -r ../requirements.txt
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -16,18 +16,19 @@ jobs:
   test:
     name: ${{ matrix.python-version }}-build
     runs-on: ubuntu-20.04
-    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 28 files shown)


leewujung:
I haven't been able to add tests for the fixed cases, but the code is ready for review.

emiliom:
> I haven't been able to add tests for the fixed cases, but the code is ready for review.

Did you intend to ask me to review? I'm not officially flagged to review this PR

leewujung:
> Did you intend to ask me to review? I'm not officially flagged to review this PR

Haha I don't know if you have time for that, but if so that'll be great!

emiliom (approved):
As far as I can tell, everything here looks good and self consistent.

leewujung:
I've now added a test for the `get_vend_filter_EK80` function that is fixed under the first bullet point. If all tests run through I think we can merge this.

It would be nice to add a test for the second bullet point in error checking for `source_Sv` in `consolidate.api.add_splitbeam_angle`, but that is better done using a mock EchoData object, so a bit out of scope for this PR.

Issue #967, opened 2023-03-09
Improve input syntax for frequency differencing operation


leewujung:
Currently the [frequency differencing implementation](https://github.com/OSOceanAcoustics/echopype/blob/a09d8f32fcd99ab9fbc7d66ef237545d9e7d2697/echopype/mask/api.py#L369) takes an awkward input that forces users to separate out the channels or frequencies from the operators (>, < , etc) that operates on them. I think the frequency differencing criteria (something like `chanA - chanB >= X dB` or `freqA - freqB >= X dB`) is something that can be done with regular expression and enhances the usability significantly.

## Requirements
- parse the input arguments (chanA/B or freqA/B (do not mix use of chan and freq), X dB, >= or other comparison operators) directly from the criteria equation
- add tests

praneethratna:
Hey @leewujung, I have a query regarding this issue - So there should be two arguments where one equation is in terms of chan and other equation in terms of freq and only one of them should be provided?

```
def frequency_differencing(
    source_Sv: Union[xr.Dataset, str, pathlib.Path],
    storage_options: Optional[dict] = {},
    freqABDiff: Optional[str] = None,
    chanABDiff: Optional[str] = None,
) -> xr.DataArray:
```
Is this the correct function definition according to the change?

leewujung:
Hey @praneethratna : no worries, thanks for contributing! I just assigned you to the issues.

Yes I was thinking to have something like this:
```
def frequency_differencing(
    source_Sv: Union[xr.Dataset, str, pathlib.Path],
    diff_eq_freq: Optional[str] = None,  # eg "38 kHz - 18 kHz >= 15 dB"
    diff_eq_chan: Optional[str] = None,  # eg "WBT 743366-15 ES38B_ES - WBT 743367-15 ES18_ES < 30 dB"
    storage_options: Optional[dict] = {},
)
```
`diff_eq_freq` or `diff_eq_chan` can be used, but not both.
In the above, strings like "WBT 743366-15 ES38B_ES" and "WBT 743367-15 ES18_ES" are values found in the `channel` dimension in the Sv dataset (there are almost always spaces in the channel strings).
I think we should also be able to deal with arbitrary spaces like "38kHz -18kHz>=15 dB".

Did the above answer your question?

praneethratna:
> Did the above answer your question?
Hey @leewujung, yes and one small doubt in the freq equation do we consider freqB as -18 khz or it is 18 khz?

[...]

PR #1106, opened 2023-07-31
Improved input syntax for `frequency_differencing` operation

praneethratna:
Addresses #967. Now input for frequency_differencing can be given directly in the form of frequency differencing criteria instead of separating out freq/chan from operators.

CC @leewujung

### DIFF
--- a/echopype/mask/api.py
+++ b/echopype/mask/api.py
@@ -8,6 +8,7 @@ import xarray as xr
 
 from ..utils.io import validate_source_ds_da
 from ..utils.prov import add_processing_level, echopype_prov_attrs, insert_input_processing_level
+from .freq_diff import _check_freq_diff_source_Sv, _parse_freq_diff_eq
 
 # lookup table with key string operator and value as corresponding Python operator
 str2ops = {
@@ -366,128 +367,11 @@ def apply_mask(
     return output_ds
 
 
-def _check_freq_diff_non_data_inputs(
-    freqAB: Optional[List[float]] = None,
-    chanAB: Optional[List[str]] = None,
-    operator: str = ">",
-    diff: Union[float, int] = None,
-) -> None:
-    """
-    Checks that the non-data related inputs of ``frequency_differencing`` (i.e. ``freqAB``,
-    ``chanAB``, ``operator``, ``diff``) were correctly provided.
-
-    Parameters
-    ----------
-    freqAB: list of float, optional
-        The pair of nominal frequencies to be used for frequency-differencing, where
-        the first element corresponds to ``freqA`` and the second element corresponds
-        to ``freqB``
-    chanAB: list of float, optional
-        The pair of channels that will be used to select the nominal frequencies to be
-        used for frequency-differencing, where the first element corresponds to ``freqA``
-        and the second element corresponds to ``freqB``
-    operator: {">", "<", "<=", ">=", "=="}
-        The operator for the frequency-differencing
-    diff: float or int
-        The threshold of Sv difference between frequencies
-    """
-
-    # check that either freqAB or chanAB are provided and they are a list of length 2
-    if (freqAB is None) and (chanAB is None):
-        raise ValueError("Either freqAB or chanAB must be given!")
-    elif (freqAB is not None) and (chanAB is not None):
-        raise ValueError("Only freqAB or chanAB must be given, but not both!")
-    elif freqAB is not None:
-        if not isinstance(freqAB, list):
-            raise TypeError("freqAB must be a list!")
-        elif len(set(freqAB)) != 2:
-            raise ValueError("freqAB must be a list of length 2 with unique elements!")
-    else:
-        if not isinstance(chanAB, list):
-            raise TypeError("chanAB must be a list!")
-        elif len(set(chanAB)) != 2:
-            raise ValueError("chanAB must be a list of length 2 with unique elements!")
-
-    # check that operator is a string and a valid operator
-    if not isinstance(operator, str):
-        raise TypeError("operator must be a string!")
-    else:
-        if operator not in [">", "<", "<=", ">=", "=="]:
-            raise ValueError("Invalid operator!")
-
-    # ensure that diff is a float or an int
-    if not isinstance(diff, (float, int)):
-        raise TypeError("diff must be a float or int!")
-
-
-def _check_source_Sv_freq_diff(
-    source_Sv: xr.Dataset,
-    freqAB: Optional[List[float]] = None,
-    chanAB: Optional[List[str]] = None,
-) -> None:
-    """
-    Ensures that ``source_Sv`` contains ``channel`` as a coordinate and
-    ``frequency_nominal`` as a variable, the provided list input
-    (``freqAB`` or ``chanAB``) are contained in the coordinate ``channel``
-    or variable ``frequency_nominal``, and ``source_Sv`` does not have
-    repeated values for ``channel`` and ``frequency_nominal``.
-
-    Parameters
-    ----------
-    source_Sv: xr.Dataset
-        A Dataset that contains the Sv data to create a mask for
-    freqAB: list of float, optional
-        The pair of nominal frequencies to be used for frequency-differencing, where
-        the first element corresponds to ``freqA`` and the second element corresponds
-        to ``freqB``
-    chanAB: list of float, optional
-        The pair of channels that will be used to select the nominal frequencies to be
-        used for frequency-differencing, where the first element corresponds to ``freqA``
-        and the second element corresponds to ``freqB``
-    """
-
-    # check that channel and frequency nominal are in source_Sv
-    if "channel" not in source_Sv.coords:
-        raise ValueError("The Dataset defined by source_Sv must have channel as a coordinate!")
-    elif "frequency_nominal" not in source_Sv.variables:
-        raise ValueError(
-            "The Dataset defined by source_Sv must have frequency_nominal as a variable!"
-        )
-
-    # make sure that the channel and frequency_nominal values are not repeated in source_Sv
-    if len(set(source_Sv.channel.values)) < source_Sv.channel.size:
-        raise ValueError(
-            "The provided source_Sv contains repeated channel values, this is not allowed!"
-        )
-
-    if len(set(source_Sv.frequency_nominal.values)) < source_Sv.frequency_nominal.size:
-        raise ValueError(
-            "The provided source_Sv contains repeated frequency_nominal "
-            "values, this is not allowed!"
-        )
-
-    # check that the elements of freqAB are in frequency_nominal
-    if (freqAB is not None) and (not all([freq in source_Sv.frequency_nominal for freq in freqAB])):
-        raise ValueError(
-            "The provided list input freqAB contains values that "
-            "are not in the frequency_nominal variable!"
-        )
-
-    # check that the elements of chanAB are in channel
-    if (chanAB is not None) and (not all([chan in source_Sv.channel for chan in chanAB])):
-        raise ValueError(
-            "The provided list input chanAB contains values that are "
-            "not in the channel coordinate!"
-        )
-
-
 def frequency_differencing(
     source_Sv: Union[xr.Dataset, str, pathlib.Path],
     storage_options: Optional[dict] = {},
-    freqAB: Optional[List[float]] = None,
-    chanAB: Optional[List[str]] = None,
-    operator: str = ">",
-    diff: Union[float, int] = None,
+    freqABEq: Optional[str] = None,
+    chanABEq: Optional[str] = None,
 ) -> xr.DataArray:
     """
     Create a mask based on the differences of Sv values using a pair of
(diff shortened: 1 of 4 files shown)


emiliom:
These improvements are very cool! 

Since you're adding `sympy` as a dependency, how "heavy" a library is it? In other words, is it likely to create dependency resolution issues? I see the dependency tree `sympy > mpmath > gmpy2`, where [gmpy2](https://github.com/aleaxit/gmpy) apparently uses a C library ("gmpy2 is an optimized, C-coded Python extension module  ...").

Are there lighter-weight alternatives to `sympy` that would meet our needs? sympy has tons of capabilities, and here I think we'd be using only a tiny subset.

requirements-dev.txt
```diff
@@ -23,3 +23,4 @@ sphinx_rtd_theme
 sphinxcontrib-mermaid
 twine
 wheel
+sympy
```
emiliom, line 26: I don't think sympy needs to go here, too. It only needs to be in `requirements.txt`
```suggestion
```

praneethratna:
> These improvements are very cool! 
> 
> Since you're adding `sympy` as a dependency, how "heavy" a library is it? In other words, is it likely to create dependency resolution issues? I see the dependency tree `sympy > mpmath > gmpy2`, where [gmpy2](https://github.com/aleaxit/gmpy) apparently uses a C library ("gmpy2 is an optimized, C-coded Python extension module  ...").

I feel `SymPy` is lightweight compated compared to other mathematical libraries found online. Also `SymPy` has a hard dependency only on mpmath but I don't it would create dependency resolution issues.

> 
> Are there lighter-weight alternatives to `sympy` that would meet our needs? sympy has tons of capabilities, and here I think we'd be using only a tiny subset.

I'm using `SymPy` for parsing numbers from strings using it's inbuilt parser. I'm not sure of any other libraries which can be used for this.

emiliom:
> Also SymPy has a hard dependency only on mpmath but I don't it would create dependency resolution issues.

Right! Scratch my comment about dependency resolution per se. I did see that the dependency was limited to mpmath > gmpy2, and we don't currently use those libraries. My apologies.

praneethratna:
> > Also SymPy has a hard dependency only on mpmath but I don't it would create dependency resolution issues.
> 
> Right! Scratch my comment about dependency resolution per se. I did see that the dependency was limited to mpmath > gmpy2, and we don't currently use those libraries. My apologies.

No worries! I understand the clarification now. Since we don't use mpmath or gmpy2 libraries, these dependencies doesn't affect us. But we need to look at alternatives for `SymPy` if that makes the package slow.

leewujung:
Since the equations are pretty predictable, one option would be to just do `re` all the way?

praneethratna:
> Since the equations are pretty predictable, one option would be to just do `re` all the way?

You mean we use a single regex expression to check for the validity of the whole frequency differencing criteria? Wouldn't that make the code a bit messy?

leewujung:
> @leewujung Do you mean we use a single regex expression to check for the validity of the whole frequency differencing criteria? Wouldn't that make the code a bit messy?

I think it would be ok? For the freqAB case, the `Hz/kHz/MHz/GHz` cases are fixed, so the number in front of that could be extracted. For the chanAB case, it is the string between the double quotes. The binary comparison operator `>, <, >=, <=, ==` are the allowed cases. And then we need to extracted the number before `dB`. This doesn't seem overly complicated?

praneethratna:
> I think it would be ok? For the freqAB case, the `Hz/kHz/MHz/GHz` cases are fixed, so the number in front of that could be extracted. For the chanAB case, it is the string between the double quotes. The binary comparison operator `>, <, >=, <=, ==` are the allowed cases. And then we need to extracted the number before `dB`. This doesn't seem overly complicated?

@leewujung I think it shouldn't be overly complicated, it's just that i split everything into individual parts for better code debugging. I'll make these changes by today.

praneethratna:
@leewujung Even if use a single `re` expression to check for validity, we again need to use separate `re`' expressions to extract freq/chan values, operators and diff or am i understanding your point in a wrong way?

praneethratna:
@leewujung how do you suggest to do this?

leewujung:
Hi @praneethratna : sorry for my super late response. What I am thinking in terms of using `re` to grab out the pattern is like the following:
- use patterns like `"(?P<freqA>\d+)(?P<unitA>\w?)Hz"` to match freqA and freqB, for example:
  ```python
  a = re.match("(?P<freq>\d+)\s*(?P<unit>\w?)Hz", "120kHz") # also matches arbitrary space between "120" and "kHz"
  a["freq"]  # get the number associated with the frequency
  a["unit"]  # get the unit associated with the frequency
  ```
  then you can use `a["unit"]` to determine the right multiplication factor to get to the numbers in `frequency_nominal`, which is with units Hz (i.e. 120kHz is 120000 in `frequency_nominal`).
- you can use the pattern above to build up the whole pattern, something like the following:
  ```python
  freqA_pattern = "(?P<freqA>\d+)\s*(?P<unitA>\w?)Hz"
  freqA_pattern = "(?P<freqB>\d+)\s*(?P<unitB>\w?)Hz"
  db_pattern = "(?P<db>\d+)\s*dB"
  cmp_pattern = "\s*(?P<cmp>\S+?)\s*"
  DIFF_MATCHER = re.compile(freqA_pattern + "\s*-\s*" + freqB_pattern + cmp_pattern + db_pattern)
  ```
  and then use outputs of the matched results with some error checking and handling (like converting `str` to `float`, k to 1000 etc) to combine with your other parts of the code.

What do you think?

praneethratna:
> Hi @praneethratna : sorry for my super late response. What I am thinking in terms of using `re` to grab out the pattern is like the following:
> 
> * use patterns like `"(?P<freqA>\d+)(?P<unitA>\w?)Hz"` to match freqA and freqB, for example:
>   ```python
>   a = re.match("(?P<freq>\d+)\s*(?P<unit>\w?)Hz", "120kHz") # also matches arbitrary space between "120" and "kHz"
>   a["freq"]  # get the number associated with the frequency
>   a["unit"]  # get the unit associated with the frequency
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   then you can use `a["unit"]` to determine the right multiplication factor to get to the numbers in `frequency_nominal`, which is with units Hz (i.e. 120kHz is 120000 in `frequency_nominal`).
> * you can use the pattern above to build up the whole pattern, something like the following:
>   ```python
>   freqA_pattern = "(?P<freqA>\d+)\s*(?P<unitA>\w?)Hz"
>   freqA_pattern = "(?P<freqB>\d+)\s*(?P<unitB>\w?)Hz"
>   db_pattern = "(?P<db>\d+)\s*dB"
>   cmp_pattern = "\s*(?P<cmp>\S+?)\s*"
>   DIFF_MATCHER = re.compile(freqA_pattern + "\s*-\s*" + freqB_pattern + cmp_pattern + db_pattern)
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   and then use outputs of the matched results with some error checking and handling (like converting `str` to `float`, k to 1000 etc) to combine with your other parts of the code.
> 
> What do you think?

@leewujung No worries! I understand it now, these suggestions helped me to refactor the code in a much simpler and clean way. I have made some assumptions while commiting the changes (related to `frequency_nominal`) and the test code which generates `frequency_nominal` which might require a change before we merge this.

leewujung:
@praneethratna : there are still places where you use `sympy` in the code. Could you remove them completely and use `re` instead? Thanks!

praneethratna:
@leewujung I have removed the usage of `sympy' completely and also squashed all the commits into one.

echopype/mask/api.py
```diff
@@ -366,58 +367,85 @@ def get_ch_shape(da):
...
+        If `freqABEq` or `chanABEq` is not a valid equation.
     """
 
-    # check that either freqAB or chanAB are provided and they are a list of length 2
-    if (freqAB is None) and (chanAB is None):
+    if (freqABEq is None) and (chanABEq is None):
         raise ValueError("Either freqAB or chanAB must be given!")
-    elif (freqAB is not None) and (chanAB is not None):
+    elif (freqABEq is not None) and (chanABEq is not None):
         raise ValueError("Only freqAB or chanAB must be given, but not both!")
```
leewujung, line 407: minor wording tweak
```suggestion
    elif (freqABEq is not None) and (chanABEq is not None):
        raise ValueError("Only one of freqAB or chanAB should be given, but not both!")
```

leewujung (commented):
@praneethratna : Thanks for this great PR!! It is awesome to have this intuitive way of putting in the freq diff equations.

I have some small suggestions for the organization and the tests:
- Since in #1142 there are more functions added to the `mask` subpackage, how about in this PR moving the frequency-differencing-related functions (`_check_source_Sv_freq_diff` and `check_freq_diff_and_parse_arguments`) to a separate module, say `echopype/mask/freq_diff.py` and keep the main API function `frequency_differencing` in `echopype/mask/api.py`? 
- for the tests, currently `test_frequency_differencing` is more of an integrated test, and in it the frequency units are all Hz so the unit matching was not tested. How about adding another set of tests to test for the function you wrote `check_freq_diff_and_parse_arguments`? This way you can test parsed element values and xfail the ones that should fail, etc.
- I suggest renaming `check_freq_diff_and_parse_arguments` to `_parse_freq_diff_eq` for the sake of clarity and also making it shorter

leewujung:
Also, I caught a bug / undesired behavior in the current `_check_source_Sv_freq_diff` function. In the current implementation, the values of `frequency_nominal` is not allowed to be duplicated:
https://github.com/OSOceanAcoustics/echopype/blob/254ee106eb580959ca85f2131168c4f17c0a285a/echopype/mask/api.py#L463-L467

However, it is possible that there are 2 channels (with different `channel` strings) that have the same `frequency_nominal` values (this is actually why we decided to use `channel` as the coordinate, instead of `frequency_nominal`, since coordinates cannot have duplicated values), and users may want to diff these 2 channels using the `chanABEq`.  I think that potentially duplicated `frequency_nominal` values should only be checked if people use `freqABEq` to put in their equation.

Looking at the code again, what do you think if we take the following approach to change this function?
- keep the first checks `# check that channel and frequency nominal are in source_Sv` intact
- for the below three checks, make it such that only `freqAB` checks are executed when people use `freqABEq`, and only `chanAB` checks are executed when people use `chanABEq`
  ```
  # make sure that the channel and frequency_nominal values are not repeated in source_Sv
  # check that the elements of freqAB are in frequency_nominal
  # check that the elements of chanAB are in channel
  ```
   ie we only check for `frequency_nominal` to not be duplicated and elements of `freqAB` are in `frequency_nominal` when people use `freqABEq`, and only check for `channel` to not be duplicated and elements of `chanAB` are in `channel` when people use `chanABEq`.

echopype/tests/mask/test_mask.py
```diff
@@ -287,33 +291,84 @@ def test_check_source_Sv_freq_diff(n: int, n_chan_freq: int, add_chan: bool, add
...
-        (5, 4, [1.0, 3.0], None, 1.0, ">", np.zeros((5, 5))),
-        (5, 4, None, ['chan1', 'chan3'], 1.0, ">", np.zeros((5, 5))),
-        (5, 4, [1.0, 3.0], None, 1.0, "<=", np.ones((5, 5))),
-        (5, 4, None, ['chan1', 'chan3'], 1.0, "<=", np.ones((5, 5))),
-        (5, 4, [1.0, 3.0], None, 1.0, "<", np.ones((5, 5)) - np.identity(5)),
-        (5, 4, None, ['chan1', 'chan3'], 1.0, "<", np.ones((5, 5)) - np.identity(5)),
+        ("1.0Hz-2.0Hz==1.0dB", None),
+        (None, '"chan1"-"chan3"==1.0dB'),
+        ("1.0 kHz - 2.0 MHz>=1.0 dB", None),
+        (None, '"chan2-12 89" - "chan4 89-12" >= 1.0 dB'),
```
leewujung, line 303: These variations are great! I can see that the current test makes sure that `_parse_freq_diff_eq` is run through without problem, though the values are not checked. Could you add a section in the test for the parsed values, to make sure the output is correct? Thanks!

echopype/mask/freq_diff.py
```diff
@@ -0,0 +1,148 @@
...
+            "The provided list input freqAB contains values that "
+            "are not in the frequency_nominal variable!"
+        )
+
+    # check that the elements of chanAB are in channel
+    if (chanAB is not None) and (not all([chan in source_Sv.channel for chan in chanAB])):
+        raise ValueError(
+            "The provided list input chanAB contains values that are "
+            "not in the channel coordinate!"
+        )
```
leewujung, line 148: I would suggest consolidating the above to improve readability:
```python
# raise error if freqAB and chanAB are both None or both have values
if ...
   raise ValueError("...")

if freqAB is not None:
   # test if frequency_nominal values are not repeated in source_Sv
   # check that the elements of freqAB are in frequency_nominal
else:  # this implies chanAB is not None
   # test if channel values are not repeated in source_Sv
   # check that the elements of chanAB are in frequency_nominal
```

echopype/tests/mask/test_mask.py
```diff
@@ -232,21 +236,21 @@ def create_input_mask(
...
                      marks=pytest.mark.xfail(strict=True,
                                              reason="This should fail because the Dataset "
                                                     "will not have the frequency_nominal variable.")),
-        pytest.param(5, 3, True, True, [1.0, 4.0], None,
+        pytest.param(5, 4, True, True, [1.0, 4.0], None,
                      marks=pytest.mark.xfail(strict=True,
                                              reason="This should fail because not all selected frequencies"
                                                     "are in the frequency_nominal variable.")),
-        pytest.param(5, 3, True, True, None, ['chan1', 'chan4'],
+        pytest.param(5, 4, True, True, None, ['chan1', 'chan9'],
```
leewujung, line 253: I'd suggest another 2 cases where you xfail duplicated `freqAB` and duplicated `chanAB` cases, just to be complete.

leewujung (changes requested):
Thanks @praneethratna for your speedy changes! I added some suggestions, and I think once they are addressed we can merge this PR. :)

leewujung (approved):
@praneethratna : Thanks a lot for your contribution. We can merge this PR now!

leewujung:
Ugh sorry just remember one thing. In here
https://github.com/OSOceanAcoustics/echopype/blob/254ee106eb580959ca85f2131168c4f17c0a285a/echopype/tests/utils/test_processinglevels_integration.py#L106
`frequency_differencing` is called (this is something the CI does not catch, unless one uses [all tests ci] (see [here](https://echopype.readthedocs.io/en/latest/contributing.html#continuous-integration-github-actions)). Could you quickly change it? Then I will merge. Thank you!

PR #1107, opened 2023-07-31
Change 2 AZFP `Beam_group1` data types & add `long_name` to Vendor attributes

emiliom:
1. Change AZFP `Beam_group1` AZFP `backscatter_r` and `gain_correction` variables from current `int` data type to `float`, to match the convention and other instruments.
2. Add `long_name` attribute for many (but not all) AZFP Vendor variables. Addresses https://github.com/OSOceanAcoustics/echopype/issues/640#issuecomment-1656369306. Descriptions for `long_name` where taken from the AZFP manual (section 2.4 and other section) and comments in `HEADER_FIELDS` in `parse_azfp.py`:
https://github.com/OSOceanAcoustics/echopype/blob/de4fb4df8c07e3579c5c700bba232d0da0bcaeff/echopype/convert/parse_azfp.py#L53-L58

Originally the scope of this PR was just `#1`, above. But It was such a small PR, and I've already created a bunch of smallish PR's, that I decided to reuse this PR by expanding its scope.

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -585,12 +585,23 @@ class SetGroupsAZFP(SetGroupsBase):
                     },
                 ),
                 "DS": (["channel"], parameters["DS"][self.freq_ind_sorted]),
-                "EL": (["channel"], parameters["EL"][self.freq_ind_sorted]),
-                "TVR": (["channel"], parameters["TVR"][self.freq_ind_sorted]),
+                "EL": (
+                    ["channel"],
+                    parameters["EL"][self.freq_ind_sorted],
+                    {"long_name": "Sound pressure at the transducer", "units": "dB"},
+                ),
+                "TVR": (
+                    ["channel"],
+                    parameters["TVR"][self.freq_ind_sorted],
+                    {
+                        "long_name": "Transmit voltage response of the transducer",
+                        "units": "dB re 1uPa/V at 1m",
+                    },
+                ),
                 "VTX": (
                     ["channel"],
                     parameters["VTX"][self.freq_ind_sorted],
-                    {"long_name": "Transmitter power supply"},
+                    {"long_name": "Amplified voltage sent to the transducer"},
                 ),
                 "Sv_offset": (["channel"], Sv_offset),
                 "number_of_samples_digitized_per_pings": (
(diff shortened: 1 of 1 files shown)


echopype/convert/set_groups_azfp.py
```diff
@@ -361,7 +361,7 @@ def set_beam(self) -> List[xr.Dataset]:
                 ),
                 "gain_correction": (
                     ["channel"],
-                    unpacked_data["gain"][self.freq_ind_sorted],
```
leewujung, line 364: I am curious why `backscatter_r` is set to float32 and `gain_correction` float64. Is this because of what the convention specifies?

emiliom: Heh, it's fairly arbitrary. I was trying to maintain consistency across echopype. `backscatter_r` is float32 in EK60/80 (probably not by design) while `gain_correction` for EK60 is float64. But we could easily change these decisions for AZFP.

The convention generally uses `float` (ie, float32) for most floating-point variables, including these two, whereas echopype typically ends up setting such variables to float64. That's a topic we could revisit in the future.

leewujung: I think we can set all of these to either float32 or float64, for self consistency?

leewujung (approved):
Looks good! I checked the types of the other variables, and they all just become either int64 or float64. Since these are not defined in the convention, I guess we just go with that?

emiliom:
> I checked the types of the other variables, and they all just become either int64 or float64. Since these are not defined in the convention, I guess we just go with that?

Sounds good to me.

echopype/convert/set_groups_azfp.py
emiliom: Do you mean `backscatter_r` and `gain_correction`, or other variables as well? There are a few other variables here and there that are also set to float32, probably not through explicit actions.

Resetting `backscatter_r` (and `backscatter_i`, for completeness) from float32 to float64 for ek60 and ek80 will likely add to the memory/size footprint (compression probably compensates to some extent). Do you think we should still go ahead and do that? Other than that, I assume there won't be any downstream consequences.

emiliom:
Thanks! Based on our side conversation, I'll merge this w/o further changes.

PR #1110, opened 2023-08-01
Changed test file name from .permission_test to UUID to make it thread safe

Sohambutala:
Thank you @lsetiawan, for this suggestion!


leewujung:
Hey @Sohambutala : Thanks for changing this! For the sake of documentation, could you edit the PR comment to put in short explanation on why this change is needed?

Issue #988, opened 2023-03-15
Unpin netCDF4 version causes failure


lsetiawan:
## Overview

This issue brings back #801. Currently we have a pin on `netCDF4`, which doesn't bring in the latest netCDF4 version `1.6.3`. After digging into this some more it seems like this is an issue with us using `zlib` as mentioned in the other PR. However, now netCDF4 actually allows for `compression` attribute to be used, see https://unidata.github.io/netcdf4-python/#Dataset.createVariable. But, the package we depend on, xarray doesn't currently have this option as mentioned in https://github.com/pydata/xarray/issues/7388. Someone has put in a PR for this fix 3 weeks ago, but it's still active, so at some point, we can potentially unpin and use the `zstd` compression like we are using with `zarr` as seen in https://github.com/OSOceanAcoustics/echopype/blob/f12322c78c0d5b802aed30a38ae7f4c1ad9e01b3/echopype/utils/coding.py#L20-L25.

emiliom:
> After digging into this some more it seems like this is an issue with us using `zlib` as mentioned in the other PR. 

This has been addressed in #1112. Now `zlib` is set to `False` for string variables.

> However, now netCDF4 actually allows for `compression` attribute to be used, see https://unidata.github.io/netcdf4-python/#Dataset.createVariable. But, the package we depend on, xarray doesn't currently have this option as mentioned in https://github.com/pydata/xarray/issues/7388. Someone has put in a PR for this fix 3 weeks ago (https://github.com/pydata/xarray/pull/7551), but it's still active, so at some point, we can potentially unpin and use the `zstd` compression like we are using with `zarr`

PR #1112 doesn't introduce the use of a `compression` attribute for netCDF4. The core issue was the failure with variable-length string variables when compression via `zlib` was set to `True`. That has been fixed by setting `zlib` to `False` for all string variables. The use of a `compression` attribute per se is left for future consideration as an enhancement (not a bug fix -- that's already done!), once it's implemented in xarray.

PR #1111, opened 2023-08-02
Unpin netcdf4 and pandas to check what is breaking [all tests ci]

ocefpaf:
I'd like to run this locally but I don't have access to the test data drive to run the full test suite.

### DIFF
--- a/echopype/utils/coding.py
+++ b/echopype/utils/coding.py
@@ -15,7 +15,7 @@ DEFAULT_TIME_ENCODING = {
 }
 
 COMPRESSION_SETTINGS = {
-    "netcdf4": {"zlib": True, "complevel": 4},
+    "netcdf4": {"zlib": False, "complevel": 4},
     # zarr compressors were chosen based on xarray results
     "zarr": {
         "float": {"compressor": zarr.Blosc(cname="zstd", clevel=3, shuffle=2)},
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,8 +6,7 @@ pynmea2
 pytz
 scipy
 xarray
-# https://github.com/pydata/xarray/pull/7650
-pandas<2
+pandas
 zarr
 fsspec
 s3fs
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 dask[array,distributed]
 jinja2
-# https://github.com/Unidata/netcdf4-python/issues/1175#issuecomment-1173142506
-netCDF4<1.6
+netCDF4>=1.6
 numpy
 pynmea2
 pytz


ocefpaf:
@lsetiawan I need your help understanding the CIs here. It seems that the tests did not run, is that correct?

emiliom:
Thanks for chiming in @ocefpaf ! First, our development branch is `dev`; I suggest running tests there rather than on `main`. In `dev` we already removed the Pandas pinning, and all is good there. But we are running into a number of issues with netCDF4. #988 describes one of them, but lately we've been running into more that we haven't documented in a GitHub issue yet.

The test data are packaged in a Docker image that's rebuilt daily. The tests can be run locally with [these instructions](https://echopype.readthedocs.io/en/stable/contributing.html#tests-and-test-infrastructure), which first deploy two docker images that include the test data.

lsetiawan:
The current failure seen in CI is exactly this one: https://github.com/Unidata/netcdf4-python/issues/1175

It is the reason we pinned 😬

ocefpaf:
@lsetiawan I could not find any other issue besides the variable length compression. If so, let me know. I guess you have everything you need to bring this one home. I'll close this b/c you know better the internals of the package to find the strings and other vlen, to remove compression from them.

PS: As we discussed on slack, https://github.com/Unidata/netcdf-c/pull/2716 will "fix this" but it is better to explicitly disable compression on variables that won't be compressed IMO. Also, that would fix this right now instead of whenever they make a new release ;-p

ocefpaf:
> The current failure seen in CI is exactly this one: [Unidata/netcdf4-python#1175](https://github.com/Unidata/netcdf4-python/issues/1175)

This one https://github.com/Unidata/netcdf4-python/issues/1205 probably describes is better and has the links to the issue upstream in netcdf-c.

ocefpaf:
> Thanks for chiming in @ocefpaf ! First, our development branch is `dev`; I suggest running tests there rather than on `main`. In `dev` we already removed the Pandas pinning, and all is good there. But we are running into a number of issues with netCDF4. #988 describes one of them, but lately we've been running into more that we haven't documented in a GitHub issue yet.
> 
> The test data are packaged in a Docker image that's rebuilt daily. The tests can be run locally with [these instructions](https://echopype.readthedocs.io/en/stable/contributing.html#tests-and-test-infrastructure), which first deploy two docker images that include the test data.

I don't have any stake on this but having non-standard branches, convoluted way to obtain data for running tests, and title coded messages to activate them, makes this less open to contributions. I would suggest to have at least a subset of the tests to run without jumping through hoops. In the long term, have the test data automatically fetched with pooch would be even better.

Anyway, thanks to @lsetiawan's guidance we figured out the problem and a fix is on the way!

PR #1112, opened 2023-08-03
feat: pin netcdf to >1.6 and add explicit encoding [all tests ci]

lsetiawan:
## Overview

This PR unpins the netCDF4 version and set it to `>1.6` so that we are getting the latest. Also, the `set_netcdf_encodings` utils has been updated so that we set `zlib=False` for strings... it looks like the older netcdf4 is already doing this under the hood.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -66,7 +66,7 @@ jobs:
       # We only want to install this on one run, because otherwise we'll have
       # duplicate annotations.
       - name: Install error reporter
-        if: ${{ matrix.runs-on }} == 'ubuntu-latest' and ${{ matrix.python-version }} == '3.9'
+        if: ${{ matrix.python-version == '3.9' }}
         run: |
           python -m pip install pytest-github-actions-annotate-failures
       - name: Install echopype
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -63,6 +63,12 @@ jobs:
       - name: Install dev tools
         run: |
           micromamba install -c conda-forge -n ${{ env.CONDA_ENV }} --yes --file requirements-dev.txt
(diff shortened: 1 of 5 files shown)


emiliom:
I've reviewed the PR, and it looks good. Next, I'll run it locally.

emiliom (approved):

emiliom:
Everything seems to be working great!

Here's the encoding on `Beam_group1.channel`, with a conda environment built with netcdf4 1.6:
```python

In [7]: ds16 = xr.open_dataset(
   'with_netcdf4v16/Summer2017-D20170615-T190214.nc', group='Sonar/Beam_group1', decode_cf=False
)

In [8]: ds16.channel.encoding
Out[8]:
{'zlib': False,
 'shuffle': False,
 'complevel': 0,
 'fletcher32': False,
 'contiguous': True,
 'chunksizes': None,
 'source': '/home/mayorga/.echopype/temp_output/with_netcdf4v16/Summer2017-D20170615-T190214.nc',
 'original_shape': (3,),
 'dtype': str}
```

With netcdf4 1.5, the encoding is identical.

File sizes are trivially different, with the netcdf4 1.6 nc file being 10 bytes smaller out of 1.4 MB.

leewujung:
Thanks @lsetiawan @emiliom !

To close the loop: the changes boiled down to set string type encoding with `"zlib" : False`. It seems this PR addresses everything in #988. Can you confirm?

emiliom:
> To close the loop: the changes boiled down to set string type encoding with "zlib" : False. 

Yup.

> It seems this PR addresses everything in https://github.com/OSOceanAcoustics/echopype/issues/988. Can you confirm?

Ultimately, yes. The core motivation for that issue has been fully addressed. But on re-reading, there are some subtleties that I'll comment on there.

PR #1114, opened 2023-08-05
Add missing sampling_offset_time for EK60 & set EK60/80 beam_direction_x/y/z to nan as needed

emiliom:
- `sampling_offset_time` is missing from EK60! I've added it. See #1097
- Set EK60/80 `beam_direction_x/y/z` to nan when all 3 are zero

### DIFF
--- a/echopype/convert/set_groups_ek60.py
+++ b/echopype/convert/set_groups_ek60.py
@@ -439,6 +439,16 @@ class SetGroupsEK60(SetGroupsBase):
                 for ch_seq in self.sorted_channel.keys()
             ]
 
+        for i, ch in enumerate(self.sorted_channel.keys()):
+            if (
+                np.isclose(beam_params["dir_x"][i], 0.00)
+                and np.isclose(beam_params["dir_y"][i], 0.00)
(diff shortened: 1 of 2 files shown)


leewujung (approved):
Thanks, these look good! 

Thinking about it again, the (0, 0, 0) combination is invalid, but all combinations that do not give a length=1 vector would also be invalid, but I could see how people might put a non-normalized vector in the (x,y,z), so the check has to handle something else or determine whether to normalize the length by default, etc, etc. I wonder if this messiness is why convention ver.2 changed the definition to use angles in the spherical coordinate system.

Anyways, please feel free to merge!

emiliom:
Thanks!

> Thinking about it again, the (0, 0, 0) combination is invalid, but all combinations that do not give a length=1 vector would also be invalid, but I could see how people might put a non-normalized vector in the (x,y,z), so the check has to handle something else or determine whether to normalize the length by default, etc, etc. 

Agreed. More generally, echopype doesn't do any conversion or checking (other than what I just added) of the ek60 and ek80 datagram values that are fed into `beam_direction_x/y/z`. That may be true for other parameters, too.

PR #1117, opened 2023-08-05
Set time arrays explicitly to datetime64[ns] in parsers; remove ZarrCombine & duplicate / old ping_time stuff [all tests ci]

emiliom:
We're now getting these run-time warnings on `open_raw`:

> .../OSOceanAcoustics/echopype/echopype/convert/set_groups_ek80.py:161: UserWarning: Converting non-nanosecond precision datetime values to nanosecond precision. This behavior can eventually be relaxed in xarray, as it is an artifact from pandas which is now beginning to support non-nanosecond precision values. This warning is caused by passing non-nanosecond np.datetime64 or np.timedelta64 values to the DataArray or Variable constructor; it can be silenced by converting the values to nanosecond precision ahead of time.
  ds = xr.Dataset(

There don't seem to be any actual, negative consequences, but it's off-putting and unnecessarily concerning to users. For EK80, there are about 20 of these warnings being generated!

The warnings are likely resulting from the unppinning of Pandas, which now results in Pandas 2 being used.

The solution is to recast or explicitly set every parsed time array to datetime64[ns], in the parsers.

- [x] AZFP
- [x] EK60
- [x] EK80


emiliom:
@leewujung let's see if the CI's all pass. Based on a limited EK80 test I ran, all the ek80 warnings are gone. But I'm puzzled why changing this "[ms]" occurrence below results in an error:

https://github.com/OSOceanAcoustics/echopype/blob/cd5264597d373b53833aa6241b9cb1cab8529b22/echopype/convert/parse_base.py#L117-L119

Please take a look at both the changes I made and whether any more changes need to be made.

leewujung:
@emiliom: so for that timestamp setting issue related to `config_datagram`, the actual issue lies in `parse_base.py::_print_status` where we print out the first timestamp of the raw file. The `.tolist` is an ugly way to getting to the underlying numbers (which is identical to results of `.astype(datetime)`, but I feel this one is a little confusing so opted for tolist), which can be formatted by strftime.

leewujung:
I ran all the convert tests locally, and there is one test failing `test_convert_ek60_duplicate_ping_times` where `"old_ping_time" in ed["Provenance"]` does not exist in ed["Provenance"]:
https://github.com/OSOceanAcoustics/echopype/blob/6b734eb86958c036919216719a8b3c6e1b41c0a1/echopype/tests/convert/test_convert_ek60.py#L170

@emiliom : do you know what this is? I don't remember anything about it and haven't tried to look into that file, but it seems that there's a duplicated ping_time and it gets correctly automatically (??) and then the original timestamps are preserved in the Provenance group?

emiliom:
Hmm. I've confirmed the test failure locally. I've been looking into it and making some headway in understanding the problem. 

But I'm also puzzled that in this PR, the test CI's for 3.9 and 3.10 are displayed as successful, but when you actually drill in to the test details (eg, https://github.com/OSOceanAcoustics/echopype/actions/runs/5843062232/job/15844826484?pr=1117#step:14:490), the error is actually there! I specifically added the `[all tests ci]` string to the PR title when I submitted the PR to ensure all tests were run.

leewujung:
Ok, I think I know what's going on. What I was able to trace include the following:
- in `set_groups_ek60.py::set_provenance`: data variable `duplicate_ping_times` is set to 0 or 1 depending on if `self.old_ping_time` is None.
- but `duplicate_ping_times` only appears in `zarr_combine.py`, which is deprecated after the new `combine_echodata` code
- `self.old_ping_time` only appears in `SetGroupsEK60.__init__`, and the operations are somewhat convoluted (though it boils down to finding if there is a duplicated `ping_time` timestamp, and if so, add +1ns to the duplicated timestamp if the backscatter data is not identical, or drop the duplicated timestamp if the backscatter data is indeed duplicated for all values.

---------------------------------

In terms of actions, my thoughts are:
- remove `zarr_combine` completely in this release, since it is not used anymore; this module is all internal, so there is no actual impact in user-facing things
- remove `self.old_ping_time` and all the related mechanisms in `SetGroupsEK60.__init__`, like below: https://github.com/OSOceanAcoustics/echopype/blob/6b734eb86958c036919216719a8b3c6e1b41c0a1/echopype/convert/set_groups_ek60.py#L84-L129
- remove `duplicate_ping_times` from the Provenance group

---------------------------------

Also, I am still not sure why the two `assert` statements are in `test_convert_ek60.py::test_convert_ek60_duplicate_ping_times`, since there is no duplicated `ping_time` timestamps in this file. Due to the various changes we just introduced in `dev` among xarray/pandas/netcdf4, I tested 2 environments, one is "old" (pandas=1.5.3, netcdf4=1.5.8, xarray=2023.7.0), and the other is "new" (pandas=2.0.3, netcdf4=1.6.4, xarray=2023.7.0), and for both environments the parsed `ping_time` does not have duplicated timestamps. I wonder if we should just remove this test altogether.

---------------------------------

I'll next look into the various `ping_time` errors/warnings revealed in the existing tests...

leewujung:
Some more messing around revealed the different timestamps resulted from `ns` or `ms` unit. It does make sense that these are different, but we need to decide what we want to do...

Using the file in `test_plot.py::test_vertical_offset_echodata` as an example, the first `ping_time` timestamp looks like the following:
- With everything set to `ns` unit (ie with the changes made in this PR):
   Screenshot 2023-08-13 at 4 37 47 PM

- With the timestamps set to `ms` unit (ie _before_ changes made in this PR):
   Screenshot 2023-08-13 at 4 39 12 PM

These values do not have practical differences, and we can make the test pass by either using index-based selection or doing `method="nearest"`.

The puzzling results of duplicated timestamp `test_convert_ek60.py::test_convert_ek60_duplicate_ping_times` is induced from the same changes:
- With everything set to `ns` unit (ie with the changes made in this PR): there is no duplicated timestamps in `ping_time`
- With the timestamps set to `ms` unit (ie _before_ changes made in this PR): there is no duplicated timestamps in `ping_time`


I wonder what we should do in this case...

emiliom:
Thanks @leewujung ! I reviewed the issues, and came to agreement with your recommendations. I opened a new issue, #1122, to give more visibility to the larger changes we're talking about here, for future reference.

I tried to do the removal of `zarr_combine.py` and the associated test in a separate PR, but ran into a problem of the duplicate - ping_time test no longer passing. I ended up including everything in this PR, though I now, just now, see that I could've just included the deletion of `zarr_combine.py` and `test_zarr_combine.py` in that PR :disappointed:. Oh well.

With respect to the deletion of the `duplicate_ping_time` and `old_ping_time` code, see my comments in #1122. I chose to remove them too, but included those changes in a single commit for easier tracking in the future, if needed.

Regarding the impact of `np.datetime64` unit (resolution), I had put together a stand-alone illustration similar to yours:

```python
dt_str = '2001-01-01T00:00:00.923654'

np.datetime64(dt_str, 's')
Out[27]: numpy.datetime64('2001-01-01T00:00:00')
np.datetime64(dt_str, 'ms')
Out[28]: numpy.datetime64('2001-01-01T00:00:00.923')
np.datetime64(dt_str, 'ns')
Out[29]: numpy.datetime64('2001-01-01T00:00:00.923654000')
```

Previously, with "ms", the original, finer resolution would've been clipped to "ms", then reset to "ns" with zero padding (+/- floating point error) in the backend to adhere to the Pandas < 2.0 limitation.

emiliom:
This is now ready to merge.

leewujung (approved):
Awesome, I think we can merge this now!

PR #1119, opened 2023-08-08
ci: fix discrepancy on test log output

lsetiawan:
## Overview

This PR fixes the discrepancy on test log output filename for PR action... currently this is not right and in summary it shows that it can't find the file.


lsetiawan:
Seems to work well now.

Issue #888, opened 2022-11-24
Further improve `compute_MVBS` in computation across ping time and channel


leewujung:
The efficiency of `compute_MVBS` was enhanced in #878, but the current implement may be inefficient if the data is very large -- although @b-reyes and I agreed that the current implementation is "good enough" for the couple use cases we are considering. Nevertheless, our thoughts are documented below.

### Binned averages across ping time
In #878 the major efficiency bottleneck for computing MVBS is resolved by efficiently distributing the first set of binned averages along `echo_range`. To avoid the dask graph getting too large, these binned averages are computed first before the second set of binned averages are compute along `ping_time`. This means that the outcomes of the first set (`er_means` in the code referenced below) are in-memory:
https://github.com/OSOceanAcoustics/echopype/blob/b0b1490a9081400943ac2606e2e6fd2deb4ae71e/echopype/preprocess/mvbs.py#L391-L393

This has 2 potential problems:
- Memory consumption may be an issue if the Sv data is very large
- The second set of binned averages are computed sequentially so potentially inefficient if the Sv data is very large

There are two possible solutions:
- Create a temporary zarr store to store `er_means` and lazy-load the data back for efficient computation.
- Manually chunk and scatter the in-memory `er_means` data into different workers to compute

### Repeat the same computation across channels
In addition, currently the computation across different channels are also done sequentially: https://github.com/OSOceanAcoustics/echopype/blob/b0b1490a9081400943ac2606e2e6fd2deb4ae71e/echopype/preprocess/mvbs.py#L444
Also something to consider, but if each frequency is already computed efficiently it is likely better to not overdo this.

lsetiawan:
Done in #1124

Issue #1002, opened 2023-03-23
compute_MVBS should propagate and bin average Sv latitude and longitude, and depth/water_level


emiliom:
When the `Sv` input to `commongrid.compute_MVBS` has `latitude` and `longitude` variables (ie, `consolidate.add_latlon` was applied after `compute_Sv`), these variables are currently not maintained in the resulting MVBS dataset. They should be propagated and bin averaged on ping_time.

emiliom:
This issue is being addressed in #1124. Once that PR is merged, this issue can be closed.

leewujung:
This can now be closed!

PR #1124, opened 2023-08-15
fix(commongrid):  improve 'compute_MVBS' using flox [all tests ci]

lsetiawan:
## Overview

This PR introduces the package [`flox`](https://flox.readthedocs.io/en/latest/), which provides strategies for fast GroupBy reductions with `dask.array`. The slides for this can be found [here](https://docs.google.com/presentation/d/1YubKrwu9zPHC_CzVBhvORuQBW-z148BvX3Ne8XcvWsQ/edit#slide=id.p). I was first introduced with this package by Deepak in scipy and seems like it's a very powerful package to essentially address https://github.com/OSOceanAcoustics/echopype/issues/888.

The current function that is implemented here is the [`xarray_reduce`](https://flox.readthedocs.io/en/latest/generated/flox.xarray.xarray_reduce.html#flox.xarray.xarray_reduce) function, which allows for groupby across multiple dimension. In this case, we'd like to group by across the `ping_time` and `echo_range`.


lsetiawan:
Adding here that xarray actually uses flox under the hood when installed for groupby operations with dask arrays!

Screenshot 2023-08-17 at 9 08 08 AM

echopype/commongrid/mvbs.py
```diff
@@ -404,9 +405,32 @@ def bin_and_mean_2d(
...
+    inverse : bool
+        If True, perform inverse transform
+
+    Returns
+    -------
+    dask.array.Array or np.ndarray
+        The transformed data
+    """
+    if inverse:
+        return 10 * dask.array.log10(data)
```
lsetiawan, line 427:
```suggestion
        return 10 * np.log10(data)
```

lsetiawan: Can just use regular numpy function even if it's dask.

echopype/commongrid/mvbs.py
```diff
@@ -417,45 +441,34 @@ def get_MVBS_along_channels(
...
+    sv = ds_Sv["Sv"].pipe(_linear_transform)
+
+    # reduce along ping_time and echo_range
+    # by binning and averaging
+    mvbs = xarray_reduce(
+        sv,
+        sv["channel"],
+        ds_Sv["ping_time"],
+        ds_Sv["echo_range"],
+        func="mean",
```
lsetiawan, line 466:
```suggestion
        func="nanmean",
```

echopype/commongrid/mvbs.py
```diff
@@ -417,45 +441,34 @@ def get_MVBS_along_channels(
...
+        expected_groups=(None, ping_interval, echo_range_interval),
+        isbin=[False, True, True],
+        method="map-reduce",
```
lsetiawan, line 469: Maybe this should be exposed to the user..? After some conversation with @leewujung, "blockwise" might be better.

leewujung:
Thanks @lsetiawan for this exciting improvement!! I did a first round of testing and added a test in https://github.com/lsetiawan/echopype/pull/2 using mock Sv datasets with irregular `echo_range` along `ping_time`, and `flox` handles it correctly. 😀

lsetiawan:
## TODO

Test this PR with various methods in flox and compare performance.

leewujung:
Not to mission creep, but do you think you can throw in one more line to do #1002 for binned averaging lat/lon along `ping_time`?

echopype/commongrid/api.py
```diff
@@ -63,10 +66,18 @@ def _set_MVBS_attrs(ds):
 
 
 @add_processing_level("L3*")
-def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
+def compute_MVBS(
+    ds_Sv: xr.Dataset,
+    range_var: Literal["echo_range", "depth"] = "echo_range",
+    range_meter_bin: int = 20,
```
leewujung, line 72: I wonder if we can make this `range_bin` to parallel the new `range_var`, and have the input pattern be of the form "20m" to parallel the `ping_time_bin`, and we parse out the number 20 while do not allow other units (it does not make sense to do MVBS for km, and probably not cm in most cases).

echopype/commongrid/api.py
```diff
@@ -76,41 +87,83 @@ def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
...
+    if range_var not in ["echo_range", "depth"]:
+        raise ValueError("range_var must be one of 'echo_range' or 'depth'.")
+
+    # Check if range_var exists in ds_Sv
+    if range_var not in ds_Sv.data_vars:
+        raise ValueError(f"range_var '{range_var}' does not exist in the input dataset.")
+
     # create bin information for echo_range
-    range_interval = np.arange(0, ds_Sv["echo_range"].max() + range_meter_bin, range_meter_bin)
+    # this computes the echo range max since there might be missing values
```
leewujung, line 127: > since there might be missing values

Not sure I understand this. Do you mean there may be NaN in `echo_range`?

echopype/commongrid/mvbs.py
```diff
@@ -404,9 +407,37 @@ def bin_and_mean_2d(
     return final
 
 
+def _linear_transform(
```
leewujung, line 410: I think we can move this function to new module `echopype/utils/compute.py` since it's something that's used in multiple functions, like `compute_NASC`.

echopype/tests/commongrid/test_mvbs.py
```diff
@@ -833,3 +421,51 @@ def test_bin_and_mean_2d(bin_and_mean_2d_params) -> None:
...
+    )
+    
+    # Check that the range_var is in the dimension
+    assert f"{range_var}_bins" in raw_MVBS.dims
+    
+    # When it's echo_range and lat_lon, the dataset should have positions
+    if range_var == "echo_range" and lat_lon is True:
+        assert raw_MVBS.attrs["has_positions"] is True
+    else:
+        assert raw_MVBS.attrs["has_positions"] is False
```
leewujung, line 471: Since the added lat/lon and depth via `ds_Sv_er_regular_w_latlon` and `ds_Sv_er_regular_w_depth` are fixed values, how about adding the expected reduced values here in the test?

leewujung (changes requested):
@lsetiawan : Thanks for this great PR and for introducing `flox`! I am very excited to thoroughly test out the performance side of things once this is in the package.

Most of my inline comments are small. But I think we'll benefit to have a unit test that tests for the actual mean values and that NaN is skipped properly in the binned average process in `get_MVBS_along_channels`. Having this confidence will go a long way in that we can truly move on to focusing on the performance side of things after this and don't have to always worry about the values. This category of tests were in place before (though in a pretty messy form) in `test_bin_and_mean_2d`, which we will remove since all the `bin_and_mean_2d` mechanisms are no longer used.

Instead of trying to figure out what was in the old tests, I'd suggest to just create new mock datasets that have Sv values that are structured in a way that expected MVBS values are straightforward to get. I would suggest including the following cases:
- every thing in regular: for example, you have repeated N columns of 1, 2, 3 (so total size along `ping_time` is N*3), and `compute_MVBS` binned-averages over N columns so you get reduced array with 1, 2, 3.
- some irregularity (this one is important): set up is similar to the above (total N*3 columns along the `ping_time` dimension), but the `ping_time_bin` used in `compute_MVBS` is smaller than N. This way you introduce irregularity on combination of values.
- some NaN: set some `echo_range` pixels to be NaN, and verify that the final MVBS output does not contain NaN, and that the binned averaged values are indeed with the NaN skipped.

These arrays should probably be small so that it is easy to craft the expected MVBS values.

echopype/tests/commongrid/test_api.py
```diff
@@ -0,0 +1,164 @@
...
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    ("er_type"),
+    [
+        ("regular"),
+        ("irregular"),
+    ],
+)
+def test_compute_MVBS_er_output(request, er_type):
```
leewujung, line 132: This should probably be changed to make it less cryptic.
```suggestion
def test_compute_MVBS_range_output(request, er_type):
```

echopype/commongrid/api.py
lsetiawan: > NaN in echo_range

Yepp exactly

lsetiawan:
> some NaN: set some echo_range pixels to be NaN, and verify that the final MVBS output does not contain NaN, and that the binned averaged values are indeed with the NaN skipped.

I was doing some testing with this and seems like NaNs are just propagated by numpy per [documentation](https://numpy.org/doc/stable/reference/generated/numpy.nanmean.html#numpy-nanmean): `Nan is returned for slices that contain only NaNs.`... do we wanna drop it?

lsetiawan:
@emiliom I need your help figuring out what's going on now with the failure on `echopype/tests/utils/test_processinglevels_integration.py` something with `processing_level` should be missing after a `compute_MVBS`? https://github.com/OSOceanAcoustics/echopype/actions/runs/6054674562/job/16432456915#step:15:603 Thanks!

emiliom:
> @emiliom I need your help figuring out what's going on now with the failure on echopype/tests/utils/test_processinglevels_integration.py something with processing_level should be missing after a compute_MVBS? https://github.com/OSOceanAcoustics/echopype/actions/runs/6054674562/job/16432456915#step:15:603 Thanks!

Darn. I'll take a look.

leewujung:
> I was doing some testing with this and seems like NaNs are just propagated by numpy per [documentation](https://numpy.org/doc/stable/reference/generated/numpy.nanmean.html#numpy-nanmean): `Nan is returned for slices that contain only NaNs.`... do we wanna drop it?

So I think it depends on the pattern of NaN you put in:
- If some 2D `ping_time` x `echo_range` bins contain entirely only NaN values, the reduced pixels from those bins will be NaN -- that is the correct behavior and we will not drop them.
- if there are just sporadic single pixels _within_ those 2D bins that are NaN, the reduced pixels from those bins will not be NaN

I hope this make sense?

echopype/tests/utils/test_processinglevels_integration.py
```diff
@@ -129,5 +129,5 @@ def _freqdiff_applymask(test_ds):
     # No product level is assigned because at present compute_MVBS drops the lat/lon data
     # associated with the input Sv dataset
     # ds = ds.rename_vars(name_dict={"Sv": "Sv_unmasked", "Sv_ch0": "Sv"})
-    mvbs_ds = ep.commongrid.compute_MVBS(ds, range_meter_bin=30, ping_time_bin='1min')
+    mvbs_ds = ep.commongrid.compute_MVBS(ds, range_meter_bin="30m", ping_time_bin='1min')
     _absence_test(mvbs_ds)
```
emiliom, line 133: Now that `compute_MVBS` can preserve lat-lon (YAY!!), a processing level attribute is expected.
```suggestion
    _presence_test(mvbs_ds, "Level 3B")
```

emiliom: I'll take the liberty of committing my "suggestion" ...

echopype/commongrid/api.py
```diff
@@ -63,10 +71,18 @@ def _set_MVBS_attrs(ds):
 
 
 @add_processing_level("L3*")
-def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
+def compute_MVBS(
+    ds_Sv: xr.Dataset,
+    range_var: Literal["echo_range", "depth"] = "echo_range",
+    range_meter_bin: str = "20m",
```
leewujung, line 77: Could you change this to `range_bin` to be parallel with `range_var`, now that we have the meter in "20m"? Thanks!

emiliom:
Chiming in on the use of `pint` for  handling `range_meter_bin` / `range_bin`. I'll paste here @leewujung's earlier comments:

> I wonder if we can make this range_bin to parallel the new range_var, and have the input pattern be of the form "20m" to parallel the ping_time_bin, and we parse out the number 20 while do not allow other units (it does not make sense to do MVBS for km, and probably not cm in most cases).

What you are saying is that this argument will only be in meters -- which to me makes sense. But in this PR @lsetiawan added functionality based on `pint` to parse, check and convert (to meters, if necessary) the units. From what I can see, `pint` is used only for that purpose (lines 125-141). But if the argument only accepts values in meters, the use of `pint` here is unnecessary. I'd suggest removing that functionality and removing `pint` from `requirements.txt`. 

I'm tickled to run into a case (not just in echopype!) where there is an obvious argument for using `pint` (I've never used it), but this doesn't seem to be it.

leewujung:
Yeah, I think strings like 5m, 5cm, 5km can be parsed easily with `re`. We have that in #1106 for parsing out the freq values and units separately. But in the use of `compute_MVBS`, we can just stick to allowing "m" only, and hence does not need to have `pint` as a dependency.

lsetiawan:
> Yeah, I think strings like 5m, 5cm, 5km can be parsed easily with re

Alrighty. I was trying to be fancy and allow for conversions and make things more unit aware, but if this is all we want then I'll remove to just have a hard case of allowing only "m". Thanks for the inputs @emiliom and @leewujung ! 😄

echopype/commongrid/api.py
```diff
@@ -76,41 +88,102 @@ def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
...
+    # normalize to lower case
+    # for range_bin
+    range_bin = range_bin.strip().lower()
+    # Only matches meters
+    match_obj = re.match(r"(\d+)(\s+)?(m)", range_bin)
+
+    # Do some checks on range meter inputs
+    if match_obj is None:
+        # This shouldn't be other units
+        raise ValueError("Found incompatible units. Must be in meters.")
```
emiliom, line 130: Making the error message less ambiguous; there are no acceptable "compatible" units. It must be in meters.
```suggestion
        raise ValueError("range_bin must be in meters (e.g., '10m').")
```

echopype/commongrid/api.py
```diff
@@ -76,41 +88,102 @@ def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
...
+        Must be one of ``echo_range`` or ``depth``.
+        Note that ``depth`` is only available if the input dataset contains
+        ``depth`` as a data variable.
+    range_bin : str
+        bin size along ``echo_range`` or ``depth`` in meters,
+        default to ``20m``
     ping_time_bin : str
         bin size along ``ping_time``, default to ``20S``
+    method: str
+        The flox strategy for reduction of dask arrays only.
```
emiliom, line 102:
```suggestion
        The flox strategy for reduction of dask arrays only, 
        default to ``map-reduce``
```

echopype/tests/utils/test_processinglevels_integration.py
```diff
@@ -129,5 +129,5 @@ def _freqdiff_applymask(test_ds):
     # No product level is assigned because at present compute_MVBS drops the lat/lon data
     # associated with the input Sv dataset
```
emiliom, line 130:
```suggestion
```

echopype/commongrid/api.py
```diff
@@ -76,41 +88,102 @@ def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
...
+    # Only matches meters
+    match_obj = re.match(r"(\d+)(\s+)?(m)", range_bin)
+
+    # Do some checks on range meter inputs
+    if match_obj is None:
+        # This shouldn't be other units
+        raise ValueError("Found incompatible units. Must be in meters.")
+
+    # Convert back to float
+    range_bin = float(match_obj.group(1))
```
emiliom, line 133: If the numerical part of the argument is a decimal number, this won't work correctly (I tested the regex at https://regex101.com/). It looks like changing the regex to `r"([\d+]*[.,]{0,1}[\d+]*)(\s+)?(m)"` would do the trick. See https://regex101.com/r/asFP2F/1

echopype/commongrid/api.py
```diff
@@ -76,41 +88,102 @@ def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
...
+        raise TypeError("range_bin must be a string")
+
+    if not isinstance(ping_time_bin, str):
+        raise TypeError("ping_time_bin must be a string")
+
+    # normalize to lower case
+    # for range_bin
+    range_bin = range_bin.strip().lower()
+    # Only matches meters
+    match_obj = re.match(r"(\d+)(\s+)?(m)", range_bin)
```
emiliom, line 125: See comment further down regarding conversion of decimal numbers to float.
```suggestion
    match_obj = re.match(r"([\d+]*[.,]{0,1}[\d+]*)(\s+)?(m)", range_bin)
```

emiliom (changes requested):
It looks like your regex would not handle decimal `range_bin` numbers. I've suggested a change to the regex that does. Other than that, I made a few suggestions to user-facing strings.
Thanks!

lsetiawan:
> It looks like your regex would not handle decimal range_bin numbers.

Ooops thanks for catching that!

emiliom (approved):
Alright, the tests are now passing. Approving!

lsetiawan:
@leewujung This is ready for another round of your review. Thanks! 😄

echopype/commongrid/api.py
```diff
@@ -76,41 +88,103 @@ def compute_MVBS(ds_Sv, range_meter_bin=20, ping_time_bin="20S"):
...
+    # Only matches meters
+    match_obj = re.match(r"([\d+]*[.,]{0,1}[\d+]*)(\s+)?(m)", range_bin)
+
+    # Do some checks on range meter inputs
+    if match_obj is None:
+        # This shouldn't be other units
+        raise ValueError("range_bin must be in meters (e.g., '10m').")
+
+    # Convert back to float
+    range_bin = float(match_obj.group(1))
```
leewujung, line 134: I think we can have this as a small utility function that can also be used in `compute_NASC` #1167.

echopype/tests/convert/utils/test_ek.py
```diff
@@ -0,0 +1,14 @@
...
+def dgram_zarr_vars():
+    return {'power': ['timestamp', 'channel'], 'angle': ['timestamp', 'channel']}
+
+def test__get_power_dims(dgram_zarr_vars):
+    power_dims = _get_power_dims(dgram_zarr_vars)
+    assert isinstance(power_dims, list)
+    assert sorted(power_dims) == sorted(['timestamp', 'channel'])
+
+def test__extract_datagram_dfs():
+    ...
```
leewujung, line 14: It seems that `test__extract_datagram_dfs` is not finished and some import elements do not exist. Are these remnants from other ongoing PRs that shouldn't be in here?

echopype/tests/commongrid/test_api.py
```diff
@@ -0,0 +1,256 @@
...
+)
+def test_compute_MVBS_bin_inputs_fail(ds_Sv_er_regular, range_bin, ping_time_bin):
+    expected_error = ValueError
+    if isinstance(range_bin, int) or isinstance(ping_time_bin, int):
+        expected_error = TypeError
+        match = r"\w+ must be a string"
+    else:
+        match = r"range_bin must be in meters"
+
+    with pytest.raises(expected_error, match=match):
```
leewujung, line 60: cool!

echopype/tests/commongrid/test_api.py
```diff
@@ -0,0 +1,256 @@
...
+                chan_expected.append([False if v > 0 else True for v in hist])
+            expected_outs.append(chan_expected)
+        return np.array(expected_outs)
+
+    range_bin = "2m"
+    ping_time_bin = "1s"
+
+    if er_type == "regular":
+        ds_Sv = request.getfixturevalue("mock_sv_dataset_regular")
+        expected_mvbs = request.getfixturevalue("mock_mvbs_array_regular")
```
leewujung, line 236: I noticed that the values to be tested against are hard-coded. I think it should be computed on the fly using a brute-force loop for the most accurate verification, sine the dataset is tiny.

I did that, and the results are different. If you compute `expected_mvbs_val` using the code below, you'll notice that it does not correspond to the values given by flox. I chased down the cause of it, and it is that along the `echo_range` axis: for the first bin (range between 0 and 2), flox seems to slice out indices that correspond to `echo_range` values (0.5, 1, 1.5, 2), instead of (0, 0.5, 1, 1.5).

Could you look into what's going on, and change the test to be against brute-force looping outputs? Thanks!
```python
    # create bin information needed for ping_time
    d_index = (
        ds_Sv["ping_time"]
        .resample(ping_time=ping_time_bin, skipna=True)
        .first()  # Not actually being used, but needed to get the bin groups
        .indexes["ping_time"]
    )
    ping_interval = d_index.union([d_index[-1] + pd.Timedelta(ping_time_bin)]).values

    # create bin information for echo_range
    # this computes the echo range max since there might NaNs in the data
    echo_range_max = ds_Sv["echo_range"].max()
    range_interval = np.arange(0, echo_range_max + 2, 2)

    sv = ep.utils.compute._log2lin(ds_Sv["Sv"].data)

    expected_mvbs_val = np.ones((2, len(ping_interval)-1, len(range_interval)-1)) * np.nan
    for ch_idx in range(2):
        for p_idx in range(len(ping_interval)-1):
            for r_idx in range(len(range_interval)-1):
                p_idx_active = np.logical_and(
                    ds_Sv["ping_time"].data >= ping_interval[p_idx],
                    ds_Sv["ping_time"].data < ping_interval[p_idx+1]
                )
                r_idx_active = np.logical_and(
                    ds_Sv["echo_range"].isel(channel=0, ping_time=0).data >= range_interval[r_idx],
                    ds_Sv["echo_range"].isel(channel=0, ping_time=0).data < range_interval[r_idx+1]
                )
                # print(p_idx_active)
                # print(r_idx_active)
                sv_tmp = sv.data[ch_idx][np.ix_(p_idx_active, r_idx_active)]
                expected_mvbs_val[ch_idx, p_idx, r_idx] = sv_tmp.mean()
    expected_mvbs_val = ep.utils.compute._lin2log(expected_mvbs_val)
```

echopype/tests/commongrid/test_mvbs.py
```diff
@@ -833,3 +423,67 @@ def test_bin_and_mean_2d(bin_and_mean_2d_params) -> None:
...
+
+        # Compute xarray reduce manually for this
+        expected_Pos = xarray_reduce(
+            ds_Sv[POSITION_VARIABLES],
+            ds_Sv["ping_time"],
+            func="nanmean",
+            expected_groups=(ping_interval),
+            isbin=True,
+            method=method,
+        )
```
leewujung, line 486: Similar to the case with testing MVBS output values, please change this to use brute-force looped results so that we know that flox is getting this right.

leewujung (changes requested):
Hey @lsetiawan : Thanks for the changes. I am finding values from flox reduce to be different from values computed with brute-force looping. See https://github.com/OSOceanAcoustics/echopype/pull/1124/files?diff=split&w=0#r1328191991. Could you look into what may be happening? If there's a problem in my loop, feel free to point it out.

Apart from that, the rest of the code looks good.

Now that the previous `bin_and_mean_2d` mechanisms are no longer used, could you remove them and the associated tests also to avoid confusion? I don't think they are used anywhere else.

Thanks!

lsetiawan:
This is ready for another round of review @leewujung. Thanks!

echopype/commongrid/api.py
```diff
@@ -62,11 +66,117 @@ def _set_MVBS_attrs(ds):
...
+    pd.IntervalIndex
+        The resulting IntervalIndex
+    """
+    return pd.IntervalIndex.from_breaks(bins, closed=closed).sort_values()
+
+
+def _parse_x_bin(x_bin: str, x_label="range_bin") -> float:
+    """
+    Parses x bin string, check unit,
+    and returns x bin in x unit.
```
leewujung, line 94: Small tweaks
```suggestion
    Parses x bin string, check unit,
    and returns x bin in the specified unit.
```

echopype/commongrid/api.py
```diff
@@ -62,11 +66,117 @@ def _set_MVBS_attrs(ds):
...
+        # This shouldn't be other units
+        raise ValueError(
+            f"{x_bin_info['name']} must be in "
+            f"{x_bin_info['unit_label']} "
+            f"(e.g., '{x_bin_info['ex']}')."
+        )
+
+    # Convert back to float
+    range_bin = float(match_obj.group(1))
+    return range_bin
```
leewujung, line 163: might be good to change `range_bin` here to `bin_value` since this is now used for `dist_bin` too.

echopype/commongrid/api.py
```diff
@@ -62,11 +66,117 @@ def _set_MVBS_attrs(ds):
...
+    # First check for bin types
+    if not isinstance(x_bin, str):
+        raise TypeError("'x_bin' must be a string")
+    # normalize to lower case
+    # for range_bin
+    x_bin = x_bin.strip().lower()
+    # Only matches meters
+    match_obj = re.match(x_bin_info["pattern"], x_bin)
+
+    # Do some checks on range meter inputs
```
leewujung, line 152:
```suggestion
    # Do some checks on x_bin inputs
```

echopype/tests/commongrid/conftest.py
```diff
@@ -0,0 +1,466 @@
...
+        depth_interval=depth_interval,
+        depth_ping_time_len=depth_ping_time_len,
+        ping_time_len=ping_time_len,
+        ping_time_interval=ping_time_interval,
+        ping_time_jitter_max_ms=0,
+        random_number_generator=random_number_generator,
+    )
+
+
+# Helper functions to generate mock Sv dataset
```
leewujung, line 322:
```suggestion
# Helper functions to generate expected MVBS outputs from mock Sv dataset
# by brute-force looping and compute the mean
```

echopype/tests/commongrid/test_api.py
```diff
@@ -0,0 +1,283 @@
...
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    ("er_type"),
+    [
+        ("regular"),
+        ("irregular"),
+    ],
+)
+def test_compute_MVBS_range_output(request, er_type):
+    """Tests the output of compute_MVBS on regular and irregular data."""
```
leewujung, line 176:
```suggestion
    """
    Tests the shape of compute_MVBS output on regular and irregular data.
    The irregularity in the input echo_range would cause some rows or columns
    of the output Sv to contain NaN.
    Here we test for the expected shape after dropping the NaNs
    for specific ping_time bins.
    """
```

echopype/tests/commongrid/conftest.py
```diff
@@ -0,0 +1,466 @@
...
+    depth_ping_time_len=[100, 300, 200],
+    ping_time_len=600,
+    ping_time_interval="0.3S",
+    ping_time_jitter_max_ms=0,
+    random_number_generator=None,
+):
+    """
+    Generate a Sv dataset with uniform echo_range across all ping_time.
+
+    ping_time_jitter_max_ms controlled jitter in milliseconds in ping_time.
```
leewujung, line 428:
```suggestion
    Parameters
    ------------
    channel_len
        number of channels
    depth_len
        number of total depth bins
    depth_interval
        depth intervals, may have multiple values
    depth_ping_time_len
        the number of pings to use each of the depth_interval
        for example, with depth_interval=[0.5, 0.32, 0.13] 
        and depth_ping_time_len=[100, 300, 200],
        the first 100 pings have echo_range with depth intervals of 0.5 m,
        the next 300 pings have echo_range with depth intervals of 0.32 m,
        and the last 200 pings have echo_range with depth intervals of 0.13 m.
    ping_time_len
        total number of ping_time
    ping_time_interval
        interval between pings
    ping_time_jitter_max_ms
        jitter of ping_time in milliseconds
```

leewujung (approved):
@lsetiawan : The changes look great! I am really happy with the peace of mind re the values being tested and we worked out all the unexpected problems. Thanks for digging into `xarray_reduce` to find out about it accepting `pd.IntervalIndex`!

My comments are only cosmetic. Some are already inline. Below I list the others that would change functions that it is probably better that they are substituted automatically in an IDE.
- change all `sv` to `Sv` in the fixture and test function names: it actually means different things (Sv in the log domain, and sv in the linear domain...)
- change all `_er_` in fixture and test function names to `_echo_range_` for clarity, e.g., `ds_Sv_er_regular` becomes `ds_Sv_echo_range_regular`, `ds_Sv_er_irregular` becomes `ds_Sv_echo_range_irregular`, etc.

I trust that the currently failing tests are not easy to fix. Feel free to merge once you have resolved them!

echopype/commongrid/api.py
lsetiawan, line 163:
```suggestion
    x_bin_value = float(match_obj.group(1))
    return x_bin_value
```

PR #1128, opened 2023-08-21
Encoding changes in to_zarr

Sohambutala:
- Reintroduced code to explicitly eliminate preferred chunks once more, due to the reoccurrence of preferred chunks in the encoding dictionary.
- Ensuring chunk alignment and encoding only when dealing with DaskArray objects.


lsetiawan (approved):
This should resolve #1194

Issue #1062, opened 2023-06-11
Add descriptions and a diagram to docs documenting vertical offsets (water level, etc)


emiliom:
From @leewujung : For the sake of clarity, could you add the definition of `water_level` and `vertical_offset` with a simple diagram of which one means which for an underwater vehicle and a ship? This should actually be in the docs, but if we have something parked here it'll be easy to plug them in afterwards.

emiliom:
Refer to #1051 for related discussions.

leewujung:
I've restructured the TOC to have a Data processing section with 2 sections:
- `data-proc-func.ipynb`: providing usage examples, especially for `cal_params` and `env_params` for calibration
- `data-proc-additional.ipynb`: a one-stop-shopping place for additional details for processing functions 

One could argue that the `cal_params` and `env_params` details should be in the `-additional` page, but since I feel the `-func` page could also use some more expansion, I decided to hold off on any further changes.

I think the `-additional` page may be the place to go for the z-axis variables and combinations, and to include an illustration.

leewujung:
This is now addressed in #1133!

Issue #790, opened 2022-08-19
Add option in `consolidate.add_depth` to use variables already in EchoData


leewujung:
Right now `consolidate.add_depth` accepts only scalar inputs for input arguments `depth_offset` and `tilt`. However, these two variables can conveniently be DataArray with a time dimension. Or we may want to use the variables already stored in the EchoData object from which the Sv is derived from. 

Relevant variables in the EchoData object are `water_depth`, `vertical_offset`, and `tilt` (AZFP only) all in the platform group.

This is required for the glider `echopype-example` notebook to plot the true depth.

## Tasks
- ~allow accepting DataArray as inputs~ -- @emiliom and @leewujung that we will remove this requirement to make this function analogous to `consolidate.add_location`
- [ ] allow using variables already stored in the corresponding EchoData object
- [ ] add tests for the above cases

emiliom:
@leewujung: AZFP data come with two tilt variables, `tilt_x` and `tilt_y`. I haven't been able to find unambiguous definitions of these two variables in ASL / AZFP documentation. What are the orientations of x and y?

`consolidate.add_depth` has a `tilt` argument. For AZFP data, how do we calculate `tilt` from `tilt_x` and `tilt_y`? Would the calculation always be unambiguous?

emiliom:
I'm in touch with Steve and Julek (ASL) and have made progress with understanding `tilt_x` and `tilt_y`.

emiliom:
Refer to #1051 for related discussions.

PR #1133, opened 2023-08-24
Add discussion on z-axis vertical coordinate system and variables

emiliom:
Addresses #1062.

Here's a sketch for a ship context, showing only variables that are relevant for calculating `backscatter_depth` (note: the first surface below PCSo is the "nominal" water level and the second one is the "actual" water level):

PXL_20230824_214531298

(sorry for the huge image!)


leewujung:
Thanks for putting in the description. There are a few things I noticed:
- for `transducer_offset_z`:
   I checked that `drop_keel_offset` in EK80 is retained as its own variable, but conceptually it seems the same as `transducer_offset_z`?  see [here](https://echopype.readthedocs.io/en/latest/data-format-raw.html#scenario-1) -- at least on Shimada, and looks like it is the case for this example data file as well, that the actual depth of the transducer is indeed following the relationship you described: `transducer_depth = transducer_offset_z - water_level - vertical_offset` if we just assign `drop_keel_offset` to `transducer_offset_z`.
- for determining `echo_range_z`:
   > the beam_direction_z variable is defined by the convention in the Sonar/Beam_groupX group as the z-axis component of a unit vector

   I think the entire unit vector _x,_y,_z values have to be taken into account to compute the angle between the transducer axis and the vertical axis. I put in my edits (edited) 

- for the diagram:
   I think `transducer_offset_z` should point from the `PCS_o` line to where you have the black triangle (right now it is from the actual water level to the black triangle)

I've added other suggestions/simplifications in this PR: https://github.com/emiliom/echopype/pull/3, and also included a digital version of the diagram. Please take a look!

docs/source/data-proc-additional.ipynb
```diff
@@ -54,7 +54,68 @@
...
-        "<!-- ### ADD FIGURE -->"
+        "(data-proc:z-variables)=\n",
+        "## Vertical coordinate (_z_-axis) variables"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "The SONAR-netCDF4 v1 convention defines several variables in the `Platform` group that are referenced to a common Platform Coordinate System (PCS) vertical (_z_-axis) origin, ${PCS}_o$. These variables, when available, enable the calculation of the depth of each echo sample from the water surface. These variables are listed below together with their definitions, with edits when appropriate."
```
emiliom, line 65: In my original text, I had this additional statement: "as well for individual instruments". That was to account for `MRU_offset_z` and `position_offset_z`, which are not necessary for the calculation of the depth of each echo sample from the water surface. I included them in the list for completeness, but deliberately excluded them from my figure sketch to focus the figure on variables used in calculating `sample_depth`. Do you think it's fine to leave this text as is, or should we bring back a statement that clarifies that not all the variables in the list are necessary to calculate `sample_depth`?

docs/source/data-proc-additional.ipynb
```diff
@@ -54,7 +54,68 @@
...
+        "- ${PCS}_o$ corresponds to the position of the pressure sensor\n",
+        "- `water_level` as defined by the convention is unnecessary and can be set to 0\n",
+        "- `vertical_offset = -pressure_sensor_depth`"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "The sketch below shows the relationship between the above vertical coordinate (_z_-axis) variables.\n",
```
emiliom, line 116: How about moving up the figure and this sentence describing it to right after the sentence "where `echo_range_z` is the _z_-axis component (positive down) of `echo_range`."? That's where all the variables have been described conceptually. Everything below that sentence is a description of specific cases or additional calculations that may be needed.

docs/source/data-proc-additional.ipynb
```diff
@@ -54,7 +54,68 @@
...
+        "```\n",
+        "\n",
+        "where `echo_range_z` is the _z_-axis component (positive down) of `echo_range`."
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "When the transducer points vertically (up or down), `echo_range_z = echo_range`. When the transducer is mounted with an angle to the vertical axis, the beam direction vector (defined by the variables `beam_direction_x`, `beam_direction_y`, `beam_direction_z` in the `Sonar/Beam_groupX`) should be used together with `echo_range` to calculate `echo_range_z`."
```
emiliom, line 99: As `echo_range_z` needs to follow the positive-down convention, I think an edit along the lines of what I've suggested would need to be made:

```suggestion
        "When the transducer points vertically, `echo_range_z = echo_range` (transducer points down) or `echo_range_z = -echo_range` (transducer points up). When the transducer is mounted with an angle to the vertical axis, the beam direction vector (defined by the variables `beam_direction_x`, `beam_direction_y`, `beam_direction_z` in the `Sonar/Beam_groupX` group) should be used together with `echo_range` to calculate `echo_range_z`."
```

emiliom:
In the figure, how about adding an arrow for `sample_depth`? I'd suggest moving the "Transducer" and "Echo sample" words to the right, and adding a `sample_depth` arrow on the far left. For completeness, we could also add `transducer_depth`, but I think that'd be clutter.

emiliom:
> - for `transducer_offset_z`:
>   I checked that `drop_keel_offset` in EK80 is retained as its own variable, but conceptually it seems the same as `transducer_offset_z`?  see [here](https://echopype.readthedocs.io/en/latest/data-format-raw.html#scenario-1) -- at least on Shimada, and looks like it is the case for this example data file as well, that the actual depth of the transducer is indeed following the relationship you described: `transducer_depth = transducer_offset_z - water_level - vertical_offset` if we just assign `drop_keel_offset` to `transducer_offset_z`.

Correct, I didn't try to match it up to a convention variable. I didn't feel I understood it well enough, or that we knew whether its usage was consistent enough to make that determination. If the origin for `drop_keel_offset` is *always* the same as for `water_level`, then I agree, `transducer_offset_z` should be to `drop_keel_offset`. In the example file you point to, `water_level` is 0.0 and `vertical_offset` hovers at a near-zero values (0.0 to -0.01). That makes it hard to determine, for me at least, whether we can say that `transducer_offset_z` and `drop_keel_offset` are always identical. It looks more like a "post-processing" determination users would make based on knowledge of their configuration.

docs/source/data-proc-additional.ipynb
leewujung: I think it is ok to leave it as is, since in theory people should be using `add_depth` to calculate depth? Speaking of that -- do you want to add a statement about it? I forgot where that issue stands at the moment. If not finished, it'd be good to point to it as issue in the working.

docs/source/data-proc-additional.ipynb
leewujung: Sounds good. Please go ahead to make the change!

docs/source/data-proc-additional.ipynb
leewujung: This sounds good to me. To try to not let the user to be even more confused (or actually doing the opposite?), how about adding a short leading in in the beginning of the sentence "To compute the sample depth, when the transducer points ..."?
```suggestion
        "To compute the sample depth, when the transducer points vertically, `echo_range_z = echo_range` (transducer points down) or `echo_range_z = -echo_range` (transducer points up). When the transducer is mounted with an angle to the vertical axis, the beam direction vector (defined by the variables `beam_direction_x`, `beam_direction_y`, `beam_direction_z` in the `Sonar/Beam_groupX`) should be used together with `echo_range` to calculate `echo_range_z`."
```

leewujung:
I updated the fig directly since there's no easy way to do it in a notebook. I made some changes to accommodate the addition of `sample_depth`. I agree it is clearer with it. :)

leewujung:
>  In the example file you point to, water_level is 0.0 and vertical_offset hovers at a near-zero values (0.0 to -0.01). That makes it hard to determine, for me at least, whether we can say that transducer_offset_z and drop_keel_offset are always identical. It looks more like a "post-processing" determination users would make based on knowledge of their configuration.

Wouldn't the small `vertical_offset` be consistent with a clam day with small heave? and `water_depth=0` means that the PCSo is defined at the nominal water level?

emiliom:
> I updated the fig directly since there's no easy way to do it in a notebook. I made some changes to accommodate the addition of sample_depth. I agree it is clearer with it. :)

Looks perfect, thanks!

docs/source/data-proc-additional.ipynb
emiliom: I like your version. I'll commit it

leewujung:
> the sample_depth arrow should start at the "Actual water surface", not at PCSo. Other than that, it does look perfect!

Geez you're right, I wasn't thinking. Will push up the edit.

docs/source/data-proc-additional.ipynb
emiliom: The new section being added in this PR plays double duty. On the one hand, it's in the Data Processing part of the documents, where the role of `consolidate.add_depth` is central. If it were solely focused on that, I'd suggest we drop `MRU_offset_z` and `position_offset_z` altogether since they don't come into play and are a distraction. But, at least as I intended it, it's also partly a summary of the convention's and echopype's storage of vertical-dimension data in a converted EchoData object.

> Speaking of that (`add_depth`) -- do you want to add a statement about it? I forgot where that issue stands at the moment. If not finished, it'd be good to point to it as issue in the working.

We pushed the issue (#790) to 0.8.1. I intend it to be my first priority after we release 0.8.0.

I can mention `add_depth` in this section, but only briefly, mentioning that it'll be more robust soon.

docs/source/data-proc-additional.ipynb
emiliom: Done in the latest commit.

docs/source/data-proc-additional.ipynb
emiliom: In the latest commit I've added a brief mention of `add_depth` in an admonition box, including a link to the open issue and upcoming improvements. 

I did not change anything vis a vis `MRU_offset_z` and `position_offset_z`

leewujung: Sounds good. Thanks!

PR #1135, opened 2023-08-26
Changed parser in `parse_azfp.py` to maintain consistency over other parsers.

### LINKED ISSUE #659: Improve parsing config XML in AZFP files
Currently only a subset of variables are parsed from the AZFP XML files. These are the ones that were parsed in the reference Matlab code from the manufacturer ASL Env Sci.

We recently found that in order to encode a unique `channel` identifier, it would be the best to use the field `Frequency Number` in the AZFP XML file. However, the current simplistic XML parsing function does not support that. Let's take this chance to improve the robustness of this XML parsing, and also to include the variable units which were not included in the current parser.

For needing a unique `channel` identifier, see #566, which is addressed by #657.

This is similar to #468.

## Tasks
- [x] improve robustness of AZFP XML parser, using something like [lxml](https://lxml.de/)
- [x] improve AZFP XML parser to include all config variables
- [ ] include variable units in the XML file to parser and store accordingly in the set_groups methods

praneethratna:
Fixes #659. Now `parse_azfp.py` uses `xml.etree.ElementTree` instead of `xm.dom.minidom` for parsing.

Update: More parameters are being parsed now, and some of those were added to the EchoData object via set_groups.

CC @leewujung @emiliom


echopype/convert/parse_azfp.py
```diff
@@ -129,10 +129,10 @@ def load_AZFP_xml(self):
 
         def get_value_by_tag_name(tag_name, element=0):
             """Returns the value in an XML tag given the tag name and the number of occurrences."""
-            return px.getElementsByTagName(tag_name)[element].childNodes[0].data
+            return root.iter(tag_name).__next__().text
 
-        xmlmap = fsspec.get_mapper(self.xml_path, **self.storage_options)
-        px = xml.dom.minidom.parse(xmlmap.fs.open(xmlmap.root))
+        et = ET.parse(self.xml_path)
```
lsetiawan, line 134: Hi @praneethratna! Thanks for this much cleaner implementation. I do have a question, does `ET.parse` allow for remote files? so if my xml is located at `http://mysite.com/myxml.xml` would this read and load that xml into an Element Tree object?

praneethratna: @lsetiawan I don't think so, i missed this point - we have to use `fsspec` and pass `xmlmap.fs.open(xmlmap.root)` as input to `ET.parse`.

emiliom: Yeah, it looks like you should retain the `xmlmap` statement (line 34), then, as you said, pass `xmlmap.fs.open(xmlmap.root)` as input to `ET.parse`. You can probably use a single, longer statement:
```python
root = ET.parse(xmlmap.fs.open(xmlmap.root)).getroot()
```

echopype/convert/parse_azfp.py
```diff
@@ -129,10 +129,10 @@ def load_AZFP_xml(self):
 
         def get_value_by_tag_name(tag_name, element=0):
             """Returns the value in an XML tag given the tag name and the number of occurrences."""
-            return px.getElementsByTagName(tag_name)[element].childNodes[0].data
+            return root.iter(tag_name).__next__().text
```
emiliom, line 132: This new statement doesn't make use of the `element` argument, which is required by the call in line 146.

praneethratna: @emiliom I removed the usage of this function for parsing `AZFP` XML files.

praneethratna:
Closing and re opening to re-run the tests.

echopype/convert/parse_azfp.py
```diff
@@ -127,25 +90,36 @@ def load_AZFP_xml(self):
...
-            self.parameters[new_name] = [
-                float(get_value_by_tag_name(old_name, ch))
-                for ch in range(self.parameters["num_freq"])
-            ]
+        root = ET.parse(xmlmap.fs.open(xmlmap.root)).getroot()
+
+        for child in root.iter():
+            if len(child.attrib) > 0:
+                for key, val in child.attrib.items():
+                    self.parameters[child.tag + key].append(val)
```
leewujung, line 99: What do you think about changing `child.tag + key` to `child.tag + key.capitalize()` so that we have the same camelcase form that is then put through the same converting camelcase to lowercase+"_" procedure, so that at the output end in `self.parameters` we have identical strings? For example, right now the first key would be `InstrumentTypestring`, and if we do the above, we will get `instrument_type_string` -- this latter one would go very well with `instrument_type` for cross referencing what is what.

echopype/convert/parse_azfp.py
```diff
@@ -127,25 +90,36 @@ def load_AZFP_xml(self):
...
+                except ValueError:
+                    val = float(child.text)
+
+            if len(child.tag) > 3 and not child.tag.startswith("VTX"):
+                words = re.findall("[A-Z]+[a-z]*", child.tag)
+                words_lower = [word.lower() for word in words]
+                if len(words) > 1:
+                    self.parameters["_".join(words_lower)].append(val)
+                elif len(words) == 1:
+                    self.parameters[words_lower[0]].append(val)
```
leewujung, line 115: you could probably wrap these in a subfunction so that you can use here and also in the above for the `child.tag + key.capitalized` fields ([this comment](https://github.com/OSOceanAcoustics/echopype/pull/1135/files?diff=unified&w=0#r1328983167)).

echopype/tests/convert/test_convert_azfp.py
```diff
@@ -172,3 +174,38 @@ def test_convert_azfp_01a_notemperature_notilt(azfp_path):
     assert "tilt_y" in echodata["Platform"]
     assert echodata["Platform"]["tilt_x"].isnull().all()
     assert echodata["Platform"]["tilt_y"].isnull().all()
+
+
+def test_load_parse_azfp_xml(azfp_path):
```
leewujung, line 179: The test looks great! Could you add more tests for the values, especially those that are across all 4 frequencies? Just so that we have a piece of mind that things are 100% correct. Thanks!

echopype/convert/parse_azfp.py
praneethratna: Sounds great!, but one query - `sensors_flag` has `PressureSensorInstalled` as one attribute, should that be converted to `sensors_flag_pressure_sensor_installed` or `sensors_flag_pressuresensorinstalled`? What do you think about this? I think the latter one should be correct.

echopype/convert/set_groups_azfp.py
```diff
@@ -659,32 +662,15 @@ def set_vendor(self) -> xr.Dataset:
...
-                    ["channel"],
-                    parameters["VTX"][self.freq_ind_sorted],
-                    {"long_name": "Amplified voltage sent to the transducer"},
-                ),
+                "EL": (["channel"], parameters["EL"][self.freq_ind_sorted]),
+                "TVR": (["channel"], parameters["TVR"][self.freq_ind_sorted]),
+                "VTX0": (["channel"], parameters["VTX0"][self.freq_ind_sorted]),
+                "VTX1": (["channel"], parameters["VTX1"][self.freq_ind_sorted]),
+                "VTX2": (["channel"], parameters["VTX2"][self.freq_ind_sorted]),
+                "VTX3": (["channel"], parameters["VTX3"][self.freq_ind_sorted]),
```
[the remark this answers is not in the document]
praneethratna: Oh sorry, I accidentally removed those changes when i encountered merge conflicts after raising the PR. Will add them back in the next commit!

leewujung (commented):
@praneethratna : Thanks for the PR, it's awesome to see this very clean handling of the parameters/parameter names!! Once the few small things I commented on are included, we can merge this.

I opened up a new issue for additional work in `set_groups_azfp` to store parameters that are  potentially varying across the "phases":  https://github.com/OSOceanAcoustics/echopype/issues/1171
Let me know if you'd like to contribute to that!

echopype/convert/parse_azfp.py
leewujung: hmmm, this is a tricky question. I think `sensors_flag_pressure_sensor_installed` is easier to read. What if we go with that one?

praneethratna: Sure, we can go with `sensors_flag_pressure_sensor_installed`. This naming choice does make it more clear and readable.

praneethratna:
> I opened up a new issue for additional work in `set_groups_azfp` to store parameters that are potentially varying across the "phases": #1171 Let me know if you'd like to contribute to that!

Sure, would love to contribute to #1171 once this is merged.

echopype/tests/convert/test_convert_azfp.py
praneethratna: Added new tests for other parameters!

echopype/convert/parse_azfp.py
```diff
@@ -127,25 +90,42 @@ def load_AZFP_xml(self):
...
+                self.parameters[camel_case_tag].append(val)
+            else:
+                self.parameters[child.tag].append(val)
+
+        # Handling the case where there is only one value for each parameter
+        for key, val in self.parameters.items():
+            if len(val) == 1:
+                self.parameters[key] = val[0]
+
+    def _to_camelCase(self, tag):
```
emiliom: Actually, you're converting *from* CamelCase *to* snake_case; so, the function name should be changed.

More generally, conversion of strings between CamelCase and snake_case is a really common need, in Python and all languages. I was curious about existing implementations in Python (I thought it might even be in a standard library, but it wasn't). But -- unsurprisingly :stuck_out_tongue_winking_eye: -- it's already done elsewhere in echopype, for the EK parser:
https://github.com/OSOceanAcoustics/echopype/blob/3ca028f04a4fd3643bb0ad350dc09a699883ee1e/echopype/convert/utils/ek_raw_parsers.py#L709

There, it's wrapped up within a class.

Later, in a follow-up PR, @leewujung I think we should create a generic, stand-alone utility function to do this, then reuse it both here and with EK's. For this PR, I think it'd extend its scope too much. In the meantime, I think the EK implementation is more comprehensive. I tested that function and the one implemented here. For cases like "firstLast" (first character is not uppercase) and "FirstSecond_last" (mixed form), your function drops the parts that are not strictly CamelCase and returns "last" and "first_second", respectively. The `from_CamelCase` EK function returns "first_last" and "first_second_last", as expected. I found an [even more compact implementation](https://rodic.fr/blog/camelcase-and-snake_case-strings-conversion-with-python/) that also behaves like `from_CamelCase`. Your implementation also throws away numbers, or worse.

My recommendation for this PR: copy the `from_CamelCase` function `convert/utils/ek_raw_parsers.py` exactly as is, here. Note that there's no need for it to have the `self` argument, since it's not used.

Having identical functions will make it easier for the next PR that will move the functions to a utility function.

leewujung:
Thanks a lot @praneethratna! I did a direct push for some cosmetic changes which you can see above, and will go ahead to merge this once the tests run through. Excited to have this upgrade in!

echopype/convert/parse_azfp.py
emiliom: Looks like I submitted my comment just as you were renaming and moving the function!

leewujung: @emiliom : I was about to merge this. Could you do what you said above, but instead of copying that function, move that function from `ek_raw_parsers.py` to `echopype/utils/misc.py`? We'll add other things to that module later, for example from #1124. After you add that I'll merge this PR. Thanks.

echopype/convert/parse_azfp.py
emiliom, line 1: Now that `PulseLen` is auto-converted to `pulse_len` and `LockOutIndex` to `lock_out_index`, it seems like other, existing usages of the old names `pulse_length` and `lockout_index` would need to be updated, too? For example, the `field_w_freq` assignments in `_split_header`.

echopype/convert/parse_azfp.py
emiliom: Will do. I'm working on it.

echopype/convert/parse_azfp.py
praneethratna: Missed that, will update them.

praneethratna: @emiliom I pushed changes for this.

emiliom: Thanks! I'm about to push the changes to the definition and use of the CamelCase > snake_case function.

praneethratna: @emiliom Cool, thanks!

emiliom:
@praneethratna see the new test errors. They look like they'll be easy to fix.

praneethratna:
> @praneethratna see the new test errors. They look like they'll be easy to fix.

I actually thought about this error but forgot while merging😅. I've fixed the error and pushed the changes!

echopype/echodata/sensor_ep_version_mapping/v05x_to_v06x.py
```diff
@@ -700,7 +700,10 @@ def _rearrange_azfp_attrs_vars(ed_obj, sensor):
             "DS",
             "EL",
             "TVR",
-            "VTX",
+            "VTX0",
+            "VTX1",
+            "VTX2",
+            "VTX3",
```
emiliom, line 706: Actually, leave this module as is, w/o changes. That's because it's intended only to address the conversion of files from a very old version of echopype (0.5) to another old version (0.6)
```suggestion
            "VTX",
```

praneethratna:
@emiliom I'm sorry for all these messed up commits, kept forgetting minor changes in code. I think this can be merged as the failing test case is not related this PR.

emiliom:
> @emiliom I'm sorry for all these messed up commits, kept forgetting minor changes in code. I think this can be merged now as the failing test is not related this PR.

Can you elaborate? I realize it's only happening with Python 3.11, and I can see that it looks like the same TypeError in `echopype/tests/convert/test_parsed_to_zarr.py`: "TypeError: 'numpy.float64' object cannot be interpreted as an integer". But if it wasn't happening before this PR, it seems like it's linked to this PR. Do you remember if it's been happening with 3.11 since you first created the PR?

Are you set up to [run tests locally](https://echopype.readthedocs.io/en/stable/contributing.html)? If not, I can try running `test_parsed_to_zarr.py` locally in a 3.11 conda environment.

praneethratna:
> > @emiliom I'm sorry for all these messed up commits, kept forgetting minor changes in code. I think this can be merged now as the failing test is not related this PR.
> 
> Can you elaborate? I realize it's only happening with Python 3.11, and I can see that it looks like the same TypeError in `echopype/tests/convert/test_parsed_to_zarr.py`: "TypeError: 'numpy.float64' object cannot be interpreted as an integer". But if it wasn't happening before this PR, it seems like it's linked to this PR. Do you remember if it's been happening with 3.11 since you first created the PR?

Yes, it's been happening since i first created this PR.

> Are you set up to [run tests locally](https://echopype.readthedocs.io/en/stable/contributing.html)? If not, I can try running `test_parsed_to_zarr.py` locally in a 3.11 conda environment.

There seems to be an issue with setting up conda in my laptop, so I don't use any conda environment while running the tests.

emiliom:
> Yes, it's been happening since i first created this PR.

Hmm, that's odd. I'll take a look locally. In the meantime, how about if you start planning for #1171?
Thanks.

emiliom:
All the tests in `echopype/tests/convert/test_parsed_to_zarr.py` run fine locally with Python 3.11, with a freshly created conda environment, on both `dev` and an older version of this PR. So, I don't know what's going on, but it looks like it's limited to the CI.

I'll take a final look at the PR to confirm that everything has been addressed.

emiliom (approved):
It looks good! I can merge it.

But first I'll modify the PR description, b/c it's become broader than just changing how the XML is parsed

PR #1142, opened 2023-08-30
Add transient and impulse noise masks

beatfactor:
### Overview
This PR introduces two new functions, `get_impulse_noise_mask` and `get_transient_noise_mask` that provide masks for impulsive noise (noise sources shorter than 1 ping) and transient noise (short, but still multi-ping noise sources). These are ported from the [echopy](https://github.com/open-ocean-sounding/echopy/) library:
[mask_impulse.py](https://github.com/open-ocean-sounding/echopy/blob/master/echopy/processing/mask_impulse.py)
[mask_transient.py](https://github.com/open-ocean-sounding/echopy/blob/master/echopy/processing/mask_transient.py)

### Methodologies

The methodology employed uses both previously published methods defined:
- _Reducing bias due to noise and attenuation in open-ocean echo integration. ICES Journal of Marine Science 72, 2482-2493_ – Ryan T. E., Downie R. A., Kloser R. J., Keith G. (2015) (`ryan` as mask generation method)
- _A noise removal algorithm for acoustic data with strong interference based on post-processing techniques CCAMLR SG-ASAM: 15/02_ – Wang et al. (2015) (`wang`, for impulse noise only - unlike in echopy, this has been modified to provide a mask rather than a modified Sv array). 
- otherwise-unpublished methods implemented by the authors of echopy (`fielding` as mask-generation method, for transient noise only).

### Key Features

1. **Impulse noise** - implements the Ryan two-sided comparison method as "ryan", an iterable modification of it that allows it to detect adjacent noise spikes, and the Wang erosion/dilation/median filtering method as "wang"
2. **Transient noise** - implements the Ryan identified Sv spikes method as "ryan" and a previously unpublished method proposed by Fielding et al, which compares the data, ping by ping, with respect to a bloc in a reference layer – should the ping median be greater than the block median by a user-defined threshold, the ping will be masked until transient noise disappears or until it gets to the minimum range allowed by the user

### Key Considerations
 In order to ensure compatibility between the `echopy` and `echopype` libraries, several important factors were taken into account while adapting the `echopy` code – responsible for noise masking:
- Data type discrepancies: the differences in input and output data types between the two libraries required adjustment.
- Mask delivery variation: in `echopy`, two separate masks are typically generated: one indicating the locations of noise (represented as `true`) and another showing where the method applied may not have been suitable for certain samples. In contrast, echopype usually delivers a single mask marking samples that are free of noise.
- Differences in the ‘pings’ and ‘range’ axes: the two libraries have divergent approaches to these axes, which needed to be addressed.

For each noise type, specific decisions were made to combine and adapt the masks appropriately within the echopype environment.

### Function Signatures

```python
def get_impulse_noise_mask(
    source_Sv: xr.Dataset,
    desired_channel: str,
    thr: Union[Tuple[float, float], int, float],
    m: Optional[Union[int, float]] = None,
    n: Optional[Union[int, Tuple[int, int]]] = None,
    erode: Optional[List[Tuple[int, int]]] = None,
    dilate: Optional[List[Tuple[int, int]]] = None,
    median: Optional[List[Tuple[int, int]]] = None,
    method: str = "ryan",
) -> xr.DataArray:

def get_transient_noise_mask(
    source_Sv: Union[xr.Dataset, str, pathlib.Path],
    desired_channel: str,
    mask_type: str = "ryan",
    **kwargs
) -> xr.DataArray:
```

### Usage
The functions can be utilised to create masks for `Sv` data based on identified single-ping or short multi-ping noise sources.


beatfactor:
> ## [Codecov](https://app.codecov.io/gh/OSOceanAcoustics/echopype/pull/1142?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=OSOceanAcoustics) Report


@mihaiboldeanu Looks like we could do a with a few more tests.

leewujung:
Hey folks, thanks for the PR!

I did a high-level review of the function/module organization and have the following suggestions. @emiliom also contributed to the discussion.
- Could you please revert all the changes in `tests/mask/test_mask.py`? These changes are due to formatting, and even though the current one is not optimal, this would cause conflicts for the ongoing other PR #1106 . 
- Based on what we have for the subpackage organization (see [here](https://github.com/OSOceanAcoustics/echopype/pull/817#issuecomment-1474564449)), these noise mask generation functions would be in the `clean` subpackage. It is true that the functions you added here are generating masks, but noise assessment is a pretty central task for data processing that having its own subpackage is logical. So, could you move them? In the comment below I'll use `clean` to avoid confusion, but let me know if it is confusing!
- Could you use [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) for the setup parts of the two test modules `test_mask_transient.py` and `test_mask_impulse_noise.py`, and put them in `tests/clean/conftest.py` so that they can be reused in the test modules. Also suggest renaming:
  - `test_mask_transient.py` to `test_clean_transient_noise.py`
  - `test_mask_impulse_noise.py` to `test_clean_transient_noise.py`
- For `get_impulse_noise_mask.py`:
  - move the function `get_impulse_noise_mask` to within `echopype/clean/api.py`
  - move all the subfunctions (`ryan`, `ryan_iterable`, `wang`, etc) to a separate module `echopype/clean/impulse_noise.py`, and add `_` to the beginning of the function call so it is clear they are private functions
- For `get_transient_noise_mask.py`:
  - move the function `get_transient_noise_mask` to within `echopype/clean/api.py`
  - move the other two functions `_get_transient_noise_mask_ryan` and `_get_transient_noise_mask_fielding` to a separate module `echopype/clean/transient_noise.py`, and rename them to just `_ryan` and `_fielding`

For the last 2 points, our convention has been to keep the main API call under a subpackage in `subpackage/api.py`, and put supporting functions into separate modules (.py files) with filenames indicating which api function they are used in.

In addition, in `clean` we currently only have one function with a generic name (`remove_noise`), even though that function implements a specific algorithm that aims to estimate the time-varying background noise. With your addition of new noise removal functions (once masked), we need to change this function name and figuring out the API pattern for noise-related functions that generate masks (like these added here) and functions that suppress some noise (the current `remove_noise`). But all that would be a separate PR.

ruxandra-valcu:
One suggestion, in that case: let's move the test setup part to the base tests folder, since some of our non-noise-detection tests (shoal detection, seabed detection etc) will also use it. So tests/conftest.py rather than tests/clean/conftest.py. What do you think?

leewujung:
> One suggestion, in that case: let's move the test setup part to the base tests folder, since some of our non-noise-detection tests (shoal detection, seabed detection etc) will also use it. So tests/conftest.py rather than tests/clean/conftest.py. What do you think?

Ah I see. Yes in that case, moving the setup to `tests/conftest.py` sounds great. Thanks!

echopype/clean/api.py
```diff
@@ -74,3 +85,206 @@ def remove_noise(ds_Sv, ping_num, range_sample_num, noise_max=None, SNR_threshol
...
+
+
+def get_transient_noise_mask(
+    source_Sv: Union[xr.Dataset, str, pathlib.Path],
+    desired_channel: str,
+    mask_type: str = "ryan",
+    **kwargs,
+) -> xr.DataArray:
+    """
+    Create a mask based on the identified signal attenuations of Sv values at 38KHz.
```
leewujung, line 97: This seems wrong?

echopype/clean/api.py
```diff
@@ -74,3 +85,206 @@ def remove_noise(ds_Sv, ping_num, range_sample_num, noise_max=None, SNR_threshol
...
+    desired_channel: str,
+    thr: Union[Tuple[float, float], int, float],
+    m: Optional[Union[int, float]] = None,
+    n: Optional[Union[int, Tuple[int, int]]] = None,
+    erode: Optional[List[Tuple[int, int]]] = None,
+    dilate: Optional[List[Tuple[int, int]]] = None,
+    median: Optional[List[Tuple[int, int]]] = None,
+    method: str = "ryan",
+) -> xr.DataArray:
+    """
```
leewujung, line 163: A couple suggestions:
- change the input argument to allow users to pass in a dict containing the parameters for the method specified in `method`
- add `frequency_desired` as an input argument to slice the Sv dataset using `frequency_nominal`

Something like below:
```python
def get_impulse_noise_mask(
    source_Sv: xr.Dataset,
    desired_channel: Union[List[str], xr.DataArray],
    method: str = "ryan",
    parameters: Dict,
) -> xr.DataArray
    """
```

echopype/clean/api.py
```diff
@@ -74,3 +85,206 @@ def remove_noise(ds_Sv, ping_num, range_sample_num, noise_max=None, SNR_threshol
...
+            source_Sv, desired_channel, m, n, thr
+        )
+        noise_free_mask = ~impulse_mask_ryan_iterable
+    elif method == "wang":
+        impulse_mask_wang = _find_impulse_mask_wang(
+            source_Sv, desired_channel, thr, erode, dilate, median
+        )
+        noise_free_mask = ~impulse_mask_wang
+    else:
+        raise ValueError(f"Unsupported method: {method}")
```
leewujung, line 215: I'd suggest the following:
- with the `parameters` dictionary added, add checks for if required parameters required by the specified `method` are provided
- create a mapping dictionary so that you can map the functions easily:
   ```
   MASK_MAP = {
      "ryan": _find_impulse_mask_ryan,
      "ryan_iterable": _find_impulse_mask_ryan_iterable,
      ...,
   }
   ```
   This way you'll be able to handle all functions in the same way and do the negation of mask in 1 line.

echopype/clean/api.py
leewujung: Also allow variations of type of `source_Sv` as in `get_attentuation_mask`

echopype/clean/api.py
```diff
@@ -74,3 +85,206 @@ def remove_noise(ds_Sv, ping_num, range_sample_num, noise_max=None, SNR_threshol
...
+
+    return noise_free_mask
+
+
+def get_attenuation_mask(
+    source_Sv: Union[xr.Dataset, str, pathlib.Path],
+    desired_channel: str,
+    mask_type: str = "ryan",
+    **kwargs,
```
leewujung, line 224: Instead of `kwargs`, suggest using the same approach as in [this comment](https://github.com/OSOceanAcoustics/echopype/pull/1142/files#r1334749088) and provide explanation and parameter requirements in the docstring.

echopype/clean/api.py
```diff
@@ -74,3 +85,206 @@ def remove_noise(ds_Sv, ping_num, range_sample_num, noise_max=None, SNR_threshol
...
+
+
+    Examples
+    --------
+
+    """
+    assert mask_type in ["ryan", "ariza"], "mask_type must be either 'ryan' or 'ariza'"
+    selected_channel_Sv = source_Sv.sel(channel=desired_channel)
+    Sv = selected_channel_Sv["Sv"].values
+    r = source_Sv["echo_range"].values[0, 0]
```
leewujung, line 269: Let's do label-aware processing in the functions and not extract just the data to be numpy array, to take advantage of distributed computing when possible.

echopype/clean/impulse_noise.py
```diff
@@ -0,0 +1,463 @@
...
+    Sv_ = mask_transformation.oned(Sv, iax, iax_, 0, log_var=True)[0]
+
+    # resample back to full resolution
+    jax = np.arange(len(Sv[0]))
+    Sv_, mask_ = mask_transformation.full(Sv_, iax_, jax, iax, jax)
+
+    # side comparison (±n)
+    dummy = np.zeros((iax.shape[0], n)) * np.nan
+    comparison_forward = Sv_ - np.c_[Sv_[:, n:], dummy]
+    comparison_backward = Sv_ - np.c_[dummy, Sv_[:, 0:-n]]
```
leewujung, line 63: [xr.DataArray.shift](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.shift.html#xarray.DataArray.shift) likely useful here for maintaining things as `xr.DataArrays`.

echopype/clean/impulse_noise.py
```diff
@@ -0,0 +1,463 @@
...
+    # get IN mask
+    comparison_forward[np.isnan(comparison_forward)] = np.inf
+    maskf = comparison_forward > thr
+    comparison_backward[np.isnan(comparison_backward)] = np.inf
+    maskb = comparison_backward > thr
+    mask = maskf & maskb
+
+    # get second mask indicating valid samples in IN mask
+    mask_[:, 0:n] = False
+    mask_[:, -n:] = False
```
leewujung, line 74: [xarray.DataArray.where](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.where.html) could be used for these.

echopype/clean/impulse_noise.py
```diff
@@ -0,0 +1,463 @@
...
+        "but does not modify them in the same way as the `wang` function from echopy."
+        "Visualizations using this mask may therefore differ from"
+        "those generated using the `wang` function from echopy. "
+        "Be aware that regions marked as True in the mask are identified as noise."
+    )
+
+    return mask_xr
+
+
+def _find_impulse_mask_ryan(
```
leewujung, line 325: Once you modify `_ryan` under the hood to use xarray operations, I think you could remove this function and just make sure output of `_ryan` is of the right dimensions.

echopype/clean/impulse_noise.py
```diff
@@ -0,0 +1,463 @@
...
+        dummy[:] = np.nan
+        forward = Sv_ - np.c_[Sv_[:, i:], dummy]
+        backward = Sv_ - np.c_[dummy, Sv_[:, 0:-i]]
+        maskf = np.ma.masked_greater(forward, thr).mask
+        maskb = np.ma.masked_greater(backward, thr).mask
+        mask = mask | (maskf & maskb)
+
+    # get second mask indicating valid samples in IN mask
+    mask_[:, 0 : max(n)] = True
+    mask_[:, -max(n) :] = True
```
leewujung, line 128: Seems that this is a loop of computing `_ryan` masks? If so, how about reusing that function?

requirements.txt
```diff
@@ -16,3 +16,4 @@ xarray-datatree==0.0.6
 psutil>=5.9.1
 more-itertools==8.13.0
 geopy
+scikit-image
```
leewujung, line 19: This is a large library, so let's make it optional so that only people who want to use the functions that contain skimage functions will need to install them.

echopype/clean/impulse_noise.py
```diff
@@ -0,0 +1,463 @@
...
+    # by running erosion cycles
+    Sv_eroded = Sv.copy()
+    for e in erode:
+        Sv_eroded = erosion(Sv_thresholded, np.ones(e))
+
+    # the last step might have turned interferences inside biology into vacant
+    # samples, this is solved by running dilation cycles
+    Sv_dilated = Sv_eroded.copy()
+    for d in dilate:
+        Sv_dilated = dilation(Sv_dilated, np.ones(d))
```
leewujung, line 183: [dask_image.ndmorph package](https://image.dask.org/en/latest/dask_image.ndmorph.html) seems to have wrapped `scipy.ndimage._morphology.binary_erosion`. Could you test if it takes delayed dask arrays correctly (so that it can deal with large, lazy-loaded datasets)?

echopype/clean/impulse_noise.py
```diff
@@ -0,0 +1,463 @@
...
+    This code is licensed under the MIT License.
+    See https://opensource.org/license/mit/ for details.
+    Original code sourced from:
+    https://github.com/open-ocean-sounding/echopy/blob/master/echopy/processing/mask_impulse.py
+
+"""
+__authors__ = [
+    "Alejandro Ariza",  # wrote ryan(), ryan_iterable(), and wang()
+    "Raluca Simedroni",  # adapted the impulse noise masking algorithms to echopype
+]
```
leewujung, line 15: Since most of my comments would be effectively rewriting the code (hence new implementations), we can modify this block once the code is in good shape.

echopype/clean/impulse_noise.py
```diff
@@ -0,0 +1,463 @@
...
+    # now corrected to corresponding Sv values before the erosion/dilation
+    Sv_corrected1 = Sv_dilated.copy()
+    mask_bio = (Sv_dilated >= thr[0]) & (Sv_dilated < thr[1])
+    Sv_corrected1[mask_bio] = Sv_thresholded[mask_bio]
+
+    # compute median convolution in Sv corrected array
+    Sv_median = Sv_corrected1.copy()
+    for m in median:
+        Sv_median = mask_transformation.log(
+            medianf(mask_transformation.lin(Sv_median), footprint=np.ones(m))
```
leewujung, line 195: Looks like median filter is wrapped by dask already: https://image.dask.org/en/latest/dask_image.ndfilters.html
Could you check if lazy-loaded dask arrays are handled?

leewujung (changes requested):
Hey @mihaiboldeanu @ruxandra-valcu @beatfactor : Thanks for re-organizing the modules.

I've gone through the impulse noise functions and added comments on places where the brute-force numpy operations can be substituted with `xarray` and `dask` functions, for scalability and readability of the code. These are things that are central to echopype's mission, since the `xarray` operations are pretty intuitive, and in the case of `skimage`, it seems `dask` already wraps them for distributed use. 

And a couple of other comments:
- It'd be useful for users to include in the docstring if the mask creating function output has a `channel` dimension, i.e., can the mask be applied across all channels, or should the mask be of exactly the same shape as the Sv dataarray in dimensions `(channel, range_sample, ping_time)` -- this is also in light of #1177
- the `_find_impulse_mask_*` functions (`_find_impulse_mask_ryan` ) could be absorbed in the `_*` functions (e.g., `_ryan`) once you changed the computation under the hood to using xarray operations (and of course absorb the few extra things that are only in the `_find_impulse_mask_*` functions

I'll review the transient noise and signal attenuation components when I am able in the next few days.

echopype/clean/impulse_noise.py
ruxandra-valcu: Modified https://github.com/OceanStreamIO/echopype/commit/4a6d80fb0519e387be003b62e43d01ea07a25ba1

echopype/clean/api.py
ruxandra-valcu: > A couple suggestions:
> * add `frequency_desired` as an input argument to slice the Sv dataset using `frequency_nominal`

Will do - could you explain to me what the logic behind this is? Would this refer to something simialr to https://echopype.readthedocs.io/en/v0.6.1/api/echopype.preprocess.swap_dims_channel_frequency.html ?

echopype/clean/impulse_noise.py
```diff
@@ -0,0 +1,463 @@
...
+    "Alejandro Ariza",  # wrote ryan(), ryan_iterable(), and wang()
+    "Raluca Simedroni",  # adapted the impulse noise masking algorithms to echopype
+]
+
+import warnings
+from typing import List, Tuple, Union
+
+import numpy as np
+import xarray as xr
+from scipy.ndimage import median_filter as medianf
```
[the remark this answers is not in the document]
ruxandra-valcu: Done

echopype/clean/impulse_noise.py
ruxandra-valcu: https://github.com/OceanStreamIO/echopype/commit/4294926d3732bc1a0bb574331bc6afcb6ffea7d4 consolidated into a single function

echopype/clean/api.py
ruxandra-valcu: https://github.com/OceanStreamIO/echopype/commit/759df07f772f7dcc624ea2565b7002267a4ed93c done

echopype/clean/impulse_noise.py
ruxandra-valcu: Unfortunately, we use grayscale erosion rather than binary erosion (and after verifying with the original paper, we need it) so this can't be ported in dask and will have to stay as numpy operations.

echopype/clean/impulse_noise.py
ruxandra-valcu: The function itself appears to handle them, but I'm not sure using it actually makes any sense, since we have to drop to numpy in the code right before it anyway, due to grey_erosion/dilation not being implemented in dask_image.ndmorph

ruxandra-valcu:
Hello @leewujung 

I've rewritten this PR (both the already-reviewed impulse noise algorithms and the transient noise and signal attenuation ones) to use xarray functionality rather than numpy.

Please let me know if there are any other improvements I should make!

ctuguinay:
Closing this since the Echopy functions were added in #1316 and #1326

PR #1143, opened 2023-08-30
Remove automatic version conversion in `open_converted`

leewujung:
## Overview

This PR removes the automatic `EchoData` format conversion when using `open_converted` on already converted data from earlier versions of echopype.

This PR does not remove the previous data format conversion mechanisms (under `echodata/sensor_ep_version_mapping`) in case we are able to reinstate this capability in the future.

The documentation is also updated.

## Context

We included a converter to automatically convert v0.5.x `EchoData` format to v0.6.x `EchoData` format. However, we have not been able to create such a functionality for all later versions, including the latest v0.8.0 which included a large number of data format changes. The existence of automatic v0.5.x to v0.6.x format conversion becomes confusing in this case since no other versions of data are able to be converted automatically in the same way, and using `open_raw` to re-convert the data is necessary. 

This is discovered by @emiliom 
> we have a problem with 0.8.0. I was trying to run and update the echopype-examples notebooks, and ran into a failure with open_converted that boils down to https://github.com/OSOceanAcoustics/echopype/blob/254ee106eb580959ca85f2131168c4f17c0a285a/echopype/echodata/sensor_ep_version_mapping/ep_version_mapper.py#L22, called by https://github.com/OSOceanAcoustics/echopype/blob/254ee106eb580959ca85f2131168c4f17c0a285a/echopype/echodata/echodata.py#L165. ie, the version tag 0.8.0 raises an error. We didn't run into this during local tests b/c the version tag was still 0.7.2xxx. But I'm puzzled why it didn't come up in the CI once the version was stamped as 0.8.0.

### DIFF
--- a/docs/source/data-format-changes.ipynb
+++ b/docs/source/data-format-changes.ipynb
@@ -41,9 +41,17 @@
         "    - Revise the dimensions of each variable to be consistent across instrument types, with dimensions deemed unnecessary dropped from some variables.\n",
         "- In the `Provenance` group: Add new attributes `combination_*` to the \"combined\" `EchoData` object, mirroring the convention-based attributes `conversion_*`.\n",
         "- In the `Vendor_specific` group: Move filter coefficients and decimation factor from attributes to variables in EK80, to facilitate consistent provenance tracking during `combine_echodata` operations.\n",
-        "- Improve the presence and use of variable attributes throughout `EchoData` groups.\n",
-        "\n",
-        "Version 0.8.0 does not incorporte the capability to read files converted by previous versions of echopype. We recommend using `open_raw` to re-convert the raw data files."
+        "- Improve the presence and use of variable attributes throughout `EchoData` groups."
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "e506a51d",
+      "metadata": {},
+      "source": [
+        ":::{Note}\n",
+        "Version 0.8.0 does not incorporte the capability to read files converted by previous versions of echopype. We recommend using `open_raw` to re-convert the raw data files.\n",
+        ":::"
       ]
     },
     {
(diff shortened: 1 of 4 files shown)


leewujung:
@emiliom : Looking at the tests, if we take the approach to just removing the auto-conversion under the hood of `open_converted`, we can just remove the entire `tests/echodata/test_echodata_structure.py` altogether. What do you think? These test have not been run for a long time now anyway...

emiliom:
> This PR removes the automatic EchoData format conversion when using open_converted on already converted data from earlier versions of echopype.
>
> This PR does not remove the previous data format conversion mechanisms (under echodata/sensor_ep_version_mapping) in case we are able to reinstate this capability in the future.

I agree with that approach. Thanks!

> @emiliom : Looking at the tests, if we take the approach to just removing the auto-conversion under the hood of open_converted, we can just remove the entire tests/echodata/test_echodata_structure.py altogether. What do you think? These test have not been run for a long time now anyway...

You're right, the only actual test in there was set to `xfail` quite a while ago. But I'm torn about whether to remove the test module file altogether ...

docs/source/data-format-changes.ipynb
```diff
@@ -41,9 +41,17 @@
...
+        "- Improve the presence and use of variable attributes throughout `EchoData` groups."
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "e506a51d",
+      "metadata": {},
+      "source": [
+        ":::{Note}\n",
+        "Version 0.8.0 does not incorporte the capability to read files converted by previous versions of echopype. We recommend using `open_raw` to re-convert the raw data files.\n",
```
emiliom, line 53: Fixing spelling mistake.
```suggestion
        "Version 0.8.0 does not incorporate the capability to read files converted by previous versions of echopype. We recommend using `open_raw` to re-convert the raw data files.\n",
```

docs/source/data-format-changes.ipynb
```diff
@@ -93,29 +101,6 @@
         "More details, including Pull Requests and discussions related to these changes, can be found in the [Release notes](whats-new.html#v0-6-0-2022-may-26)."
       ]
     },
-    {
```
emiliom, line 96: Good call. I agree with removing this stuff.

emiliom (approved):
Your changes look good, thanks. 

I can't decide if I'd rather keep `test_echodata_structure.py` even though it's not doing anything. You decide!

leewujung:
> I can't decide if I'd rather keep test_echodata_structure.py even though it's not doing anything. You decide!

Haha, ok. I decided that we'll keep it under a different filename (`test_echodata_version_convert`, instead of `test_echodata_structure`), and I've added notes on the reason behind this decision -- I  think these may be useful to keep as reference, just like how we are keeping `echodata/sensor_ep_version_mapping.py`.

I'll merge this now!

Issue #1144, opened 2023-08-31
Bump python requirement to python 3.9 and above


lsetiawan:
## Overview

I've found that xarray has pinned to python >=3.9 since February 2023 as reflected in release [`v2023.02.0`](https://github.com/pydata/xarray/blob/v2023.02.0/setup.cfg#L76). Currently there are failures for Python 3.8 in PR #1124 because it's using xarray version `v2023.01.0` with Pandas > 2. So I think it would be good to bump up the python minimum requirement since we are using xarray heavily.

emiliom:
Makes sense.

PR #1145, opened 2023-08-31
test: Remove NASC import causing failure

lsetiawan:
Commented out import for 'compute_NASC' causing the only error in CI at the moment seen in https://github.com/OSOceanAcoustics/echopype/actions/runs/6032689378/job/16368234856#step:11:1539 and https://github.com/OSOceanAcoustics/echopype/actions/runs/6032689378/job/16368234931#step:11:1539

PR #1146, opened 2023-08-31
chore(python): Bump python min to 3.9 [all tests ci]

lsetiawan:
Bumping python requirement to >=3.9 following xarray.

Ref: #1144

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
         runs-on: [ubuntu-latest]
         experimental: [false]
         # TODO: Uncomment when netcdf is unpinned
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
         runs-on: [ubuntu-latest]
         experimental: [false]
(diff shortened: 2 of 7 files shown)


lsetiawan:
This will fail since PR https://github.com/OSOceanAcoustics/echopype/pull/1145 is not merged yet.

.github/workflows/build.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
         runs-on: [ubuntu-latest]
         experimental: [false]
         # TODO: Uncomment when netcdf is unpinned
```
emiliom, line 27: This is unrelated to this PR per say, but is this TODO something we should act on now that netcdf is unpinned?

.github/workflows/pr.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
         runs-on: [ubuntu-latest]
         experimental: [false]
         # TODO: Uncomment when netcdf is unpinned
```
emiliom, line 23: This is unrelated to this PR per say, but is this TODO something we should act on now that netcdf is unpinned?

emiliom (approved):
Looks good to me (assuming the CI runs successfully). Thanks!

.github/workflows/build.yaml
lsetiawan: Thanks for catching that! I'll add to a separate PR

lsetiawan:
Currently CI will most likely fail: https://github.com/OSOceanAcoustics/echopype/pull/1146#issuecomment-1701575398

Issue #1147, opened 2023-08-31
Activate Python 3.11 CI testing


lsetiawan:
## Overview

@emiliom pointed out that there's a TODO in Github action for python 3.11 testing that now should be activated since we have unpinned netcdf4.

_Originally posted by @emiliom in https://github.com/OSOceanAcoustics/echopype/pull/1146#discussion_r1312070181_

emiliom:
And note that in #1146 you didn't add Python 3.11 to the suite of Python versions. I didn't comment on that, but I think it was appropriate to leave that addition to a separate, follow up PR, and maybe *after* the upcoming release if it introduces new issues.

PR #1148, opened 2023-08-31
ci: Add Python 3.11 testing [all tests ci]

lsetiawan:
Uncommented matrix for python-version 3.11 to start experimenting with Python 3.11 to see if it works.

------

Ref: #1147


leewujung (approved):
Thanks @lsetiawan , this is great!

Issue #1127, opened 2023-08-17
Further improve 'compute_NASC' to use flox


lsetiawan:
## Overview

To add depth and location see functions in `consolidate` module.

lsetiawan:
Done in #1167

PR #1150, opened 2023-09-01
Patch/nasc

simedroniraluca:
## NASC (Nautical Area Scattering Coefficient) Computation Correction

This PR is opened to propose a solution for https://github.com/OSOceanAcoustics/echopype/pull/1136

### Changes Made:

1. **Conversion Factors:** Conversion factors have been incorporated to facilitate the conversion from backscattering cross-section to scattering cross-section and from meters to nautical miles (nmi).

2. **Computation Approach:**
   - Cells Selection: Cells are chosen based on their depth and distance. The distance calculation is based on latitude and longitude coordinates.
   - NASC Calculation: For each channel and cell, the NASC is computed using the following steps:
     - Multiply the Sv (in linear units) with the height of the samples. The sample height is determined as the difference between consecutive depth values.
     - Compute the mean of the obtained values.
     - Multiply the mean value by the number of samples within the selected cell per the number of pings in the same cell.
     - Further multiply the result by $4 \times \pi \times 1852^2$
     - Finally, divide the obtained result by the percentage of non-NaN samples present in the analysis.

### DIFF
--- a/echopype/commongrid/__init__.py
+++ b/echopype/commongrid/__init__.py
@@ -1,6 +1,7 @@
-from .api import compute_MVBS, compute_MVBS_index_binning
+from .api import compute_MVBS, compute_MVBS_index_binning, compute_NASC
 
 __all__ = [
     "compute_MVBS",
     "compute_MVBS_index_binning",
+    "compute_NASC",
 ]
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -1,12 +1,20 @@
 """
 Functions for enhancing the spatial and temporal coherence of data.
 """
+from typing import Union
+
 import numpy as np
 import pandas as pd
 import xarray as xr
 
 from ..utils.prov import add_processing_level, echopype_prov_attrs, insert_input_processing_level
 from .mvbs import get_MVBS_along_channels
+from .nasc import (
+    check_identical_depth,
+    get_depth_bin_info,
+    get_dist_bin_info,
+    get_distance_from_latlon,
+)
 
 
 def _set_var_attrs(da, long_name, units, round_digits, standard_name=None):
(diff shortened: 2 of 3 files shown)


leewujung:
Hey @simedroniraluca : Thanks very much for pointing out the bug and sending in the above! I did some investigation and comparison and put the results in [this gist](https://gist.github.com/leewujung/876174a6f8b8e9da2d0c79c04058b11c). In it I compare the `echopy` implementation with the `Echoview` one, and an adaption that we'll use in `echopype` using `flox`.

For the bug in #1136 and the different implementations, it boils down to:
- for some reason I completely forgot about the scaling to the reference units of nautical miles and the mean "height" of the pixels when trying to implement the `Echoview` NASC in the first implementation
- `echopy` handles NaN samples in Sv data by a scaling, which is probably up for debate
- `echopy` computes ping-by-ping NASC, while `Echoview` computes NASC in an average sense across the defined cell depth range and distance range

To make `compute_NASC` more efficient we are moving away from the looping approach (#1127). I'll send in a separate and **incomplete** PR in a moment that has the barebone stuff using `flox`, which @lsetiawan will carry forward to finish it.

leewujung:
For visibility, here I copy-paste some notes from [the gist](https://gist.github.com/leewujung/876174a6f8b8e9da2d0c79c04058b11c):

### Regarding `echopy` implementation
> The implementation in `echopy` integrated Sv between the range limits `r0` and `r1`.
> 
> There are a few specifics to note:
> - if `method="mean"`, the function takes the mean of `sv*sh` and multiply that by the number of vertical samples, and treat this number as the integration output
> - if `method=""sum`, the function just takes the sum of `sv*sh` to be the integration output
> - the resulting integrated output is "corrected" by scaling according to the number of valid samples in each column (`distance_nmi`). For example, if there are only 2 non-NaN samples out of 3 total samples along depth, then the integrated output `NASC_raw` is scaled by 3/2: `NASC_corrected = NASC_raw / (2/3)`
> - the code also assumes that the "height" of the samples are conserved across ping, which may not be the case unless the `Sv` data has been regridded
> 
> I am not sure about the final correction of the NASC based on the missing non-NaN values. This would result in inflating the NASC number if there are a lot of NaNs in `sv`, and would not be accurate if the goal is to obtain `NASC` from masked `Sv` dataset.


### Regarding `echopy` vs `Echoview` differences
> - the Echoview implementation computes NASC in a sense of a mean across multiple pings
> - the Echopy implementation computes ping-by-ping NASC with some scaling to compensate for potential NaN samples along depth in some pings

leewujung:
Hey @simedroniraluca: I think we can close this issue and move our discussions to #1167, so that everything is in one place. Thanks again for your contribution!!

Issue #1153, opened 2023-09-01
xarray warnings when EK60/80 `Platform` `time1` contains a `NaT` value


emiliom:
While running the echopype-examples notebooks, I noticed three warnings that probably have no known impacts, but can be of concern to users who see them:

1. ... site-packages/xarray/coding/times.py:254: RuntimeWarning: invalid value encountered in cast flat_num_dates_ns_int = (flat_num_dates * _NS_PER_TIME_DELTA[delta]).astype(
2. ... site-packages/xarray/coding/times.py:618: RuntimeWarning: invalid value encountered in cast int_num = np.asarray(num, dtype=np.int64)
3. ... site-packages/xarray/core/duck_array_ops.py:188: RuntimeWarning: invalid value encountered in cast  return data.astype(dtype, **kwargs)

In [notebook 3 (OOI data)](https://osoceanacoustics.github.io/echopype-examples/ms_OOI_EK60_mooringtimeseries.html), they're happening with `open_raw` and `to_zarr` (all 3 warnings, but especially 1 & 2) with every one of the 19 raw files that are handled, and with `open_converted` (warning 1 only) with only 3 of the converted files. In [notebook 2 (hake data)](https://osoceanacoustics.github.io/echopype-examples/ms_PacificHake_EK60_cruisetracks.html) it's harder to isolate, but I'm seeing only warning 3, and they only happen 3 times, which I assume means 3 files out of the total 170 raw files that are handled. In notebook 1, handles only 5 raw files, none of them occur.

This [PR](https://github.com/OSOceanAcoustics/echopype-examples/pull/37) shows the notebooks with their warnings.

After some digging with a couple of the OOI files, it's clear that the source of all 3 warnings is the presence of a `Platform` `time1` coordinate that has a `NaT` value (actually, it is length one and its value is `NaT`). It looks to me like xarray's behavior has changed (ie, a previous version was not emitting these warnings), b/c I'm testing the single file that's examined initially in echopype-examples notebook 3. On the current version online, `Platform` `time1` does have a single NaT value, but the warning was not emitted. See [here](https://osoceanacoustics.github.io/echopype-examples/ms_OOI_EK60_mooringtimeseries.html#converting-from-raw-data-files-to-a-standardized-data-format). Now, the `open_raw` statement that creates the EchoData object emits the warning (see PR #37); `to_zarr` and `to_netcdf` on that file emit the other warnings.

[...]

PR #1154, opened 2023-09-01
Set EK60/80 `Platform` and `NMEA` nan timestamp to first `ping_time` value [all tests ci]

emiliom:
Addresses the first two xarray warnings identified in #1153. Implements the solution described there, for EK60.

Setting single-valued nan `Platform.time1` cases to the first `ping_time` value is consistent with the approach currently used with AZFP. EK80 may benefit from this change, too, but I'm not aware of a test case raw file, plus we're short on time.

See discussion in #1153.

**[@leewujung edits for future reference]**: Saildrone EK80 data by default does not have the lat/lon data since GPS is not plugged into the echosounder itself.

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -158,15 +158,22 @@ class SetGroupsAZFP(SetGroupsBase):
         """Set the Platform group."""
         platform_dict = {"platform_name": "", "platform_type": "", "platform_code_ICES": ""}
         unpacked_data = self.parser_obj.unpacked_data
-        time2 = self.parser_obj.ping_time
-        time1 = [time2[0]]
+
+        # Create nan time coordinate for lat/lon (lat/lon do not exist in AZFP 01A data)
+        time1 = [np.nan]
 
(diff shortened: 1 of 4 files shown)


leewujung:
> EK80 may benefit from this change, too, but I'm not aware of a test case raw file, plus we're short on time.

The saildrone EK80 data by default does not have the lat/lon data since GPS is not plugged into the echosounder itself, so those would be the example raw files.

leewujung:
Ok, I ended up discovering more cases like this using the EK80 file. To summarize, these are:
- EK60: Platform.time1
- EK80: Platform.time1, Platform.time2
- Platform/NMEA.time (this one did not show up for the OOI files because they happen to have a single, non-GPS NMEA message)

Since they all can be handled in the same way (@emiliom I changed your EK60 implementation to account for cases where the channels may not ping simultaneously), I added a new private method `SetGroupsBase._nan_timestamp_handler` for correcting a single nan value timestamp.

leewujung:
I'll self-merge this now to check everything is in good order for releasing v0.8.1.

PR #1156, opened 2023-09-02
ci: Temporarily remove Python 3.11 [all tests ci]

leewujung:
This PR temporarily removes Python 3.11 testing and packaging until `parsed2zarr` overhaul is completed, which likely would either remove `more_itertools` as a dependency or unpin it (currently pinned to `more_itertools=8.13.0`).


leewujung:
Hey @lsetiawan : I am very puzzled by why in the "experimental" 3.11 testing runs tests like `echopype/tests/convert/test_parsed_to_zarr.py::test_raw2zarr[noaa_offloaded]` did not fail, but when 3.11 is tested for real, the same test would fail.

All of the failure is related to the use of `more_itertools` in the current `parsed2zarr` module. That library was pinned to 8.13.0, but the latest is already 10.1.0. I am hope in #1070 you can either completely remove this dependency or unpin it. For v0.8.1, I'll temporarily remove 3.11 just to be safe. Thanks!

PR #1167, opened 2023-09-17
fix(commongrid): fix bugs and improve `compute_NASC` using flox

leewujung:
## Overview

@simedroniraluca found out about bugs in the previous `commongrid.compute_NASC` function (see #1150 for discussions), so we temporarily removed it before releasing 0.8.0 (#1136). This PR:
1. fix the bugs in the previous implementation
2. improve performance using `flox`

## Tasks
- [x] debug and compare NASC implementation -- see [this gist](https://gist.github.com/leewujung/3b058ab63c3b897b273b33b907b62f6d)
- [x] revive `compute_NASC` and add barebone flox implementation and TODOs
- [x] factor out core NASC computation and mean over lat/lon and ping_time to `commongrid/nasc.py`, as has been done in `compute_MVBS`
- [x] fix old tests
- [x] add new tests using mock data that test for both output NASC dimensions and values
- [x] add tests to compare with Echoview NASC outputs

## Notes
@lsetiawan: The tests will fail, because the `compute_NASC` function is not complete (I have not add in `channel` in the flox statement), but I think the changes give you enough to work with to finish the whole thing. Feel free to direct push to my branch.

I pulled in your changes in #1124 so that I can reuse some functions, so the commit messages and files changed tab are a mess right now, but obviously we'll pull in `dev` once #1124 is merged to clean this up. All my changes in this PR are only in `commongrid/api.py`.

For the last item on the task list: I will see if it's possible to get NOAA colleagues' help to generate from test data from Echoview, since I don't have a dongle/license to do it myself.


leewujung:
@lsetiawan : when testing a prototype of this I ran into the following FutureWarning. Would be good to take care of it here:
```
FutureWarning: Series.fillna with 'method' is deprecated and will raise in a future version. Use obj.ffill() or obj.bfill() instead.
  df_pos["dist"] = df_pos["dist"].fillna(method="ffill").fillna(method="bfill")
```
This is in the function `nasc.py::_get_distance_from_latlon`

lsetiawan:
https://github.com/OSOceanAcoustics/echopype/pull/1167#issuecomment-1732651809 Fixed in https://github.com/OSOceanAcoustics/echopype/pull/1167/commits/7661fee8ec543eeece3b926d368687787631c692

echopype/tests/commongrid/test_api.py
```diff
@@ -32,11 +41,150 @@ def test__parse_x_bin(x_bin, x_label, expected_result):
...
+
 # NASC Tests
 @pytest.mark.integration
-@pytest.mark.skip(reason="NASC is not implemented yet")
-def test_compute_NASC(test_data_samples):
-    pass
+@pytest.mark.parametrize("compute_mvbs", [True, False])
+def test_compute_NASC(request, test_data_samples, compute_mvbs):
+    if any(request.node.callspec.id.startswith(id) for id in ["ek80", "azfp"]):
+        pytest.skip("Skipping NASC test for ek80 and azfp, no data available")
```
lsetiawan, line 117: Will need valid test files that have lat lon

echopype/commongrid/api.py
```diff
@@ -163,6 +168,242 @@ def _parse_x_bin(x_bin: str, x_label="range_bin") -> float:
...
+
+    if x_var == "ping_time":
+        # This is MVBS computation
+        # apply inverse mapping to get back to the original domain and store values
+        da_MVBS = sv_mean.pipe(_lin2log)
+        return xr.merge([ds_Pos, da_MVBS])
+    else:
+        # Get mean ping_time along distance_nmi
+        # this is a feature only available for NASC
+        # computation
```
leewujung, line 359: Just for clarification:
```suggestion
        # Get mean ping_time along distance_nmi
        # this is only done for NASC computation,
        # since for MVBS the ping_time is used for binning already.
```

echopype/commongrid/api.py
```diff
@@ -163,6 +168,242 @@ def _parse_x_bin(x_bin: str, x_label="range_bin") -> float:
...
+    if all(v in ds_Sv for v in POSITION_VARIABLES):
+        ds_Pos = xarray_reduce(
+            ds_Sv[POSITION_VARIABLES],
+            ds_Sv[x_var],
+            func="nanmean",
+            expected_groups=(x_interval),
+            isbin=True,
+            method=method,
+        )
+        ds_Pos.attrs["has_positions"] = True
```
leewujung, line 334: Let's move this block to a separate function called in `compute_MVBS` or `compute_NASC`, so that `get_x_along_channels` does not have a "hidden" handling of lat/lon positions.

echopype/commongrid/api.py
```diff
@@ -415,131 +644,151 @@ def compute_MVBS_index_binning(ds_Sv, range_sample_num=100, ping_num=100):
...
+            range_var: np.array([v.left for v in raw_NASC[f"{range_var}_bins"].values]),
+        },
+    )
+
+    # "has_positions" attribute is inserted in get_x_along_channels
+    # when the dataset has position information
+    # propagate this to the final MVBS dataset
+    if raw_NASC.attrs.get("has_positions", False):
+        for var in POSITION_VARIABLES:
+            ds_NASC[var] = (["distance"], raw_NASC[var].data, ds_Sv[var].attrs)
```
leewujung, line 761: Use new function per https://github.com/OSOceanAcoustics/echopype/pull/1167/files#r1341591543

echopype/commongrid/api.py
```diff
@@ -415,131 +644,151 @@ def compute_MVBS_index_binning(ds_Sv, range_sample_num=100, ping_num=100):
...
-#     )
-#     ds_NASC.attrs["geospatial_lat_min"] = round(float(ds_Sv["latitude"].min().values), 5)
-#     ds_NASC.attrs["geospatial_lat_max"] = round(float(ds_Sv["latitude"].max().values), 5)
-#     ds_NASC.attrs["geospatial_lon_min"] = round(float(ds_Sv["longitude"].min().values), 5)
-#     ds_NASC.attrs["geospatial_lon_max"] = round(float(ds_Sv["longitude"].max().values), 5)
-
-#     return ds_NASC
+def compute_NASC(
+    ds_Sv: xr.Dataset,
+    range_bin: str = "20m",
```
leewujung, line 649:
```suggestion
    range_bin: str = "10m",
```

echopype/commongrid/api.py
```diff
@@ -415,131 +644,151 @@ def compute_MVBS_index_binning(ds_Sv, range_sample_num=100, ping_num=100):
...
+) -> xr.Dataset:
+    """
+    Compute Nautical Areal Scattering Coefficient (NASC) from an Sv dataset.
+
+    Parameters
+    ----------
+    ds_Sv : xr.Dataset
+        A dataset containing Sv data.
+        The Sv dataset must contain ``latitude``, ``longitude``, and ``depth`` as data variables.
+    range_bin : str, default '20m'
```
leewujung, line 663:
```suggestion
    range_bin : str, default '10m'
```

echopype/commongrid/api.py
```diff
@@ -415,131 +644,151 @@ def compute_MVBS_index_binning(ds_Sv, range_sample_num=100, ping_num=100):
...
+    xr.Dataset
+        A dataset containing NASC
+
+    Notes
+    -----
+    The NASC computation implemented here generally corresponds to the Echoview algorithm PRC_NASC
+    https://support.echoview.com/WebHelp/Reference/Algorithms/Analysis_Variables/PRC_ABC_and_PRC_NASC.htm#PRC_NASC  # noqa
+    The difference is that since in echopype masking of the Sv dataset is done explicitly using
+    functions in the ``mask`` subpackage, the computation only involves computing the
+    mean Sv and the mean height within each cell.
```
leewujung, line 688:
```suggestion
    functions in the ``mask`` subpackage, the computation only involves computing the
    mean Sv and the mean height within each cell, where some Sv "pixels" may have been
    masked as NaN.
```

leewujung (commented):
Hey @lsetiawan : Thanks for the great changes! My comments are all pretty minor, mostly tidying up kind. Please see below. 

I got some potential test files that are exports from Echoview, but since along the distance dimension, they could only do the binning either based on a variable called "interval" that is externally set, or based on the number of pings, it's a little different than the distance-based binning we can do automatically here. I'll look into that some more and scoping out how to add it later. For now, let's get this PR merged! 😀

Comments:
- I like the creation of `get_x_along_channels`! I would suggest changing the triaging to be explicit based on `x`: `x="MVBS"` or `x="NASC"`, which then selects `x_var` under the hood to be `ping_time` or `distance_nmi`. The current triaging works but can be pretty confusing in the future.

- except for `get_distance_from_latlon`, everything in `nasc.py` are no longer used. How creating a `utils.py` or `helper.py` and move all functions other than `compute_MVBS`, `compute_MVBS_index_binning`, `compute_NASC`, and `regrid` there?

- remove `_set_MVBS_attrs` and use `_set_var_attrs` directly here: https://github.com/OSOceanAcoustics/echopype/blob/7ec7a1c7d641a49bd703b76241d1917698dfe043/echopype/commongrid/api.py#L145

- remove `test_nasc.py` since it's now empty :P

echopype/commongrid/api.py
leewujung: Ok! :D

lsetiawan:
This should now be ready for 2nd review:

> I like the creation of get_x_along_channels! I would suggest changing the triaging to be explicit based on x: x="MVBS" or x="NASC", which then selects x_var under the hood to be ping_time or distance_nmi. The current triaging works but can be pretty confusing in the future.

I've broken this up to `_groupby_x_along_channels`, `compute_raw_MVBS`, and `compute_raw_NASC` to be more explicit for each step. Additionally, now for position variables propagation, I added a `_get_reduced_positions` function that will add it or not, so this is more visible, rather than hidden.

> except for get_distance_from_latlon, everything in nasc.py are no longer used. How creating a utils.py or helper.py and move all functions other than compute_MVBS, compute_MVBS_index_binning, compute_NASC, and regrid there?

Agreed. Now all the helper functions have been moved.

> remove _set_MVBS_attrs and use _set_var_attrs directly here

Since the `_set_MVBS_attrs` is used twice in compute MVBS and the MVBS binning index thing, I think I'm gonna leave this alone to not repeat the 2 steps in there.

> remove test_nasc.py since it's now empty :P

Done.

leewujung:
@lsetiawan : Thanks for the changes, they look great! I love how clean the code is now. I just added the processing level decorator for L4* to `compute_NASC`. Once the tests all run through I'll merge this PR. Super excited!

> Since the _set_MVBS_attrs is used twice in compute MVBS and the MVBS binning index thing, I think I'm gonna leave this alone to not repeat the 2 steps in there.

I agree with this choice.

PR #1169, opened 2023-09-18
Add seabed masks

ruxandra-valcu:
### Overview:

This PR introduce a new function, get_seabed_mask, that provide masks for seabed and false seabed detections. These are ported from the echopy library:

### Methodologies:

The methology employed uses the following methods: 
- maxSv - Initially detects the seabed as the ping sample with the strongest Sv value, as long as it exceeds a dB threshold. Then it searches up along the ping until Sv falls below a secondary (lower) dB threshold, where the final seabed is set.
- deltaSv - Examines the difference in Sv over a 2-samples moving window along every ping, and returns the range of the first value that exceeded  a user-defined dB threshold (likely, the seabed).
- blackwell - Detects and masks false seabed echoes using the split-beam angle and Sv, based in  "Blackwell et al (2019), Aliased seabed detection in fisheries acoustic data". Complete article here: https://arxiv.org/abs/1904.10736
- blackwell_mod - a modified version of the original algorithm, including extra arguments to evaluate whether aliased seabed items can occur, give, the true seabed detection range, and the possibility of tuning the percentile's rank.
- experimental - Mask Sv above a threshold to get a potential seabed mask. Then, the mask is  dilated to fill seabed breaches, and small objects are removed to prevent  masking high Sv features that are not seabed (e.g. fish schools or spikes).  Once this is done, the mask is built up until Sv falls below a 2nd threshold, Finally, the mask is extended all the way down.
- ariza -  Mask Sv above a threshold to get potential seabed features. These features are eroded first to get rid of fake seabeds (spikes, schools, etc.) and  dilated afterwards to fill in seabed breaches. Seabed detection is coarser than other methods (it removes water nearby the seabed) but the seabed line  never drops when a breach occurs. Suitable for pelagic assessments and reconmended for non-supervised processing.


leewujung:
Closing this as new function will be added.

PR #1170, opened 2023-09-18
Add shoal mask

ruxandra-valcu:
### Overview:

This PR introduce a new function, get_shoal_mask, that provides masks for shoal detections. These are ported from the echopy library:

### Methodologies:

Detects and masks shoals following the algorithm described in:

        "Will et al. (1993): MOVIES-B — an acoustic detection description
        software . Application to shoal species' classification".

    Contiguous Sv samples above a given threshold will be considered as the
    same shoal, as long as it meets the contiguity criteria described by Will
    et al. (1993):

        * Vertical contiguity: along-ping gaps are allowed to some extent.
        Typically, no more than the number of samples equivalent to half of the
        pulse length.

        * Horizontal contiguity: above-threshold features from contiguous pings
        will be regarded as the same shoal if there is at least one sample in
        each ping at the same range depth.

    Although the default settings strictly complies with Weill's contiguity
    criteria, other contiguity arguments has been enabled in this function to
    increase operability. For instance, the possibility to allow gaps in the
    horizontal, or to set minimum vertical and horizontal lengths for a feature
    to be regarded as a shoal. These settings are set to zero (not applicable)
    by default.


leewujung:
Closing this as new function will be added.

Issue #1171, opened 2023-09-18
Add storage for XML parameters from >1 phases in AZFP


leewujung:
With the new XML parser in #1135, we can now handle more than 1 phase in AZFP data. However, currently only the phase 1 parameters parsed from XML are stored in the `Vendor_specific` group. I think we can add an additional dimension `phase` to the XML parameters that are under the different phases to store all the information.

emiliom:
Ideally, we should get an AZFP file that does have >1 phase, for testing. Currently it doesn't look like we do.

leewujung:
I added one actually, it's already in the CI. I've pointed @praneethratna to it. It is the one Steve Pearce provided.

emiliom:
I just checked the `CI_test_data` Google Drive folder. Are you referring to `azfp/23081211.XML`? If so, Steve only sent us the XML file, not the data (01A) file. The folder contains three XML files but only two .01A files.

leewujung:
Yes, that is the file. I think for testing parsing the XML that is sufficient, but maybe you're thinking about going all the way to the `set_vendor` part? Maybe we can ask Steve for an example 01A file?

emiliom:
> but maybe you're thinking about going all the way to the set_vendor part? 

Yup

> Maybe we can ask Steve for an example 01A file?

Yeah, hopefully he has the .01A file that goes with the XML file, or another complete set.

praneethratna:
> I added one actually, it's already in the CI. I've pointed @praneethratna to it. It is the one Steve Pearce provided.

Yes, I'll actually change the test to use this new file for better tests of new code.

PR #1182, opened 2023-09-24
added support for AZFP multiple phase attributes parsing

praneethratna:
Addresses #1171 and now XML parameters from multiple phases are stored in the `Vendor_Specific` group.

### DIFF
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -91,26 +91,32 @@ class ParseAZFP(ParseBase):
         """
 
         xmlmap = fsspec.get_mapper(self.xml_path, **self.storage_options)
-        root = ET.parse(xmlmap.fs.open(xmlmap.root)).getroot()
+        phase_number = None
+        for event, child in ET.iterparse(xmlmap.fs.open(xmlmap.root), events=("start", "end")):
+            if event == "end" and child.tag == "Phases":
+                phase_number = None
+            if event == "start":
+                if len(child.tag) > 3 and not child.tag.startswith("VTX"):
+                    camel_case_tag = camelcase2snakecase(child.tag)
+                else:
+                    camel_case_tag = child.tag
 
-        for child in root.iter():
-            if len(child.tag) > 3 and not child.tag.startswith("VTX"):
-                camel_case_tag = camelcase2snakecase(child.tag)
-            else:
-                camel_case_tag = child.tag
-            if len(child.attrib) > 0:
-                for key, val in child.attrib.items():
-                    self.parameters[camel_case_tag + "_" + camelcase2snakecase(key)].append(val)
+                if len(child.attrib) > 0:
+                    for key, val in child.attrib.items():
+                        self.parameters[camel_case_tag + "_" + camelcase2snakecase(key)].append(val)
+                        if child.tag == "Phase":
+                            phase_number = val
 
-            if all(char == "\n" for char in child.text):
-                continue
-            else:
-                try:
-                    val = int(child.text)
-                except ValueError:
-                    val = float(child.text)
-
-            self.parameters[camel_case_tag].append(val)
+                if all(char == "\n" for char in child.text):
+                    continue
+                else:
+                    try:
+                        val = int(child.text)
+                    except ValueError:
+                        val = float(child.text)
+                    if phase_number is not None and camel_case_tag != "phase":
+                        camel_case_tag += f"_phase{phase_number}"
+                    self.parameters[camel_case_tag].append(val)
 
         # Handling the case where there is only one value for each parameter
         for key, val in self.parameters.items():
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -103,7 +103,10 @@ class ParseAZFP(ParseBase):
 
                 if len(child.attrib) > 0:
                     for key, val in child.attrib.items():
-                        self.parameters[camel_case_tag + "_" + camelcase2snakecase(key)].append(val)
+                        attrib_tag = camel_case_tag + "_" + camelcase2snakecase(key)
+                        if phase_number is not None and camel_case_tag != "phase":
+                            attrib_tag += f"_phase{phase_number}"
+                        self.parameters[attrib_tag].append(val)
                         if child.tag == "Phase":
                             phase_number = val
 
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -815,7 +815,7 @@ class SetGroupsAZFP(SetGroupsBase):
                 "ad_len": (["ad_len"], list(range(len(unpacked_data["ad"][0])))),
                 "phase_number": (
                     ["phase_number"],
-                    sorted([int(num) for num in parameters["phase_number"]], reverse=False),
+                    sorted([int(num) for num in parameters["phase_number"]]),
                 ),
             },
         )
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -9,6 +9,18 @@ import xarray as xr
 from ..utils.coding import set_time_encodings
 from .set_groups_base import SetGroupsBase
 
+PHASE_PARAMS = ["burst_interval", "pings_per_burst", "average_burst_pings"]
+
+PHASE_FREQ_PARAMS = [
+    "acquire_frequency",
+    "dig_rate",
+    "range_samples",
+    "range_averaging_samples",
+    "lock_out_index",
+    "gain",
+    "storage_format",
+]
+
 
 class SetGroupsAZFP(SetGroupsBase):
     """Class for saving groups to netcdf or zarr from AZFP data files."""
@@ -500,7 +512,18 @@ class SetGroupsAZFP(SetGroupsBase):
         unpacked_data = self.parser_obj.unpacked_data
         parameters = self.parser_obj.parameters
         ping_time = self.parser_obj.ping_time
-        tdn = parameters["pulse_len"][self.freq_ind_sorted] / 1e6
+        tdn = []
+        for num in parameters["phase_number"]:
+            tdn.append(parameters["pulse_len_phase{}".format(num)][self.freq_ind_sorted] / 1e6)
+        tdn = np.array(tdn)
+        for param in PHASE_FREQ_PARAMS:
+            for num in parameters["phase_number"]:
+                parameters[param].append(
+                    parameters[param + "_phase{}".format(num)][self.freq_ind_sorted]
+                )
+        for param in PHASE_PARAMS:
+            for num in parameters["phase_number"]:
+                parameters[param].append(parameters[param + "_phase{}".format(num)])
         anc = np.array(unpacked_data["ancillary"])  # convert to np array for easy slicing
 
         # Build variables in the output xarray Dataset
@@ -639,13 +662,13 @@ class SetGroupsAZFP(SetGroupsBase):
                 ),
                 # parameters with channel dimension from XML file
                 "XML_transmit_duration_nominal": (
-                    ["channel"],
+                    ["phase_number", "channel"],
                     tdn,
                     {"long_name": "(From XML file) Nominal bandwidth of transmitted pulse"},
                 ),  # tdn comes from parameters
                 "XML_gain_correction": (
-                    ["channel"],
-                    parameters["gain"][self.freq_ind_sorted],
+                    ["phase_number", "channel"],
+                    parameters["gain"],
                     {"long_name": "(From XML file) Gain correction"},
                 ),
                 "instrument_type": parameters["instrument_type"][0],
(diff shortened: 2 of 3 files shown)


echopype/convert/set_groups_azfp.py
```diff
@@ -500,7 +512,18 @@ def set_vendor(self) -> xr.Dataset:
         unpacked_data = self.parser_obj.unpacked_data
         parameters = self.parser_obj.parameters
         ping_time = self.parser_obj.ping_time
-        tdn = parameters["pulse_len"][self.freq_ind_sorted] / 1e6
+        tdn = []
+        for num in parameters["phase_number"]:
+            tdn.append(parameters["pulse_len_phase{}".format(num)][self.freq_ind_sorted] / 1e6)
```
emiliom, line 517: I recommend using f string formatting rather than `.format()`. It's typically easier and to read and the intention is clearer. Here's the same line using f string formatting:
```suggestion
            tdn.append(parameters[f"pulse_len_phase{num}"][self.freq_ind_sorted] / 1e6)
```

echopype/convert/set_groups_azfp.py
```diff
@@ -9,6 +9,18 @@
 from ..utils.coding import set_time_encodings
 from .set_groups_base import SetGroupsBase
 
+PHASE_PARAMS = ["burst_interval", "pings_per_burst", "average_burst_pings"]
+
+PHASE_FREQ_PARAMS = [
```
emiliom, line 14: These two lists are used only in `set_vendor`. Having them here, with a global, module-wide scope, seems a bit misleading. Plus, I think it'd be more helpful for readability if they were in fact declared at the start of `set_vendor`, because that's where they'll be used.

echopype/convert/set_groups_azfp.py
```diff
@@ -9,6 +9,18 @@
...
+    "acquire_frequency",
```
emiliom, line 15: This parameter is not used, is not saved in `set_vendor`. Drop, for now.

echopype/convert/set_groups_azfp.py
```diff
@@ -789,6 +818,10 @@ def set_vendor(self) -> xr.Dataset:
                     list(range(len(unpacked_data["ancillary"][0]))),
                 ),
                 "ad_len": (["ad_len"], list(range(len(unpacked_data["ad"][0])))),
+                "phase_number": (
+                    ["phase_number"],
+                    sorted([int(num) for num in parameters["phase_number"]], reverse=False),
```
emiliom, line 823: `reverse=False` is unnecessary, since that's the default.
```suggestion
                    sorted([int(num) for num in parameters["phase_number"]]),
```

echopype/convert/parse_azfp.py
```diff
@@ -91,26 +91,32 @@ def load_AZFP_xml(self):
...
 
-        for child in root.iter():
-            if len(child.tag) > 3 and not child.tag.startswith("VTX"):
-                camel_case_tag = camelcase2snakecase(child.tag)
-            else:
-                camel_case_tag = child.tag
-            if len(child.attrib) > 0:
-                for key, val in child.attrib.items():
-                    self.parameters[camel_case_tag + "_" + camelcase2snakecase(key)].append(val)
+                if len(child.attrib) > 0:
```
emiliom, line 104: Parameters created from attributes of phase-dependent elements are not being assigned a phase number string token (eg, "phase1"). This leads to the inconsistency of phase-dependent parameters where the element values are written into separate parameters (eg, `pings_per_burst_phase1 = 1`, `pings_per_burst_phase2 = 1`) but their attributes are written as an array of a single parameter (eg, `pings_per_burst_units = ['pings', 'pings']`). The change I'm suggesting here and in the next suggestion would make the handling of element attributes similar to that of the element values.

But, note that this doesn't seem to have any actual impact on data that are saved to the EchoData object! Apparently none of the phase-dependent attribute values are being used. Still, having a consistent structure for the parsing of element values and attributes into parameters will avoid confusion in the future.

```suggestion
                if phase_number is not None and camel_case_tag != "phase":
                    camel_case_tag += f"_phase{phase_number}"

                if len(child.attrib) > 0:
```

echopype/convert/parse_azfp.py
```diff
@@ -91,26 +91,32 @@ def load_AZFP_xml(self):
...
-            self.parameters[camel_case_tag].append(val)
+                if all(char == "\n" for char in child.text):
+                    continue
+                else:
+                    try:
+                        val = int(child.text)
+                    except ValueError:
+                        val = float(child.text)
+                    if phase_number is not None and camel_case_tag != "phase":
+                        camel_case_tag += f"_phase{phase_number}"
```
emiliom, line 118: Deleting this block goes together with the comment and suggested change above.
```suggestion

```

echopype/tests/convert/test_convert_azfp.py
emiliom, line 1: The changes to this test look fine. But how about changing the XML file that is tested, to use one that actually has >1 `<Phase>` element? To do that, change `azfp_xml_path = azfp_path / '17030815.XML'` to `azfp_xml_path = azfp_path / '23081211.XML'`.

BTW, `azfp_01a_path` is not actually needed! (I know that was there all along). I suggest removing the `azfp_01a_path` declaration and changing `parseAZFP = ParseAZFP(str(azfp_01a_path), str(azfp_xml_path))` to `parseAZFP = ParseAZFP(None, str(azfp_xml_path))`. I've tested this.

Once you change the test to  '23081211.XML', you'll have to add all the `*_phase2` parameter names to the test.

emiliom (changes requested):
See my comments and suggestions. The PR does what it's intended to do -- thank you!! It has no bugs. My suggestions are all just improvements or clarifications.

echopype/convert/set_groups_azfp.py
```diff
@@ -512,18 +500,25 @@ def set_vendor(self) -> xr.Dataset:
...
         for num in parameters["phase_number"]:
-            tdn.append(parameters["pulse_len_phase{}".format(num)][self.freq_ind_sorted] / 1e6)
+            tdn.append(parameters[f"pulse_len_phase{num}"][self.freq_ind_sorted] / 1e6)
         tdn = np.array(tdn)
         for param in PHASE_FREQ_PARAMS:
             for num in parameters["phase_number"]:
-                parameters[param].append(
-                    parameters[param + "_phase{}".format(num)][self.freq_ind_sorted]
-                )
+                parameters[param].append(parameters[param + f"_phase{num}"][self.freq_ind_sorted])
```
emiliom, line 518: Use f strings for entire string you're composing, not just part of it:
```suggestion
                parameters[param].append(parameters[f"{param}_phase{num}"][self.freq_ind_sorted])
```

echopype/convert/set_groups_azfp.py
```diff
@@ -512,18 +500,25 @@ def set_vendor(self) -> xr.Dataset:
...
         for param in PHASE_PARAMS:
             for num in parameters["phase_number"]:
-                parameters[param].append(parameters[param + "_phase{}".format(num)])
+                parameters[param].append(parameters[param + f"_phase{num}"])
```
emiliom, line 521: Use f strings for entire string you're composing, not just part of it:
```suggestion
                parameters[param].append(parameters[f"{param}_phase{num}"])
```

echopype/convert/set_groups_azfp.py
```diff
@@ -93,7 +81,7 @@ def _create_unique_channel_name(self):
         """
 
         serial_number = self.parser_obj.unpacked_data["serial_number"]
-        frequency_number = self.parser_obj.parameters["frequency_number"]
+        frequency_number = self.parser_obj.parameters["frequency_number_phase1"]
```
emiliom, line 84: Was this an oversight that you caught now, or a change required by your other recent changes?

echopype/convert/set_groups_azfp.py
```diff
@@ -512,18 +500,25 @@ def set_vendor(self) -> xr.Dataset:
...
         ping_time = self.parser_obj.ping_time
+        PHASE_PARAMS = ["burst_interval", "pings_per_burst", "average_burst_pings"]
+        PHASE_FREQ_PARAMS = [
+            "dig_rate",
+            "range_samples",
+            "range_averaging_samples",
+            "lock_out_index",
+            "gain",
+            "storage_format",
+        ]
```
emiliom, line 511: Since the scope of these two variables is no longer global to the module, I suggest using lower-case names instead.

echopype/convert/set_groups_azfp.py
praneethratna: I caught this now after incorporating `phase number` to a child attribute names as well.

echopype/tests/convert/test_convert_azfp.py
emiliom, line 1: Starting with line 228, most of the tests (asserts) done on both `_phase1` and `_phase2` parameters are identical for the two phases. Currently the tests duplicates the code fort the two phases. I suggest minimizing this duplication by setting up the tests that are identical across phases in a for loop (eg, `for phase in [1,2]:`), then using f strings to compose the parameter names (eg, `f"pulse_len_phase{phase}"`). That'll make the code more readable and more maintainable.

emiliom (changes requested):
Thanks for the updates! I have one or two questions and some suggestions, all easy to implement. The suggestion in the test is the only one that involves substantial changes, but they should be easy to implement, too.

echopype/convert/set_groups_azfp.py
emiliom: Ah, that makes sense. Thanks.

echopype/tests/convert/test_convert_azfp.py
praneethratna: Updated!

echopype/convert/set_groups_azfp.py
praneethratna: Updated, changed the names to lower-case.

echopype/convert/set_groups_azfp.py
praneethratna: Updated!

echopype/convert/set_groups_azfp.py
praneethratna: Updated!

echopype/tests/convert/test_convert_azfp.py
emiliom: Looks perfect, thank you!

emiliom:
Hmm, now all the tests are failing :cry: . @praneethratna can you look into the error messages to see what's going on?

praneethratna:
> Hmm, now some tests are failing 😢 . @praneethratna can you look into the error messages to see what's going on?

Yeah sure, but they doesn't seem to be related to my changes, but anyway closing and reopening the PR so that tests will run again.

emiliom:
> Yeah sure, but they doesn't seem to be related to my changes, but anyway closing and reopening the PR so that tests will run again.

Ah. Thanks for rerunning the tests. They're still failing. I'll try to take a closer look tomorrow. Your changes impact how AZFP files are processed; so, depending on which tests are failing, and on what data files, it *is* possible that the changes are causing this. But it sounds like you've already looked closely and discarded that possibility?

praneethratna:
> > Yeah sure, but they doesn't seem to be related to my changes, but anyway closing and reopening the PR so that tests will run again.
> 
> Ah. Thanks for rerunning the tests. They're still failing. I'll try to take a closer look tomorrow. Your changes impact how AZFP files are processed; so, depending on which tests are failing, and on what data files, it *is* possible that the changes are causing this. But it sounds like you've already looked closely and discarded that possibility?

Yes, I've did a tear-down of docker containers and re deployed containers on my laptop and the tests are running fine. I'm not sure why they are failing here.

emiliom:
Ah, thanks! I didn't know you were now set up to run the test suite locally. Fantastic!

praneethratna:
> Ah, thanks! I didn't know you were now set up to run the test suite locally. Fantastic!

Ah, yes! I did set up the test suite locally during the first PR itself but I've had problems with conda on my laptop.

emiliom:
FYI I ran the tests locally with Python 3.10. There were 3 errors. Yeah, they do look to be unrelated to this PR. All 3 errors involve `echopype/tests/convert/test_parsed_to_zarr.py::TestParsed2Zarr::test__create_zarr_info`. 

I'll follow up with that elsewhere. We may end up merging this PR before we resolve those other problems, but for now I'd like to investigate further.

emiliom:
I am persuaded that this PR is fine and can be merged, and the CI errors are cause by an unrelated problem. So, I'll go ahead and merge it now. Thanks @praneethratna !!

I'll open a new issue to address the CI problem.

emiliom (approved):

emiliom:
I am persuaded that this PR is fine and can be merged, and the CI errors are cause by an unrelated problem. So, I'll go ahead and merge it now. Thanks @praneethratna !!

I'll open a new issue to address the CI problem.

Issue #1179, opened 2023-09-21
Replace the current `parsed2zarr` mechanism


leewujung:
This issue is to pick up from #1070 to replace the current `parsed2zarr` mechanism to ensure code is bug-free and easy to maintain in the future.

I am referencing my https://github.com/OSOceanAcoustics/echopype/pull/1070#issuecomment-1730079171 here:

> The reason why i think the current complicated approach is unnecessary is because there are 2 stages of padding that is happening:
> 1. the "padding" for each channel to become rectangular (so to pad the shorter pings so that all pings within a channel have the same length) is in `pad_shorter_ping`. That is the first NaN-padding expansion.
> 2. The second NaN-padding expansion is when some channels lack a certain type of data, so NaN is padded for the entire "slice" of that channel to match the size of the other channels.
> 
> Expansion 1 happens for all data without xarray/zarr/dask (which can potentially be sped up or make more efficient?). After that, if we take note of the shape of variables in `self.ping_data_dict["angle"][ch_id]` for each `ch_id` , we would know exactly whether or not expansion 2 is needed. There, we can decide what is the strategy to use to write to the zarr store -- can we append sequentially along segments of `ping_time` since we know all the pings are same length? and for the channels that lack any data at all, creating a NaN slice for each sections (along `ping_time` segments) would likely let us hover under the memory limit.

@lsetiawan and I have done some investigation and determined a clean path forward.

leewujung:
Notes while going through code:

Right now the power, angle, and complex samples are all directed to be saved in `parser.zarr_datagrams` after `parser.parse_raw` -- I think this is unnecessary and messy. It helps with data integrity and therefore later maintenance to keep these variables along with others in `parser.ping_data_dict`.
- To change this, remove the handling specific to `zarr_vars` in all occurrence of the use of `_append_channel_ping_data`. 
- Better to keep a declaration up front for which variables should be handled with `swap=True` and select from `parser.ping_data_dict`

lsetiawan:
This Issue is related to #966

Issue #489, opened 2021-11-12
Large memory expansion when data are converted and organized into groups/EchoData object


leewujung:
This issue is split from #407 since the symptom is similar but the underlying cause is different.

Tagging @oftfrfbf, @lsetiawan @emiliom so that we can continue the discussion here. 😃 

This issue focuses on the problem that there is sometimes very large expansion of memory use when echopype is converting a file of moderate size.
(vs #407 focuses on the problem when the data file itself is too large to fit into memory.)

## Description

Based on @oftfrfbf and @lsetiawan 's investigations (quoted below), there may very well be two different types of things happening (or more? depending on the exact form of data):

- for [this 725 MB OOI file](https://rawdata-west.oceanobservatories.org/files/CE04OSPS/PC01B/ZPLSCB102_10.33.10.143/2018/08/07/OOI-D20180807-T000000.raw): 
  - 1 of the 3 channels is split-beam and produces `alongship_angle` and `athwartship_angle` data, but the other 2 single-beam channels do not produce this data.
  - The data are first parsed into lists and then organized/merged into xarray DataArray/Dataset. During the second stage @lsetiawan pointed out that there is an expansion of memory.
  - I suspect this is due to trying to pad NaNs to the (correctly) missing angle data slices in order to assemble a data cube with dimensions `frequency` (channel), `range_bin` and `ping_time`.

- for [the 95 MB file](https://noaa-wcsd-pds.s3.amazonaws.com/data/raw/Albatross_Iv/AL0409/EK60/L0003-D20040909-T161906-EK60.raw) shared by @oftfrfbf: 
  Below is @oftfrfbf's profiling result (quoted from [here](https://github.com/OSOceanAcoustics/echopype/issues/407#issuecomment-962110675)):
  > <img alt="Memory Profile Parsing Raw and Creating EchoData" width="700" src="https://user-images.githubusercontent.com/7818564/140557000-2035cac4-45cf-4049-9972-7d3e28f0fdb8.png">
  > 
  > Profiling for parser.parse_raw() (shown in red) only uses about 3 GB of memory while lines 424 through 445, accounting for the creation of the EchoData model (shown in blue), brought that value up to a total of 10 GB.

  - I wonder whether the recorded range within this file was changed dramatically between ping to ping? I haven't had a change to look into the file but will try to do so soon and report back.
[...]

PR #1183, opened 2023-09-27
Fixtures to generate mock data for clean up `use_swap` handling

leewujung:
@lsetiawan : in this PR you'll find functions that generate mock data for testing the `use_swap` functionality for #1179.

I made it such that the output is in identical format with `parser.ping_data_dict`, for `timestamp`, `power`, and `angle`.

See my comment here for more info: https://github.com/OSOceanAcoustics/echopype/issues/1179#issuecomment-1737733489

Because of the possible variations, I think using both the actual data for integration tests and mock data for unittest will be a good idea.

In terms of actual .raw data that we need to handle, see my previous investigation https://github.com/OSOceanAcoustics/echopype/issues/489#issuecomment-968340670 and https://github.com/OSOceanAcoustics/echopype/issues/489#issuecomment-968342133.

### DIFF
--- a/echopype/convert/parse_base.py
+++ b/echopype/convert/parse_base.py
@@ -282,7 +282,7 @@ class ParseEK(ParseBase):
                 self.ping_time[new_datagram["channel"]].append(new_datagram["timestamp"])
 
                 # Append ping by ping data
-                self._append_channel_ping_data(new_datagram)
+                self._append_channel_ping_data(new_datagram, zarr_vars=False)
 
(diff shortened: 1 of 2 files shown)


leewujung:
BTW, the tests WILL fail because I only did the short-circuiting to make it easy to see how data looks like without the `zarr_datagrams` business. I did not try to fix anything downstream, because they need to be removed.

lsetiawan:
Merged in #1185

PR #1185, opened 2023-10-02
refactor(convert): Full refactor for the usage of disk during memory expansion [all tests ci]

lsetiawan:
## Overview

This PR does a full refactor for the "swap" functionality by largely removing the whole usage of `Parsed2Zarr` objects in favor of a simpler direct usage of `zarr` library features to create arrays.

## Issues Resolved

- https://github.com/OSOceanAcoustics/echopype/issues/1179
- https://github.com/OSOceanAcoustics/echopype/issues/795
- https://github.com/OSOceanAcoustics/echopype/issues/782
- https://github.com/OSOceanAcoustics/echopype/issues/777


lsetiawan:
Note to self: Look into the potential use of context manager for `open_raw`? https://realpython.com/python-with-statement/

lsetiawan:
@leewujung I got all the unit test in that uses your mock data. It took me a bit to make sure that I'm taking consideration various things such as the shape and values in there. I was able to get the old integration test in there though I found some issues with one of the EK80 sample 😞. Also, I need to bring back the s3 side of testing... I might need one more day to get those in. But please do start reviewing when you get a chance to at least get familiar with how it is and see if anything need to change or confusing. Thanks.

lsetiawan:
> Also, I need to bring back the s3 side of testing... I might need one more day to get those in.

Thinking about this... I feel maybe for a "swap" it would be less complexity and just keep it on disk... since this is really just for temporary space when opening a file and it's not that large. We know that zarr write to s3 can be very slow, especially when it's such a small file, I am not sure if it's even worth it. Let me know what you think. Thanks.

leewujung:
> But please do start reviewing when you get a chance to at least get familiar with how it is and see if anything need to change or confusing. 
> We know that zarr write to s3 can be very slow, especially when it's such a small file, I am not sure if it's even worth it.

Let's connect Monday afternoon to go through the approach you are taking and the s3 consideration you have -- I think that's going to be more efficient than going back and forth here. Thanks.

echopype/convert/api.py
```diff
@@ -315,10 +313,9 @@ def open_raw(
     xml_path: Optional["PathHint"] = None,
     convert_params: Optional[Dict[str, str]] = None,
     storage_options: Optional[Dict[str, str]] = None,
-    use_swap: bool = False,
-    destination_path: Optional[str] = None,
+    destination_path: Optional[str] = "auto",
     destination_storage_options: Optional[Dict[str, str]] = None,
-    max_mb: int = 100,
+    max_chunk_size: str = "100MB",
 ) -> Optional[EchoData]:
```
lsetiawan, line 319: Let's not make this optional return

echopype/convert/api.py
```diff
@@ -347,10 +344,6 @@ def open_raw(
         and need to be added to the converted file
     storage_options : dict, optional
         options for cloud storage
-    use_swap: bool
-        **DEPRECATED: This flag is ignored**
-        If True, variables with a large memory footprint will be
-        written to a temporary zarr store at ``~/.echopype/temp_output/parsed2zarr_temp_files``
     destination_path: str
```
lsetiawan, line 347: Let's go back to `use_swap`: "auto", True, False

echopype/convert/api.py
```diff
@@ -347,19 +343,11 @@ def open_raw(
...
         options for cloud storage
-    use_swap: bool
-        **DEPRECATED: This flag is ignored**
-        If True, variables with a large memory footprint will be
-        written to a temporary zarr store at ``~/.echopype/temp_output/parsed2zarr_temp_files``
-    destination_path: str
-        The path to a swap directory in a case of a large memory footprint.
-        This path can be a remote path like s3://bucket/swap_dir.
+    use_swap: "auto" | bool
+        Flag to use disk swap in a case of a large memory footprint.
```
leewujung, line 347: Somehow I could not make suggestions on the lines below since it includes deleted lines, so I put the whole thing here:
```suggestion
    use_swap: "auto" | bool
        Flag to use disk swap in case of a large memory footprint. 
        When set to True (or when set to "auto" and large memory footprint is needed, 
        this function will create a temporary zarr store at the operating system's 
        temporary directory.
```

echopype/convert/api.py
```diff
@@ -385,20 +373,15 @@ def open_raw(
     In a case of a large memory footprint, the program will determine if using
```
leewujung, line 373:
```suggestion
    In case of a large memory footprint, the program will determine if using
```

echopype/convert/parse_base.py
```diff
@@ -69,54 +76,274 @@ def _print_status(self):
...
+
+    def _get_data_shapes(self) -> dict:
+        """Get all the expanded data shapes"""
+        all_data_shapes = {}
+        # Get all data type shapes
+        for raw_type in self.raw_types:
+            ping_data_dict = (
+                self.ping_data_dict_tx if raw_type == "transmit" else self.ping_data_dict
+            )
+            data_type_shapes = calc_final_shapes(self.data_types, ping_data_dict)
```
leewujung, line 104: Just adding some comments for readability as I go through the code
```suggestion
            # data_types: ["power", "angle", "complex"]
            data_type_shapes = calc_final_shapes(self.data_types, ping_data_dict)
```

echopype/convert/parse_base.py
```diff
@@ -69,54 +76,274 @@ def _print_status(self):
...
-                ):  # if no data in a particular channel
-                    self.ping_data_dict[data_type][k] = None
-                else:
-                    # Sort complex and power/angle channels and pad NaN
-                    self.ch_ids[data_type].append(k)
-                    self.ping_data_dict[data_type][k] = self.pad_shorter_ping(v)
-            # Transmit data
-            self.rectangularize_transmit_ping_data(data_type)
+        # Compute the final expansion shapes for each data type
+        expanded_data_shapes = self._get_data_shapes()
```
leewujung, line 164: `self._get_data_shapes()` is also used in `self.__should_use_swap()`, so `expanded_data_shapes` could be calculated first and passed in to `__should_use_swap` as an argument.

echopype/convert/parse_base.py
```diff
@@ -69,54 +76,274 @@ def _print_status(self):
...
+            # for "receive" raw type only
+            if raw_type == "receive":
+                self.ch_ids[data_type].append(ch_id)
+
+            # Pad shorter ping with NaN
+            # do this for each channel
+            # this is the first small expansion
+            padded_arr = self.pad_shorter_ping(arr_list)
+
+            # Multiply power data by conversion factor
```
leewujung, line 274: Adding this for our future selves
```suggestion
            # data_type="angle" does not require extra manipulation, so below
            # only handles power and complex data
            
            # Multiply power data by conversion factor
```

echopype/convert/set_groups_ek60.py
```diff
@@ -656,78 +621,63 @@ def set_beam(self) -> List[xr.Dataset]:
...
-                    },
-                )
+            var_dict["backscatter_r"] = (
+                ["ping_time", "range_sample"],
+                self.parser_obj.ping_data_dict["power"][ch],
+                {
+                    "long_name": self._varattrs["beam_var_default"]["backscatter_r"]["long_name"],
+                    "units": "dB",
+                },
+            )
```
leewujung, line 631: This is a trivial comment: to move the setting of this variable into the above dict with all the other variables

echopype/utils/io.py
```diff
@@ -424,3 +433,74 @@ def validate_source_ds_da(
         check_file_existence(file_path=source_ds_da, storage_options=storage_options)
 
     return source_ds_da, file_type
+
+
+# Utilities for creating temporary swap zarr files -------------------------------------
+def _create_zarr_store_map(path: str, storage_options: dict) -> FSMap:
```
leewujung, line 439: I couldn't find any use of this in the code base, did I miss anything?

lsetiawan: You're right! Thanks for catching that. This was an old function. Just cleaned up.

echopype/echodata/echodata.py
```diff
@@ -67,27 +67,40 @@ def __init__(
...
-            # remove temporary directory created by Parsed2Zarr, if it exists
-            if p2z_temp_file.fs.exists(p2z_temp_file.root):
-                p2z_temp_file.fs.rm(p2z_temp_file.root, recursive=True)
+    def cleanup_swap_files(self):
+        """
+        Clean up the swap files during raw data conversion
+        """
+        sonar_group = "Sonar"
+        beam_group_var = "beam_group"
+        for beam_group in self[sonar_group][beam_group_var].to_numpy():
```
leewujung, line 81: This is just a question: why is `.to_numpy` needed here?

lsetiawan: This is getting the `beam_group` labels such as `Beam_group1`, `Beam_group2`, etc, so the values are needed here.

echopype/convert/parse_base.py
```diff
@@ -69,54 +76,283 @@ def _print_status(self):
...
+            # Get the final data shape
+            data_shape = data_type_shapes[data_type]
+
+            # Auto chunk on first dimension
+            # since this is ping time
+            chunks = ("auto",) + (data_shape[1],)
+            chunks = auto_chunks(
+                chunks=chunks, shape=data_shape, limit=max_chunk_size, dtype=np.dtype("float64")
+            )
+            chunks = chunks + data_shape[2:]
```
leewujung, line 252: Adding some explanation -- is this correct?
```suggestion
            chunks = chunks + data_shape[2:]  # angle data has >2 dimensions
```

echopype/tests/convert/test_convert_api.py
```diff
@@ -0,0 +1,167 @@
...
+            "L0003-D20040909-T161906-EK60.raw",
+            "EK60",
+            False,
+            marks=pytest.mark.xfail(
+                run=False,
+                reason="Expected out of memory error. See https://github.com/OSOceanAcoustics/echopype/issues/489",
+            ),
+        ),
+    ],
+    ids=["noaa_offloaded", "noaa_not_offloaded"],
```
leewujung, line 72: since now the argument is `use_swap`, maybe we change these too:
```suggestion
    ids=["noaa_use_swap_true", "noaa_use_swap_false"],
```

echopype/convert/parse_base.py
leewujung: I think this part does not need to change if complex data is reshaped in the parser and `data_shape[2:]` in that case would contain the right shape, i.e.:
- angle data: (number_of_pings, length_of_range_sample_dim, 2)
- complex data: (number_of_pings, length_of_range_sample_dim, n_sector), where `n_sector` is 3 or 4

leewujung (approved):
Hey @lsetiawan : THANK YOU! This PR looks great! I see that you've resolved most of my comments yesterday. The remaining ones are pretty small. Also awesome to see the use of mocker in the tests -- it sets a couple great examples to add more tests to cover parser and set_groupers in the future. 

My only significant comment is the one related to the reshaping of complex variables. It is a small change, but since this PR is already very large, and touching the parser usually takes extra care, I think it's better to take it in another PR. We have a couple tests that test for the actual parsed values that would be good for catching any errors.

echopype/convert/parse_base.py
lsetiawan: The comment is almost correct, it can also cover complex data like you mentioned. The code snippet above determines the chunk sizes for each dimension:

- L248: Create chunks variable like ("auto", length_of_range_sample_dim)
- L249: Dask `auto_chunks` determine the most optimized chunk size for the "auto" dimension, in this case ping time
- L252: This one would take chunks determined by dask and we append to the rest of the dimension sizes, since angle and complex might have more that 2D, then this would be captured and the resulting chunks might look something like `((100,), (5000,), 2)`. This why we have L253 that grabs the first value of chunks to create something like `(100, 5000, 2)`. 
 
> I think this part does not need to change if complex data is reshaped in the parser and data_shape[2:] in that case would contain the right shape,

Yea, you're right in this comment that no matter what the rest of the dimension is, it will just keep the chunk as the full dimension sizes.

lsetiawan, line 252:
```suggestion
            chunks = chunks + data_shape[2:] # Get the n dimension sizes if > 2D
```

lsetiawan:
@leewujung Thanks for this extensive review! I'm glad that the review went well and that you're satisfied with the changes 😄 

Regarding your last comment:
> My only significant comment is the one related to the reshaping of complex variables.

Would you be able to create a new issue for this so that it can be tracked? Thanks!

lsetiawan:
> Would you be able to create a new issue for this so that it can be tracked? Thanks!

Nvm, just quickly did this. See https://github.com/OSOceanAcoustics/echopype/issues/1213

Issue #1181, opened 2023-09-24
Parse AZFP pressure data when available


emiliom:
AZFP deployments can contain a pressure sensor. However, the echopype AZFP parser currently does not look for and read pressure data, as it does for temperature. Add this.

When read, it should be used to populate `Platform.vertical_offset` while also setting `Platform.water_level` to 0. See https://echopype.readthedocs.io/en/stable/data-proc-additional.html#vertical-coordinate-z-axis-variables

We'll need to obtain a sample data file that includes pressure data.

emiliom:
After parsing the pressure data, the data array needs to be written into the EchoData object.

I think there are two steps here:
1. Store `pressure` in the `Environment` group, via the `set_env` class method. Populate its variable attributes. Base this on what's done for temperature, but modifying attribute values to correspond to pressure. I can provide those attributes here, later.
2. Use `pressure` and a standard seawater equation ("equations of state") to calculate depth (meters). Then set `Platform.vertical_offset` to this depth array, multiplied by -1 (see [here](https://echopype.readthedocs.io/en/stable/data-proc-additional.html#vertical-coordinate-z-axis-variables)). `LoadAZFP.m` has this conversion in the `computeDepth` function. It doesn't look like we have a seawater density function in echopype, but we can incorporate one. The "ideal" package would be https://github.com/TEOS-10/GSW-Python, but unfortunately it depends on a C library. Note that `GSW-Python` already includes a function to go from pressure to depth (see [gsw.conversions.z_from_p](https://teos-10.github.io/GSW-Python/conversions.html)), but it requires at least one parameter that may not be available: latitude. @leewujung should we use simplifications for density (we'll have to use a default salinity anyways) and omit a latitude adjustment?

emiliom:
PR #1189 addressed item 1 above: parsing and storing pressure data, plus tests.

For item 2, we'll need to decide what conversion formula / algorithm / library to use to convert from pressure to depth.

emiliom:
For reference, here's the functionality used by the AZFP Matlab toolbox to calculate density (The links below are to a private repo):

- In `LoadAZFP.m`, [computeDepth function](https://github.com/OSOceanAcoustics/AZFP_matlab/blob/main/matlab_toolbox/LoadAZFP.m#L540):
  ```matlab
  g = 9.815;
  dens = sw_dens(Parameters.Salinity,T,Pressure);
  Depth = Pressure*10000 / (dens * g);
  ```
[...]

PR #1189, opened 2023-10-05
`parse_azfp` now parses AZFP pressure data according to given matlab code

praneethratna:
Addresses #1181 and now `parse_azfp.py` parses AZFP pressure data according to Matlab file shared by Steve.

CC @leewujung

### DIFF
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -197,6 +197,25 @@ class ParseAZFP(ParseBase):
 
         return N * USL5_BAT_CONSTANT
 
+    def _computer_pressure(self, ping_num, is_valid):
+        """
+        Compute pressure in decibar
+
+        Parameters
+        ----------
+        ping_num
+            ping number
+        is_valid
+            whether the associated parameters have valid values
+        """
+        if self.parameters["sensors_flag_pressure_sensor_installed"] == "no" or not is_valid:
+            return np.nan
+
+        counts = self.unpacked_data["ancillary"][ping_num][3]
+        v_in = 2.5 * (counts / 65535)
+        P = v_in * self.parameters["a1"] + self.parameters["a0"] - 10.125
+        return P
+
     def parse_raw(self):
         """
         Parse raw data file from AZFP echosounder.
@@ -216,6 +235,7 @@ class ParseAZFP(ParseBase):
         temperature_is_valid = _test_valid_params(["ka", "kb", "kc"])
         tilt_x_is_valid = _test_valid_params(["X_a", "X_b", "X_c"])
         tilt_y_is_valid = _test_valid_params(["Y_a", "Y_b", "Y_c"])
+        pressure_is_valid = _test_valid_params(["a0", "a1"])
 
         with fmap.fs.open(fmap.root, "rb") as file:
             ping_num = 0
@@ -242,6 +262,10 @@ class ParseAZFP(ParseBase):
                         # Compute y tilt from unpacked_data[ii]['ancillary][1]
                         self.unpacked_data["tilt_y"].append(
                             self._compute_tilt(ping_num, "Y", tilt_y_is_valid)
+                        )
+                        # Compute pressure from unpacked_data[ii]['ancillary'][3]
+                        self.unpacked_data["pressure"].append(
+                            self._computer_pressure(ping_num, pressure_is_valid)
                         )
                         # Compute cos tilt magnitude from tilt x and y values
                         self.unpacked_data["cos_tilt_mag"].append(
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -197,7 +197,7 @@ class ParseAZFP(ParseBase):
 
         return N * USL5_BAT_CONSTANT
 
-    def _computer_pressure(self, ping_num, is_valid):
+    def _compute_pressure(self, ping_num, is_valid):
         """
         Compute pressure in decibar
 
@@ -208,7 +208,7 @@ class ParseAZFP(ParseBase):
         is_valid
             whether the associated parameters have valid values
         """
-        if self.parameters["sensors_flag_pressure_sensor_installed"] == "no" or not is_valid:
+        if not is_valid or self.parameters["sensors_flag_pressure_sensor_installed"] == "no":
             return np.nan
 
         counts = self.unpacked_data["ancillary"][ping_num][3]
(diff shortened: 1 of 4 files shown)


praneethratna:
@emiliom I have added the fix for failing CI tests here. It is caused due to `charset-normalizer` dependency version error.

echopype/convert/parse_azfp.py
```diff
@@ -197,6 +197,25 @@ def _compute_battery(self, ping_num, battery_type):
 
         return N * USL5_BAT_CONSTANT
 
+    def _computer_pressure(self, ping_num, is_valid):
```
emiliom, line 200: Rename the function:
```suggestion
    def _compute_pressure(self, ping_num, is_valid):
```

echopype/convert/parse_azfp.py
```diff
@@ -216,6 +235,7 @@ def _test_valid_params(params):
         temperature_is_valid = _test_valid_params(["ka", "kb", "kc"])
         tilt_x_is_valid = _test_valid_params(["X_a", "X_b", "X_c"])
         tilt_y_is_valid = _test_valid_params(["Y_a", "Y_b", "Y_c"])
+        pressure_is_valid = _test_valid_params(["a0", "a1"])
```
emiliom, line 238: A cosmetic change: move `pressure_is_valid` to be right after `temperature_is_valid`. Temperature and pressure come from external sensors that are sometimes not present (especially pressure). In that sense, I think it's a bit more logical to have those two variables be together.

echopype/convert/parse_azfp.py
```diff
@@ -243,6 +263,10 @@ def _test_valid_params(params):
                         self.unpacked_data["tilt_y"].append(
                             self._compute_tilt(ping_num, "Y", tilt_y_is_valid)
                         )
+                        # Compute pressure from unpacked_data[ii]['ancillary'][3]
```
emiliom, line 266: A cosmetic change: move the `self.unpacked_data["pressure"]` statement to be right after `self.unpacked_data["temperature"]` statement. Temperature and pressure come from external sensors that are sometimes not present (especially pressure). In that sense, I think it's a bit more logical to have those two variables be together. Also, here the pressure statement is in between statements involving tilt, which should remain as a continuous block.

echopype/convert/parse_azfp.py
```diff
@@ -197,6 +197,25 @@ def _compute_battery(self, ping_num, battery_type):
...
+        Compute pressure in decibar
+
+        Parameters
+        ----------
+        ping_num
+            ping number
+        is_valid
+            whether the associated parameters have valid values
+        """
+        if self.parameters["sensors_flag_pressure_sensor_installed"] == "no" or not is_valid:
```
emiliom, line 211: Is the parameter "sensors_flag_pressure_sensor_installed" always present? In case it isn't, I suggest reversing the elements in this test:
```suggestion
        if not is_valid or self.parameters["sensors_flag_pressure_sensor_installed"] == "no":
```
Or maybe the "sensors_flag_pressure_sensor_installed" test isn't needed and may be misleading, since `pressure_is_valid` already carries the information we need: if "a0" or "a1" are not present or valid, the pressure computation should not be carried out.

emiliom:
@praneethratna I've reviewed this PR and it looks great. See my comments about a spelling mistake and small, cosmetic changes.

Let's wait until we get an AZFP data file that does have pressure data, then you can add a test.

The open question then will be whether to add to this PR the parts that create and populate variables in `set_groups_azfp.py`, including `pressure` and `vertical_offset`; or do that in a separate PR. Adding it to this PR is seems fine.

echopype/convert/parse_azfp.py
praneethratna: Resolved.

echopype/convert/parse_azfp.py
praneethratna: Updated!

echopype/convert/parse_azfp.py
praneethratna: Updated! Moved `pressure_is_valid` check right after `temeperature_is_valid`.

echopype/convert/parse_azfp.py
praneethratna: I think the parameter `sensors_flag_pressure_sensor_installed` will always exist. The test will make sure that even if `a0` and `a1` are present but if sensor flag is not installed, then the values are incorrect.

praneethratna:
> The open question then will be whether to add to this PR the parts that create and populate variables in `set_groups_azfp.py`, including `pressure` and `vertical_offset`; or do that in a separate PR. Adding it to this PR is seems fine.

@emiliom I have added the changes for including `pressure` attribute into `Environment` group in this PR. I think the changes for 2nd point that you've mentioned [here](https://github.com/OSOceanAcoustics/echopype/issues/1181#issuecomment-1763217291) can be added in a separate PR.

echopype/convert/parse_azfp.py
emiliom: Thanks. I've looked at a sample XML file that doesn't have pressure data, and you're right, it does have this XML element:
```xml
<SensorsFlag PressureSensorInstalled="no" ParosInstalled="no">0</SensorsFlag>
```

emiliom:
@praneethratna everything looks good! The new test looks good. But I'd like to run this locally before approving the PR. I'll do that later, hopefully today.

I do have a couple of suggestions, though, to improve the tests in `test_convert_azfp.py` that are related to pressure:
- Assuming the data file contains temperature too, let's expand the scope of your new test to do an analogous test for temperature. That should be pretty easy to add. Then rename the test to `def test_convert_azfp_01a_pressure_temperature`, and update the docstring.
- Modify `test_convert_azfp_01a_notemperature_notilt` to test for the pressure variable being present and all-nan, just like `temperature`. Literally, copy and adapt the two lines from the temperature test! The new test for pressure will be run on the same data file being used there. Then rename the test to, say, `test_convert_azfp_01a_no_temperature_pressure_tilt`

praneethratna:
> I do have a couple of suggestions, though, to improve the tests in `test_convert_azfp.py` that are related to pressure:
> 
> * Assuming the data file contains temperature too, let's expand the scope of your new test to do an analogous test for temperature. That should be pretty easy to add. Then rename the test to `def test_convert_azfp_01a_pressure_temperature`, and update the docstring.
> * Modify `test_convert_azfp_01a_notemperature_notilt` to test for the pressure variable being present and all-nan, just like `temperature`. Literally, copy and adapt the two lines from the temperature test! The new test for pressure will be run on the same data file being used there. Then rename the test to, say, `test_convert_azfp_01a_no_temperature_pressure_tilt`

@emiliom The suggestions seems a lot relevant and beneficial for testing the new `pressure` attribute. I pushed the changes for suggestions that you have mentioned in the latest commit!

echopype/tests/convert/test_convert_azfp.py
```diff
@@ -156,7 +156,7 @@ def test_convert_azfp_01a_different_ranges(azfp_path):
     check_platform_required_scalar_vars(echodata)
 
 
-def test_convert_azfp_01a_notemperature_notilt(azfp_path):
+def test_convert_azfp_01a_no_temperature_pressure_tilt(azfp_path):
     """Test converting file with no valid temperature or tilt data."""
```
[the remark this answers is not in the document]
emiliom: I'll go ahead and commit my suggestion

emiliom:
Thanks! Your changes look good. I made an inline commit that just updated a test docstring.

Once the CI completes, assuming they're successful, I'll go ahead and approve and merge the PR.

emiliom (approved):
The CI passed, so it looks good! I'll merge the PR. Thanks!!

PR #1191, opened 2023-10-08
Fixed failing CI tests due to dependency version error [all tests ci]

praneethratna:
Fixed failing CI tests which is caused due to `charset-normalizer` dependency version error for which version should be explicitly set to `3.1.0`. 

Discussions regarding the error and its fix - [Stackoverflow](https://stackoverflow.com/questions/75501048/how-to-fix-attributeerror-partially-initialized-module-charset-normalizer-has), [Github](https://github.com/microsoft/TaskMatrix/issues/242).

CC @emiliom

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ xarray-datatree==0.0.6
 psutil>=5.9.1
 geopy
 flox>=0.7.2,<1.0.0
+charset-normalizer==3.1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,4 +16,4 @@ xarray-datatree==0.0.6
 psutil>=5.9.1
 geopy
 flox>=0.7.2,<1.0.0
-charset-normalizer==3.1.0
+charset-normalizer<3.2


emiliom:
Thanks so much for this PR and for the links! Great detective work. `charset-normalizer` is clearly the source of the problem.

The [current version of `charset-normalizer`](https://github.com/Ousret/charset_normalizer/releases) is 3.3.0. This is the version that was being installed last week. In fact, that version was released on 9-30, so there is a good chance that that version is the one that actually caused the problem.

The discussions you linked to are from early this year, and they recommend 3.1.0 because at the time that was the latest and the problems being reported were based on older versions.

In general, whenever possible, I think it's best to avoid pinning to a specific version. My recommendation is to change your addition to `charset-normalizer<3.3`

requirements.txt
```diff
@@ -16,3 +16,4 @@ xarray-datatree==0.0.6
 psutil>=5.9.1
 geopy
 flox>=0.7.2,<1.0.0
+charset-normalizer==3.1.0
```
emiliom, line 19: See my comment about this.
```suggestion
charset-normalizer<3.3
```

emiliom:
Hmm. `charset-normalizer` [3.2.0 was installed in the CI](https://github.com/OSOceanAcoustics/echopype/actions/runs/6452751746/job/17515199029?pr=1191#step:10:41), as expected. 3.2.0 [was released](https://github.com/Ousret/charset_normalizer/releases/tag/3.2.0) on June 7, and was the dependency version we likely used successfully for a little less than 4 months.

But now the CI is failing again, with the same error. It would seem `charset-normalizer` 3.1.0 does solve the problem (though there were some hiccups in the CI), but 3.2.0 doesn't. That doesn't make sense to me, given our recent experience. I would hold off on merging this PR until we have a better handle of this.

BTW, there's a [pip installation error in the CI log](https://github.com/OSOceanAcoustics/echopype/actions/runs/6452751746/job/17515199029?pr=1191#step:10:136) involving the `botocore` version that was installed -- but only for Py3.11!

praneethratna:
> But now the CI is failing again, with the same error. It would seem `charset-normalizer` 3.1.0 does solve the problem (though there were some hiccups in the CI), but 3.2.0 doesn't. That doesn't make sense to me, given our recent experience. I would hold off on merging this PR until we have a better handle of this.

I think some of dependencies require `cchartdet` and some need `charset-normalizer` which is causing circular imports in `charset-normalizer` >3.1 versions and that might be due to an update to dependency that requires `cchardet`.
> BTW, there's a [pip installation error in the CI log](https://github.com/OSOceanAcoustics/echopype/actions/runs/6452751746/job/17515199029?pr=1191#step:10:136) involving the `botocore` version that was installed -- but only for Py3.11!

I think this might be a one time issue since when i re ran the tests, that error did not occur again.

lsetiawan:
Hi @praneethratna! Thank you so much for this very detailed investigation. This is such great detective work. After looking through the CI, I've narrowed down the problem to version mismatch during the CI setup. Currently this system is quite convoluted as we're installing pip packages and conda packages causing mismatch on dependencies and the `charset-normalizer` being installed twice with the conda package being `3.3.0` and the pip package is `3.2.0`. Your pinning above did work, which was really great. However, this would introduce a pinned dependency to an upstream package that we don't directly use. Having this pinning could potentially cause other dependency resolving issues in the future, so instead I've removed the conda piece in PR https://github.com/OSOceanAcoustics/echopype/pull/1192. This has been something in the back burner and looks like this is time to make those changes. Now that the package is using a single package manager, all looks good and I'm closing this PR. Thanks again!

PR #1192, opened 2023-10-09
ci: Update CI to barebone python [all tests ci]

lsetiawan:
This PR is a follow up to https://github.com/OSOceanAcoustics/echopype/pull/1191 since even new PR's are failing!

Conda is now removed from CI since there isn't any special packages that we need from there anymore and we're basically already installing via pip through conda anyways. Plus this was a recommendation from Filipe to do over the summer.

### DIFF
--- a/echopype/tests/convert/test_parsed_to_zarr.py
+++ b/echopype/tests/convert/test_parsed_to_zarr.py
@@ -15,6 +15,8 @@ from echopype.convert.parsed_to_zarr_ek60 import Parsed2ZarrEK60
 from echopype.echodata.convention import sonarnetcdf_1
 from echopype.convert.api import _check_file, SONAR_MODELS
 
+pytestmark = pytest.mark.skip(reason="Removed Parsed2Zarr")
+
 test_bucket_name = "echopype-test"
 port = 5555
 endpoint_uri = "http://127.0.0.1:%s/" % port
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -72,6 +72,8 @@ jobs:
           python -m pip install -e .[plot]
       - name: Print installed packages
         run: |
+          which pip
+          pip list
           micromamba list -n ${{ env.CONDA_ENV }}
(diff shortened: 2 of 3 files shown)


echopype/tests/convert/test_parsed_to_zarr.py
```diff
@@ -15,6 +15,8 @@
 from echopype.echodata.convention import sonarnetcdf_1
 from echopype.convert.api import _check_file, SONAR_MODELS
 
+pytestmark = pytest.mark.skip(reason="Removed Parsed2Zarr")
```
emiliom, line 18: Is this relevant to this PR?

emiliom:
Looks good, thanks! I made one small inline comment. My only other comment is that it looks like the files `.ci_helpers/py${{ matrix.python-version }}.yaml` are no longer used; or are they actually used in other actions? If not, let's delete them, right?

echopype/tests/convert/test_parsed_to_zarr.py
lsetiawan: > Is this relevant to this PR?

Yepp, without the skip, it'll have a botocore error because it's spinning up moto server multiple times... I have to figure that out in PR https://github.com/OSOceanAcoustics/echopype/pull/1185. For now, it's best to skip this test since lots of changes a coming!

lsetiawan:
> re they actually used in other actions?

Looks like they're still being used in the `windows-utils-test` shenanigans that I'm not sure what it's doing. Probably good to have a tests clean up soon that include this stuff.

echopype/tests/convert/test_parsed_to_zarr.py
emiliom: Got it

emiliom:
> Looks like they're still being used in the windows-utils-test shenanigans that I'm not sure what it's doing. Probably good to have a tests clean up soon that include this stuff.

Sounds good. Merge away, and thanks!!

emiliom (approved):

PR #1193, opened 2023-10-11
Add qc to top level init

ctuguinay:
This PR addresses the fact that `qc` doesn't exist in init and `qc` functions cannot be imported. Super small fix.


ctuguinay:
Closing this since I'm not seemingly able to import qc functions without this addition.

PR #1198, opened 2023-10-23
Optimize Frequency Differencing with Dask

### LINKED ISSUE #1165: Check to see if this masking `xr.where` parallelize
This [`xr.where`](https://docs.xarray.dev/en/stable/generated/xarray.where.html) needs to be investigated to see if it can parallelize with the use of dask.

https://github.com/OSOceanAcoustics/echopype/blob/3ca028f04a4fd3643bb0ad350dc09a699883ee1e/echopype/mask/api.py#L500

anantmittal:
Fixes #1165

### DIFF
--- a/echopype/mask/api.py
+++ b/echopype/mask/api.py
@@ -523,8 +523,8 @@ def frequency_differencing(
 
         da_list = []
 
-        # Tranform Sv blocks into a contiguous flattened array and
-        # iterate over them to apply the transformation
+        # Transform Sv blocks into a contiguous flattened array
+        # and iterate over them to apply the transformation
         for Sv_block in source_Sv["Sv"].data.blocks.ravel():
             # Apply and delay the _get_lhs and _create_mask function on Sv_block
             lhs = dask.delayed(_get_lhs)(Sv_block, chanA_idx, chanB_idx)
--- a/echopype/mask/api.py
+++ b/echopype/mask/api.py
@@ -500,7 +500,7 @@ def frequency_differencing(
         freqB_pos = np.argwhere(source_Sv.channel.values == chanAB[1]).flatten()[0]
 
     # If Sv data is not dask array
-    if source_Sv["Sv"].chunks == None:
+    if source_Sv["Sv"].chunks is None:
         # get the left-hand side of condition
         lhs = source_Sv["Sv"].sel(channel=chanA) - source_Sv["Sv"].sel(channel=chanB)
 
--- a/echopype/mask/api.py
+++ b/echopype/mask/api.py
@@ -5,6 +5,9 @@ from typing import List, Optional, Union
 
 import numpy as np
 import xarray as xr
+import dask
+import dask.array
+
 
 from ..utils.io import validate_source_ds_da
 from ..utils.prov import add_processing_level, echopype_prov_attrs, insert_input_processing_level
@@ -451,8 +454,8 @@ def frequency_differencing(
     >>> Sv_ds = xr.Dataset(data_vars={"Sv": Sv_da, "frequency_nominal": freq_nom})
     ...
     >>> # compute frequency-differencing mask using channel names
-    >>> echopype.mask.frequency_differencing(source_Sv=mock_Sv_ds, storage_options={},
-    ...                                      freqABEq=None, chanABEq = '"chan1" - "chan2">=10.0')
+    >>> echopype.mask.frequency_differencing(source_Sv=Sv_ds, storage_options={},
+    ...                                      freqABEq=None, chanABEq = '"chan1" - "chan2">=10.0dB')
     <xarray.DataArray 'mask' (ping_time: 5, range_sample: 5)>
     array([[False, False, False, False, False],
            [False, False, False, False, False],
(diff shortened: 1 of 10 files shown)


echopype/mask/api.py
```diff
@@ -493,23 +495,72 @@ def frequency_differencing(
...
 
-    # get the left-hand side of condition
-    lhs = source_Sv["Sv"].sel(channel=chanA) - source_Sv["Sv"].sel(channel=chanB)
+        freqA_pos = np.argwhere(source_Sv.channel.values == chanAB[0]).flatten()[0]
+        freqB_pos = np.argwhere(source_Sv.channel.values == chanAB[1]).flatten()[0]
 
-    # create mask using operator lookup table
-    da = xr.where(str2ops[operator](lhs, diff), True, False)
+    # If Sv data is not dask array
+    if source_Sv["Sv"].chunks is None:
```
lsetiawan, line 502:
```suggestion
    if isinstance(source_Sv["Sv"].variable._data, dask.array.Array):
```

echopype/mask/api.py
```diff
@@ -493,23 +495,72 @@ def frequency_differencing(
...
+            coords={
+                "ping_time": source_Sv["Sv"].coords["ping_time"].values,
+                "range_sample": source_Sv["Sv"].coords["range_sample"].values,
+            },
+            attrs={
+                "mask_type": "frequency differencing",
+                "history": f"{datetime.datetime.utcnow()} +00:00. "
+                "Mask created by mask.frequency_differencing. "
+                f"Operation: Sv['{chanA}'] - Sv['{chanB}'] {operator} {diff}",
+            },
```
lsetiawan, line 565: This is similar to assign attrs above

https://github.com/OSOceanAcoustics/echopype/pull/1198/files#diff-36c0af5a3cf3a0fa59c1880a35e59e23e136740d1b963ffda24d8c714989879dR512-R520

I suggest extracting this out from the `if-else`

anantmittal: I extracted that out. 

I think `datetime.datetime.utcnow()` might be different than what was being computed earlier because that value will be calculated before the masking operations since it's outside of `if-else.`

echopype/mask/api.py
```diff
@@ -493,23 +495,67 @@ def frequency_differencing(
...
 
-    da = da.assign_attrs({**mask_attrs, **{"history": history_attr}})
+        # assign provenance attributes
+        da = da.assign_attrs(xr_dataarray_attrs)
 
-    return da
+        return da
+    # If Sv data is dask array
+    else:
+        num_column_blocks = int(np.ceil(source_Sv["Sv"].shape[2] / source_Sv["Sv"].chunks[2][0]))
```
lsetiawan, line 525: This grabbing of shape is very particular and assumes that the right dimension is located at `2`. What dimension name are you trying to grab here? `range_sample`?

echopype/mask/api.py
```diff
@@ -493,23 +506,34 @@ def frequency_differencing(
...
-    da = xr.where(str2ops[operator](lhs, diff), True, False)
+    # If Sv data is not dask array
+    if not isinstance(source_Sv["Sv"].variable._data, dask.array.Array):
+        da = _frequency_diff_and_mask(source_Sv["Sv"], chanA, chanB, diff)
+    # If Sv data is dask array
+    else:
+        # Get the final data array template
+        template = source_Sv["Sv"].isel(channel=0).drop_vars("channel")
+        # Iterate over all the chunks
+        da = source_Sv["Sv"].map_blocks(
```
lsetiawan, line 517: @anantmittal I've updated your code to use the built-in xarray's `map_blocks` to map over the blocks, which will return the template data array. This is exactly the same way like you had, but now we don't need to convert to dask array, then convert back 😄 I just realized that this function exists!

lsetiawan:
@anantmittal Could you please fix the pre-commit failure. Thanks!

lsetiawan (approved):

lsetiawan:
@leewujung I know this is quite different from what you were looking for in https://github.com/OSOceanAcoustics/echopype/issues/1165#issuecomment-1776360813. I've created a separate issue for that in https://github.com/OSOceanAcoustics/echopype/issues/1200. Would be able to review this and see for this specific case that it's all good? Thanks!

leewujung (commented):
@anantmittal @lsetiawan : Thanks for the PR! Everything looks great, and my main comments are regarding [the resolved comment](https://github.com/OSOceanAcoustics/echopype/pull/1198#discussion_r1372030224) from @lsetiawan.
- rather than assuming the 3rd dimension is `range_sample`, if the intention is to grab `range_sample` dimension, what about actually finding its index and use it to be error-proof?
- in terms of traversing the blocks, for larger combined datasets, usually the `ping_time` dimension will be the largest. I wonder if it is worthwhile to do this blocking along the `ping_time` dimension, especially if we're looking to change to use `map_block` in the future?

lsetiawan:
> rather than assuming the 3rd dimension is range_sample, if the intention is to grab range_sample dimension, what about actually finding its index and use it to be error-proof?
    in terms of traversing the blocks, for larger combined datasets, usually the ping_time dimension will be the largest. I wonder if it is worthwhile to do this blocking along the ping_time dimension, especially if we're looking to change to use map_block in the future?


@leewujung I think I've addressed your concerns now, please let me know what you think. Thanks! 😄

lsetiawan:
@leewujung is this good to go?

leewujung:
@lsetiawan : Looks good. Thanks for the changes!

PR #1202, opened 2023-10-27
AZFP `platform` group will now include `vertical_offset` data by calculating `depth`

praneethratna:
Addresses 2nd step of https://github.com/OSOceanAcoustics/echopype/issues/1181#issuecomment-1763217291, by including calculations for `depth` value mentioned in https://github.com/OSOceanAcoustics/echopype/issues/1181#issuecomment-1781696782 which is used to set the `vertical_offset` variable. 

CC @emiliom

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -7,6 +7,7 @@ import numpy as np
 import xarray as xr
 
 from ..utils.coding import set_time_encodings
+from ..utils.misc import depth_from_pressure
 from .set_groups_base import SetGroupsBase
 
 
@@ -167,6 +168,7 @@ class SetGroupsAZFP(SetGroupsBase):
         """Set the Platform group."""
         platform_dict = {"platform_name": "", "platform_type": "", "platform_code_ICES": ""}
         unpacked_data = self.parser_obj.unpacked_data
+        depth = depth_from_pressure(unpacked_data["pressure"], latitude=30, atm_pres_surf=0)
 
         # Create nan time coordinate for lat/lon (lat/lon do not exist in AZFP 01A data)
         time1 = [np.nan]
@@ -177,7 +179,9 @@ class SetGroupsAZFP(SetGroupsBase):
         tilt_y = [np.nan] if np.isnan(unpacked_data["tilt_y"]).all() else unpacked_data["tilt_y"]
         if (len(tilt_x) == 1 and np.isnan(tilt_x)) and (len(tilt_y) == 1 and np.isnan(tilt_y)):
             time2 = [self.parser_obj.ping_time[0]]
+            depth = [depth[0]]
         else:
+            depth = -1 * depth
             time2 = self.parser_obj.ping_time
 
         # Handle potential nan timestamp for time1 and time2
@@ -208,7 +212,7 @@ class SetGroupsAZFP(SetGroupsBase):
                 ),
                 "vertical_offset": (
                     ["time2"],
-                    [np.nan] * len(time2),
+                    depth,
                     self._varattrs["platform_var_default"]["vertical_offset"],
                 ),
                 "water_level": (
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -170,13 +170,13 @@ class SetGroupsAZFP(SetGroupsBase):
         unpacked_data = self.parser_obj.unpacked_data
         time2 = self.parser_obj.ping_time
         if np.isnan(unpacked_data["pressure"]).all():
-            depth = [np.nan] * len(time2)
-            water_level = [np.nan]
+            vertical_offset = [np.nan] * len(time2)
+            water_level = np.nan
         else:
-            depth = -1 * depth_from_pressure(
+            vertical_offset = -1 * depth_from_pressure(
                 unpacked_data["pressure"], latitude=30, atm_pres_surf=0
             )
-            water_level = [0.0]
+            water_level = 0.0
 
         # Create nan time coordinate for lat/lon (lat/lon do not exist in AZFP 01A data)
         time1 = [np.nan]
@@ -212,11 +212,11 @@ class SetGroupsAZFP(SetGroupsBase):
                 ),
                 "vertical_offset": (
                     ["time2"],
-                    depth,
+                    vertical_offset,
                     self._varattrs["platform_var_default"]["vertical_offset"],
                 ),
                 "water_level": (
-                    ["time1"],
+                    [],
                     water_level,
                     self._varattrs["platform_var_default"]["water_level"],
                 ),
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -183,6 +183,11 @@ class SetGroupsAZFP(SetGroupsBase):
 
         tilt_x = unpacked_data["tilt_x"]
         tilt_y = unpacked_data["tilt_y"]
+        if np.isnan(tilt_x).all() and np.isnan(tilt_y).all() and np.isnan(vertical_offset).all():
+            tilt_x = [np.nan]
+            tilt_y = [np.nan]
+            vertical_offset = [np.nan]
+            time2 = [self.parser_obj.ping_time[0]]
 
         # Handle potential nan timestamp for time1 and time2
         time1 = self._nan_timestamp_handler(time1)
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -168,21 +168,21 @@ class SetGroupsAZFP(SetGroupsBase):
         """Set the Platform group."""
         platform_dict = {"platform_name": "", "platform_type": "", "platform_code_ICES": ""}
         unpacked_data = self.parser_obj.unpacked_data
-        depth = depth_from_pressure(unpacked_data["pressure"], latitude=30, atm_pres_surf=0)
+        time2 = self.parser_obj.ping_time
+        if np.isnan(unpacked_data["pressure"]).all():
+            depth = [np.nan] * len(time2)
+            water_level = [np.nan]
+        else:
+            depth = -1 * depth_from_pressure(
+                unpacked_data["pressure"], latitude=30, atm_pres_surf=0
+            )
+            water_level = [0.0]
 
         # Create nan time coordinate for lat/lon (lat/lon do not exist in AZFP 01A data)
         time1 = [np.nan]
 
-        # If tilt_x and/or tilt_y are all nan, create single-value time2 dimension
-        # and single-value (np.nan) tilt_x and tilt_y
-        tilt_x = [np.nan] if np.isnan(unpacked_data["tilt_x"]).all() else unpacked_data["tilt_x"]
-        tilt_y = [np.nan] if np.isnan(unpacked_data["tilt_y"]).all() else unpacked_data["tilt_y"]
-        if (len(tilt_x) == 1 and np.isnan(tilt_x)) and (len(tilt_y) == 1 and np.isnan(tilt_y)):
-            time2 = [self.parser_obj.ping_time[0]]
-            depth = [depth[0]]
-        else:
-            depth = -1 * depth
-            time2 = self.parser_obj.ping_time
+        tilt_x = unpacked_data["tilt_x"]
+        tilt_y = unpacked_data["tilt_y"]
 
         # Handle potential nan timestamp for time1 and time2
         time1 = self._nan_timestamp_handler(time1)
@@ -216,8 +216,8 @@ class SetGroupsAZFP(SetGroupsBase):
                     self._varattrs["platform_var_default"]["vertical_offset"],
                 ),
                 "water_level": (
-                    [],
-                    np.nan,
+                    ["time1"],
+                    water_level,
                     self._varattrs["platform_var_default"]["water_level"],
                 ),
                 "tilt_x": (
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -213,7 +213,7 @@ class ParseAZFP(ParseBase):
 
         counts = self.unpacked_data["ancillary"][ping_num][3]
         v_in = 2.5 * (counts / 65535)
-        P = v_in * self.parameters["a1"] + self.parameters["a0"] - 10.125
+        P = v_in * self.parameters["a1"] + self.parameters["a0"] - 10.1325
         return P
 
     def parse_raw(self):
--- a/echopype/tests/convert/test_convert_azfp.py
+++ b/echopype/tests/convert/test_convert_azfp.py
@@ -156,8 +156,8 @@ def test_convert_azfp_01a_different_ranges(azfp_path):
     check_platform_required_scalar_vars(echodata)
 
 
-def test_convert_azfp_01a_no_temperature_pressure_tilt(azfp_path):
-    """Test converting file with no valid temperature, pressure and tilt data."""
+def test_convert_azfp_01a_no_temperature_pressure_tilt_verticaloffset(azfp_path):
+    """Test converting file with no valid temperature, pressure, tilt and vertical_offset data."""
     azfp_01a_path = azfp_path / 'rutgers_glider_notemperature/22052500.01A'
     azfp_xml_path = azfp_path / 'rutgers_glider_notemperature/22052501.XML'
 
@@ -173,6 +173,10 @@ def test_convert_azfp_01a_no_temperature_pressure_tilt(azfp_path):
     assert "pressure" in echodata["Environment"]
     assert echodata["Environment"]["pressure"].isnull().all()
 
+    # Vertical offset variable is present in the Platform group and its values are all nan
+    assert "vertical_offset" in echodata["Platform"]
+    assert echodata["Platform"]["vertical_offset"].isnull().all()
+
     # Tilt variables are present in the Platform group and their values are all nan
     assert "tilt_x" in echodata["Platform"]
     assert "tilt_y" in echodata["Platform"]
@@ -180,8 +184,8 @@ def test_convert_azfp_01a_no_temperature_pressure_tilt(azfp_path):
     assert echodata["Platform"]["tilt_y"].isnull().all()
 
 
-def test_convert_azfp_01a_pressure_temperature(azfp_path):
-    """Test converting file with valid pressure and temperature data."""
+def test_convert_azfp_01a_pressure_temperature_verticaloffset(azfp_path):
+    """Test converting file with valid pressure, temperature and vertical_offset data."""
     azfp_01a_path = azfp_path / 'pressure' / '22042221.01A'
     azfp_xml_path = azfp_path / 'pressure' / '22042220.XML'
 
@@ -196,6 +200,10 @@ def test_convert_azfp_01a_pressure_temperature(azfp_path):
     # Temperature variable is present in the Environment group and its values are not all nan
     assert "temperature" in echodata["Environment"]
     assert not echodata["Environment"]["temperature"].isnull().all()
+
+    # Vertical offset variable is present in the Platform group and its values are not all nan
+    assert "vertical_offset" in echodata["Platform"]
+    assert not echodata["Platform"]["vertical_offset"].isnull().all()
 
 
 def test_load_parse_azfp_xml(azfp_path):


echopype/utils/misc.py
```diff
@@ -11,3 +14,43 @@ def camelcase2snakecase(camel_case_str):
...
+    # Calculate depth
+    x = np.sin(np.deg2rad(latitude))
+    pressure = np.subtract(pressure, atm_pres_surf)
+    depth_w_g = (
+        -k1 * np.power(pressure, 4)
+        + k2 * np.power(pressure, 3)
+        - k3 * np.power(pressure, 2)
+        + k4 * pressure
+    )
+    gravity = g * (1.0 + k5 * np.power(x, 2) + k6 * np.power(x, 4)) + k7 * pressure
```
emiliom, line 54: A few style suggestions, for readability:
- You can use the `-` operator with numpy arrays and scalars, instead of `np.subtract`
- The `**` operator works equally well as `np.power`, but is easier to read
- Moved `x` declaration to just before `gravity`, where it's first used

```suggestion
    pressure = pressure - atm_pres_surf
    depth_w_g = -k1 * pressure**4 + k2 * pressure**3 - k3 * pressure**2 + k4 * pressure
    x = np.sin(np.deg2rad(latitude))
    gravity = g * (1.0 + k5 * x**2 + k6 * x**4) + k7 * pressure
```

echopype/utils/misc.py
```diff
@@ -11,3 +14,43 @@ def camelcase2snakecase(camel_case_str):
...
+    latitude : float
+        Latitude in degrees
+    atm_pres_surf : float
+        Atmospheric pressure at the surface in dbar
+    Returns
+    -------
+    depth : array_like
+        Depth in meters
+    """
+    # Constants
```
emiliom, line 35: For consistency with UNESCO 1983, please use the names for constants used there. Your k1 - k4 should be mapped to c4 to c1. For your k5 - k7, UNESCO 1983 doesn't define constant names. How about k1 - k4?

emiliom: I also suggest using the signs for constants c1 - c4 from UNESCO. Then, the terms in the `depth_w_g` will all be additions.

emiliom: FYI I double checked the values of all those constants, and they're correct. The signs for some of them will change -- see my comment, above.

echopype/utils/misc.py
```diff
@@ -11,3 +14,43 @@ def camelcase2snakecase(camel_case_str):
...
+
+def depth_from_pressure(pressure, latitude=30, atm_pres_surf=0):
+    """
+    Convert pressure to depth using UNESCO equation.
+    Parameters
+    ----------
+    pressure : array_like
+        Pressure in dbar
+    latitude : float
+        Latitude in degrees
```
emiliom, line 27:
```suggestion
        Latitude in decimal degrees
```

echopype/utils/misc.py
```diff
@@ -11,3 +14,43 @@ def camelcase2snakecase(camel_case_str):
...
+    atm_pres_surf : float
+        Atmospheric pressure at the surface in dbar
+    Returns
```
emiliom, line 30:
```suggestion
    Convert pressure to depth using UNESCO 1983 algorithm.
 
    UNESCO. 1983. Algorithms for computation of fundamental properties of seawater (Pressure to Depth conversion, pages 25-27). Prepared by Fofonoff, N.P. and Millard, R.C. UNESCO technical papers in marine science, 44. http://unesdoc.unesco.org/images/0005/000598/059832eb.pdf
    
    Parameters
    ----------
    pressure : array_like
        Pressure in dbar
    latitude : float
        Latitude in decimal degrees
    atm_pres_surf : float
        Atmospheric pressure at the surface in dbar
    
    Returns
```

echopype/convert/set_groups_azfp.py
```diff
@@ -167,6 +168,7 @@ def set_platform(self) -> xr.Dataset:
         """Set the Platform group."""
         platform_dict = {"platform_name": "", "platform_type": "", "platform_code_ICES": ""}
         unpacked_data = self.parser_obj.unpacked_data
+        depth = depth_from_pressure(unpacked_data["pressure"], latitude=30, atm_pres_surf=0)
```
emiliom, line 171: pressure will rarely be available in AZFP data. So, here, there needs to be a handler to check if `"pressure" in unpacked_data` (I can't remember, but I assume that if there was no pressure data present, `unpacked_data` doesn't contain a pressure entry?). If pressure is not present, set `depth = [np.nan] * len(time2)`, as was done before. Except `time2` hasn't been declared here yet, so you should move this `depth` block to just before the `ds = xr.Dataset` statement.

echopype/convert/set_groups_azfp.py
```diff
@@ -208,7 +212,7 @@ def set_platform(self) -> xr.Dataset:
                 ),
                 "vertical_offset": (
                     ["time2"],
-                    [np.nan] * len(time2),
+                    depth,
                     self._varattrs["platform_var_default"]["vertical_offset"],
                 ),
                 "water_level": (
```
emiliom, line 218: `water_level` should be set to 0.0 (a scalar value) if pressure is present. If pressure is not present, it stays as it is now, `np.nan`. See https://github.com/OSOceanAcoustics/echopype/issues/1181#issue-1910068233

echopype/convert/set_groups_azfp.py
```diff
@@ -177,7 +179,9 @@ def set_platform(self) -> xr.Dataset:
         tilt_y = [np.nan] if np.isnan(unpacked_data["tilt_y"]).all() else unpacked_data["tilt_y"]
         if (len(tilt_x) == 1 and np.isnan(tilt_x)) and (len(tilt_y) == 1 and np.isnan(tilt_y)):
             time2 = [self.parser_obj.ping_time[0]]
+            depth = [depth[0]]
         else:
+            depth = -1 * depth
             time2 = self.parser_obj.ping_time
```
emiliom, line 185: Hmm, I'm just realizing that `pressure` may have its own time dimension that's different from `time2`! Let me think about this ...

emiliom: I also need to think about the `-1` factor, and our statement in the echopype documentation that `vertical_offset = -pressure_sensor_depth` ....

One thing we need to account for is whether the echosounder is pointing up or down. I need to think about that, too.

echopype/utils/misc.py
praneethratna: Updated!

echopype/convert/set_groups_azfp.py
praneethratna: `pressure` data will always be available in the parsed data and if in case the required parameters for calculating `pressure` aren't available, then the value be set to `np.nan`.So, I don't think there will be a case with `pressure` data not being present.

echopype/convert/set_groups_azfp.py
praneethratna: Should this be if `pressure` value is equal to `np.nan` then `water_level` should also be `np.nan`? Since, if  `pressure` value cannot be calculated due to required parameters not being present in the XML file then it would be set to `np.nan`.

echopype/convert/set_groups_azfp.py
praneethratna: > Hmm, I'm just realizing that `pressure` may have its own time dimension that's different from `time2`! Let me think about this ...

The code i just wrote is a work around since when both `tilt_x` and `tilt_y` are all `np.nan` then we set `time2` dimension as 1  but `pressure` data might be present when they `np.nan`.I think `pressure` might have it's own dimension?

echopype/utils/misc.py
praneethratna: Updated the doctsring!

echopype/utils/misc.py
praneethratna: Updated, removed the usage of `np.power` and `np.subtract` and also moved `x` declaration just before `gravity` calculation!

echopype/convert/set_groups_azfp.py
emiliom: > Should this be if `pressure` value is equal to `np.nan` then `water_level` should also be `np.nan`? 

Correct. Then, if pressure values are not all nan (there are valid values), `water_level` should be set to 0.0, a scalar. `water_level` is a variable that is separate but related to `vertical_offset` (see [here](https://echopype.readthedocs.io/en/stable/data-proc-additional.html#vertical-coordinate-z-axis-variables)). For AZFP, we're allowing `water_level` to only be either `np.nan` or `0.0`.

echopype/utils/misc.py
```diff
@@ -11,3 +14,44 @@ def camelcase2snakecase(camel_case_str):
...
+    c2 = -2.2512e-5
+    c3 = 2.279e-10
+    c4 = -1.82e-15
+    k1 = 5.2788e-3
+    k2 = 2.36e-5
+    k3 = 1.092e-6
+
+    # Calculate depth
+    pressure = pressure - atm_pres_surf
+    depth_w_g = +c4 * pressure**4 + c3 * pressure**3 + c2 * pressure**2 + c1 * pressure
```
emiliom, line 53: Remove unnecessary `+` sign, and follow the order of terms used in UNESCO 1983:
```suggestion
    depth_w_g = c1 * pressure + c2 * pressure**2 + c3 * pressure**3 + c4 * pressure**4
```

echopype/convert/set_groups_azfp.py
emiliom: Ok. I think I was unclear, but I understand what you mean. `unpacked_data` will always have a `pressure` entry, but it may be set to `np.nan`.

In that case, though, it seems like an unnecessary step to apply the function `depth_from_pressure` if we know in advance that pressure is np.nan. So, how about something like:
```suggestion
        if unpacked_data["pressure"] == np.nan:
            depth = np.nan
        else:
            depth = depth_from_pressure(unpacked_data["pressure"], latitude=30, atm_pres_surf=0)
```

Also, as a side note, I need to ask others about pressure measurements and whether it's typical that `atm_pres_surf` (atmospheric pressure) is already removed.

echopype/convert/set_groups_azfp.py
emiliom: Dealing with the time dimensions here may be a bit tricky. Yes, `pressure` may have its own time dimension, but I'll need to ask about this.

praneethratna:
@emiliom I have raised a separate PR #1207 for addition of `depth_from_pressure` method and along with tests for that method as suggested by you. Once that is merged, I'll re-work on the commits here to include the changes only for setting `vertical_offset` value in this PR. Thanks!

echopype/utils/misc.py
praneethratna: Updated in PR #1207!

emiliom:
There are some open questions that I've raised, some of which were on me to follow up on. I'm compiling and summarizing all of them here, with some updates:

1. Is atmospheric pressure (and potentially other overlying pressure effects) typically already removed from pressure measurements?
    - It's not unusual to remove atmospheric pressure in the sensor calibration, so that pressure = 0 represents the surface. We've accounted for this in PR #1207 in the eneric `depth_from_pressure` function by setting the default value of the `atm_pres_surf` to zero.
    - Now the question is whether an **AZFP pressure sensor** is hard wired to always remove atmospheric pressure, to never remove it, or if it's a user configuration. That will determine whether to pass `atm_pres_surf` a value of 0 or 10.1325 (dbar) in the call to `depth_from_pressure`, or whether it's undetermined.
2. Assigning a value to `water_level`
    - For AZFP, we're allowing `water_level` to only be either `np.nan` or `0.0`.
    - If pressure values are not all `np.nan` (there are valid values), `water_level` should be set to 0.0. Otherwise, if pressure is all `np.nan`,  `water_level` will be `np.nan`
3. Based on the definitions in the [echopype documentation](https://echopype.readthedocs.io/en/stable/data-proc-additional.html#vertical-coordinate-z-axis-variables), the platform coordinate system origin corresponding to the position of the pressure sensor, and `water_level` being set to zero, the assignment `vertical_offset = -pressure_sensor_depth` is correct, and `vertical_offset` will be negative.
4. The echosounder pointing direction (up or down) doesn't impact the calculations and assumptions used here for setting `vertical_offset` and `water_level`.
5. Time dimension for the pressure variable, and therefore for the assigned `vertical_offset` variable
    - If `tilt_x` and `tilt_y` have valid values, they will already be using a `time2` dimension. Do we know if `pressure` actually shares the same clock, so we can reuse that dimension?
    - If `tilt_x` and `tilt_y` are all `np.nan`, they're pre-assigned with length-1 `time2` dimension. `pitch` and `roll` will also use this dummy, "place holder" dimension. Should pressure be assigned a new dimension, `time3`? Or should its dimension be called `time2`, then `tilt_x`, `tilt_y`, `pitch` and `roll` will become 
  arrays of `np.nan` with length >1?
6. Test for `unpacked_data["pressure"] == np.nan`. If `True`, skip the the call to `depth_from_pressure`.

emiliom:
Regarding point 5 above: I've confirmed in `parse_azfp.py` (see the `parse_raw` method) that `pressure`, when present, will be on the same "clock" as `tilt_x` and `tilt_y`. So, when present and used to generate `vertical_offset`, `vertical_offset` can reuse the `time2` time dimension used by `tilt_x` and `tilt_y`. If `tilt_x` and `tilt_y` have valid data, their length and time values will be identical to those of `vertical_offset`. If `tilt_x` and `tilt_y` are all nan and set to be of length 1, I guess we'll just extend them to the full time dimension of `vertical_offset`.

echopype/convert/set_groups_azfp.py
praneethratna: Updated!

echopype/convert/set_groups_azfp.py
praneethratna: Updated!

praneethratna:
@emiliom I have pushed the changes which addresses most of the points mentioned except one where we are still not sure whether to pass `atm_surf_pres` a value of 0 or 10.1325 (dbar).

emiliom:
> I have pushed the changes which addresses most of the points mentioned 

Thanks! I'll look over your commits today.

> except one where we are still not sure whether to pass atm_surf_pres a value of 0 or 10.1325 (dbar).

We haven't heard back from Steve on that question (I believe you were cc'd on that email). I'll follow up again today.

Also, I just realized that I forgot to mention here a related PR I submitted, #1226. If approved, the main impact on this PR is that the `pressure` variable will be created in `set_groups_azfp.py` only when the AZFP data file actually contains pressure data. Right now it's always created and it's filled with np.nan when the data file doesn't have pressure data.

Finally, @leewujung made a comment related to this PR (https://github.com/OSOceanAcoustics/echopype/pull/1222#discussion_r1394848585):

> I am a bit hesitate regarding converting pressure to depth directly, because of the uncertainty in latitude. I'll comment on it in https://github.com/OSOceanAcoustics/echopype/pull/1202

echopype/convert/set_groups_azfp.py
```diff
@@ -208,12 +212,12 @@ def set_platform(self) -> xr.Dataset:
...
                 "vertical_offset": (
                     ["time2"],
-                    [np.nan] * len(time2),
+                    depth,
                     self._varattrs["platform_var_default"]["vertical_offset"],
                 ),
                 "water_level": (
-                    [],
-                    np.nan,
+                    ["time1"],
```
emiliom, line 219: water_level is a scalar variable. It has no dimension.
```suggestion
                    [],
```

echopype/convert/set_groups_azfp.py
```diff
@@ -167,18 +168,21 @@ def set_platform(self) -> xr.Dataset:
         """Set the Platform group."""
         platform_dict = {"platform_name": "", "platform_type": "", "platform_code_ICES": ""}
         unpacked_data = self.parser_obj.unpacked_data
+        time2 = self.parser_obj.ping_time
+        if np.isnan(unpacked_data["pressure"]).all():
+            depth = [np.nan] * len(time2)
+            water_level = [np.nan]
```
emiliom, line 174: `water_level` is a scalar variable, so its value should not be a list.
```suggestion
            water_level = np.nan
```

echopype/convert/set_groups_azfp.py
```diff
@@ -167,18 +168,21 @@ def set_platform(self) -> xr.Dataset:
...
+        else:
+            depth = -1 * depth_from_pressure(
+                unpacked_data["pressure"], latitude=30, atm_pres_surf=0
+            )
+            water_level = [0.0]
```
emiliom, line 179: `water_level` is a scalar variable, so its value should not be a list.
```suggestion
            water_level = 0.0
```

echopype/convert/set_groups_azfp.py
```diff
@@ -167,18 +168,21 @@ def set_platform(self) -> xr.Dataset:
         """Set the Platform group."""
         platform_dict = {"platform_name": "", "platform_type": "", "platform_code_ICES": ""}
         unpacked_data = self.parser_obj.unpacked_data
+        time2 = self.parser_obj.ping_time
+        if np.isnan(unpacked_data["pressure"]).all():
+            depth = [np.nan] * len(time2)
+            water_level = [np.nan]
+        else:
+            depth = -1 * depth_from_pressure(
```
emiliom, line 176: `depth` is a positive value (0 at the surface). It's the variable `vertical_offset` that is set to -1 * depth. So, to avoid confusion, let's rename the variable `depth` to `vertical_offset` here, above in the if block, and below in the `vertical_offset` dataarray assignment.

echopype/convert/set_groups_azfp.py
```diff
@@ -167,18 +168,21 @@ def set_platform(self) -> xr.Dataset:
...
-        # If tilt_x and/or tilt_y are all nan, create single-value time2 dimension
-        # and single-value (np.nan) tilt_x and tilt_y
-        tilt_x = [np.nan] if np.isnan(unpacked_data["tilt_x"]).all() else unpacked_data["tilt_x"]
-        tilt_y = [np.nan] if np.isnan(unpacked_data["tilt_y"]).all() else unpacked_data["tilt_y"]
-        if (len(tilt_x) == 1 and np.isnan(tilt_x)) and (len(tilt_y) == 1 and np.isnan(tilt_y)):
-            time2 = [self.parser_obj.ping_time[0]]
-        else:
-            time2 = self.parser_obj.ping_time
+        tilt_x = unpacked_data["tilt_x"]
+        tilt_y = unpacked_data["tilt_y"]
```
emiliom, line 185: You removed the block that checks if `unpacked_data["tilt_x"]` and `unpacked_data["tilt_y"]` are arrays with all nan, sets them to be length-1 nan arrays, and creates `time2` to be a length-1 array. But let's examine the possible scenarios:

tilt_x & tilt_y | vertical_offset | results for tilt_x, tilt_y, and vertical_offset
---- | ---- | ---- 
valid data | valid data | "full-length" arrays with valid data, with an unmodified `time2` dimension
valid data | invalid data (all nan) | full-length arrays where tilt_x & tilt_y have valid data and vertical_offset is all nan
invalid data (all nan) | valid data | full-length arrays where tilt_x & tilt_y are all nan and vertical_offset has valid data
invalid data (all nan) | invalid data (all nan) | length-1 arrays set to nan values, where `time2` dimension is set to `[self.parser_obj.ping_time[0]]`

For the 4th scenario (all invalid, nan data), I don't think your simplification would result in length-1 arrays for `time2`, `tilt_x`, `tilt_y`, and `vertical_offset`. Or does it?

emiliom:
> except one where we are still not sure whether to pass atm_surf_pres a value of 0 or 10.1325 (dbar).

We heard back from Steve about this! In summary, it sounds like their pressure measurements are almost always, if not always, *uncorrected* ("uncompensated") for atmospheric pressure. You already apply this atmospheric correction in `parse_azfp.py::_compute_pressure`, so using `atm_surf_pres a value=0` in the call to `depth_from_pressure` is correct. However, the value you used in `_compute_pressure` is the one from the AZFP Matlab code, 10.125. Let's use the [officially accepted value](https://en.wikipedia.org/wiki/Atmospheric_pressure) instead, 10.1325 dbar.

echopype/convert/set_groups_azfp.py
praneethratna: > For the 4th scenario (all invalid, nan data), I don't think your simplification would result in length-1 arrays for `time2`, `tilt_x`, `tilt_y`, and `vertical_offset`. Or does it?

Nope, but i thought even if `tilt_x`, `tilt_y` and `vertical_offset` are all `np.nan` we don't make them length - 1 arrays and instead keep them with their original dimension as it is. I think that should be modified?

echopype/convert/set_groups_azfp.py
praneethratna: Updated!

echopype/convert/set_groups_azfp.py
praneethratna: Updated!

echopype/convert/set_groups_azfp.py
praneethratna: Updated!

echopype/convert/set_groups_azfp.py
praneethratna: Updated!

emiliom:
Thanks @praneethratna ! All your changes look good.

echopype/convert/set_groups_azfp.py
emiliom: If `tilt_x`, `tilt_y` and `vertical_offset` are all nan, they don't really have an "original" dimension. But the SONAR-netCDF4 convention that we follow defines `vertical_offset` to have a time dimension, which in our case is `time2` (`tilt_x` and `tilt_y` are not defined by the convention). For AZFP files, no other variable in the Platform group that has valid data would use that dimension. So, setting `time2` to be length-1 reduces data storage, minimizes ambiguity, and allows us to maintain the time dimension in `vertical_offset`.

praneethratna: thanks, got it. Updated the code!

leewujung:
If I understand the various emails and comments surrounding this PR correctly, I don’t think we should add `vertical_offset` based on depth calculated by pressure sensor sensor reading by assigning latitude=30. IMHO we should leave the user only with the pressure sensor values, so they can decide what they want to do to convert it to depth or vertical_offset. I think adding a `vertical_offset` value automatically would likely be misleading, especially if people do not carefully consider where the value comes from. 

Since the pressure to depth function already exists (thank you!), we can add to the docs to let users know that this function exists if they want to convert the pressure reading to depth by themselves. 

On the time stamp: tilt_x/y and pressure and many other sensor readings are read from the same time base, because they are wrapped in the same data packet.

emiliom:
That's reasonable. That means we won't merge this PR, since the updating of `vertical_offset` (and the related `water_level`)  was the sole focus of the PR. I suggest we leave the PR open for the time being, though, in case we want to revisit the issue in the future. In the meantime, I'll change it to a "Draft".

As reference, it's worth noting that in the AZFP Matlab toolbox, the [computeDepth](https://github.com/OSOceanAcoustics/AZFP_matlab/blob/main/matlab_toolbox/LoadAZFP.m#L540) function in `LoadAZFP.m` also doesn't account for latitude -- for the same reason, since latitude normally would not be available.

> Since the pressure to depth function already exists (thank you!), we can add to the docs to let users know that this function exists if they want to convert the pressure reading to depth by themselves.

Sounds good!

leewujung:
I'll close this PR now since adding `vertical_offset` based on the pressure value could be misleading without knowing what the actual latitude value is.

Issue #857, opened 2022-10-15
Remove remnant unused "ALL" param in parser object


leewujung:
We decided (and implemented) a few releases ago that we will NOT allow selective parsing of only a subset of EK60/EK80 datagrams. However the old selector `self.data_Type = "ALL"` is still present in the code and causing confusion. Let's remove it.

emiliom:
Is this as simple as removing these two blocks?

https://github.com/OSOceanAcoustics/echopype/blob/f8082cac32bfbee2b21c62d266507491d221182d/echopype/convert/parse_ek80.py#L15-L16

https://github.com/OSOceanAcoustics/echopype/blob/f8082cac32bfbee2b21c62d266507491d221182d/echopype/convert/parse_ek60.py#L12-L13

leewujung:
I am not sure, but it likely involves checking some functions that used ALL as a param. I think those were meant to try to selectively parse only some datagram types. Not sure if the mechanisms were entirely removed.

PR #1203, opened 2023-10-27
Removed the unused `ALL` param in parser object [all tests ci] 

praneethratna:
Addresses #857 by removing unused "ALL" param in parser object.

CC @leewujung


praneethratna:
Closing and re-opening the PR to rerun all the tests.

echopype/convert/parse_base.py
```diff
@@ -159,11 +146,8 @@ def parse_raw(self):
             # Read the rest of datagrams
             self._read_datagrams(fid)
 
-        if "ALL" in self.data_type:
-            # Convert ping time to 1D numpy array, stored in dict indexed by channel,
-            #  this will help merge data from all channels into a cube
-            for ch, val in self.ping_time.items():
-                self.ping_time[ch] = np.array(val, dtype="datetime64[ns]")
+        for ch, val in self.ping_time.items():
+            self.ping_time[ch] = np.array(val, dtype="datetime64[ns]")
```
leewujung, line 150: Let's keep the comments that was originally on top of this:
```suggestion
        # Convert ping time to 1D numpy array, stored in dict indexed by channel,
        #  this will help merge data from all channels into a cube
        for ch, val in self.ping_time.items():
            self.ping_time[ch] = np.array(val, dtype="datetime64[ns]")
```

echopype/convert/api.py
```diff
@@ -430,10 +430,7 @@ def open_raw(
 
     # TODO: the if-else below only works for the AZFP vs EK contrast,
     #  but is brittle since it is abusing params by using it implicitly
-    if SONAR_MODELS[sonar_model]["xml"]:
-        params = xml_chk
-    else:
-        params = "ALL"  # reserved to control if only wants to parse a certain type of datagram
+    params = xml_chk
```
leewujung, line 433: I think we can remove these altogether and change the line 441 assignment to be just `params=xml_chk`, and add a comment to say that `xml_chk` is empty if not AZFP files

leewujung: Looking into the code again, it seems that this `params` argument is really only used for passing AZFP XML file. Maybe we should rename it to be something like `file_meta`? (this will go with the `meta_source_filenames` that are populated for AZFP like [here](https://echopype.readthedocs.io/en/latest/data-format-raw.html#asl-environmental-sciences-azfp-data)).

Could you please make this change, and also add a comment at line 492 to say currently this is only used for AZFP XML file?

echopype/convert/parse_base.py
```diff
@@ -132,19 +131,7 @@ def parse_raw(self):
...
-                elif "CONFIG" in self.data_type:
-                    xml_type = "configuration"
-                logger.info(f"exporting {xml_type} XML file")
-                # Don't parse anything else if only the config xml is required.
-                if "CONFIG" in self.data_type:
-                    return
-            # If not exporting to XML, print the usual converting message
-            else:
-                self._print_status()
+            self._print_status()
```
leewujung, line 134: Tiny addition of a comment:
```suggestion
            # print the usual converting message
            self._print_status()
```

leewujung (commented):
@praneethratna : Thanks for the PR, the changes look great! I just have [one comment](https://github.com/OSOceanAcoustics/echopype/pull/1203/files#r1382431126) about the current argument `params` to the parser functions, about renaming it since it is really only used for AZFP XML file at the moment. 

Along with this change, could you please also clean up the unused `params` argument in the `__init__` for ParseEK60/EK80/AD2CP?

Also, a note that in #1185 the whole mechanisms around `dgram_zarr_vars` is removed and replaced with a much cleaner set of code. Because that PR touches so many places in the parser code for the EK files, do you mind that we keep this current PR unmerged until that one is merged, and ask you to bring in the changes into this PR before merging?

Thanks again for your contribution! Really appreciate it!

echopype/convert/api.py
praneethratna: I don't understand this comment, do you mean to rename `params` to `file_meta` in line 433 and populate that value in `_set_convert_params` method?

leewujung: Somehow the line number got messed up, maybe between different commits/branches. I meant this line here: https://github.com/OSOceanAcoustics/echopype/blob/f98e825e808a98dfdf43f66c8be242347d353e3b/echopype/convert/api.py#L444

I meant to say to rename this `params` to `file_meta`, since for AZFP this XML file name eventually becomes `meta_source_filenames` in the set_group functions.

`_set_convert_params` should be removed since we don't use it anymore, but I think we can do that in another PR, so that this one is not too overloaded.

praneethratna: Understood, thanks! Will make another PR for removal of `_set_convert_params` after this.

leewujung: Thank you!

echopype/convert/api.py
```diff
@@ -428,20 +428,16 @@ def open_raw(
...
 
     # obtain dict associated with directly writing to zarr
     dgram_zarr_vars = SONAR_MODELS[sonar_model]["dgram_zarr_vars"]
 
     # Parse raw file and organize data into groups
     parser = SONAR_MODELS[sonar_model]["parser"](
         file_chk,
-        params=params,
+        # Currently used only for AZFP XML file
+        file_meta=params,
```
leewujung, line 440: How about just changing this to `file_meta=xml_chk` and the line (`params = xml_chk`) above?

leewujung:
@praneethratna : I just have one minor comment above. 

Also: #1185 is just merged! The merge conflicts are because of the removal of everything related to `dgram_zarr_vars`. I took a quick look and thought it should be straightforward to resolve the conflicts, but please let me know if you run into cases you are unsure about. Once the merge is done I think we can merge this PR. Thank you!!

echopype/convert/api.py
praneethratna: Updated!

praneethratna:
@leewujung I messed up this PR while resolving the merge conflicts, so I've made a new pull request #1214 with all the changes and all the tests passing, please do check that! I'm closing this PR. Thanks!

Issue #1152, opened 2023-09-01
`consolidate.add_depth` outputs to work with `compute_MVBS`


leewujung:
In the overhaul of `compute_MVBS` (currently in the working in #1124) the variable `water_depth` is propagated down in the MVBS dataset. We should review what variables are needed to be propagated down, now that we have standardized the z-axis variables in v0.8.0, and some more improvements/refinement of `add_depth` will be introduced soon by @emiliom.

Relevant code section:
https://github.com/lsetiawan/echopype/blob/5a963bd397980583827bac2b6ba0b4bdd6ccf8cc/echopype/commongrid/api.py#L164-L166
This is in the PR code so may change in the future.

emiliom:
Note: the variable that's now propagated is called `water_level`, not `water_depth`.

leewujung:
This will be addressed potentially under #1205 or another PR that addresses the same thing.

PR #1205, opened 2023-10-28
Enable use of data from EchoData object in `consolidate.add_depth`

emiliom:
Overhaul `consolidate.add_depth` to enable using data from EchoData object. Addresses #790. See the echopype documentation for [reference on several of the variables](https://echopype.readthedocs.io/en/stable/data-proc-additional.html#vertical-coordinate-z-axis-variables).

Rotations are performed using [scipy.spatial.transform.Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html).

Notes and caveats:
- See the TODOs in the code.
- If both an EchoData object and external data (as scalar arguments) are passed, only data from the EchoData object is used.
- I'm seeing too many nans in the resulting `depth` variable for the sample EK80 file, so I suspect something isn't working properly.
- AZFP `tilt_x/y` variables are not handled. Only standard SONAR-netCDF4 variables are used. It assumes that tilt_x/y variables have already been transformed into pitch & roll. I'll create a new issue for this.
- I learned recently (from the EK80 reference manual) that EK80 `TransducerAlphaX/Y/Z` variables are in degrees. We're currently interpreting them as a unit vector when mapping them to `beam_direction_x/y/z`. A conversion will need to be implemented in `set_groups_ek80` (and ek60?). I'll create a new issue for this. For the sample file (below), it's not an issue b/c all `TransducerAlphaX/Y/Z` values are zero. I couldn't find an EK80 file with `TransducerAlphaX/Y/Z` values that are not all zero.
- We don't have a sample EK60/80 raw file that has all relevant variables populated with what appear to be valid values. The one file I'm using, from 2021 Hake survey ("ek80/ncei-wcsd/SH2106/EK80/Reduced_Hake-D20210701-T131621.raw"), only has some of the variables with apparently valid, non zero data: `pitch`, `roll`, and `vertical_offset`. `vertical_offset` (based on `heave`) is also suspicious b/c values are close to zero. `water_level` (0) + `vertical_offset` is supposed to be the positive distance from the ship origin.

### DIFF
--- a/echopype/consolidate/api.py
+++ b/echopype/consolidate/api.py
@@ -53,7 +53,7 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
 def add_depth(
     ds: xr.Dataset,
     echodata: Optional[EchoData] = None,
-    echodata_strict=False,
+    echodata_strict: bool = False,
     depth_offset: float = 0,
     tilt: float = 0,
     downward: bool = True,
--- a/echopype/consolidate/api.py
+++ b/echopype/consolidate/api.py
@@ -70,7 +70,7 @@ def add_depth(
     ds : xr.Dataset
         Source Sv dataset to which a depth variable will be added.
         Must contain `echo_range`.
-    echodata : EchoData, default=None
+    echodata : EchoData, optional
         ``EchoData`` object from which the ``Sv` dataset originated.
         It must contain transducer position and orientation information.
     echodata_strict : bool, default=False
--- a/echopype/consolidate/api.py
+++ b/echopype/consolidate/api.py
@@ -102,13 +102,9 @@ def add_depth(
     """
 
     if echodata is None:
-        # depth_offset, tilt and downward are scalars passed as arguments to add_depth,
-        # and echodata is None
+        # depth_offset and tilt scalars passed as arguments to add_depth,
         transducer_depth = depth_offset
-
-        # Multiplication factor depending on if transducers are pointing downward
-        orientation_mult = 1 if downward else -1
-        echo_range_z_scaling = orientation_mult * np.cos(np.deg2rad(tilt))
+        echo_range_z_scaling = np.cos(np.deg2rad(tilt))
     else:
         # TODO: TODOs and notes for AZFP:
         #  - Assume tilt_x/y data has been translated into pitch & roll variables.
@@ -236,8 +232,10 @@ def add_depth(
     #  instead of echodata variables?
 
     # Compute depth
+    # Multiplication factor depending on if transducers are pointing downward
+    orientation_mult = 1 if downward else -1
     # ds["echo_range"] dimensions: (channel, ping_time, range_sample)
-    ds["depth"] = transducer_depth + ds["echo_range"] * echo_range_z_scaling
+    ds["depth"] = transducer_depth + orientation_mult * ds["echo_range"] * echo_range_z_scaling
 
     # Add attributes, including history attribute
     # TODO: In history_attr, specify whether the offset & angle data originated in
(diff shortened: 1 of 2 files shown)


emiliom:
Notes to self for creating mock `EchoData` data for testing the new functionality. Here's the minimal code needed to create a bare `EchoData` object that has just the elements needed by `consolidate.add_depth`: a `sonar_model` attribute and the "Platform" and "Sonar/Beam_group1" groups.

```python
from datatree import DataTree
import xarray as xr
from echopype.echodata.echodata import EchoData

ep = EchoData(sonar_model="EK80")

ds1 = xr.Dataset(dict(bar=("x", [1, 2])))
ds2 = xr.Dataset(dict(bar=("x", [10, 20])))
tree = DataTree.from_dict({'Platform': ds1, 'Sonar/Beam_group1': ds2}, name='root')
ep._set_tree(tree)
```

echopype/consolidate/api.py
```diff
@@ -51,6 +52,8 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
 
 def add_depth(
     ds: xr.Dataset,
+    echodata: EchoData = None,
```
lsetiawan, line 55:
```suggestion
    echodata: Optional[EchoData] = None,
```

echopype/consolidate/api.py
```diff
@@ -67,66 +70,183 @@ def add_depth(
...
-    #     if tilt is None:
-    #         if "tilt" in ds:
-    #             tilt = ds["tilt"]
-    #         else:
-    #             raise ValueError(
-    #                 "tilt not found in dataset and needs to be supplied by the user. "
-    #                 "Required when vertical=False"
-    #             )
-    # else:
-    #     tilt = 0
```
lsetiawan, line 103:
```suggestion
    del echodata_strict

```
Looks like pre-commit is currently not throwing error for unused argument! Interesting... anyhow, maybe it's better to just remove it here for now as a better practice: https://pycodequ.al/docs/pylint-messages/w0613-unused-argument.html#how-to-fix

echopype/consolidate/api.py
```diff
@@ -51,6 +52,8 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
 
 def add_depth(
     ds: xr.Dataset,
+    echodata: EchoData = None,
+    echodata_strict=False,
```
lsetiawan, line 56:
```suggestion
    echodata_strict: bool = False,
```

echopype/consolidate/api.py
emiliom: Thanks

echopype/consolidate/api.py
emiliom: Right now that argument is more for discussion; ie, whether we need something like it. If we end up seeing value in it, it will be implemented (used). Otherwise, it'll be dropped. So, I won't change anything yet.

echopype/consolidate/api.py
```diff
@@ -67,66 +70,183 @@ def add_depth(
     ds : xr.Dataset
         Source Sv dataset to which a depth variable will be added.
         Must contain `echo_range`.
-    depth_offset : float
+    echodata : EchoData, default=None
```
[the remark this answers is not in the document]
emiliom: Got it. Will merge.

emiliom:
Added a test function to generate mock EchoData object with only the variables (and one attribute) required by `consolidate.add_depth`. Added a test that uses this mock data, together with a mock Sv Dataset.

leewujung:
@ctuguinay : This is also related to #1152. Let's talk through this.

leewujung:
We'll close this now because it'll be superseded by #1319 .

PR #1207, opened 2023-10-31
Added `depth_from_pressure` method required for the calculation of `vertical_offset` value

praneethratna:
Partial fix for 2nd step of https://github.com/OSOceanAcoustics/echopype/issues/1181#issuecomment-1763217291, by including calculations for `depth` value as mentioned in https://github.com/OSOceanAcoustics/echopype/issues/1181#issuecomment-1781696782.

### DIFF
--- a/echopype/utils/misc.py
+++ b/echopype/utils/misc.py
@@ -40,7 +40,10 @@ def depth_from_pressure(
     latitude : Union[float, FloatSequence], default=30.0
         Latitude in decimal degrees.
     atm_pres_surf : Union[float, FloatSequence], default=0.0
-        Atmospheric pressure at the surface in dbar
+        Atmospheric pressure at the surface in dbar.
+        Use the default 0.0 value if pressure is corrected to be 0 at the surface.
+        Otherwise, enter a correction for pressure due to air, sea ice and any other 
+        medium that may be present
 
     Returns
     -------
--- a/echopype/utils/misc.py
+++ b/echopype/utils/misc.py
@@ -1,4 +1,9 @@
+from typing import List, Optional, Tuple, Union
+
 import numpy as np
+from numpy.typing import NDArray
+
+FloatSequence = Union[List[float], Tuple[float], NDArray[float]]
 
 
 def camelcase2snakecase(camel_case_str):
@@ -16,7 +21,11 @@ def camelcase2snakecase(camel_case_str):
     return camel_case_str.lower()
 
 
-def depth_from_pressure(pressure, latitude=30, atm_pres_surf=0):
+def depth_from_pressure(
+    pressure: Union[float, FloatSequence],
+    latitude: Optional[Union[float, FloatSequence]] = 30.0,
+    atm_pres_surf: Optional[Union[float, FloatSequence]] = 0.0,
+) -> NDArray[float]:
     """
     Convert pressure to depth using UNESCO 1983 algorithm.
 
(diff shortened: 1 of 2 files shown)


emiliom:
Thanks @praneethratna ! I'm working on this.

emiliom:
See the PR I submitted against your PR, with enhancements. This will be on your echopype fork.

emiliom:
@praneethratna I'm moving [your comment](https://github.com/praneethratna/echopype/pull/2#discussion_r1379570199) to the main PR, here, since it's a broader comment and not specific to my edits to your PR (I know, it's confusing!):

> I just now remember that in the `_compute_pressure` method we subtract 10.125 from `pressure` value everytime. Should we remove this `atm_pres_surf` argument then?

For reference, here's the Matlab function you're referring to: [LoadAZFP.m, computeDepth](https://github.com/OSOceanAcoustics/AZFP_matlab/blob/main/matlab_toolbox/LoadAZFP.m#L540). 

And here's the statement in `depth_from_pressure` that subtracts  `atm_pres_surf` from `pressure`:
```python
pressure = pressure - atm_pres_surf
```

emiliom:
I don't think we should remove `atm_pres_surf` argument. The broader question is whether pressure measurements that a user may pass to `depth_from_pressure` would ever have atmospheric pressure at the surface *already* removed. If that's the case, then the function should not apply a correction for a constant atmospheric pressure value.

That's why we're including `atm_pres_surf` as a function argument, but also why it's optional and has a default value. I think the main question is what default value would be most useful: 0, or 10.1325 dbar. Right now it's 0. I'll ask around to come to a decision, but let's discuss that here, not in my "PR to your PR", since I didn't change the `atm_pres_surf` value in that PR.

praneethratna:
> I don't think we should remove the `atm_pres_surf` argument. The broader question is whether pressure measurements that a user may pass to `depth_from_pressure` would ever have atmospheric pressure at the surface *already* removed. If that's the case, then the function should not apply a correction for a constant atmospheric pressure value.

I was actually thinking only for use case of setting `vertical_offset`. I missed the point that the function can be used by a user independently as well.

> That's why we're including `atm_pres_surf` as a function argument, but also why it's optional and has a default value. I think the main question is what default value would be most useful: 0, or 10.1325 dbar. Right now it's 0. I'll ask around to come to a decision, but let's discuss that here, not in my "PR to your PR", since I didn't change the `atm_pres_surf` value in that PR. 

I think it should be 10.1325 dbar? Since we take pressure in dbar.

emiliom:
I've confirmed that pressure sensors can be set (calibrated) to return a value of 0 at the surface. That's probably not an unusual configuration. So, let's keep the default `atm_pres_surf` value at 0. I will make an inline suggestion with edits to the doc string to make this clearer. Then, we can merge this PR!

echopype/utils/misc.py
```diff
@@ -11,3 +19,66 @@ def camelcase2snakecase(camel_case_str):
...
+    papers in marine science, 44. http://unesdoc.unesco.org/images/0005/000598/059832eb.pdf
+
+    Parameters
+    ----------
+    pressure : Union[float, FloatSequence]
+        Pressure in dbar
+    latitude : Union[float, FloatSequence], default=30.0
+        Latitude in decimal degrees.
+    atm_pres_surf : Union[float, FloatSequence], default=0.0
+        Atmospheric pressure at the surface in dbar
```
emiliom, line 43:
```suggestion
    atm_pres_surf : Union[float, FloatSequence], default=0.0
        Atmospheric pressure at the surface in dbar.
        Use the default 0.0 value if pressure is corrected to be 0 at the surface.
        Otherwise, enter a correction for pressure due to air, sea ice and any other 
        medium that may be present
```

emiliom: I'll merge this commit suggestion myself.

Issue #1164, opened 2023-09-13
Does `signal.convolve` use dask under the hood?


lsetiawan:
Within the `ek80_complex` module we use [`signal.convolve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.convolve.html), which might not be optimized for xarray dataset. Some investigation is needed to see if the code below would properly perform FFT using [dask](https://docs.dask.org/en/stable/) in a parallel fashion.

https://github.com/OSOceanAcoustics/echopype/blob/3ca028f04a4fd3643bb0ad350dc09a699883ee1e/echopype/calibrate/ek80_complex.py#L275-L293

lsetiawan:
Look into potentially using: https://xrft.readthedocs.io/en/latest/

anantmittal:
Confirm if there is a test function for `compress_pulse` function in `echopype/echopype/calibrate/ek80_complex.py`.

lsetiawan:
@anantmittal You should be able to use the same test data like in the `test_ek80_BB_power_Sv` test: https://github.com/OSOceanAcoustics/echopype/blob/9dfe5ba9cd656eb2f5dc9007fcb960231108cf1e/echopype/tests/calibrate/test_calibrate_ek80.py#L155-L156

leewujung:
Addressed by #1208.

PR #1208, opened 2023-11-01
Update compress_pulse to Parallelize Convolution

anantmittal:
Explores #1164


echopype/calibrate/ek80_complex.py
```diff
@@ -273,21 +274,34 @@ def compress_pulse(backscatter: xr.DataArray, chirp: Dict) -> xr.DataArray:
...
+
     for chan in backscatter["channel"]:
+        # Select channel `chan` and drop the specific beam dimension if all of the values are nan.
         backscatter_chan = backscatter.sel(channel=chan).dropna(dim="beam", how="all")
 
         tx = chirp[str(chan.values)]
         replica = np.flipud(np.conj(tx))
+
+        def _signal_convolve(m):
+            return signal.convolve(m, replica, mode="full")[replica.size - 1 :]
```
lsetiawan, line 286: Should the `replica.size` be `tx.size`? Why did this change?

lsetiawan (approved):

lsetiawan:
This current PR seem to make a difference in terms of parallelization. @leewujung would you be able to go through the changes and see if this is what you're looking for for this case? Thank you!

leewujung:
@anantmittal: Thanks for the PR! The change here does make it possible to perform convolution on a dask array, which was previously not allowed, so that's great.

I tested this using a not very large dataset (`test_data/ek80/D20170912-T234910.raw`, ~95 MB), and found that by forcing the `backscatter_r/i` variables to be dask arrays, the parallelization didn't seem to consistently improve the speed compare to straight in-memory computation. But this is probably not a fair comparison since the data is small. Were you able to try this on a much larger mock data (e.g. randomly generate `backscatter_chan`) to see what the difference may be?

leewujung (commented):
@anantmittal: Thanks for the PR! The change here does make it possible to perform convolution on a dask array, which was previously not allowed, so that's great.

I tested this using a not very large dataset (`test_data/ek80/D20170912-T234910.raw`, ~95 MB), and found that by forcing the `backscatter_r/i` variables to be dask arrays, the parallelization didn't seem to consistently improve the speed compare to straight in-memory computation. But this is probably not a fair comparison since the data is small. Were you able to try this on a much larger mock data (e.g. randomly generate `backscatter_chan`) to see what the difference may be?

echopype/tests/calibrate/test_calibrate_ek80.py
```diff
@@ -151,34 +178,70 @@ def test_ek80_BB_range(ek80_cal_path, ek80_ext_path):
...
+    pyecholab_data_path,
+    pyecholab_file_path,
+    dask_array,
+    request,
+):
+    raw_data_path = request.getfixturevalue(raw_data_path)
+    ek80_raw_path = raw_data_path / raw_file_name  # rx impedance / rx fs / tcvr type
+
+    if dask_array:
+        ed = ep.open_raw(ek80_raw_path, sonar_model="EK80", destination_path="swap")
```
leewujung, line 212: Due to changes in #1185, `destination_path="swap"`  should be changed to `use_swap=True`

lsetiawan:
## Benchmarking test

I did some simple benchmarking test with a test file in the `ek80/D20170912-T234910.raw`. The test does the following:
1. I read in this file and then duplicate the data across additional ping times, which expands the data from `271.2 MB` to `5.3 GB`.
2. From here, I chunk the data across ping time and save to a file for consistency
3. I read in the data back and run this PR's `compress_pulse` function that have been modified and looks like when run in parallel with dask, we do see some speed up compared to not. 

## Results
W/O Dask the run took 58.1 s
With Dask the run took 24.3 s

Looking at the dask dashboard, the computation happens in parallel

Screenshot 2023-12-15 at 3 32 30 PM


For reproducibility, you can see my run in the following gist: https://gist.github.com/lsetiawan/20c09c3eaa76508a741e31ff28a83402

lsetiawan:
> I tested this using a not very large dataset (test_data/ek80/D20170912-T234910.raw, ~95 MB), and found that by forcing the backscatter_r/i variables to be dask arrays, the parallelization didn't seem to consistently improve the speed compare to straight in-memory computation.

For a small dataset, due to the dask overhead for creating task graph, it's probably better to load things in memory first then do the computation. When I did try this with the original data, sure enough a dask array is slower than just in memory computation.

The approach suggested in this PR works for both in-memory and dask array, which is really powerful since when the data is big, we can parallelize the convolution! This I think satisfy the issue in https://github.com/OSOceanAcoustics/echopype/issues/1164. Further optimization can probably happen, but it's a first start 😄

lsetiawan:
@leewujung is this good to go?

leewujung (approved):

leewujung:
Yep, I think this is good now. Thanks for doing the benchmarking tests!

Issue #1195, opened 2023-10-17
Cannot open simrad ES60 raw files


merin-joseph:
### General description of problem
Cannot open simrad ES60 raw files

### Computing environment

- Echopype version:
-echopype-0.8.1 
### Minimum example

The following code reproduces the errors I encountered:
```
ed = open_raw('L0012-D20190319-T192316-ES60.raw',sonar_model='EK60') 
ed
```

### Error message printouts

Below is the error messages I received when running the above code:
```
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In[5], line 1
----> 1 ed = open_raw('L0012-D20190319-T192316-ES60.raw',sonar_model='EK60') # EK80
      2 ed

File /srv/conda/envs/notebook/lib/python3.10/site-packages/echopype/utils/prov.py:237, in add_processing_level.<locals>.wrapper.<locals>.inner(*args, **kwargs)
    235 @functools.wraps(func)
    236 def inner(*args, **kwargs):
--> 237     dataobj = func(*args, **kwargs)
    238     if is_echodata:
    239         ed = dataobj

File /srv/conda/envs/notebook/lib/python3.10/site-packages/echopype/convert/api.py:429, in open_raw(raw_file, sonar_model, xml_path, convert_params, storage_options, use_swap, max_mb)
    424 # Parse raw file and organize data into groups
    425 parser = SONAR_MODELS[sonar_model]["parser"](
    426     file_chk, params=params, storage_options=storage_options, dgram_zarr_vars=dgram_zarr_vars
    427 )
--> 429 parser.parse_raw()
    431 # Direct offload to zarr and rectangularization only available for some sonar models
    432 if sonar_model in ["EK60", "ES70", "EK80", "ES80", "EA640"]:
    433     # Create sonar_model-specific p2z object

File /srv/conda/envs/notebook/lib/python3.10/site-packages/echopype/convert/parse_base.py:154, in ParseEK.parse_raw(self)
    148         self.CON1_datagram = None
    150     # IDs of the channels found in the dataset
    151     # self.ch_ids = list(self.config_datagram['configuration'].keys())
    152 
    153     # Read the rest of datagrams
--> 154     self._read_datagrams(fid)
    156 if "ALL" in self.data_type:
    157     # Convert ping time to 1D numpy array, stored in dict indexed by channel,
    158     #  this will help merge data from all channels into a cube
    159     for ch, val in self.ping_time.items():

[...]

PR #1215, opened 2023-11-08
Fixes `UnicodeDecodeError` for ES60 files [all tests ci]

praneethratna:
Addresses #1195 by changing encoding to `unicode_escape` instead of using default encoding `utf-8`.

CC @leewujung

### DIFF
--- a/echopype/convert/utils/ek_raw_parsers.py
+++ b/echopype/convert/utils/ek_raw_parsers.py
@@ -1541,7 +1541,7 @@ class SimradRawParser(_SimradDatagramParser):
         for indx, field in enumerate(self.header_fields(version)):
             data[field] = header_values[indx]
             if isinstance(data[field], bytes):
-                data[field] = data[field].decode()
+                data[field] = data[field].decode(encoding="unicode_escape")
 
         data["timestamp"] = nt_to_unix((data["low_date"], data["high_date"]))
         data["bytes_read"] = bytes_read


emiliom:
I'll add "[all tests ci]" to the PR title and closing-reopening the PR to force all tests to be run, to ensure there are no negative consequences with other EK/ES data. As it stands, it looks like none of the tests were actually run in the CI.

praneethratna:
Ping @emiliom, should we add a new test case using the data file provided here - https://github.com/OSOceanAcoustics/echopype/issues/1195#issuecomment-1788674130?

leewujung:
@praneethratna: thanks for looking in to this! I checked the example ES60 file and this happens in the `spare0` field and with `encoding="unicode_escape"` it returns `'7B\x00\x00\x90A'`. With a regular EK60 file, the same `spare0` field returns `'\x00\x00\x00\x00\x00\x00'`. I think this is fine since this field doesn't actually encode anything useful.

I am a little reluctant to add a 100MB file into the test data set. Let's ping the person who raised this issue to see if they could help provide a smaller file.

We can merge this once we decide whether we would add a smaller test data file into this (if such file exists). If not, we can add a comment that this error occurred for an ES60 file.

emiliom:
@praneethratna I've gone ahead and uploaded the new small test file (https://github.com/OSOceanAcoustics/echopype/issues/1195#issuecomment-1805255054) to our test folder on Google Drive. I created a new top-level folder, `es60`. BTW, that really is a tiny file -- just 11KB! I haven't tried opening it.

I think you can go ahead and create a test for it. Thanks!

leewujung:
@praneethratna: I've just verified that the 11KB file does contain the type of byte string that was not previously decoded. Please go ahead to add a test for this file under `test_convert_ek60.py`. Thank you!

leewujung (approved):

Issue #1129, opened 2023-08-22
`consolidate` subpackage functions need to accept both in-memory or stored datasets


leewujung:
From discussion with @Sohambutala:
Some functions in the subpackage `consolidate` such as `add_location` and `add_splitbeam_angle` have inconsistent support for whether or not non-in-memory `Sv` or `EchoData` datasets are allowed as inputs. In a workflow sense it is much better to allow stored netcdf or zarr files that contains these datasets, so we should add the support uniformly for functions in the `consolidate`.

Some specific notes:
- `add_splitbeam_angle` currently accepts in-memory or stored `Sv` datasets, but not for `EchoData`
- `add_location` currently does not accept stored `Sv` or `EchoData` datasets

PR #1216, opened 2023-11-08
Added support for `consolidate` subpackage functions to accept both in-memory or stored datasets

praneethratna:
Addresses #1129, #1237 and now `consolidate` subpackage functions can accept both in-memory or stored datasets.

CC @leewujung

### DIFF
--- a/echopype/consolidate/api.py
+++ b/echopype/consolidate/api.py
@@ -302,6 +302,10 @@ def add_splitbeam_angle(
             "so that the split-beam angles can be written to disk!"
         )
 
+    # obtain the file format of source_Sv if it is a path
+    if isinstance(source_Sv, (str, Path)):
+        source_Sv_type = get_file_format(source_Sv)
+
     source_Sv = open_source(source_Sv, "dataset", storage_options)
     echodata = open_source(echodata, "echodata", storage_options)
 
@@ -380,7 +384,6 @@ def add_splitbeam_angle(
     source_Sv["angle_alongship"] = theta
     source_Sv["angle_athwartship"] = phi
     if to_disk:
-        source_Sv_type = get_file_format(source_Sv)
         if source_Sv_type == "netcdf4":
             source_Sv.to_netcdf(mode="a", **storage_options)
         else:
--- a/echopype/consolidate/api.py
+++ b/echopype/consolidate/api.py
@@ -8,14 +8,14 @@ import xarray as xr
 from ..calibrate.ek80_complex import get_filter_coeff
 from ..echodata import EchoData
 from ..echodata.simrad import retrieve_correct_beam_group
-from ..utils.io import validate_source_ds_da
+from ..utils.io import open_ds_da_ed_path, validate_source_ds_da_ed
 from ..utils.prov import add_processing_level
 from .split_beam_angle import add_angle_to_ds, get_angle_complex_samples, get_angle_power_samples
 
 POSITION_VARIABLES = ["latitude", "longitude"]
 
 
-def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
+def swap_dims_channel_frequency(ds: Union[xr.Dataset, str, pathlib.Path]) -> xr.Dataset:
     """
     Use frequency_nominal in place of channel to be dataset dimension and coorindate.
 
@@ -24,8 +24,9 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
 
     Parameters
     ----------
-    ds : xr.Dataset
-        Dataset for which the dimension will be swapped
+    ds : xr.Dataset or str or pathlib.Path
+        Dataset or path to a file containing the Dataset
+        for which the dimension will be swapped
 
     Returns
     -------
@@ -35,6 +36,7 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
     -----
     This operation is only possible when there are no duplicated frequencies present in the file.
     """
+    ds, _ = open_ds_da_ed_path(ds, None, {})
     # Only possible if no duplicated frequencies
     if np.unique(ds["frequency_nominal"]).size == ds["frequency_nominal"].size:
         return (
@@ -50,7 +52,7 @@ def swap_dims_channel_frequency(ds: xr.Dataset) -> xr.Dataset:
 
 
 def add_depth(
-    ds: xr.Dataset,
+    ds: Union[xr.Dataset, str, pathlib.Path],
     depth_offset: float = 0,
     tilt: float = 0,
     downward: bool = True,
@@ -64,8 +66,9 @@ def add_depth(
 
     Parameters
     ----------
-    ds : xr.Dataset
-        Source Sv dataset to which a depth variable will be added.
+    ds : xr.Dataset or str or pathlib.Path
+        Source Sv dataset or path to a file containing the Source Sv dataset
+        to which a depth variable will be added.
         Must contain `echo_range`.
     depth_offset : float
         Offset along the vertical (depth) dimension to account for actual transducer
@@ -114,6 +117,7 @@ def add_depth(
     # else:
     #     tilt = 0
 
+    ds, _ = open_ds_da_ed_path(ds, None, {})
     # Multiplication factor depending on if transducers are pointing downward
     mult = 1 if downward else -1
 
@@ -132,7 +136,11 @@ def add_depth(
 
 
 @add_processing_level("L2A")
-def add_location(ds: xr.Dataset, echodata: EchoData = None, nmea_sentence: Optional[str] = None):
+def add_location(
+    ds: Union[xr.Dataset, str, pathlib.Path],
+    echodata: Optional[Union[EchoData, str, pathlib.Path]],
+    nmea_sentence: Optional[str] = None,
+):
     """
     Add geographical location (latitude/longitude) to the Sv dataset.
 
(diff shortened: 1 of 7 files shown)


echopype/echodata/echodata.py
```diff
@@ -93,6 +92,8 @@ def cleanup_swap_files(self):
                         v.store for k, v in dask_graph.items() if "original-from-zarr" in k
                     ]
                     fs = zarr_stores[0].fs
+                    from ..utils.io import delete_zarr_store
+
```
leewujung, line 96: Just curious why you decided to move these imports to within the function? Is it because they are only used once?

praneethratna: @leewujung they were causing cyclical import and also the reason you've mentioned since anyway they're being used only once, I don't think we need to define them globally in the file.

echopype/utils/io.py
```diff
@@ -414,24 +417,46 @@ def validate_source_ds_da(
...
+        check_file_existence(file_path=source_ds_da_ed, storage_options=storage_options)
+
+    return source_ds_da_ed, file_type
+
+
+def open_ds_da_ed_path(
+    source_ds_da: Optional[Union[xr.Dataset, xr.DataArray, str, Path]],
+    source_ed: Optional[Union[EchoData, str, Path]],
+    storage_options: Optional[dict],
+) -> Tuple[Optional[Union[xr.Dataset, xr.DataArray]], Optional[EchoData]]:
```
leewujung, line 446: Somehow I feel forcing `source_ed` to always be in the input argument is a little awkward, when you only need to deal with `source_ds_da`. Do you think it will work if the function is changed to take in `source`, `source_type`, and `storage_options`, and depending on `source_type` different open function is used.

echopype/utils/io.py
```diff
@@ -380,29 +382,30 @@ def env_indep_joinpath(*args: Tuple[str, ...]) -> str:
     return joined_path
 
 
-def validate_source_ds_da(
-    source_ds_da: Union[xr.Dataset, xr.DataArray, str, Path], storage_options: Optional[dict]
-) -> Tuple[Union[xr.Dataset, str, xr.DataArray], Optional[str]]:
+def validate_source_ds_da_ed(
+    source_ds_da_ed: Union[xr.Dataset, xr.DataArray, EchoData, str, Path],
```
leewujung, line 386: We can probably change `source_ds_da_ed` to just be `source`? Since the type is pretty clear here.

leewujung (commented):
Hey @praneethratna : Thanks for the PR! I only have pretty small comments and a question here.

While going through the changes, I noticed that the use of `source_Sv_path` does not make much sense () and that results in the mess in `add_splitbeam_angle` that `source_Sv` is separately validated outside of `open_ds_da_ed_path`. We should probably deal with it in a separate PR, to keep this one tidy.

echopype/utils/io.py
praneethratna: > Do you think it will work if the function is changed to take in `source`, `source_type`, and `storage_options`, and depending on `source_type` different open function is used.

@leewujung Yes, I have updated the method which will use different open function depending on `source_type`.

praneethratna:
Hey @leewujung I have addressed the above comments and also included the changes for #1237 in this PR since they are based on these changes. Thanks!

leewujung:
Hey @praneethratna : The last changes you made removed the part to save the added `theta` and `phi` variables to an existing dataset. Are you thinking that these variables would just be live in `source_Sv`? Maybe we just remove the `add_angle_to_ds` function and put the content (with necessary modifications) in `add_splitbeam_angle`?

leewujung (changes requested):
Hey @praneethratna : The last changes you made removed the part to save the added `theta` and `phi` variables to an existing dataset. Are you thinking that these variables would just be live in `source_Sv`?

Maybe we just remove the `add_angle_to_ds` function and put the content (with necessary modifications) in `add_splitbeam_angle`?

praneethratna:
> Hey @praneethratna : The last changes you made removed the part to save the added `theta` and `phi` variables to an existing dataset. Are you thinking that these variables would just be live in `source_Sv`?
> 
> Maybe we just remove the `add_angle_to_ds` function and put the content (with necessary modifications) in `add_splitbeam_angle`?

Hey @leewujung I think they should live, but anyway I have moved the content to `add_splitbeam_angle` itself and also removed the `return_dataset` parameter!

leewujung:
> Hey @leewujung I think they should live, but anyway I have moved the content to `add_splitbeam_angle` itself and also removed the `return_dataset` parameter!

Ah I see what the confusion may be. What I meant is that the functionality of saving the new variables `phi` and `theta` was removed in your changes, but we actually need the functionality (to update the on-disk dataset with the added variables  `phi` and `theta`, while still being able to access them as `angle_alongship` and `angle_athwartship`).

How about we do the following:
- remove `return_dataset`
- add a new argument `to_disk` (default to `True`), where `to_disk=True` means that the angle will be saved to disk via `mode="a"` and the `source_Sv` dataset lazy-loaded back and returned, and `to_disk=False` means that the angle will just be added to `source_Sv` and returned but not saved to disk.

leewujung (changes requested):
> Hey @leewujung I think they should live, but anyway I have moved the content to `add_splitbeam_angle` itself and also removed the `return_dataset` parameter!

Ah I see what the confusion may be. What I meant is that the functionality of saving the new variables `phi` and `theta` was removed in your changes, but we actually need the functionality (to update the on-disk dataset with the added variables  `phi` and `theta`, while still being able to access them as `angle_alongship` and `angle_athwartship`).

How about we do the following:
- remove `return_dataset`
- add a new argument `to_disk` (default to `True`), where `to_disk=True` means that the angle will be saved to disk via `mode="a"` and the `source_Sv` dataset lazy-loaded back and returned, and `to_disk=False` means that the angle will just be added to `source_Sv` and returned but not saved to disk.

echopype/consolidate/api.py
```diff
@@ -240,19 +255,20 @@ def add_splitbeam_angle(
...
 
     Returns
     -------
     xr.Dataset or None
-        If ``return_dataset=False``, nothing will be returned.
-        If ``return_dataset=True``, either the input dataset ``source_Sv``
+        If ``to_disk=False``, nothing will be returned.
+        If ``to_disk=True``, either the input dataset ``source_Sv``
         or a lazy-loaded Dataset (from the path ``source_Sv``)
         with split-beam angles added will be returned.
```
leewujung, line 269: I think the True and False conditions here are reverse.

> to_disk=True means that the angle will be saved to disk via mode="a" and the source_Sv dataset lazy-loaded back and returned, and to_disk=False means that the angle will just be added to source_Sv and returned but not saved to disk.

echopype/consolidate/api.py
```diff
@@ -364,9 +369,19 @@ def add_splitbeam_angle(
...
-        theta, phi, source_Sv, return_dataset, source_Sv_path, file_type, storage_options
-    )
+    theta.attrs["long_name"] = "split-beam alongship angle"
+    phi.attrs["long_name"] = "split-beam athwartship angle"
+
+    # add the split-beam angles to the provided Dataset
+    source_Sv["angle_alongship"] = theta
+    source_Sv["angle_athwartship"] = phi
+    if to_disk:
+        if source_Sv_type is not None:
```
leewujung, line 379: I think there should be a check up front, right after `source_Sv_type = get_file_format(source_Sv)`, where you fail the case where `to_disk=True and source_Sv_type is None`, otherwise it is not obvious what's going on if people put `to_disk=True` but the to_netcdf/to_zarr are not invoked due to `source_Sv_type` being None.

leewujung: Actually, looking at `get_file_format`, it does not seem like the output would ever be None. Could you rework the logic up front? What I wanted to say in this and the previous comment is that, if people put in a `source_Sv` that is an in-memory object instead of a path, an error should be raised if they also set `to_disk=True`, because there is no place the data can be saved to.

leewujung (changes requested):
@praneethratna : I put in a couple comments about changes needed in the logic of checking the combination `to_disk` and `souce_Sv` type. Thanks!

echopype/consolidate/api.py
praneethratna: @leewujung Understood, an error will be raised if `source_Sv` is not a path and `to_disk` is true.

echopype/consolidate/api.py
praneethratna: Updated!

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -340,7 +340,7 @@ def _tests(ds_test, location_type, nmea_sentence=None):
                 'splitbeam/DY1801_EK60-D20180211-T164025_angles_T4.mat',
                 'splitbeam/DY1801_EK60-D20180211-T164025_angles_T5.mat'
             ],
-            "CW", "power", False, True
+            "CW", "power", False, False
```
leewujung, line 343: So this has to be `True` because here we need to test whether both `to_disk=True` and `to_disk=False` are working. This is the only case where `to_disk=True` is tested.

The reason why this test fails right now is because you are doing `source_Sv_type = get_file_format(source_Sv)` after `source_Sv = open_source(source_Sv, "dataset", storage_options)`, so `source_Sv` is no longer a path/str when you do `get_file_formate`.

Good to be careful about changing existing testing parameter combination, since they are usually there for a reason.

Also suggest changing the `write_Sv_to_file` parameter to just `to_disk` to make the code easy to read.

leewujung (changes requested):
@praneethratna : I found a bug in the latest change and commented on it. Should be pretty straightforward to fix.

For the testing case where `to_disk=True`, please also add the below to make sure that it is actually loading data from the temp zarr file (in which case the `.data` is a dask array):
```python
assert isinstance(source_Sv["angle_alongship"].data, dask.array.core.Array)
assert isinstance(source_Sv["angle_athwartship"].data, dask.array.core.Array)
```

Thanks!

echopype/tests/consolidate/test_consolidate_integration.py
praneethratna: @leewujung Ah, sorry, misunderstood the code point there. Updated now!

leewujung (approved):
Awesome, thanks for the changes! I'll merge this now!

Issue #1217, opened 2023-11-09
Error upon trying calibration of Simrad EK80 .nc converted file: ValueError: zero-size array to reduction operation maximum which has no identity


icaro-00:
Hello all! I am currently trying to get sv values from Simrad EK80 echosounder. I converted the .raw file to .nc (netCDF4) files. I am now moving on the second step, where I am processing the .nc converted file and calibrate to get sv values. However, if I run the script provided, I run into the error:

ValueError: zero-size array to reduction operation maximum which has no identity

Does anyone know how to proceed to fix it? Or am I missing some mid-steps prior to calibration? Or?

**The code I run:**

nc_path = r"C:\Users\Runner\Desktop\Echopype\unpacked_files\20230925--D20230925-T082637.nc"  

echodata = ep.open_converted(nc_path) 

ds_Sv = ep.calibrate.compute_Sv(echodata, waveform_mode="CW", encode_mode="complex")

-------------------------------------------------------

**And here below the error I get:**

ValueError                                Traceback (most recent call last)
Cell In[16], line 5
      2 echodata = ep.open_converted(nc_path)  # create an EchoData object
      4 # For EK80 data, you need to specify waveform_mode and encode_mode
----> 5 ds_Sv = ep.calibrate.compute_Sv(echodata, waveform_mode="CW", encode_mode="complex")

File ~\anaconda3\envs\Echopype\lib\site-packages\echopype\calibrate\api.py:206, in compute_Sv(echodata, **kwargs)
    120 def compute_Sv(echodata: EchoData, **kwargs) -> xr.Dataset:
    121     """
    122     Compute volume backscattering strength (Sv) from raw data.
    123 
   (...)
    204     it must be set using `EchoData.update_platform()`.
    205     """
--> 206     return _compute_cal(cal_type="Sv", echodata=echodata, **kwargs)

File ~\anaconda3\envs\Echopype\lib\site-packages\echopype\calibrate\api.py:60, in _compute_cal(cal_type, echodata, env_params, cal_params, ecs_file, waveform_mode, encode_mode)
     58 # Perform calibration
     59 if cal_type == "Sv":
---> 60     cal_ds = cal_obj.compute_Sv()
     61 elif cal_type == "TS":
     62     cal_ds = cal_obj.compute_TS()

File ~\anaconda3\envs\Echopype\lib\site-packages\echopype\calibrate\calibrate_ek.py:628, in CalibrateEK80.compute_Sv(self)
[...]

PR #1219, opened 2023-11-09
Handle missing `receiver_sampling_frequency` in EK80 data [all tests ci]

praneethratna:
Addresses #1217 by adding a check for `receiver_sampling_frequency` not being equal to 0 or else setting it to its default value.

CC @leewujung

### DIFF
--- a/echopype/calibrate/cal_params.py
+++ b/echopype/calibrate/cal_params.py
@@ -396,7 +396,10 @@ def get_cal_params_EK(
     # Private function to get fs
     def _get_fs():
         # If receiver_sampling_frequency recorded, use it
-        if "receiver_sampling_frequency" in vend:
+        if (
+            "receiver_sampling_frequency" in vend
+            and not np.isclose(vend["receiver_sampling_frequency"], 0).all()
+        ):
             return vend["receiver_sampling_frequency"]
         else:
             # If receiver_sampling_frequency not recorded, use default value


leewujung:
@praneethratna : This looks great! I verified the file @icaro-00 and those that we already have in the test data, and confirmed that even if `receiver_sampling_frequency` exists it could be 0, which is a case we haven't seen previously. Before we have only seen files that do not contain this field altogether, from earlier versions of the EK80 software.

Please feel free to merge! Thanks!

leewujung (approved):

leewujung:
BTW, @icaro-00 : Do you have a file that is much smaller that we could include in our test data set (e.g., a few MB)? We typically do not include large files because that lengthens the time of our test runs significantly. Thanks!

leewujung (approved):

icaro-00:
> BTW, @icaro-00 : Do you have a file that is much smaller that we could include in our test data set (e.g., a few MB)? We typically do not include large files because that lengthens the time of our test runs significantly. Thanks!

Hi @leewujung, the smallest find I could find is 300MB. Here is the link: https://wetransfer.com/downloads/1e20766c3f67ddccda9cfbcecb20edec20231116101129/4d5f8f 

Is the errror fixed and available by updating Echopype? Or should I still wait for the 0.8.2 release to be completed 100 percent?

Thank you!

icaro-00:
I actually managed to find another smaller file, 20MB.

Here the link: 
https://wetransfer.com/downloads/f54d5fd680e8d724926cd7849dfe782020231116120052/2d0d2b

leewujung:
Thanks Vogt the files @icaro-00 !

You could try to pip install the dev branch that had the fix included, like shown [here].(https://www.w3docs.com/snippets/python/pip-install-from-git-repo-branch.html)

Our goal is to get v0.8.2 out on Nov 20, so you could also wait for that if it’s not too far away for you.

leewujung:
@praneethratna : I added the 20MB file into the google folder: `ek80/D20230804-T083032.raw`. Please add a test for testing for the stored `receiver_sampling_frequency` value when you have a chance. Thank you!

icaro-00:
> Thanks Vogt the files @icaro-00 !
> 
> You could try to pip install the dev branch that had the fix included, like shown [here](https://www.w3docs.com/snippets/python/pip-install-from-git-repo-branch.html)
> 
> Our goal is to get v0.8.2 out on Nov 20, so you could also wait for that if it’s not too far away for you.

Thank you for the info @leewujung, I will wait for Nov 20, and thanks for helping on this!

Issue #1225, opened 2023-11-15
Overhaul AZFP Environnment group


emiliom:
Currently the AZFP `Environment` group is created with `temperature` and `pressure` variables, even if the variables have no valid data. Since `pressure` data are almost never present, and these variables are not defined by SONAR-netCDF4, always creating these variables is unnecessary and wasteful.

Also, the mandatory variables `absorption_indicative` and `sound_speed_indicative` are not created.

- [ ] Always create `absorption_indicative` and `sound_speed_indicative` variables (float type). In SONAR-netCDF4 they have a `frequency` dimension; here, they will have a `channel` dimension. At this time I don't know where the values would come from, so set to them to `np.nan`
- [ ] Create an associated `frequency_nominal(channel)` variable
- [ ] `temperature` will be created only if `self.parser_obj.unpacked_data["temperature"]` contains valid data
- [ ] `pressure` will be created only if `self.parser_obj.unpacked_data["pressure"]` contains valid data

The attributes for `absorption_indicative` and `sound_speed_indicative` are taken from the convention:

```python
# absorption_indicative
{
    "long_name": "Indicative acoustic absorption",
    "units": "dB/m",
    "valid_min": 0.0,
}
# sound_speed_indicative
{
    "long_name": "Indicative sound speed",
    "standard_name": "speed_of_sound_in_sea_water",
    "units": "m/s",
    "valid_min": 0.0,
}
````

Note that if neither  `temperature` nor `pressure` contain valid data, the variables won't be created, and the `time` dimension and coordinate variable should not be created either.

leewujung:
This can be closed now since #1226 is merged.

PR #1226, opened 2023-11-15
Overhaul AZFP `Environment` group

emiliom:
Fully addresses #1225.

I'm not aware of a source of data to populate the `absorption_indicative` & `sound_speed_indicative` variables, so I'm simply filling them with `np.nan`. @leewujung do you know if there's something in the AZFP raw data & XML that could be used? If there isn't, or it's not straightforward, I suggest leaving the variables in with `np.nan`; after all, the variables were not present before this PR :sweat_smile:

### DIFF
--- a/echopype/convert/set_groups_azfp.py
+++ b/echopype/convert/set_groups_azfp.py
@@ -101,43 +101,89 @@ class SetGroupsAZFP(SetGroupsBase):
 
     def set_env(self) -> xr.Dataset:
         """Set the Environment group."""
-        # TODO Look at why this cannot be encoded without the modifications
-        # @ngkavin: what modification?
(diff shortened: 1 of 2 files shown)


leewujung (approved):
Hey @emiliom : Thanks for the PR. LGTM!

And yeah, `absorption_indicative` and `sound_speed_indicative` need to be calculated for AZFP for calibrating the data.

emiliom:
> And yeah, `absorption_indicative` and `sound_speed_indicative` need to be calculated for AZFP for calibrating the data.

Ah. Maybe in the next release, the calculation of those parameters can be moved to `open_raw`, if appropriate.

I'll merge the PR now. Thanks!

PR #1228, opened 2023-11-16
Fixes build issue with `minioci.dockerfile` [all tests ci]

praneethratna:
Fixes build issue with `minioci.dockerfile` due to `microdnf` command not found.

### DIFF
--- a/.ci_helpers/docker/http.dockerfile
+++ b/.ci_helpers/docker/http.dockerfile
@@ -1,4 +1,7 @@
 FROM httpd:2.4
 ARG TARGETPLATFORM
 
+RUN apt update && apt-get update
+RUN apt install -y git
+
 COPY echopype/test_data /usr/local/apache2/htdocs/data
--- a/.ci_helpers/docker/http.dockerfile
+++ b/.ci_helpers/docker/http.dockerfile
@@ -1,7 +1,4 @@
 FROM httpd:2.4
 ARG TARGETPLATFORM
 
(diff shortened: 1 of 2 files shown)


leewujung:
Hey @praneethratna : Thanks for looking into this. There are other more recent releases and one with the `latest` tag too. It seems better to not pin on a specific version, unless there are very good reasons to do so. Do you know what the core issues were for the previous failing, and is there a more general fix that we could use?

praneethratna:
> Hey @praneethratna : Thanks for looking into this. There are other more recent releases and one with the `latest` tag too. It seems better to not pin on a specific version, unless there are very good reasons to do so. Do you know what the core issues were for the previous failing, and is there a more general fix that we could use?

Hi @leewujung I tried to build using `latest` tag but that also failed due to similar error. The core issue being `microdnf` command was not found which can be due to `microdnf` not being used post Nov 1st release, i'm not very sure of it. 
I have a workaround for this which is - 
```FROM minio/minio AS base
FROM debian AS builder
ARG TARGETPLATFORM

RUN apt-get update && apt-get install -y git
FROM minio/minio
COPY --from=builder /usr/bin/git /usr/bin/git

# Your other commands
CMD ["minio", "server", "/data"]
``` 
What do you think about this?

praneethratna:
Closing and re-opening the PR to rerun all the tests!

leewujung (approved):

Issue #1224, opened 2023-11-13
Allow multi-channel mask for `apply_mask` and simplify allowed `fill_value`


leewujung:
## Overview

Currently `mask.apply_mask` only accepts masks (a single mask or a list of masks) that do not have a `channel` dimension, and this mask will be applied to all channels. However, it would be more flexible for `mask.apply_mask` to be able to also accept masks that contain a channel dimension. For example for noise masks, it is likely that the mask for each channel would be different.

This does create the potentially problematic case where the `source_ds` and elements of the list of masks may have different channels. For example:
1. `mask=[mask1, mask2]`, where `mask1` has `channel=["chA"]` and `mask2` has `channel=["chB"]`
2. `source_ds` have `channel=["chA", "chB", "chC"]` (i.e., the combined masks would only have 2 of the 3 channels), so we need to decide what to do with `chC`

To handle the above:
1. Looks like doing [`xr.broadcast`](https://docs.xarray.dev/en/stable/generated/xarray.broadcast.html) before multiplying elements in the list of masks together would work (i.e., in the example above, the combined mask would have `channel=["chA", "chB"]`)
2. It seems more intuitive for channels that exist in `source_ds` but are missing in `mask` be left as is -- i.e. the mask would be all 1s for these channels. But perhaps it is the best to leave this as an option.

Also, looking at the code again, allowing `fill_value` to be of type `np.ndarray` would be really messy if we allow multi-channel masks, because there will be no way to match up which channel is which. I think we should just remove this and do `fill_value: Union[int, float, xr.DataArray] = np.nan`.

## Tasks
If we take the above approach, the tasks are:
- [ ] Remove `np.ndarray` from allowed type for `fill_value`
- [ ] If `fill_value` is of type `xr.DataArray`, it must have the same dimension as `source_ds`
- [ ] Check for the channel dimension for all elements in `mask` to ensure that they have the same shape in `ping_time` and `range_sample` (and `depth`, see #1204) -- this check should already be in place, but the code could use some checking and potentially simplifcation
- [ ] Make sure to broadcast all input masks together before combining them (element-wise multiplication)
- [ ] Add an input argument `keep_unmasked_channel: bool = True`
  - when `True`: channels that are not in `mask` will be left as is, i.e. no masking will be applied to the data in `chC`
  - when `False`: channels that are not in `mask` will be masked, i.e. all data in `chC` will be masked with values in `fill_value`.

[...]

Issue #1204, opened 2023-10-28
Add `depth` to allowed dimension in `apply_mask`


leewujung:
Right now `mask.apply_mask` only allows datasets with `range_sample` as one of the dimensions along with `ping_time` and `channel` (optional). However outputs of `compute_MVBS` for example would have dimension `depth` in place of `range_sample` in the along-range dimension. We need to add `depth` into the allowed dimension in `apply_mask`.

ctuguinay:
Closed with the merging of https://github.com/OSOceanAcoustics/echopype/pull/1230

PR #1230, opened 2023-11-22
Apply Mask Changes: Multichannel, Allow Depth, Simplify Fill Value

ctuguinay:
This PR is meant to address #1204 and #1224.

### DIFF
--- a/echopype/mask/api.py
+++ b/echopype/mask/api.py
@@ -269,14 +269,17 @@ def apply_mask(
 
     The code allows for these 3 cases of `source_ds` and `mask` dimensions:
 
-    1) No channel in both source ds and mask, but they have matching ping time and depth
-    dimensions.
-    2) Source ds and mask both have matching channel, ping time, and depth dimensions.
-    3) Source ds has channel dimension and mask doesn't, but they have matching ping
-    time and depth dimensions.
+    1) No channel in both `source_ds` and `mask`,
+    but they have matching `ping_time` and
+    `depth` (or `range_sample`) dimensions.
+    2) `source_ds` and `mask` both have matching `channel`,
+    `ping_time`, and `depth` (or `range_sample`) dimensions.
+    3) `source_ds` has the channel dimension and `mask` doesn't,
+    but they have matching
+    `ping_time` and `depth` (or `range_sample`) dimensions.
 
-    If a user only wants to apply masks to a subset of the channels in source ds,
-    they could put 1s/0s for all data in the channels that are not masked.
+    If a user only wants to apply masks to a subset of the channels in `source_ds`,
+    they could put 1s to allow all data entries in the other channels.
 
     Parameters
     ----------
--- a/echopype/mask/api.py
+++ b/echopype/mask/api.py
@@ -317,26 +317,30 @@ def apply_mask(
 
     # Obtain final mask to be applied to var_name
     if isinstance(mask, list):
-        # perform a logical AND element-wise operation across the masks
-        final_mask = np.logical_and.reduce(mask)
+        # Broadcast all input masks together before combining them (element-wise multiplication)
+        broadcasted_masks = xr.broadcast(*mask)
+
+        # Perform a logical AND element-wise operation across the masks
+        final_mask = np.logical_and.reduce(broadcasted_masks)
 
         # xr.where has issues with attrs when final_mask is an array, so we make it a DataArray
-        final_mask = xr.DataArray(final_mask, coords=mask[0].coords)
+        final_mask = xr.DataArray(final_mask, coords=broadcasted_masks[0].coords)
     else:
         final_mask = mask
 
-    # Sanity check: final_mask should be of the same shape as source_ds[var_name]
-    #               along the ping_time and range_sample dimensions
-    def get_ch_shape(da):
-        return da.isel(channel=0).shape if "channel" in da.dims else da.shape
-
-    # Below operate on the actual data array to be masked
+    # Operate on the actual data array to be masked
     source_da = source_ds[var_name]
 
-    source_da_shape = get_ch_shape(source_da)
-    final_mask_shape = get_ch_shape(final_mask)
+    # Sanity check: final_mask should be of the same shape as source_ds[var_name]
+    # along the ping_time and range_sample dimensions.
+    source_da_chan_shape = (
+        source_da.isel(channel=0).shape if "channel" in source_da.dims else source_da.shape
+    )
+    final_mask_chan_shape = (
+        final_mask.isel(channel=0).shape if "channel" in final_mask.dims else final_mask.shape
+    )
 
-    if final_mask_shape != source_da_shape:
+    if final_mask_chan_shape != source_da_chan_shape:
         raise ValueError(
             f"The final constructed mask is not of the same shape as source_ds[{var_name}] "
             "along the ping_time and range_sample dimensions!"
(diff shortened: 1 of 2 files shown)


ctuguinay:
Remove np.ndarray from allowed type for fill_value

- [ ]  If fill_value is of type xr.DataArray, it must have the same dimension as source_ds

- [ ]  Check for the channel dimension for all elements in mask to ensure that they have the same shape in ping_time and range_sample (and depth, see https://github.com/OSOceanAcoustics/echopype/issues/1204) -- this check should already be in place, but the code could use some checking and potentially simplifcation

- [ ] Make sure to broadcast all input masks together before combining them (element-wise multiplication)

- [ ]  Add an input argument keep_unmasked_channel: bool = True
    when True: channels that are not in mask will be left as is, i.e. no masking will be applied to the data in chC
    when False: channels that are not in mask will be masked, i.e. all data in chC will be masked with values in fill_value.

echopype/mask/api.py
```diff
@@ -262,23 +274,28 @@ def apply_mask(
         Points to a Dataset that contains the variable the mask should be applied to
     mask: xr.DataArray, str, pathlib.Path, or a list of these datatypes
         The mask(s) to be applied.
-        Can be a single input or list that corresponds to a DataArray or a path.
-        Each entry in the list must have dimensions ``('ping_time', 'range_sample')``.
-        Multi-channel masks are not currently supported.
+        Can be a individual input or list that corresponds to a DataArray or a path.
```
leewujung, line 277: small fix
```suggestion
        Can be a individual input or a list that corresponds to a DataArray or a path.
```

echopype/mask/api.py
```diff
@@ -262,23 +274,28 @@ def apply_mask(
...
         mask that will be applied to ``var_name``.
     var_name: str, default="Sv"
         The Sv variable name in ``source_ds`` that the mask should be applied to.
-        This variable needs to have coordinates ``ping_time`` and ``range_sample``,
-        and can optionally also have coordinate ``channel``.
+        This variable needs to have coordinates ``('ping_time', 'range_sample')`` or
+        coordinates ``('ping_time', 'depth')``, and can optionally also have coordinate
+        ``channel``.
         In the case of a multi-channel Sv data variable, the ``mask`` will be broadcast
         to all channels.
```
leewujung, line 291: I can see how the original implementation is trying to get to _all_ possible cases, including when the Sv data variable does not have a `channel` dimension/coordinate. This is not the case for echopype-generated Sv though, and I think we could require that, and also require that for mask(s) that have a `channel` dimension, the length of all dimensions should match between the masks and the Sv data variable -- i.e. we expect that only the relevant channels are provided as input arguments, say via slicing. I'll expand on this in the overall comment too.

leewujung (commented):
Thanks for the changes @ctuguinay! The core changes for expanding this function to allow data variables with the `depth` coordination (in addition to `range_sample`) looks good and to simplify fill value specification look good.

I have a more overarching comment about the what we expect in the input arguments. Right now the code is set up to be super flexible and catch as much cases as possible if there are mismatches between `source_ds` and `mask_list`. I think that we can reduce the code complexity significantly if we impose some expectations on the input arguments to be responsibility of the users. This is also because I think the current code does not actually catch all the things that could go wrong in the mismatches, so it may be better to not try to accommodate some scenarios but not others.

My suggestions are the following:
- If the mask(s) in `mask_list` have the `channel` dimension, we require that `source_ds` and all elements in `mask_list` have the same `channel` coordinates and the shapes are the same (across `channel`, `ping_time`, and `range_sample` or `depth`). 
- If the mask(s) in `mask_list` do not have the `channel` dimension, we require that `source_ds` and all elements in `mask_list` have the same shape (across `ping_time` and `range_sample` or `depth`). 
- The above requires users to slice their masks and/or Sv dataset at the input.
- This will remove the need for the `keep_unmasked_channel` argument. I think allowing this variable would potentially make the data processing pipeline very messy, since the shape of the dataset could change somewhat unpredictably depending on this argument `keep_unmasked_channel`. If people only want to apply masks to a subset of the channels, they could put 1s for all data in the channels that are not masked.

Let me know what you think.

ctuguinay:
Thanks for the review @leewujung!

Yeah, I agree with you on there being too many cases that this code tries to accommodate. I also agree with you on all suggestions you posed.

Just to get the cases in order, once we get to the post-broadcast post-logical-reduce stage (once we get to here https://github.com/ctuguinay/echopype/blob/e52e05f33a82d1700d24f5c58d072fe15d319136/echopype/mask/api.py#L335), I should pass only the following:
- No channel in both source and mask, but they have matching ping time and depth
- Source and mask both have matching channel, ping time, and depth
- Source has channel and mask doesn't, but they have matching ping time and depth

ctuguinay:
@leewujung This PR should be ready to review again

echopype/mask/api.py
```diff
@@ -260,21 +260,31 @@ def apply_mask(
...
+
+    1) No channel in both source ds and mask, but they have matching ping time and depth
+    dimensions.
+    2) Source ds and mask both have matching channel, ping time, and depth dimensions.
+    3) Source ds has channel dimension and mask doesn't, but they have matching ping
+    time and depth dimensions.
+
+    If a user only wants to apply masks to a subset of the channels in source ds,
+    they could put 1s/0s for all data in the channels that are not masked.
+
```
leewujung, line 280:
```suggestion
    The code allows for these 3 cases of `source_ds` and `mask` dimensions:

    1) No channel in both `source_ds` and `mask`, but they have matching ping time and depth
    dimensions.
    2) `source_ds` and `mask` both have matching channel, ping time, and depth dimensions.
    3) `source_ds` has channel dimension and `mask` doesn't, but they have matching ping
    time and depth dimensions.

    If a user only wants to apply masks to a subset of the channels in `source_ds`,
    they could put 1s to allow all data entries in the other channels.

```

echopype/mask/api.py
```diff
@@ -347,44 +354,25 @@ def apply_mask(
...
-    # Apply the mask to var_name
-    if "channel" in final_mask.dims and "channel" in source_da.dims:
-        # Identify common channels
-        common_channels = set(final_mask.coords["channel"].values).intersection(
-            set(source_da.coords["channel"].values)
+    # If final_mask has dim channel then source_da must have dim channel
+    if "channel" in final_mask.dims and "channel" not in source_da.dims:
+        raise ValueError(
+            "The final constructed mask has dim channel, "
+            f"so source_ds[{var_name}] must have dim channel."
```
leewujung, line 363: Good to write error messages with clear text so users can have clear guidance. Things that are understandable to us may not be necessarily understandable widely.
```suggestion
            "The final constructed mask has the channel dimension, "
            f"so source_ds[{var_name}] must also have the channel dimension."
```

echopype/mask/api.py
```diff
@@ -347,44 +354,25 @@ def apply_mask(
...
         )
+    # If final_mask and source_da both have channel dimension, then they must
+    # have the same number of channels.
+    elif "channel" in final_mask.dims and "channel" in source_da.dims:
+        if len(final_mask["channel"]) != len(source_da["channel"]):
+            raise ValueError(
+                f"If both the final constructed mask and source_ds[{var_name}] "
+                "have channel then they must have matching channel dimensions."
```
leewujung, line 371:
```suggestion
                f"If both the final constructed mask and source_ds[{var_name}] "
                "have the channel dimension, that dimension should match between the two."
```

leewujung (approved):
The changes look good. I suggested some small edits for the docstring and error messages. It is important to be as clear as possible about these things. I think this is ready to be merged -- thanks for the changes!

ctuguinay:
@leewujung I just committed your suggestions, and this should be ready to merge. Thank you for making the docstring more precise! I think I wrote that right after I got tests working so I was at the point of complete familiarity, and I didn't stop to think whether or not someone new could understand it. I will work on that 👍

leewujung:
Thanks @ctuguinay , I'll merge this now! 🎉

leewujung:
Oops, I just realized that you target `main` and not `dev`, and I used squash and merge. I may have to do hard reset and ask you to another PR...

leewujung:
Looks like there's no compounding changes from other PRs that have been merged to `dev`, so this should be ok. We want to simplify our dev workflow anyway #1275 so this forces us to do it, ha.

PR #1233, opened 2023-11-23
Add test for ES60 spare field decoding issue

praneethratna:
Added test case to validate the fix for #1195 which was fixed in #1215 .

CC @leewujung


leewujung (approved):
@praneethratna : I verified that the test did run and all looks great! Thanks, I'll merge this now!

Issue #1239, opened 2023-12-04
[File conversion] No parser available for type MRU version 1


jmjech:
### General description of problem
_Required_
I have EK80 files collected within the past month using a Kongsberg WBT Tube that give me an error when I attempt to convert from .raw to .nc4. Data were collected on the uncrewed surface vehicle, Drix.

### Computing environment
_Required_
- Echopype version: 0.8.2 (installed on about 1 December 2023, but I did not uninstall a previous version)
- How echopype was installed (e.g., conda or pip): pip
- Operating system: Ubuntu 22.04.3 LTS

### Minimum example
_Required: A minimum code snippet that reproduces your problem._
The following code reproduces the errors I encountered:
```python
import echopype as ep
from echopype import open_raw
from pathlib import Path
filename = Path('/home/mjech/Desktop/WindDevelopment/EK80_Data/2023120_BIWF_WTG-3_Spiral-D20231021-T003735.raw')
filelist = [filename]
EKmodel = 'EK80'   #EK60
for f in filelist:
    ed = open_raw(str(f), sonar_model=EKmodel)

```

### Error message printouts
_Required: Copy-paste the entire error messages you encounter here._
Below is the error messages I received when running the above code:
```python
In [3]: %run Convert_EK80-to-netCDF.py
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/NOAA_Gdrive/sonarpros/Python_Programs/EK_ES/Convert_EK80-to-netCDF.py in <module>
     93 # convert each to netCDF4
     94 for f in filelist:
---> 95     ed = open_raw(str(f), sonar_model=EKmodel)
     96     # Henry B. Bigelow ICES code is 33HH
     97     ed['Platform']['platform_name'] = 'Drix18'

~/.local/lib/python3.10/site-packages/echopype/utils/prov.py in inner(*args, **kwargs)
    235             @functools.wraps(func)
    236             def inner(*args, **kwargs):
--> 237                 dataobj = func(*args, **kwargs)
    238                 if is_echodata:
    239                     ed = dataobj

~/.local/lib/python3.10/site-packages/echopype/convert/api.py in open_raw(raw_file, sonar_model, xml_path, convert_params, storage_options, use_swap, max_chunk_size)
    421     )
    422     # Actually parse the raw datagrams from source file
--> 423     parser.parse_raw()
    424 
    425     # Direct offload to zarr and rectangularization only available for some sonar models

[...]

PR #1242, opened 2023-12-10
Added parser support for EK80 MRU1

praneethratna:
Addresses #1239 and now EK80 files of type MRU version 1 can be parsed.

CC @leewujung

### DIFF
--- a/echopype/convert/utils/ek_raw_parsers.py
+++ b/echopype/convert/utils/ek_raw_parsers.py
@@ -502,6 +502,16 @@ class SimradMRUParser(_SimradDatagramParser):
         pitch:        float
         heading:      float
 
+        type:         string == 'MRU1'
+        low_date:     long uint representing LSBytes of 64bit NT date
+        high_date:    long uint representing MSBytes of 64bit NT date
+        timestamp:    datetime.datetime object of NT date, assumed to be UTC
+        heave:        float
+        roll :        float
+        pitch:        float
+        heading:      float
+        length:       long uint
+
     The following methods are defined:
 
         from_string(str):    parse a raw ER60 NMEA datagram
@@ -522,7 +532,17 @@ class SimradMRUParser(_SimradDatagramParser):
                 ("roll", "f"),
                 ("pitch", "f"),
                 ("heading", "f"),
-            ]
+            ],
+            1: [
+                ("type", "4s"),
+                ("low_date", "L"),
+                ("high_date", "L"),
+                ("heave", "f"),
+                ("roll", "f"),
+                ("pitch", "f"),
+                ("heading", "f"),
+                ("length", "L"),
+            ],
         }
 
         _SimradDatagramParser.__init__(self, "MRU", headers)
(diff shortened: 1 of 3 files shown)


echopype/convert/utils/ek_raw_parsers.py
```diff
@@ -522,7 +522,17 @@ def __init__(self):
                 ("roll", "f"),
                 ("pitch", "f"),
                 ("heading", "f"),
-            ]
+            ],
+            1: [
```
leewujung, line 526: Could you add in the docstring regarding MRU1 type in addition to MRU0?

leewujung (commented):
@praneethratna : Thanks for putting this in! I just have a minor comment about adding MRU1 in the docstring. I verified that the rest of `open_raw` runs through fine with the pitch, roll, heave data stored in the Platform group as pitch, roll, and _vertical_offset_.

But the parsed `heading` variable is not stored. Let me take a look at the convention to see if there's a variable we can store it, and if not, we can put it in the Platform group just with `heading` as the name. I will get back to you on this!

It'll be nice to have a test file in the CI for this.

@jmjech: Do you have a file that is smaller (say a few MB) that we can use as a test file? It adds to the time for tests to run through if the test files are large, so we usually try to include only small files.

Thanks both!

jmjech:
I e-mailed a 2 MB file to Wu-Jung. Let me know if it works for you. I have other "smaller" files.
For some reason it is not allowing me to add the data files here.

leewujung:
@jmjech: Thank you! The file works and does contain MRU1 datagrams.

@praneethratna: I have added this file (`20231016_Cal_-D20231016-T220322.raw`) under `test_data/ek80`. Could you please add a test for parsing it? In looking into this file I also checked for `np.all(echodata["Platform"]["pitch"].data == np.array(parser.mru["pitch"]))` and the same thing for `heave` (compare with `vertical_offset` in Platform) and `roll` (compare with `roll` in Platform). Would be good if we have these in the tests, by using the parser object https://github.com/OSOceanAcoustics/echopype/blob/03e3b52bb522222f5cdf887b6f519e47d4aae6bf/echopype/convert/api.py#L423

leewujung:
Also, on storing the `heading` variable. It is indeed in the ver.1 convention as below:

Screenshot 2023-12-14 at 11 43 23 AM

@praneethratna: Could you please add this to the set group so that it gets stored together with the other variables? As you have seen, they are on the same timebase since they are all from the MRU* datagrams. Thanks!

echopype/convert/utils/ek_raw_parsers.py
praneethratna: Updated!

praneethratna:
@leewujung I have added docstring for new `MRU1` datagram and also added a new test case using the file given by @jmjech. Also, now the `heading` variable will be stored in `Platform` group with the attributes which you've mentioned. Thanks!

leewujung (approved):
@praneethratna : Thanks for adding `heading` and test. I'll merge this now. :)

PR #1252, opened 2024-01-02
feat: add new parser for parsing BI500 files

### LINKED ISSUE #1258: BI500 `open_raw` outputs
This is from discussions with @praneethratna and @jmjech for adding support for BI500 data. For context, after BI500 support is added we could move forward to the closely related EK500 data. 

Different from the raw data contained in EK60 and EK80, BI500 files contain nominally "calibrated" Sv data, which came from applying a suite of parameters (through a parameter file) that users put in to configure EK500, and the echosounder applies these parameters to the raw data to output BI500 files. 

This creates a situation where the parsed BI500 data are not "raw" in itself and are more equivalent to the Sv datasets generated from `compute_Sv`. In theory we could leverage the SONAR-netCDF4 [v2.0 generalization](https://nbviewer.org/github/ices-publications/SONAR-netCDF4/blob/v2.0/Formatted_docs/crr341.pdf#page=70) to store the nominally calibrated Sv data into the `backscatter_r` variable, but this may cause confusion with other echosounders, for which `backscatter_r` contains uncalibrated data.

For the first shot we can do the following:
```python
echodata, ds_cal = open_raw(raw_file="BI500_FILEPATH", ..., metadata_file="METADATA. ekprm")
```
The key things are:
- Generate 2 sets of outputs from BI500 data: `echodata` and `ds_cal` (instead of only the single `echodata` output for all other sonar models): 
    - `echodata` contain all raw data variables and all user-input config parameters from the parameter file that are used in calibration
    -  `ds_cal` contains the Sv data recorded in the BI500 files, and all user-input config parameters from the parameter file that are used in calibration. These are what are in Sv datasets generated by `compute_Sv`

- Change the previous `xml_path` (which was only used for AZFP data to read the XML file associated with the 01A file) to `metadata_file` (which can then be used to pass in both AZFP XML file or the BI500 parameter file)

- _Not_ part of this issue: #1259

@praneethratna and @jmjech: Please feel free to point out anything incorrect and other things I may have missed!

praneethratna:
This PR adds support for parsing BI500 files.
CC @leewujung @jmjech


jmjech:
Hi @praneethratna & @leewujung:
Thanks for the initial start on reading BI500 data! Great start.

I copied your parse_bi500.py and core.py to my ~/echopype/convert/ and ~/echopype directories and ran my code:
```
import echopype as ep
from echopype import open_raw
from pathlib import Path

filename = Path('/home/mjech/NOAA_Gdrive/sonarpros/IDL_Programs/testdata/singlefile/N031-S445-S2000008-F011990-T01-D20000913-T103336-Data')
EKmodel = 'BI500' 
ed = open_raw(str(filename), sonar_model=EKmodel)

```
You'll notice that I use pathlib's Path for my files. I like it, but to get the filename into echopype, I need to cast it as a string. The reason I mention this is that I ran into some errors based on the file name in parse_bi500.py code and core.py. I think they came up because the BI500 files do not have a "true" suffix, which seems necessary for some of the initial file name and file path checking that is done in parse_bi500 and maybe `ParseBase`.

Here is what I did to get around these errors:
1. In core.py, I modified the line in the BI500 sonar model dict entry `"validate_ext": validate_ext(''),` rather than the `None` type that was used. You'll notice I use single quotes rather than double quotes so you would have `.validate_ext("")`. Anyway, that prevented an error that "None" types were not valid.
2. In parse_bi500.py I needed to
a. insert `from pathlib import Path`
b. reorganize the self declarations in the __init__ section to:
```
        self.file_types = FILE_TYPES
        self.timestamp_pattern = FILENAME_DATETIME_BI500
        self.file_type_map = defaultdict(None)

        self.parameters = defaultdict(list)
        self.ping_counts = defaultdict(list)
        self.vlog_counts = defaultdict(list)
        self.index_counts = defaultdict(list)
        self.unpacked_data = defaultdict(list)
        self.sonar_type = "BI500"

        self.fsmap = self._validate_folder_path(file)
        self.index_file = self._get_index_file(self.fsmap)

```
You'll notice that I needed to set all the different parameters before going to the "file" lines (the last two lines) because the parameters didn't get set and those "file" functions needed them set.

In addition, in the `_validate_folder_path(self, folder_path)` function, I used Path to get the parent folder:
```
   def _validate_folder_path(self, folder_path):
        """Validate the folder path."""
        folder_path = str(Path(folder_path).parent)
        fsmap = fsspec.get_mapper(folder_path, **self.storage_options)
        try:
            all_files = fsmap.fs.ls(folder_path)
        except NotADirectoryError:
            raise ValueError(
                "Expecting a folder containing at least '-Data' and '-Info' files, "
                f"but got {folder_path}"
            )

```
I seems I needed to do this because the BI500 files don't have a suffix.

I've gotten over the initial humps, but now I get some other errors that may be better for you to look into. Here are the error messages:
error                                     Traceback (most recent call last)
~/NOAA_Gdrive/sonarpros/Python_Programs/EK_ES/test_BI500.py in <module>
     15 filename = Path('/home/mjech/NOAA_Gdrive/sonarpros/IDL_Programs/testdata/singlefile/N031-S445-S2000008-F011990-T01-D20000913-T103336-Data')
     16 EKmodel = 'BI500'
---> 17 ed = open_raw(str(filename), sonar_model=EKmodel)
     18 
     19 '''

~/.local/lib/python3.10/site-packages/echopype/utils/prov.py in inner(*args, **kwargs)
    235             @functools.wraps(func)
    236             def inner(*args, **kwargs):
--> 237                 dataobj = func(*args, **kwargs)
    238                 if is_echodata:
    239                     ed = dataobj

~/.local/lib/python3.10/site-packages/echopype/convert/api.py in open_raw(raw_file, sonar_model, xml_path, convert_params, storage_options, use_swap, max_chunk_size)
    421     )
    422     # Actually parse the raw datagrams from source file
--> 423     parser.parse_raw()
    424 
    425     # Direct offload to zarr and rectangularization only available for some sonar models

~/.local/lib/python3.10/site-packages/echopype/convert/parse_bi500.py in parse_raw(self)
    240                 self.unpacked_data["pelagic"].append(unpacked_data[:PELAGIC_COUNT])
    241                 self.unpacked_data["bottom"].append(
--> 242                     unpacked_data[PELAGIC_COUNT : PELAGIC_COUNT + BOTTOM_COUNT]
    243                 )
    244                 for trace_num in range(TRACE_COUNT):

error: unpack requires a buffer of 1324 bytes

These errors seem to be associated with the actual reading and parsing of the data file.

If this is unclear, I can send my revised parse_bi500.py and core.py code.

Thanks!
mike

praneethratna:
Hey @jmjech Thanks for testing out the code locally on sample data and informing me about the errors. 
1. The reason i have used `None` earlier since as mentioned by you the BI500 files don't have a true suffix and it isn't necessar for a check. I have changed it to `validate_ext("")` as suggested and it works fine now.
2. I have also re-arranged the lines `__init__` to solve the errors regarding the initialisation.
3. The error caused after that is due to offset and count values being used from `-Vlog` instead of `-Ping` while unpacking `-Data` file and also due a mistake in `START_FORMAT` value which i have rectified. 
You can now pull the latest code changes and everything should work fine on the parser part. Since, we don't have `set_groups_bi500.py` setup yet we can test the parser as follows:
```
>>> from echopype.convert.parse_bi500 import ParseBI500
>>> parser = ParseBI500(str('/home/praneeth/echopype/test_ek500'), None)
>>> parser.parse_raw()
```
where `test_ek500` is a folder containing `-Data`, `-Ping`, `-Vlog`, `-Info`, `-Work`, `-Snap` corresponding to prefix `N031-S445-S2000008-F011990-T01-D20000913-T103336`. You can access the parsed data of `-Data` files using `parser.unpacked_data` and rest of the files using `parser.parameters`.

jmjech:
Hi @praneethratna. I successfully imported BI500 data and was able to produce a couple of echograms! One thing that needs to be done is convert the values that have been read from the data file to dB. You do that by multiplying by 10*log_base_10(2)/256.

leewujung:
Hey @praneethratna : Thanks for getting back to this PR! I tested the `open_raw` function using your test example and also looked into the variables, and also discussed it with @LOCEANlloydizard (I think he's looking into your code in more detail as well).

After thinking for some time, I still came back to the same conclusion as I had from before in https://github.com/echostack-org/echopype/issues/1258#issue-2120710747, that we would be better off to keep the use the same `open_raw` call but output 2 datasets. I would still suggest:
```python
echodata, ds_cal = open_raw(
    raw_file="BI500_FILEPATH",
    sonar_model="EK500",
    metadata_file="METADATA. ekprm"  # see last section
)
```
in [this google spreadsheet](https://docs.google.com/spreadsheets/d/11AhRzHj-vymT6telZEYd0zmH7iHfhNybEdbnJk2VOvU/edit?gid=352240964#gid=352240964) under the column "object name" the "EchoData" corresponds to `echodata` above, and "ds_calibrated" corresponds to `ds_cal` above. 

Essentially my thought boils down to keeping the EchoData object as the container for holding various parameters and metadata type of information from both the BI500 -* series of files as well as the `.ekprm` file, and `ds_cal` holds calibrated quantities BI500 generates directly, including Sv, Sp (uncompensated TS), and TS (compensated TS).

The alternative is to output 4 objects 
```python
echodata, ds_Sv, ds_Sp, ds_TS = open_raw(
    raw_file="BI500_FILEPATH",
    sonar_model="EK500",
    metadata_file="METADATA. ekprm"  # see last section
)
```
This would be fine too, but seems too messy?

-----------------------------
### Variable/dimension/coordinate names

@LOCEANlloydizard @praneethratna One linkage I wanted to make sure we make is with @LOCEANlloydizard's PR https://github.com/echostack-org/echopype/pull/1588, to make sure that we use the same variable, dimension, and coordinate name for the TS and Sp datasets from single target detection. The sonar-netCDF4 convention v2 does specify [a single_target group](https://nbviewer.org/github/ices-publications/SONAR-netCDF4/blob/v2.0/Formatted_docs/crr341.pdf#page=43). In general I don't agree with putting processed data in the some object as raw data, but since there's an existing definition that's quite similar to what we have as outputs, it would be a good idea to just follow it.

How about in both here and #1588 (so outputs from `compute_Sp` and `compute_TS`), we use the definitions from the above link as much as we can?

-----------------------------

### `.ekprm` file
I'll write up a separate issue for the `.ekprm` file that contains BI500 settings (such as single target detection threshold etc), since it's more a question for @jmjech who has actually operated this instruments (and who provided these test files).

LOCEANlloydizard:
**UPDATED** 

Thx @leewujung for the description and infos!
@praneethratna I updated the BI500 implementation to follow the structure (as I have the other PR open also..):

- open_raw() now returns (echodata, ds_cal);
- echodata keeps the metadata and ancillary BI500 information
- ds_cal contains the calibrated BI500 products (Sv, Sv_bottom, echo_range, echo_range_bottom) together with the stored single-target detections (uncompensated_TS, compensated_TS, single_target_range, split-beam angles, etc.), following the existing echopype calibrated output structure and SONAR-netCDF naming as closely as possible
- updated the single-target variables, dimensions, and coordinates to follow the SONAR-netCDF naming convention as closely as possible (single_target, single_target_range, single_target_alongship_angle, single_target_athwartship_angle, uncompensated_TS, compensated_TS, etc.), consistent with the work in #1588
- added test

If you have time could you have a look and let me know what you think? It would also be great to add a few more tests if you see any good cases to cover? I'll have another look over the weekend or early next week, but we can also chat on Slack if that's easier, cheers!

LOCEANlloydizard:
Update: I added a BI500 example notebook to showcase the current workflow (PR: https://github.com/echostack-org/echopype-examples/pull/113)

also added echo_range and echo_range_bottom (with range_sample and range_sample_bottom as the sample-index dimensions) so that the BI500 calibrated outputs follow the same structure as compute_Sv()

I just wanted to confirm the interpretation of the BI500 fields PelagicUpper, PelagicLower, BottomUpper, and BottomLower:

- do PelagicUpper/PelagicLower represent range from the transducer or absolute depth?
- do they define the bin edges or the centres of the first and last samples?

cheers!

praneethratna:
> * do PelagicUpper/PelagicLower represent range from the transducer or absolute depth?

They represent the absolute depth below the sea surface and are independent of the transducer depth.

> * do they define the bin edges or the centres of the first and last samples?

They define the bin edges.

leewujung:
To make sure we have all info easily accessible:

Here are some [EK500/BI500 documentations](https://drive.google.com/drive/folders/18k7uUVqHS_XCHSKp_vD5Bq8YEjWnXGsU) from @jmjech back in 2023 that has the definitions for all variables.

In the same [folder](https://drive.google.com/drive/folders/1XS6gDm2BYrk8bVWJDnbKg2xd2OH6x88D) we have the [spreadsheet](https://docs.google.com/spreadsheets/d/11AhRzHj-vymT6telZEYd0zmH7iHfhNybEdbnJk2VOvU/edit?usp=sharing) and [example data](https://drive.google.com/drive/folders/1tQ7IN4G0fNFuOF0hx7HTgVNQoglNZ7bW).

LOCEANlloydizard:
oh right, yes thank you!

LOCEANlloydizard (approved):
hey @praneethratna @leewujung, I updated the BI500 outputs to use depth / depth_bottom following the clarification that the BI500 bounds are absolute depth bin edges!

I also:

- added multi-channel acquisition handling
- kept the calibrated Sv and single-target products in ds_cal
- left beam_type unset for now, since we don't have a reliable source for it in the BI500 files (??)
- added/expanded tests for the calibrated outputs, depth, single targets, and multi-channel grouping

I think we should be good to merge this as a first stable implementation and build on it from there. In particular, .ekprm handling can stay for the separate follow-up issue discussed above, and we can add beam_type later once we know how to populate it?

I'm also doing a small re-read of the example notebook right now, but I think it shows the workflow well as a [BI500 getting-started example](https://github.com/echostack-org/echopype-examples/pull/113), including the multi-channel output and downstream processing! let me know what you think, cheers!

Issue #1268, opened 2024-02-29
Add `func` argument to `compute_MVBS`


ctuguinay:
The addition of the `func` argument will enable users to specify the aggregation function used by the underlying `flox.xarray.xarray_reduce`. This `func` parameter determines how the aggregation is performed on the Sv data array. By default, this function is set to `nanmean`, but users can opt to use other aggregate functions like `mean` in the case where they want to NaN any MVBS bin that 'contains' NaN values in the original Sv grid.

One can find all available options for func listed here: https://flox.readthedocs.io/en/latest/generated/flox.xarray.xarray_reduce.html.

leewujung:
We close this now with #1269 merged. Also that since compute_MVBS is supposed to be a mean or nanmean function, we don't need to include the flexibility of allowing different `func`. But the intermediate function under the hood does allow that and we'll likely use them for other processing functions later.

PR #1269, opened 2024-02-29
expose func arg and skip_na to compute_MVBS users

ctuguinay:
PR for #1268

In addition, modified `test_compute_MVBS_values` to include a test for adding NaNs to the `ds_Sv` to illustrate how that would affect the computation of `compute_MVBS` using both `func="nanmean"` and `func="mean"`.

### DIFF
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -64,8 +64,8 @@ def compute_MVBS(
         See flox `documentation <https://flox.readthedocs.io/en/latest/implementation.html>`_
         for more details.
     skipna: bool, default True
-        If true, mean function skips NaN values.
-        Else, mean function includes NaN values.
+        If true, the mean operation skips NaN values.
+        Else, the mean operation includes NaN values.
     closed: {'left', 'right'}, default 'left'
         Which side of bin interval is closed.
     **flox_kwargs
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -34,6 +34,7 @@ def compute_MVBS(
     ping_time_bin: str = "20S",
     method="map-reduce",
     func="nanmean",
+    skipna=True,
     closed: Literal["left", "right"] = "left",
     **flox_kwargs,
 ):
@@ -67,6 +68,11 @@ def compute_MVBS(
         The flox aggregation function used for reducing the data array.
         By default, 'nanmean' is used. Other options can be found in the flox `documentation
         <https://flox.readthedocs.io/en/latest/generated/flox.xarray.xarray_reduce.html>`_.
+    skipna: bool, default True
+        If true, aggregation function skips NaN values.
+        Else, aggregation function includes NaN values.
+        Note that if ``func`` is set to 'mean' and ``skipna`` is set to True, then aggregation
+        will have the same behavior as if func is set to 'nanmean'.
     closed: {'left', 'right'}, default 'left'
         Which side of bin interval is closed.
     **flox_kwargs
@@ -111,6 +117,7 @@ def compute_MVBS(
         range_var=range_var,
         method=method,
         func=func,
+        skipna=skipna,
         **flox_kwargs,
     )
 
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -33,6 +33,7 @@ def compute_MVBS(
     range_bin: str = "20m",
     ping_time_bin: str = "20S",
     method="map-reduce",
+    func="nanmean",
     closed: Literal["left", "right"] = "left",
     **flox_kwargs,
 ):
@@ -62,6 +63,10 @@ def compute_MVBS(
         The flox strategy for reduction of dask arrays only.
         See flox `documentation <https://flox.readthedocs.io/en/latest/implementation.html>`_
         for more details.
+    func: str, default 'nanmean'
+        The flox aggregation function used for reducing the data array.
+        By default, 'nanmean' is used. Other options can be found in the flox `documentation
+        <https://flox.readthedocs.io/en/latest/generated/flox.xarray.xarray_reduce.html>`_.
     closed: {'left', 'right'}, default 'left'
         Which side of bin interval is closed.
     **flox_kwargs
@@ -105,6 +110,7 @@ def compute_MVBS(
         ping_interval,
         range_var=range_var,
         method=method,
+        func=func,
         **flox_kwargs,
     )
 
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -31,7 +31,7 @@ def compute_MVBS(
     ds_Sv: xr.Dataset,
     range_var: Literal["echo_range", "depth"] = "echo_range",
     range_bin: str = "20m",
-    ping_time_bin: str = "20S",
+    ping_time_bin: str = "20s",
     method="map-reduce",
     skipna=True,
     closed: Literal["left", "right"] = "left",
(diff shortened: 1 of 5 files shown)


leewujung (commented):
Hey @ctuguinay : Thanks for the investigation and changes! I was thinking through the use of `func` and `skipna`, and am thinking that we should probably avoid the confusion (which stemmed from the way [xarray_reduce](https://flox.readthedocs.io/en/latest/generated/flox.xarray.xarray_reduce.html) of using `func` and `skipna` at the same time.

For `compute_MVBS`, since it is supposed to be a mean operation, we do not have to allow for other `func`, but just need the `skipna` parameter. The same for `compute_raw_MVBS`. The place where we really need the `func` is `_groupby_x_along_channels`. So, I think we can have:
- `compute_MVBS` and `compute_raw_MVBS` will only have `skipna` as a parameter
- `_groupby_x_along_channels` have both `func` and `skipna`, and since it is under the hood, we can make sure `func="nanmean"`+`skipna=True` or `func="mean"`+`skipna=False` are used with it.

The `compute_NASC` computation comes to mind naturally, that we could allow the same `skipna` argument, and under the hood use either `func="nansum"`+`skipna=True`  or `func="sum"` + `skipna=False`

Thoughts?

ctuguinay:
Hey @leewujung, thanks for the review!

I agree with you on the implementation details. I also think that not all combinations should be allowed (ie the combinations `func = nansum` and `skipna = False` is a bit contradictory) and I was gonna add some restriction as to the combinations that the user could pass in, but this solution is much cleaner.

Would you also like me to make this same change of exposing `skipna` and using `func` under the hood to `compute_NASC`?

leewujung:
> Would you also like me to make this same change of exposing `skipna` and using `func` under the hood to `compute_NASC`?

Sounds good if you could add this for `compute_NASC` in this PR also. Thanks!

ctuguinay:
Hey @leewujung, I made the changes, but only the changes for `compute_MVBS`.

I do have one question though: In `compute_raw_NASC`, the operation change should only be in the setting of mean/nanmean in the  `sv_mean = _groupby_x_along_channels(...,func=func="nanmean" if skipna else "mean",...)` right? Not sum/nansum? Since `sv_mean` and not `sv_sum` is being calculated here. Let me know if this implementation is incorrect.

I do also see the `sum` operations being used to calculate `h_mean_denom` and `h_mean_num` and I'm not sure if in those computations I should do anything with `nansum`. The same applies with `ds_ping_time` and it using `nanmean` currently.  Should these three computations' aggregate functions interact at all with the `skipna` being passed into `compute_raw_NASC`?

leewujung:
> In `compute_raw_NASC`, the operation change should only be in the setting of mean/nanmean in the `sv_mean = _groupby_x_along_channels(...,func=func="nanmean" if skipna else "mean",...)` right? Not sum/nansum? Since `sv_mean` and not `sv_sum` is being calculated here. Let me know if this implementation is incorrect, since your comment above said `nansum` /`sum`.

Oh right, you're right that it would still be mean/nanmean in the `compute_NASC` function the way it is implemented.

> But, I do also see the `sum` operations being used to calculate `h_mean_denom` and `h_mean_num` and I'm not sure if in those computations I should do anything with `nansum`. The same applies with `ds_ping_time` and it using `nanmean` currently. Should these three computations' aggregate functions interact at all with the `skipna` being passed into `compute_raw_NASC`?

For `h_mean_*` and `ds_ping_time`, I think we can hard-code it to use `nanmean`/`nansum` and `skipna=False`, because:
- the `h_mean_*` provides the "height" or the steps in the integral, so NaN comes from the Sv side but not the steps
- `ds_ping_time` is used to come up with a new coordinate along the time or distance dimension, so even if the values end up being NaN, the coordinate should not be

ctuguinay:
Sounds good, thanks!

ctuguinay:
I'll put this one on hold for a bit since I am getting some weird NaN values in the `compute_NASC` tests when I shouldn't be getting them (using nanmean/nansum). I'll come back to this after addressing a few of the Echopype user issues.

leewujung:
> I think if I take a very thorough look at the bins that are being created, that will allow me to make sense of this...I think.

Yeah, and making the test data with smaller dimensions may help with figuring out what is going on. For example computing NASC using just 2 pings of data should give you some flexibility to look into where NaNs should be created and where it should not.

ctuguinay:
Looking back at that compute MVBS test that I modified, I think I'll revert the change I made for that and have two separate tests (1 for compute MVBS and 1 for compute NASC) with smaller data arrays to do as you suggested above. The `_parse_nans` may hide some unintentional NaNs in the `ds_MVBS` and it's just so messy working with such a large dataset.

echopype/commongrid/api.py
```diff
@@ -58,21 +57,15 @@ def compute_MVBS(
...
-        The flox aggregation function used for reducing the data array.
-        By default, 'nanmean' is used. Other options can be found in the flox `documentation
-        <https://flox.readthedocs.io/en/latest/generated/flox.xarray.xarray_reduce.html>`_.
     skipna: bool, default True
-        If true, aggregation function skips NaN values.
-        Else, aggregation function includes NaN values.
-        Note that if ``func`` is set to 'mean' and ``skipna`` is set to True, then aggregation
-        will have the same behavior as if func is set to 'nanmean'.
+        If true, mean function skips NaN values.
+        Else, mean function includes NaN values.
```
leewujung, line 68:
```suggestion
        If true, the mean operation skips NaN values.
        Else, the mean operation includes NaN values.
```

echopype/commongrid/api.py
```diff
@@ -300,6 +293,9 @@ def compute_NASC(
         The flox strategy for reduction of dask arrays only.
         See flox `documentation <https://flox.readthedocs.io/en/latest/implementation.html>`_
         for more details.
+    skipna: bool, default True
+        If true, mean function skips NaN values.
+        Else, mean function includes NaN values.
```
leewujung, line 298:
```suggestion
        If true, the mean operation skips NaN values.
        Else, the mean operation includes NaN values.
```

echopype/commongrid/utils.py
```diff
@@ -113,6 +107,9 @@ def compute_raw_NASC(
         The flox strategy for reduction of dask arrays only.
         See flox `documentation <https://flox.readthedocs.io/en/latest/implementation.html>`_
         for more details.
+    skipna: bool, default True
+        If true, mean function skips NaN values.
+        Else, mean function includes NaN values.
```
leewujung, line 112:
```suggestion
        If true, the mean operation skips NaN values.
        Else, the mean operation includes NaN values.
```

leewujung:
@ctuguinay : I direct-pushed some changes to fix `mock_Sv_dataset_irregular` outputs, see below:
- for Sv datasets generated by `compute_Sv`, the `Sv` and the `echo_range` dataarrays will have identical NaN elements. You can see it in at the end of the `compute_Sv` code. The original `mock_Sv_dataset_irregular` fixture only set some `echo_range` elements to NaN but not the corresponding `Sv` elements.
- once I fixed that, I removed 2 lines in your `test_compute_MVBS_NASC_skipna_nan_and_non_nan_values` to set the NaN elements
- and then I fixed `test_compute_NASC_values` by adding `skipna=True` in the `compute_NASC` call, and a few fixtures called in it in `conftest.py` and `testing.py`

echopype/tests/commongrid/test_commongrid_api.py
```diff
@@ -438,11 +438,45 @@ def test_compute_NASC_values(request, er_type):
...
+        ("NASC", False),
+    ],
+)
+def test_compute_MVBS_NASC_skipna_nan_and_non_nan_values(request, operation, skipna):
+    # Create subset dataset with 2 channels, 2 ping times, 2 depth values:
+
+    # Get fixture for irregular Sv
+    ds_Sv = request.getfixturevalue("mock_Sv_dataset_irregular")
+    # Already has 2 channels, so subset for only ping time and range sample
+    subset_ds_Sv = ds_Sv.isel(ping_time=slice(0,2), range_sample=slice(0,2))
```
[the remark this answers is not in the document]
ctuguinay: I agree. I'll make these changes.

leewujung (approved):

Issue #702, opened 2022-05-23
Reduced sampling rate Power/Angle samples


marianpena:
Hi, 
I've been testing echopype with EK80 data recorded in CW mode and reduced sampling rate Power/Angle samples; backscatter looks fine but Sv is full of NaN values. I used ep.calibrate.compute_Sv(ed,waveform_mode="CW",encode_mode='power')

Echopype specifies 2 modes ("complex"` and `"power") but three data formats are possible:

The Stored sampled data for WBTs running CW options can be used to reduce the amount of recorded data when you are using a Wide Band Transceiver (WBT) with CW pulses.
• 	Select Complex samples to use the default data format.
• 	Select Power/Angle to reduce the file sizes.
• 	Select Reduced sampling rate Power/Angle samples to minimize the file sizes.

Is this last one supported?

Thanks!

marianpena:
Ok I found the error and how to fix it. Temperature, salinity and depth had the same length as ping_time but only the first value was filled (the rest as NaN). I fixed it by specifying:
env_params = {
    'temperature': ed.environment.temperature[0].values,   # temperature in degree Celsius
    'salinity': ed.environment.salinity[0].values,     # salinity in PSU
    'pressure': ed.environment.depth[0].values,     # pressure in dbar
}
and then ep.calibrate.compute_Sv(ed,waveform_mode="CW",encode_mode=encode_mode, env_params=env_params)

leewujung:
Hey @marianpena : I have been doing some refactoring of calibration code, which contains particularly parts about `cal_params` and `env_params` intake. I think my changes fixed the issues you exposed. Could you point me a sample file and a snippet of code to test? I don't have access to an EK80 system right now to produce reduced sampling rate data. Thanks!

PR #1270, opened 2024-02-29
test echo range for EK80 reduced sample rate

### LINKED ISSUE #969: Add test for `echo_range` for EK80 reduced sampling rate data
This is a leftover from #968 where the `echo_range` output dimension should also be checked for EK80 reduced sampling rate data.

This may be related to #702.

MohamedNasser8:
Fixes #969 
I don't know if the mean by reduced sampling rate is the reduced file size, if it is, The **fifth** test case do the same thing as me.
Or I should resample the data?

### DIFF
--- a/echopype/tests/calibrate/test_range_integration.py
+++ b/echopype/tests/calibrate/test_range_integration.py
@@ -19,6 +19,7 @@ import echopype as ep
         # EK80 CW power
         ("EK80", "EK80", "Summer2018--D20180905-T033113.raw", None, None, None, "CW", "power"),
         # TODO: EK80 reduced sampling rate
+        ("EK80", "EK80", "Summer2018--D20180905-T033113.raw", None, None, None, "BB", "complex"),
     ],
     ids=[
         "azfp",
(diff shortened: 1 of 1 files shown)


leewujung:
@MohamedNasser8 : My bad, I did not put in enough info the original description in #969.

Here the reduced sampling rate refers to a specific setup the EK80 instrument allows. That setup would make the discretization along the `range_sample` axis coarser in the `EchoData` object (from the `open_raw` function), which affects the `echo_range` values in the output of `compute_Sv`. The intention of #969 is to check the calculation against the expected values from another program. I got some sample data from a colleague from #702, and will put that somewhere for you to download, and the expected values it should be tested against.

MohamedNasser8:
@leewujung Thanks for your illustration, I'm waiting for the data to test it.

MohamedNasser8:
Hello @leewujung, I just wanted to remind you about the data I can test with.

ctuguinay:
I'm going to close this. I think this should probably be tested when we get in-lab EK80 systems in conjunction with the many other  possible configurations @leewujung

leewujung:
yeah but probably should keep this open as a reminder?

ctuguinay:
Ah yeah I see

leewujung:
Looking at this again I agree with you @ctuguinay that we should just close this. We have #969 as a reminder.

PR #1278, opened 2024-03-13
Fix error parsing integer in frequency-differencing equation

leewujung:
This PR fixes the bug that the current equation parsing regex pattern does not parse frequencies expressed as integers and not floats.

Two tests re added: one with mixed integers and floats, and one with all integers.


praneethratna (approved):
Looks good, thanks!

PR #1285, opened 2024-03-20
Optimize get vend cal params power

anujsinha3:
* Refactor function to avoid expand_dims and sortby if required

* test: increase co-ordinates to 200 for ping_time dimension

### DIFF
--- a/echopype/tests/calibrate/test_cal_params.py
+++ b/echopype/tests/calibrate/test_cal_params.py
@@ -14,7 +14,7 @@ from echopype.calibrate.cal_params import (
     get_vend_cal_params_power,
 )
 
-DATA = np.ones((2, 200)) * 2000
+DATA = np.random.rand(2, 200)
 TIME_COORDINATES = np.ones(200) * 1000
 
 @pytest.fixture
--- a/echopype/tests/calibrate/test_cal_params.py
+++ b/echopype/tests/calibrate/test_cal_params.py
@@ -360,7 +360,7 @@ def test_sanitize_user_cal_dict(sonar_type, user_dict, channel, out_dict):
                 coords={"ping_time": TIME_COORDINATES, "channel": ["chA", "chB"]},
             ),
             xr.DataArray(
-                np.vstack([np.full(200, 2000), np.full(200, np.nan)]),
+                np.vstack([DATA[0,:], np.full(200, np.nan)]),
                 dims=["channel", "ping_time"],
                 coords={"ping_time": TIME_COORDINATES, "channel": ["chA", "chB"]},
             ),
--- a/echopype/tests/calibrate/test_cal_params.py
+++ b/echopype/tests/calibrate/test_cal_params.py
@@ -18,14 +18,6 @@ DATA = np.ones((2, 200)) * 2000
 TIME_COORDINATES = np.ones(200) * 1000
 
 @pytest.fixture
-def freq_center_interp():
-    return xr.DataArray(
-        [[25, 55]],
-        dims=["ping_time", "channel"],
-        coords={"channel": ["chA", "chB"], "ping_time": [1]},
-    )
-
-@pytest.fixture
 def freq_center():
     return xr.DataArray(
         DATA,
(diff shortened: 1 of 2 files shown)


echopype/tests/calibrate/test_cal_params.py
```diff
@@ -2,19 +2,35 @@
...
+TIME_COORDINATES = np.ones(200) * 1000
 
 @pytest.fixture
-def freq_center():
+def freq_center_interp():
     return xr.DataArray(
         [[25, 55]],
         dims=["ping_time", "channel"],
-        coords={"channel": ["chA", "chB"], "ping_time": [1]}
+        coords={"channel": ["chA", "chB"], "ping_time": [1]},
```
lsetiawan, line 25: I notice that this `ping_time` is still only 1 value, can you update this to be like `freq_center`? Or if this is the same as `freq_center` then just update the test using this fixture to use `freq_center`? Thanks!

anujsinha3: Done. This was being used for `test_get_interp_da`(). I have updated the test and assertions to use `freq_center`

lsetiawan:
@leewujung could you do a review on this, it looks good to me, just wanted to make sure that it's all good to you as well. Thanks! The test data is now bigger.

lsetiawan (approved):

echopype/tests/calibrate/test_cal_params.py
```diff
@@ -252,41 +338,39 @@ def test_sanitize_user_cal_dict(sonar_type, user_dict, channel, out_dict):
...
             xr.DataArray(
-                np.array([[100], [200]]),
+                DATA,
                 dims=["channel", "ping_time"],
-                coords={"ping_time": [1], "channel": ["chA", "chB"]},
+                coords={"ping_time": TIME_COORDINATES, "channel": ["chA", "chB"]},
             ),
             xr.DataArray(
-                [[100], [5.5]],
+                np.vstack([np.full(200, 2000), np.full(200, np.nan)]),
```
leewujung, line 363:
```suggestion
                np.vstack([[DATA[0,:], np.full(200, np.nan)]),
```

echopype/tests/calibrate/test_cal_params.py
```diff
@@ -2,19 +2,27 @@
...
+    CAL_PARAMS,
+    param2da,
+    sanitize_user_cal_dict,
+    _get_interp_da,
+    get_cal_params_AZFP,
+    get_cal_params_EK,
+    get_vend_cal_params_power,
 )
 
+DATA = np.ones((2, 200)) * 2000
```
leewujung, line 17:
```suggestion
DATA = np.random.rand(2, 200)
```

leewujung (commented):
@anujsinha3 : Thanks for the changes! I like how you avoided the expansion. Expanding the test data dimension is also great.

I suggested a small change to `DATA` to just make things more randomized, since the functions are supposed to be flexible to deal with arbitrary numbers. See what you think.

echopype/tests/calibrate/test_cal_params.py
anujsinha3: Hi @leewujung, 
Initially, I had used the np.random.rand() as you suggested, but that caused the test assertions to fail as the test outputs also got randomized (especially for interpolation test cases). Hence I switched to np.ones().

Happy to adapt further suggestions.

anujsinha3:
Hi @leewujung,
Initially, I had used the np.random.rand() as you suggested, but that caused the test assertions to fail as the test outputs also got randomized (especially for interpolation test cases). Hence I switched to np.ones().

Happy to adapt further suggestions.

leewujung:
@anujsinha3 : Locally the tests I suggested worked. Could you point to where the interpolation and failure was? It was hard for me to tell which parts actually got changed due to the reformatting. Even if the values are random, wouldn't it be possible to put the expected interpolated results based on `DATA` into the assertion?

anujsinha3:
Ah, I see the error now. I was multiplying the `np.random.rand()` with a constant, which caused the test case failures. I have fixed that now. Apologies for the confusion.

leewujung:
@anujsinha3 : No problem! Feel free to merge this then!

leewujung:
Alright, I'll merge this now.

Issue #1294, opened 2024-04-03
`add_location` fails with duplicated timestamps in GPS data


ashtonflinders:
Not sure if this is a code issue or my not understanding the required variables....

```
import echopype as ep
from echopype import open_raw

raw_file_s3path = "s3://noaa-wcsd-pds/data/raw/Alaska_Knight/NEBS21AK/ES60/L0008-D20210531-T044412-ES60.raw"
ed = open_raw(
   raw_file_s3path, sonar_model='EK60',
   storage_options={'anon': True}
)

Sv = ep.calibrate.compute_Sv(ed)
ep.consolidate.add_location(Sv, ed)
```

---------------------------------------------------------------------------
InvalidIndexError                         Traceback (most recent call last)
Cell In[25], line 1
----> 1 ep.consolidate.add_location(Sv, ed)

File ~/opt/anaconda3/envs/echopype/lib/python3.9/site-packages/echopype/utils/prov.py:237, in add_processing_level.<locals>.wrapper.<locals>.inner(*args, **kwargs)
    235 @functools.wraps(func)
    236 def inner(*args, **kwargs):
--> 237     dataobj = func(*args, **kwargs)
    238     if is_echodata:
    239         ed = dataobj

File ~/opt/anaconda3/envs/echopype/lib/python3.9/site-packages/echopype/consolidate/api.py:196, in add_location(ds, echodata, nmea_sentence)
    194 interp_ds = ds.copy()
    195 time_dim_name = list(echodata["Platform"]["longitude"].dims)[0]
--> 196 interp_ds["latitude"] = sel_interp("latitude", time_dim_name)
    197 interp_ds["longitude"] = sel_interp("longitude", time_dim_name)
    198 # Most attributes are attached automatically via interpolation
    199 # here we add the history

File ~/opt/anaconda3/envs/echopype/lib/python3.9/site-packages/echopype/consolidate/api.py:186, in add_location.<locals>.sel_interp(var, time_dim_name)
    179     return xr.DataArray(
    180         data=position_var.values[0] * np.ones(len(ds["ping_time"]), dtype=np.float64),
    181         dims=["ping_time"],
    182         attrs=position_var.attrs,
    183     )
    184 else:
    185     # Values may be nan if there are ping_time values outside the time_dim_name range
--> 186     return position_var.interp(**{time_dim_name: ds["ping_time"]})

[...]

Issue #1286, opened 2024-03-20
Strange GPS coordinates appearing with add_location() method


oftfrfbf:
Hello, I am trying to use echopype to match up Sv data with its associated geospatial information and I am getting strange GPS values when using "add_location()". I have noticed this happening with a number of different raw files, below I include one example.

This is using python 3.10.12 and echopype==0.8.3.

This first piece of code is my manual method for pairing up NMEA datagram latitude and longitudes with the nearest ping_time and this works as expected:

```
!pip install s3fs boto3==1.34.51 numpy==1.24.4 pandas==1.5.3 xarray==2022.12.0 geopandas==0.14.3 folium==0.16.0 typing_extensions==4.8.0 echopype==0.8.3 mapclassify
import s3fs
import folium
import os
import boto3
import numpy as np
import echopype as ep
import geopandas
import folium
import mapclassify as mc
import pandas as pd
import xarray as xr
from botocore import UNSIGNED
from botocore.config import Config
import matplotlib.pyplot as plt

s3 = boto3.client('s3', config=Config(signature_version=UNSIGNED))
s3_file_system = s3fs.S3FileSystem(anon=True)
print(f"echopype version: {ep.__version__}")

bucket_name = 'noaa-wcsd-pds'
ship_name = 'Albatross_Iv'
cruise_name = 'AL0403'
sensor_name = 'EK60'
file_name = "L0010-D20040416-T094042-EK60.raw"

raw_file_s3_path = f"s3://{bucket_name}/data/raw/{ship_name}/{cruise_name}/{sensor_name}/{file_name}"
echodata = ep.open_raw(raw_file_s3_path, sonar_model=sensor_name, use_swap=True, storage_options={'anon': True})
latitude = echodata.platform.latitude.values
latitude_values = latitude.copy()
latitude_values.sort()
print(latitude_values)
# [ 0.        0.       43.68361  ... 43.811595 43.811595 43.811595] <-- Note sorted latitude values

longitude = echodata.platform.longitude.values
nmea_times = np.sort(echodata.platform.time1.values)
time1 = echodata.environment.time1.values
indices = np.searchsorted(a=nmea_times, v=time1, side="right") - 1
lat = latitude[indices]
lat[indices < 0] = np.nan  # values recorded before indexing are set to nan
lon = longitude[indices]
lon[indices < 0] = np.nan

[...]

PR #1296, opened 2024-04-04
Add warnings and raised errors for `add_location`

ctuguinay:
Add warnings for `add_location` for NaN/0 lat/lon and duplicated timestamps to address the following issues:
[Strange GPS coordinates appearing with add_location() method #1286](https://github.com/OSOceanAcoustics/echopype/issues/1286)
[add_location fails with duplicated timestamps in GPS data #1294](https://github.com/OSOceanAcoustics/echopype/issues/1294)


ctuguinay:
@leewujung This is ready to be looked at.

ctuguinay:
Oh wait, I'll add some tests to cover these scenarios.

ctuguinay:
@leewujung This PR should be ready for review now

echopype/consolidate/api.py
```diff
@@ -191,10 +194,35 @@ def sel_interp(var, time_dim_name):
...
+
+    # Check if there are duplicates in time_dim_name
+    if len(np.unique(echodata["Platform"][time_dim_name].data)) != len(
+        echodata["Platform"][time_dim_name].data
+    ):
+        logger.warning(
+            f'The echodata["Platform"]["{time_dim_name}"] array contains duplicate values. '
+            "Interpolation expects unique time values."
+        )
+
```
leewujung, line 222: I think this should raise an error instead of a warning? Because the underlying code would actually raise a warning. This is as opposed to the GPS NaN/0 cases that the underlying code would run through without erroring out.

ctuguinay: Ah, I see. I think I misinterpreted your original statement here: https://github.com/OSOceanAcoustics/echopype/issues/1294#issuecomment-2037745091. I assumed that meant letting Pandas call the error and we provide a warning message prior to showing the user where the error occurred

leewujung: Oh I see, no problem!

ctuguinay:
@leewujung Made the switch from warning to raised value error. This should be ready for a quick review again.

ctuguinay:
### Explanation of the current warning (from slack conversation)

Here's the current order of events in the code that is causing this warning: (https://github.com/pydata/xarray/blob/1d43672574332615f225089d69f95a9f8d81d912/xarray/coding/times.py#L723C1-L803C32):


1. The computation converts nanosecond resolution time to int64

2. Computes needed time time_delta (represents the time unit granularity based on the specified units), and needed_time_delta (represents the minimum unit granularity required for lossless integer encoding of the dates)

3. (Optional) If time_delta > needed_time_delta, and we specify None dtype encoding we get the not faithful serialization message, since with integer bit 64 encoding, we will lose precision.

4. The operation to finalize the encoding is then done in int64.

5. (Optional) Then, if we actually specify dtype, we cast to dtype (if numpy deems safe).

Now, I am not quite sure at this point why float64 should silence this warning: The casting to float64 is a bit of a 'moot point', since the casting operation is done post-encoding computation, which we know will result in lossful integer encoding. (edited) 
This is probably why `None` and `float64` encoding result in the same values. The `float64` `dtype` does practically nothing to change the encoded array, since it is done at the very last step. Additionally, the `time_delta` > `needed_time_delta` also is why we lose precision even when there is no warning (in the case where we set `dtype` set to `float64`).

echopype/consolidate/api.py
```diff
@@ -204,7 +204,7 @@ def sel_interp(var, time_dim_name):
     ).any()
     interp_msg = (
         "Interpolation may be negatively impacted, "
-        "consider handling these values before calling add_location."
+        "consider handling these values before calling ``add_location``."
```
leewujung, line 207: Do you know if the error/earning messages would be formatted if one uses ``x`` ? I know it would be rendered in docstrings, but not sure about the printout messages. I guess this is useful regardless because users would know the thing in the quote is a variable?

leewujung (approved):
Looks good! Feel free to merge this.

Issue #1290, opened 2024-03-28
Serialization warning on timestamp resolution


leewujung:
This came from #1232:

There is a serialization warning with timestamp resolution. This seems something that came up after pandas 2.0 update in xarray. It would be good to resolve this, since if not suppressed, a huge bulk of warnings would show up when mass-converting files. It is also good to make sure we know if there is any resolution issue.

Sample file:
[SCAN-D20231122-T172241.zip](https://github.com/OSOceanAcoustics/echopype/files/13457137/SCAN-D20231122-T172241.zip)

Warning message:
```
/lib/python3.9/site-packages/echopype/utils/coding.py:87: 
UserWarning: Times can't be serialized faithfully to int64 with requested units 'seconds since 1900-01-01T00:00:00+00:00'. 
Resolution of 'microseconds' needed. 
Serializing times to floating point instead. 
Set encoding['dtype'] to integer dtype to serialize to int64. 
Set encoding['dtype'] to floating point dtype to silence this warning.
  encoded_data, _, _ = coding.times.encode_cf_datetime(
```

ctuguinay:
* Look through issues in Xarray to see if this is addressed and if not raise an issue
* Just filter it out for now

ctuguinay:
@leewujung 
So this warning pops up here where xarray enforces that if the `dtype` passed in is `None` then a 'not faithful' serialization warning is passed in: https://github.com/pydata/xarray/blob/97d3a3aaa071fa5341132331abe90ec39f914b52/xarray/coding/times.py#L772C1-L778C18. Echopype calls the function that this code is in in `_encode_dataarray`: https://github.com/OSOceanAcoustics/echopype/blob/4278dd6d67389d04aa0a4f28f282d78c5f823a4e/echopype/utils/coding.py#L74C1-L91C74

This change stems from this PR that was created to fix a series of cf datetime time issues that arose with Pandas 2.0: https://github.com/pydata/xarray/commit/b6d4bf72fd56014c0bd142bc438d6204d8dfd1ef

[...]

PR #1299, opened 2024-04-11
Fix time encoding to use int64 and "nanoseconds since 1970-01-01 00:00:00Z"

ctuguinay:
This PR addresses the not faithful time serialization warning from issue #1290.

### DIFF
--- a/echopype/calibrate/ek80_complex.py
+++ b/echopype/calibrate/ek80_complex.py
@@ -1,5 +1,4 @@
 from collections import defaultdict
-from functools import partial
 from typing import Dict, Literal, Optional, Union
 
 import numpy as np
@@ -227,10 +226,11 @@ def get_transmit_signal(
     # but keeping this here for use as standalone function
     if waveform_mode == "BB" and np.all(beam["transmit_type"] == "CW"):
         raise TypeError("File does not contain BB mode complex samples!")
+
     # Generate all transmit replica
     y_all = {}
     y_time_all = {}
-    # TODO: expand to deal with the case with varying non-NaN tx param across ping_time
+    # TODO: expand to deal with the case with varying tx param across ping_time
     tx_param_names = [
         "transmit_duration_nominal",
         "slope",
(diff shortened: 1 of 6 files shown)


leewujung:
@ctuguinay : The issue in #1290 is only on timestamps, right? I am not sure why you have to change things that are not related to timestamps. Those seem to result into test failing. The time encoding thing will only result in a warning when data are converted, but the calibration will still pass. 

This is a different issue from the NaN values in the calibration code.

ctuguinay:
### Explanation of the current warning (from slack conversation)

@leewujung

Here's the current order of events in the code that is causing this warning: (https://github.com/pydata/xarray/blob/1d43672574332615f225089d69f95a9f8d81d912/xarray/coding/times.py#L723C1-L803C32):


1. The computation converts nanosecond resolution time to `int64`.

2. Computes needed time time_delta (represents the time unit granularity based on the specified units), and needed_time_delta (represents the minimum unit granularity required for lossless integer encoding of the dates).

3. (Optional) If `time_delta` > `needed_time_delta`, and we specify None `dtype` encoding we get the not faithful serialization message, since with integer bit 64 encoding, we will lose precision.

4. The operation to finalize the encoding is then done in `int64`.

5. (Optional) Then, if we actually specify `dtype`, we cast to `dtype` (if `numpy` deems safe).

Now, I am not quite sure at this point why `float64` should silence this warning: The casting to `float64` is a bit of a 'moot point', since the casting operation is done post-encoding computation, which we know will result in lossful integer encoding. 
This is probably why `None` and `float64` encoding result in the same values. The `float64` `dtype` does practically nothing to change the encoded array, since it is done at the very last step. Additionally, the `time_delta` > `needed_time_delta` also is why we lose precision even when there is no warning (in the case where we set `dtype` set to `float64`).

Perhaps this is an issue I bring up in `xarray`? This still doesn't make sense to me, but I do think it'd be good to silence this error as soon as possible since it can be very disheartening seeing so many time serialization warnings on your first use of  Echopype.

On our own end, do you think adding a warning in the logger for this would be a good idea? That way those users that have logger verbose turned on can still get the message.

leewujung (dismissed):
@ctuguinay : Looks great! I am glad we got this resolved after much digging and  trial and error!

I only have one comment: What do you think of changing the function name `_encode_dataarray` to `_encode_time_dataarray`? Since it is only used for `set_time_encodings` and the content does not make sense for other types of dataarray?

Otherwise I think this is good to go.

ctuguinay:
@leewujung Made that change. This should be ready to merge now

PR #1302, opened 2024-04-13
Support for EK80 Broadband Complex Multiplex Data

ctuguinay:
Addresses #743 and #1287.

### Calibration Code Changes:
* In `compress_pulse`, the convolution operation handles `NaNs` by turning them into `0s` and turning them back into `NaNs` post-convolution.
* In `get_transmit_signal`, filtered out NaNs from `tx_params`
    * The code itself still doesn't accept multiple unique non-`NaN` `tx_params`, but that should be a separate PR.
* In `harmonize_env_param_time`, separated out the single `time1` when drop `NaN` case from single `time1` without drop `NaN` case.
    * This was done specifically for the case where EK80 multiplex data has 1 `time1` dimension and has alternating `NaN` / non-`NaN` pattern. When drop `NaN` on the `time1` dimension is called on the multiplex data array, it will drop ALL values of the array, which is not the desired behavior.
* In gain compensation operation, zero out `NaNs` that appear due to 1) multiplex ping patterns and 2) single beam transducer systems that don't contain angle offset information.

### Small Change
* Set Xarray DataArray/Dataset `drop` to `drop_vars`, since `drop` is now deprecated

### DIFF
--- a/echopype/calibrate/ek80_complex.py
+++ b/echopype/calibrate/ek80_complex.py
@@ -259,9 +259,13 @@ def get_transmit_signal(
     return y_all, y_time_all
 
 
-def _nan_check_convolve(m, replica):
+def _convolve_nans_efficiently(m, replica):
     """
-    Convolve two arrays while handling NaN values efficiently.
+    Convolve two arrays while handling NaN values efficiently:
+
+    If all elements in ``m`` are NaN, the function returns ``m`` unchanged.
+    If there are NaN values in ``m``, they are temporarily set to 0 before convolution
+    and restored back to NaN in the resulting array after convolution.
 
     Parameters
     ----------
@@ -272,18 +276,30 @@ def _nan_check_convolve(m, replica):
 
     Returns
     -------
-    array-like
+    conv_result : array-like
         Convolved array.
-
-    Notes
-    -----
-    If all elements in ``m`` are NaN, the function returns ``m``.
-    Otherwise, we allow ``convolve`` to choose whether FFT or discrete convolution is used.
     """
-    if np.all(np.isnan(m)):
+    # Create a mask for NaN values
+    nan_mask = np.isnan(m)
+
+    # Check if all elements in m are NaN
+    if np.all(nan_mask):
         return m
-    else:
-        return signal.convolve(m, replica, mode="full")[replica.size - 1 :]
+
+    # Create a copy of m to modify
+    m_copy = m.copy()
+
+    # Replace NaNs with 0s
+    m_copy[nan_mask] = 0
+
+    # Perform convolution
+    # Scipy chooses whether FFT or discrete convolution is fastest and uses faster method
+    conv_result = signal.convolve(m_copy, replica, mode="full")[replica.size - 1 :]
+
+    # Restore NaN values in the resulting array
+    conv_result[nan_mask] = np.nan
+
+    return conv_result
 
 
 def compress_pulse(backscatter: xr.DataArray, chirp: Dict) -> xr.DataArray:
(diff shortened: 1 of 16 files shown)


echopype/calibrate/ek80_complex.py
```diff
@@ -278,11 +323,18 @@ def compress_pulse(backscatter: xr.DataArray, chirp: Dict) -> xr.DataArray:
...
+        # Compute complex conjugate of transmit values and reverse order of elements
         replica = np.flipud(np.conj(tx))
 
+        # Create partial of _convolve_nans_efficiently with passed in replica
+        _convolve_nans_efficiently_partial = partial(_convolve_nans_efficiently, replica=replica)
+
+        # Apply convolve on backscatter (along range sample dimension) and replica
         pc = xr.apply_ufunc(
-            lambda m: (signal.convolve(m, replica, mode="full")[replica.size - 1 :]),
+            _convolve_nans_efficiently_partial,
```
leewujung, line 337: I am a little confused by the new function here. Wouldn't it work as well with the following sequence:
- set all NaN elements in `backscatter_chan` to 0
- run the original `xr.apply_ufunc` block
- set all 0 elements in `pc` to NaN

echopype/tests/calibrate/test_calibrate_ek80.py
```diff
@@ -371,10 +370,8 @@ def test_ek80_BB_power_echoview(ek80_path):
     # Below only compare the first ping
     ev_vals = df_real.values[:, :]
     ep_vals = pc_mean.values.real[:, :]
-    assert np.allclose(ev_vals[:, 69:8284], ep_vals[:, 69:], atol=1e-4)
-    assert np.allclose(ev_vals[:, 90:8284], ep_vals[:, 90:], atol=1e-5)
+    assert np.allclose(ev_vals[:, 69:], ep_vals[:, 69:], atol=1e-4)
```
leewujung, line 374: Could you remind me why you removed this second assert?

echopype/calibrate/ek80_complex.py
ctuguinay: Oh yup, you're right, I'll do that instead 👍

echopype/calibrate/ek80_complex.py
```diff
@@ -323,18 +279,21 @@ def compress_pulse(backscatter: xr.DataArray, chirp: Dict) -> xr.DataArray:
         # Select channel `chan` and drop the specific beam dimension if all of the values are nan.
         backscatter_chan = backscatter.sel(channel=chan).dropna(dim="beam", how="all")
 
+        # Create NaN mask
+        nan_mask = np.isnan(backscatter_chan.compute())
```
leewujung, line 283: Do we need `.compute` here?

ctuguinay: Just tested it and unfortunately no. Without the `.compute()`, the `xr.where` at line 286, and the one at line 306 doesn't force compute it. Without computing prior, line 306 will turn `pc` from something computed to something not computed.

leewujung: If you only have .compute() in the line with xr.where, does it not work?

leewujung: I guess it does not really matter in this situation, where we have .compute somewhere in the middle of the function... maybe when we need to figure out how to distribute efficiently....

ctuguinay: It does work, made that change👍

leewujung: What I was thinking was to do 
```
pc=xr.where(nan_mask, np.nan,pc)
pc_all.append(pc.compute())
```
But looking at the code again, I think it'll actually be fine to keep things not computed until users explicitly want it to be computed. In a sense it is how it's supposed to be, like how we have it in `combine_echodata`. We can address that in another go.

leewujung:
Ok, after some fiddling I found that many other parts of the calibration code does not currently support a fully lazy-loaded EchoData object, mostly due to "Indexing with a boolean dask array is not allowed". Those boolean indexing places are all pretty small under the current constraint that the transmit signal type are identical across `ping_time` (so can change across `channel`), so it may be fine to force compute them, but we need to do some more experiment and benchmarking.

That is outside the scope of this PR though, so I'll merge this PR first. Thanks for the investigations and changes to address the alternating channel issue!

Issue #1300, opened 2024-04-12
Add additional `Sv_offset` values to AZFP calibration


leewujung:
From communication with Steve Pearce at ASL Env Science, we can now add the following `Sv_offset` values for calibrating AZFP data.
- 67 kHz: 500 us (1.1 dB)
- 125 kHz: 150 us (1.4 dB), 250 us (1.3 dB)
- 200 kHz: 150 us (1.4 dB), 250 us (1.3 dB)
- 455 kHz: 250 us (1.3 dB)
- 769 kHz: 150 us (1.4 dB)

ctuguinay:
Closed by #1304

PR #1304, opened 2024-04-17
Update `_calc_Sv_offset` for AZFP parsing

leewujung:
This PR includes the following:
- adds additional values in the Sv_offset dictionary based on communication with ASL Evn Sci
- updates `_calc_Sv_offset` to use a param dictionary instead of nested if-else

One existing test needs to be skipped because the pulse length is not in the Sv_offset dictionary.
We need to request a new file from ASL or Rutgers folks to make sure we still have tests covering files without temperature/pressure/tilt data (those on glider).

This is to address #1300.

### DIFF
--- a/echopype/convert/parse_azfp.py
+++ b/echopype/convert/parse_azfp.py
@@ -13,6 +13,42 @@ from .parse_base import ParseBase
 
 FILENAME_DATETIME_AZFP = "\\w+.01A"
 
+# Common Sv_offset values for frequency > 38 kHz
+SV_OFFSET_HF = {
+    300: 1.1,
+    500: 0.8,
+    700: 0.5,
+    900: 0.3,
+    1000: 0.3,
+}
+SV_OFFSET_LF = {
+    500: 1.1,
+    1000: 0.7,
+}
+SV_OFFSET = {
+    38000.0: {**SV_OFFSET_LF},
+    67000.0: {
+        500: 1.1,
+        **SV_OFFSET_HF,
+    },
+    125000.0: {
+        150: 1.4,
+        250: 1.3,
+        **SV_OFFSET_HF,
+    },
+    200000.0: {
+        150: 1.4,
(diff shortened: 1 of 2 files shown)


ctuguinay (approved):
I think the changes look good. I am a bit curious about its usage though.

I'm looking at the code found [here](https://github.com/OSOceanAcoustics/echopype/blob/d7a6e7dc1dc31a76a4b4e2ef72fb12232d6f589f/echopype/convert/set_groups_azfp.py#L580C1-L587C1) in `parse_groups_azfp.py`  `set_vendor` function:

```
        # Build variables in the output xarray Dataset
        Sv_offset = np.zeros_like(self.freq_sorted)
        for ind, ich in enumerate(self.freq_ind_sorted):
            # TODO: should not access the private function, better to compute Sv_offset in parser
            Sv_offset[ind] = self.parser_obj._calc_Sv_offset(
                self.freq_sorted[ind], unpacked_data["pulse_len"][ich]
            )
```

Perhaps it can be an independent function, but still in `parse_azfp.py`, or we can just (as the TODO says), create the `Sv_offset` in the parser and pack it in `parser_obj.unpacked_data`? I think the latter would be much nicer since `set_vendor` should just be packing the vendor datasets. Maybe a separate PR once the test data has arrived.

leewujung:
> I think the latter would be much nicer since `set_vendor` should just be packing the vendor datasets, and it also gives the function an actual purpose within the parser class, rather than just being in it for encapsulation.

Yep, it'll be nice to have that in the parser so that we can keep the set group methods clean. Do you want to make a PR for that change? 😛

ctuguinay:
@leewujung Yeah I can do that. Perhaps, I do that PR in conjunction with when the test data arrives, so it's a little more than a few lines changed?

leewujung:
Sure, let me email ASL and cc you.

PR #1311, opened 2024-04-24
Support lazy-loaded EK80 broadband-complex data

leewujung:
This PR addresses the issues with for "Indexing with a boolean dask array is not allowed" encountered when calibrating EK80 complex data that is lazy-loaded.

This issue essentially calls for forcing to compute in the places where such indexing occurs. The cases I've seen (and "fixed") so far are using boolean indexing to select channels (based on data from the first ping), so the impact on performance is likely small. Right now this is fine since we restrict the use case to when the transmit signals are identical across all `ping_time` (but can vary across `channel`). Performance issue may arise when we want to accommodate transmit signal change across `ping _time` (#925).

I also adjusted one place in the convolution code to allow dask to function in parallel, by rechunking the `range_sample` dimension into 1 chunk.

**Note: this is not yet ready for review.**

### DIFF
--- a/echopype/calibrate/ek80_complex.py
+++ b/echopype/calibrate/ek80_complex.py
@@ -294,20 +294,27 @@ def compress_pulse(backscatter: xr.DataArray, chirp: Dict) -> xr.DataArray:
         replica = np.flipud(np.conj(tx))
 
         # Apply convolve on backscatter (along range sample dimension) and replica
+        # Rechunking backscatter_chan is needed to avoid the following ValueError:
+        #    ValueError: dimension range_sample on 0th function argument to apply_ufunc 
+        #    with dask='parallelized' consists of multiple chunks, but is also a core dimension.
+        #    To fix, either rechunk into a single array chunk along this dimension,
+        #    i.e., ``.chunk(dict(range_sample=-1))``,
+        #    or pass ``allow_rechunk=True`` in ``dask_gufunc_kwargs`` 
+        #    but beware that this may significantly increase memory usage.
         pc = xr.apply_ufunc(
             lambda m: (signal.convolve(m, replica, mode="full")[replica.size - 1 :]),
-            backscatter_chan,
+            backscatter_chan.chunk({"range_sample": -1}),
             input_core_dims=[["range_sample"]],
             output_core_dims=[["range_sample"]],
             dask="parallelized",
             vectorize=True,
             output_dtypes=[np.complex64],
-        ).compute()
+        )#.compute()
 
         # Restore NaN values in the resulting array.
         # Computing of `nan_mask` here is necessary in the case when `nan_mask` is lazy loaded
         # or else the resulting `pc` will also be lazy loaded.
-        pc = xr.where(nan_mask.compute(), np.nan, pc)
+        pc = xr.where(nan_mask, np.nan, pc)
 
         pc_all.append(pc)
 
--- a/echopype/calibrate/ek80_complex.py
+++ b/echopype/calibrate/ek80_complex.py
@@ -280,13 +280,13 @@ def _convolve_per_channel(backscatter_subset: np.ndarray, replica_dict: dict, ch
         # Create zeros like array from `backscatter_subset`
         convolved = np.zeros_like(backscatter_subset, dtype=np.complex64)
         # Iterate over channels
-        for i, channel in enumerate(channels):
+        for ch_seq, channel in enumerate(channels):
             # Extract replica values
             replica = replica_dict[str(channel.values)]
             # Convolve backscatter and chirp replica
-            convolved[:, i] = signal.convolve(backscatter_subset[:, i], replica, mode="full")[
-                replica.size - 1 :
-            ]
+            convolved[:, ch_seq] = signal.convolve(
+                backscatter_subset[:, ch_seq], replica, mode="full"
+            )[replica.size - 1 :]
         return convolved
 
 
@@ -326,9 +326,13 @@ def compress_pulse(backscatter: xr.DataArray, chirp: Dict) -> xr.DataArray:
     # Apply convolve on backscatter and replica (along range sample and channel dimension):
     # To enable parallelized computation with `dask='parallelized'`, we rechunk to ensure that
     #  the data is chunked with only one chunk along the core dimensions.
+    if backscatter_with_zeroed_nans.chunks is not None:
+        backscatter_with_zeroed_nans = backscatter_with_zeroed_nans.chunk(
+            {"range_sample": -1, "channel": -1}
+        )
     pc = xr.apply_ufunc(
         _convolve_per_channel_partial,
-        backscatter_with_zeroed_nans.chunk({"range_sample": -1, "channel": -1}),
+        backscatter_with_zeroed_nans,
         input_core_dims=[["range_sample", "channel"]],
         output_core_dims=[["range_sample", "channel"]],
         dask="parallelized",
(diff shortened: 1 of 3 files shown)


leewujung:
The failed tests are a little weird. I've just looked into one of them: `test_ek80_BB_power_from_complex`. My changes didn't actually change the type of operations being performed on the data, but the convolution output seems different.

ctuguinay:
@leewujung if everything here looks good, could you merge this?

echopype/calibrate/ek80_complex.py
```diff
@@ -273,47 +285,53 @@ def compress_pulse(backscatter: xr.DataArray, chirp: Dict) -> xr.DataArray:
...
+        backscatter_NaN_beam_drop = backscatter.sel(channel=channel).dropna(dim="beam", how="all")
+        backscatter_NaN_beam_drop_all.append(backscatter_NaN_beam_drop)
+
+        # Extract tx
+        tx = chirp[str(channel.values)]
         # Compute complex conjugate of transmit values and reverse order of elements
-        replica = np.flipud(np.conj(tx))
+        replica_dict[str(channel.values)] = np.flipud(np.conj(tx))
+    # Concatenate backscatter channels with dropped NaN beam dimensions.
+    backscatter_NaN_beam_drop_all = xr.concat(backscatter_NaN_beam_drop_all, dim="channel")
```
leewujung, line 303: Question: I think the original intention to dropna with `dim="beam"` is to remove the padded sections along the `range_sample` dimension, since some channels contain data that are shorter. So, if you dropna for a particular channel, but other channels are longer, wouldn't this `xr.concat` recreate the dropped NaN values?

ctuguinay: Oh yeah, I think you're right. Let me think and tinker with this.

ctuguinay: CODE:

```
data = np.array([
    [np.nan, np.nan], [4, 5], [np.nan, 6], [8, 9]
])
da = xr.DataArray(
    data,
    dims=["channel", "beam"],
    coords={
        "channel": np.arange(4),
        "beam": np.arange(2)
    })
dropped_nan_beam_da_list = []
for channel_index in range(len(da["channel"])):
    dropped_da = da.isel(channel=channel_index).dropna("beam", how="all")
    print(f"`dropped_da` shape (chan idx {channel_index}): {dropped_da.shape[0]}")
    dropped_nan_beam_da_list.append(dropped_da)
print("concatenated da:", xr.concat(dropped_nan_beam_da_list, dim="channel"))
```

OUTPUT:

```
`dropped_da` shape (chan idx 0): 0
`dropped_da` shape (chan idx 1): 2
`dropped_da` shape (chan idx 2): 1
`dropped_da` shape (chan idx 3): 2
concatenated da: <xarray.DataArray (channel: 4, beam: 2)> Size: 64B
array([[nan, nan],
       [ 4.,  5.],
       [nan,  6.],
       [ 8.,  9.]])
Coordinates:
  * beam     (beam) int64 16B 0 1
  * channel  (channel) int64 32B 0 1 2 3
```

ctuguinay: So yeah you're right, this padding does happen. Let me tinker with the implementation that's currently in this PR. There's probably a clever way skip to skip this concat.

leewujung: Oh, I think your code above is missing the `range_sample` dimension, which is where that padding happens.

But yeah, I think the easiest is to just not dropna here. We have 2 options:
- try drop the NaNs in the `_convolve_per_channel` function, since there you do have a loop that is channel by channel. The only thing is that you have to drop the NaN and add them back to ensure the output length is the same
- just pay the computational price

Maybe you could try seeing how the speed differ for these 2 approaches? It is possible that with all the optimization people have put into convolution, the speed difference is small (and hopefully the memory consumption is reasonable too).

ctuguinay: That sounds good, I'll try it out. I have a suspicion that the 2nd option of just not dropping NaNs might be faster, since the time needed to convolve on all NaNs may be faster than the time needed to remove and add back NaNs.

leewujung: Yeah, I agree, it very possibly would be that way!

ctuguinay:
Recent changes will come from: https://github.com/leewujung/echopype/pull/37

echopype/calibrate/ek80_complex.py
```diff
@@ -258,6 +259,37 @@ def get_transmit_signal(
...
+        # Create zeros like array from `backscatter_subset`
+        convolved = np.zeros_like(backscatter_subset, dtype=np.complex64)
+        # Iterate over channels
+        for i, channel in enumerate(channels):
+            # Extract replica values
+            replica = replica_dict[str(channel.values)]
+            # Convolve backscatter and chirp replica
+            convolved[:, i] = signal.convolve(backscatter_subset[:, i], replica, mode="full")[
+                replica.size - 1 :
+            ]
```
leewujung, line 289: This is tiny thing but I wonder it may be good to change `i` to maybe `ch_seq` or something, so that it is clear what that is, and also does not run into the confusion with the imaginary part being i or j, etc.

echopype/calibrate/ek80_complex.py
```diff
@@ -273,47 +305,41 @@ def compress_pulse(backscatter: xr.DataArray, chirp: Dict) -> xr.DataArray:
...
+        channels=backscatter_with_zeroed_nans["channel"],
+    )
 
-        pc_all.append(pc)
+    # Apply convolve on backscatter and replica (along range sample and channel dimension):
+    # To enable parallelized computation with `dask='parallelized'`, we rechunk to ensure that
+    #  the data is chunked with only one chunk along the core dimensions.
+    pc = xr.apply_ufunc(
+        _convolve_per_channel_partial,
+        backscatter_with_zeroed_nans.chunk({"range_sample": -1, "channel": -1}),
```
leewujung, line 331: Here I wonder if it makes difference if you make one chunk per channel, like how you did it in the noise filter case.

echopype/tests/calibrate/test_calibrate_ek80.py
```diff
@@ -306,10 +306,7 @@ def test_ek80_BB_power_compute_Sv(
         encode_mode=encode_mode,
     )
     pyel_vals = pyel_BB_p_data["sv_data"]
-    if dask_array:
-        ep_vals = ds_Sv["Sv"].sel(channel=ch_sel).squeeze().data.compute()
-    else:
-        ep_vals = ds_Sv["Sv"].sel(channel=ch_sel).squeeze().data
+    ep_vals = ds_Sv["Sv"].sel(channel=ch_sel).squeeze().data.compute()
```
leewujung, line 309: Does this mean that right now with the `dask_array=False` test case the data is still lazy-loaded? Asking because I seem to remember the original intention of this is to see if lazy-loading works explicitly, but I could also just be misremembering things.

echopype/calibrate/ek80_complex.py
ctuguinay: Since it's a core dimension in `apply_ufunc`, I have to chunk it into a single chunk. It is enforced in this case.

echopype/calibrate/ek80_complex.py
ctuguinay: I agree I'll make this change

echopype/tests/calibrate/test_calibrate_ek80.py
ctuguinay: Ah yeah, it's because of the `.chunk({"range_sample": -1, "channel": -1})` that's now in the convolution. Let me make this chunk conditional on whether or not the backscatter variable is already chunked.

ctuguinay:
@leewujung Thanks! Should be ready for another short review haha

echopype/calibrate/ek80_complex.py
leewujung: Ah I see, ok. Thanks! We can see how things look with large datasets when we get to that.

leewujung:
Awesome! Thanks @ctuguinay, this looks great. I'll merge it now.

Issue #1314, opened 2024-04-30
Implement Noise and Attenuation Functions from Ryan et. al. 2015


ctuguinay:
From [Reducing bias due to noise and attenuation in open-ocean echo integration data](https://academic.oup.com/icesjms/article/72/8/2482/2459093), implement the following noise filters:

- [x] IN filter
- [x] AS filter
- [x] TN filter

Implementation of some (if not all) of these filters can be found in [echopy](https://github.com/open-ocean-sounding/echopy) written in pure numpy. We want to implement these filters using Xarray, and if Dask Delaying is not too involved, implement that as well. If not, we can write a separate issue for that.

ctuguinay:
Related to #1211, since the existing noise functions will have to be renamed to something more specific.

PR #1316, opened 2024-05-04
Add noise and attenuation functions from Ryan et. al. 2015 and echopy

ctuguinay:
Addresses #1314


ctuguinay:
@leewujung I might think of more stuff to change over the weekend, but I think this is at a good place now. I also added `np.digitize` for upsampling the impulse noise array.

echopype/clean/api.py
```diff
@@ -2,74 +2,380 @@
...
+    Locate and create a mask for impulse noise using a ping-wise two-sided comparison.
+
+    Parameters
+    ----------
+    ds_Sv : xarray.Dataset
+        Calibrated Sv data with depth data variable.
+    depth_bin : str, default `5`m
+        Donwsampling bin size along ``depth`` in meters.
+    num_side_pings : int, default `1`
+        Number of side pings to look at for the two-side comparison.
```
leewujung, line 118: the paper recommended setting the default to 2 -- check their Table 2. Maybe we use 2?

echopype/clean/utils.py
```diff
@@ -0,0 +1,228 @@
...
+            (ping_time_idx - num_pings < 0)
+            | (ping_time_idx + num_pings > Sv.shape[0] - 1)
+            | np.all(np.isnan(Sv[ping_time_idx, up:lw]))
+        ):
+            # Compare ping and block medians, and mask ping if difference greater than
+            # threshold.
+            pingmedian = _lin2log(np.nanmedian(_log2lin(Sv[ping_time_idx, up:lw])))
+            blockmedian = _lin2log(
+                np.nanmedian(
+                    _log2lin(Sv[(ping_time_idx - num_pings) : (ping_time_idx + num_pings), up:lw])
```
leewujung, line 198: so here `num_pings` actually corresponds to $n/2$ in the $S_{v_n}$ in equation 2 -- is this "inherited" from echopy?

leewujung:
Hey @ctuguinay : Thanks for the PR! Looking at the paper again, I feel that the none of the operations is really so locked in with the requirement of finding the accurate depth (or ping time) bin edges, which is the source of the performance bottleneck, especially the TN one as you pointed out.

All of the criteria are built on the concept of comparing with neighbors, and the binning from accurate bin edges (from the coordinate labels) vs inaccurate bin edges (from just the indices) would likely result in pretty small differences. If the noise needs to be masked, it would get masked either way; if the noise is borderline, the masking is likely be at least equally sensitive to other independent parameter changes (such as changing the averaging bin size along depth or across pings).

What do you think about keeping what you have, but implement a "fast" version that is based on indices? That can be another PR so we can get this merged first?

echopype/clean/utils.py
```diff
@@ -0,0 +1,228 @@
...
+        upsample_mapping.compute(),
+        input_core_dims=[["depth_bins"], ["range_sample"]],
+        output_core_dims=[["range_sample"]],
+        exclude_dims=set(["depth_bins", "range_sample"]),
+        dask="parallelized",
+        vectorize=True,
+        output_dtypes=[np.float64],
+    )
+    # Set range sample coords
+    upsampled_Sv = upsampled_Sv.assign_coords(range_sample=ds_Sv["range_sample"])
```
leewujung, line 145: I think you can use `xr.reindex_like` to achieve this, in the same way as what we talked about the other day, for the section in remove background noise to do forward fill after changing the coordinate of the downsampled array to the first index of each bin (which you have from `range_interval`).

Roughly (i have not tested these, but i am sure you get the gist):
```python
# Downsample Sv along range sample
downsampled_Sv = ...

# Substitute coordinate value in the downsampled Sv to be first element of each bin
downsampled_Sv.assign_coord({"depth": [FIRST_ELEMENT_OF_EACH_BIN]})

# Upsample via reindex_like
upsampled_Sv = downsampled_Sv.reindex_like(ds_Sv["Sv"], method="ffill")
```

echopype/tests/clean/test_noise.py
```diff
@@ -2,11 +2,433 @@
...
+        pooled_Sv.where(~within_mask)
+    )
+
+    # Check that NaN is the only pool boundary unique value
+    assert np.isclose(unique_pool_boundary_values, np.array([np.nan]), equal_nan=True)
+
+    # Compute Sv
+    ds_Sv["Sv"] = ds_Sv["Sv"].compute()
+
+    # Manually check correct binning and aggregation values
```
leewujung, line 180: For the checks here: are you doing spot-check only because the test would run for too long? In the above you actually calculated `pooled_Sv`, so I am just curious why you didn't end up check all values.

leewujung:
I just went through the tests and they look in good shape, I just have some small comments. One thing I didn't find was that for TN, there isn't an equivalent for testing the output mask like for IN and AS. It would be good to have that, to confirm that the function works as intended. if computation speed is an issue, you can "crop" the test dataset to only a subset of pings that go across a section containing transient noise, so that it'll run much faster.

echopype/clean/api.py
ctuguinay: Ah yeah I see that n=2 note below their table 2. I'll use that insetad.

echopype/clean/utils.py
ctuguinay: Ah yeah, I see what you mean. I can use the fact that depth is monotonic to `ffill`. I'll add that.

echopype/clean/utils.py
ctuguinay: Yeah, this is inherited directly from `echopy`. Maybe instead of `num_ping` I'll rename this to `num_side_pings`, following the convention from transient noise masking. I think it's kind of hard to work with this `n/2` since we would have to enforce even numbers and the user would have to do that `n/2` calculation in their head.

echopype/tests/clean/test_noise.py
ctuguinay: Oh, I am doing spot-checking because the code I wrote for computing `pooled_Sv` was the simplest implementation I could think of for computing it programmatically at the time. So my thought process then was to just manually check a few values. But now that I look at it, the array is small enough that I can manually check all values, which I'll add in this revision.

echopype/clean/utils.py
ctuguinay: Since we are assuming that depth and subsequently depth bins are not uniform, I had to do this `reindex` upsampling per channel, per ping time, in a nested for loop.

ctuguinay:
@leewujung Thanks for the review! This should be ready to look at again. The only suggestion I deviated a bit from was the suggestion for upsampling using `reindex`, since `range_sample` assignment is not uniform across channel and ping time (since depth and depth bins aren't either), I had to reindex per channel, per ping time.

I also created issue #1321 for the simple index version of these functions.

echopype/clean/utils.py
leewujung: I don't think the ffill approach will be a problem for non-uniform depths. After all it is jus the left bin edges (which is the first index you put into that `[FIRST_ELEMENT_OF_EACH_BIN]` list) that matters when you do ffill.

leewujung: oh i just realized what you mean -- it is when across ping_time the depth is different, then that becomes a problem. Ok, fair enough. But this is a weird messy situation, sigh. OK let's leave it as is, and just speed up in the assume-uniform (index only) case in a separate PR.

ctuguinay: Pretty unfortunate :/Should be much faster (and simpler) using reindex after any coarsen operation

ctuguinay: I wonder what your opinion is on if what I had initially is cleaner/better, with the ufunc and the index assignment mapping, or what I have now with reindex nested for-loop

leewujung: I am not sure... because I don't remember what the last one was. You can benchmark the performance? It would be good to know also.

leewujung (approved):
Looks good! Will merge this now.

PR #1319, opened 2024-05-22
Use Echodata `Platform` and `Beam_group1` Variables to Compute Depth for EK60/EK80

ctuguinay:
Based on Emilio's #1205 PR. Addresses #790 and #1152.


ctuguinay:
@leewujung This should be ready for review. The only changes I made from the last time we talked were how to deal with choosing `Beam_group1` vs `Beam_group2`, and interpolating transducer depth `time2` to `ping_time` if and only if they are not equal.

echopype/consolidate/api.py
```diff
@@ -55,84 +60,123 @@ def swap_dims_channel_frequency(ds: Union[xr.Dataset, str, pathlib.Path]) -> xr.
...
+        If True, use Echodata Platform group vertical offset values to compute transducer depth.
+        Currently only implemented for EK60/EK80 sonar models.
+    use_platform_angles: bool, `False`
+        If True, use Echodata Platform group angle values to compute `echo_range` scaling values.
+        Currently only implemented for EK60/EK80 sonar models.
+        Cannot be used in tandem with Echodata Beam group angles.
+    use_beam_angles: bool `False`
+        If True, use Echodata Beam group angle values to compute `echo_range` scaling values.
+        Currently only implemented for EK60/EK80 sonar models.
+        Cannot be used in tandem with Echodata Platform group angles.
```
leewujung, line 103: is the following correct?
```suggestion
    use_platform_angles: bool, `False`
        If True, use Echodata Platform group angle values to compute `echo_range` scaling values.
        Currently only implemented for EK60/EK80 sonar models.
        In the current implementation cannot be used in tandem with `use_beam_angles`.
    use_beam_angles: bool `False`
        If True, use Echodata Beam group angle values to compute `echo_range` scaling values.
        Currently only implemented for EK60/EK80 sonar models.
        In the current implementation cannot be used in tandem with `use_platform_angles`.
```

echopype/consolidate/api.py
```diff
@@ -55,84 +60,123 @@ def swap_dims_channel_frequency(ds: Union[xr.Dataset, str, pathlib.Path]) -> xr.
...
-    Notes
-    -----
-    Currently this function only scalar inputs of depth_offset and tilt angle.
-    In future expansion we plan to add the following options:
+    # Raise `ValueError` if `echodata` is needed but not passed in
+    if (not echodata) and (use_platform_vertical_offsets or use_platform_angles or use_beam_angles):
+        raise ValueError(
+            "If any of `use_platform_vertical_offsets` or `use_platform_angles` is `True` "
+            + "then `echodata` cannot be `None`."
+        )
```
leewujung, line 117:
```suggestion
    # Raise `ValueError` if `echodata` is needed but not passed in
    if (not echodata) and (use_platform_vertical_offsets or use_platform_angles or use_beam_angles):
        raise ValueError(
            "If any of `use_platform_vertical_offsets`, "
            + "`use_platform_angles` "
            + "or `use_beam_angles` is `True`, "
            + "then `echodata` cannot be `None`."
        )
```

echopype/consolidate/api.py
```diff
@@ -55,84 +60,123 @@ def swap_dims_channel_frequency(ds: Union[xr.Dataset, str, pathlib.Path]) -> xr.
...
+        # Open Echodata
+        echodata = open_source(echodata, "echodata", {})
 
-    # Compute depth
-    ds["depth"] = mult * ds["echo_range"] * np.cos(tilt / 180 * np.pi) + depth_offset
-    ds["depth"].attrs = {"long_name": "Depth", "standard_name": "depth", "units": "m"}
+        # Grab sonar model
+        sonar_model = echodata["Sonar"].attrs["sonar_model"]
+
+        # Raise value error if sonar model is supported for `use_platform/beam_...` arguments
```
leewujung, line 133: In the comment, do you mean "not supported"?

echopype/consolidate/api.py
```diff
@@ -55,84 +60,123 @@ def swap_dims_channel_frequency(ds: Union[xr.Dataset, str, pathlib.Path]) -> xr.
...
-    tilt : float
-        Transducer tilt angle [degree].
-        Default is 0 (transducer mounted vertically).
-    downward : bool
-        Whether or not the transducers point downward.
-        Default to True.
+        Will only be used if `use_platform_vertical_offsets=False`.
+    tilt : float, default `0.0`
+        Transducer tilt angle [degree]. 0 corresponds to a transducer pointing vertically.
+        Will only be  used if both `use_platform_angles=False` and `use_beam_angles=False`.
```
leewujung, line 90: I have a bit of different opinion here -- I feel that `tilt` should overwrite what's in the EchoData Platform/Beam group, just like how the `cal_params` or `env_params` given by the users would overwrite that's in the EchoData Environment group during calibration.

echopype/consolidate/api.py
```diff
@@ -55,84 +60,123 @@ def swap_dims_channel_frequency(ds: Union[xr.Dataset, str, pathlib.Path]) -> xr.
...
-    #                 "Required when vertical=False"
-    #             )
-    # else:
-    #     tilt = 0
+    # Raise `NotImplementedError` if `use_platform_angles` and `use_beam_angles` are
+    # both true.
+    if use_platform_angles and use_beam_angles:
+        raise NotImplementedError(
+            "Computing depth with both platform and beam angles is not implemented yet."
+        )
```
leewujung, line 124: I think we should also add an explicit check for `tilt` here if either `use_platform_angles` or `use_beam_angles` are `True` when `tilt` is also given. Also see my comment above about `tilt` taking precedence over `use_platform_angles` or `use_beam_angles`.

echopype/consolidate/api.py
```diff
@@ -55,84 +60,123 @@ def swap_dims_channel_frequency(ds: Union[xr.Dataset, str, pathlib.Path]) -> xr.
...
-        to which a depth variable will be added.
-        Must contain `echo_range`.
-    depth_offset : float
+        Source Sv dataset to which a depth variable will be added.
+    echodata : EchoData or str or pathlib.Path, optional, default `None`
+        `EchoData` object from which the `Sv` dataset originated.
+    depth_offset : float, default `0.0`
         Offset along the vertical (depth) dimension to account for actual transducer
         position in water, since `echo_range` is counted from transducer surface.
-        Default is 0.
```
leewujung, line 80: Similar to `tilt`, I think we should also let `depth_offset` taking precedence over `use_platform_vertical_offsets`.

echopype/consolidate/api.py
```diff
@@ -55,84 +60,123 @@ def swap_dims_channel_frequency(ds: Union[xr.Dataset, str, pathlib.Path]) -> xr.
...
+        if echodata["Sonar/Beam_group1"]["channel"].equals(ds["channel"]):
+            beam_group_name = "Beam_group1"
+        else:
+            beam_group_name = "Beam_group2"
+
+        # Compute echo range scaling in EK systems using beam angle data
+        echo_range_scaling = ek_use_beam_angles(echodata[f"Sonar/{beam_group_name}"], ds["channel"])
+    else:
+        # Compute echo range scaling from user input tilt
+        echo_range_scaling = np.cos(np.deg2rad(tilt))
```
leewujung, line 164: So a few thoughts:
- I think you can remove those conditions since you've checked `sonar_model` being one of EK60/EK80.
- Since you already checked the values of `tilt` and `use_platform_angles` and `use_beam_angles` and issue error or warnings, I think the structure could be simplified to be just `if` and no `else`, which makes the code easier to read.

With `depth_offset` and `tilt` taking precedence over the vertical offset and angles in the EchoData object, I think the logic here needs some rework. I'd suggest something like below as a clean decision tree structure:
```python
if depth_offset:
  # use depth offset since it takes precedence
else:
  # use vertical offset in EchoData if that's what the user wants and sonar_model is EK60/80
  if sonar_model in ["EK60", "EK80"]:
    if use_platform_angles:
      ...
    if use_beam_angles:
      ...
  else:
    ...

if tilt:
  # use tilt since it takes precedence
else:
  # use angles in EchoData if that's what the user wants and sonar_model is EK60/80
  if sonar_model in ["EK60", "EK80"]:
    ...
  else:
    ...

```
I've not tested the above so please check the tree is in good order.

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -147,36 +151,430 @@ def _build_ds_Sv(channel, range_sample, ping_time, sample_interval):
     )
 
 
-def test_add_depth():
+@pytest.mark.integration
+@pytest.mark.parametrize("file, sonar_model, compute_Sv_kwaargs", [
```
leewujung, line 155: compute_Sv_kwaargs --> compute_Sv_kwargs

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -147,36 +151,430 @@ def _build_ds_Sv(channel, range_sample, ping_time, sample_interval):
...
+])
+def test_add_depth_EK_with_platform_angles(file, sonar_model, compute_Sv_kwaargs):
+    """Test `depth` values when using EK Platform angles to compute it."""
+    # Open EK Raw file and Compute Sv
+    ed = ep.open_raw(file, sonar_model=sonar_model)
+    ds_Sv = ep.calibrate.compute_Sv(ed, **compute_Sv_kwaargs)
+
+    # Replace any Beam Angle NaN values with 0
+    ed["Platform"]["pitch"] = ed["Platform"]["pitch"].fillna(0)
+    ed["Platform"]["roll"] = ed["Platform"]["roll"].fillna(0)
```
leewujung, line 433: Are these just adding them so that the add_depth computation would go through?

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -147,36 +151,430 @@ def _build_ds_Sv(channel, range_sample, ping_time, sample_interval):
...
+def test_add_depth_EK_with_beam_angles(file, sonar_model, compute_Sv_kwaargs):
+    """Test `depth` values when using EK Beam angles to compute it."""
+    # Open EK Raw file and Compute Sv
+    ed = ep.open_raw(file, sonar_model=sonar_model)
+    ds_Sv = ep.calibrate.compute_Sv(ed, **compute_Sv_kwaargs)
+
+    # Replace any Beam Angle NaN values with 0
+    ed["Sonar/Beam_group1"]["beam_direction_x"] = ed["Sonar/Beam_group1"]["beam_direction_x"].fillna(0)
+    ed["Sonar/Beam_group1"]["beam_direction_y"] = ed["Sonar/Beam_group1"]["beam_direction_y"].fillna(0)
+    ed["Sonar/Beam_group1"]["beam_direction_z"] = ed["Sonar/Beam_group1"]["beam_direction_z"].fillna(0)
```
leewujung, line 503: I suspect that you use these so that the add_depth computation would go through? Different from the pitch, roll angles though, a beam direction vector (0, 0, 0) does not make sense as a direction vector, right? The length of (0, 0, 0) would be 0, but a direction vector should have a length 1, right?

leewujung (changes requested):
Hey @ctuguinay : Thanks for the PR. Sorry it took a while to get to this. I commented on quite a few things, and can take another pass once you've made those changes. 

One thing that I think we can address in another PR down the road is that all these tests are integration tests, because we are forced to do `open_raw` and `compute_Sv`. I think Don had put in mechanisms to mock an EchoData object so that we can properly do unit test of these downstream functions. It is relatively easy to mock an Sv dataset. We can use mocker patch to mock things too. But yeah too much for this PR, and let's just keep these integration tests as they are. I'll add an issue to track this for later.

echopype/consolidate/api.py
ctuguinay: Yup I'll add this

echopype/consolidate/api.py
ctuguinay: I agree I'll make this change 👍

echopype/tests/consolidate/test_consolidate_integration.py
ctuguinay: Yeah, else the values would just be NaN.

echopype/tests/consolidate/test_consolidate_integration.py
ctuguinay: I agree yeah that doesn't make sense. I'll add a test to check that `ek_use_beam_angles` outputs what it should output.

echopype/consolidate/api.py
ctuguinay: Yup I'll make that change 👍

echopype/consolidate/api.py
ctuguinay: I agree 👍

ctuguinay:
@leewujung Thanks for the review! This should be ready to review again

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -147,36 +151,523 @@ def _build_ds_Sv(channel, range_sample, ping_time, sample_interval):
...
+            ),
+            "roll": xr.DataArray(
+                [0, 90, 0],
+                dims=("ping_time")
+            ),
+        },
+        coords={"ping_time": ping_time_da}
+    )
+    echo_range_scaling = ep.consolidate.ek_depth_utils.ek_use_platform_angles(platform_ds, ping_time_da)
+    assert np.allclose(echo_range_scaling.values, np.array([0.0, 0.0, 1.0]))
```
leewujung, line 181:
```suggestion
    # In ping time 1 and 2, the platform is completely sideways so the echo range scaling
    # should be 0 (i.e zeros out the entire depth).
    # In ping time 3, the platform is completely vertical so the echo range scaling should
    # be 1 (i.e no change).
    # In ping time 4, the platform is tilted by 45 deg so the echo range scaling should
    # be 1/sqrt(2).
    platform_ds = xr.Dataset(
        {
            "pitch": xr.DataArray(
                [-90, 0, 0, -45],
                dims=("ping_time")
            ),
            "roll": xr.DataArray(
                [0, 90, 0, 0],
                dims=("ping_time")
            ),
        },
        coords={"ping_time": ping_time_da}
    )
    echo_range_scaling = ep.consolidate.ek_depth_utils.ek_use_platform_angles(platform_ds, ping_time_da)
    assert np.allclose(echo_range_scaling.values, np.array([0.0, 0.0, 1.0, 1/np.sqrt(2)]))
```

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -147,36 +151,523 @@ def _build_ds_Sv(channel, range_sample, ping_time, sample_interval):
...
+    beam_ds = xr.Dataset(
+        {
+            "beam_direction_x": xr.DataArray([1, 0, 0], dims=("channel")),
+            "beam_direction_y": xr.DataArray([0, 1, 0], dims=("channel")),
+            "beam_direction_z": xr.DataArray([0, 0, 1], dims=("channel")),
+        },
+        coords={"channel": channel_da}
+    )
+    echo_range_scaling = ep.consolidate.ek_depth_utils.ek_use_beam_angles(beam_ds)
+    assert np.allclose(echo_range_scaling.values, np.array([0.0, 0.0, 1.0]))
```
leewujung, line 204:
```suggestion
    # Create a Dataset with channel-specific beam direction vectors
    # In channels 1 and 2, the transducers are not pointing vertically at all so the echo
    # range scaling should be 0 (i.e zeros out the entire depth).
    # In channel 3, the transducer is completely vertical so the echo range scaling should
    # be 1 (i.e no change).
    # In channel 4, the transducer is tilted to the x direction by 30 deg, so the
    # echo range scaling should be sqrt(3)/2.
    channel_da = xr.DataArray(["chan1", "chan2", "chan3"], dims=("channel"))
    beam_ds = xr.Dataset(
        {
            "beam_direction_x": xr.DataArray([1, 0, 0, 1/2], dims=("channel")),
            "beam_direction_y": xr.DataArray([0, 1, 0, 0], dims=("channel")),
            "beam_direction_z": xr.DataArray([0, 0, 1, np.sqrt(3)/2], dims=("channel")),
        },
        coords={"channel": channel_da}
    )
    echo_range_scaling = ep.consolidate.ek_depth_utils.ek_use_beam_angles(beam_ds)
    assert np.allclose(echo_range_scaling.values, np.array([0.0, 0.0, 1.0, np.sqrt(3)/2]))
```

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -147,36 +151,523 @@ def _build_ds_Sv(channel, range_sample, ping_time, sample_interval):
...
+    )
+
+    # Check if the expected warnings are logged
+    depth_offset_warning = (
+        "When `depth_offset` is specified, platform vertical offset "
+        "variables will not be used."
+    )
+    tilt_warning = (
+        "When `tilt` is specified, beam/platform angle variables will " "not be used."
+    )
```
leewujung, line 351:
```suggestion
    tilt_warning = (
        "When `tilt` is specified, beam/platform angle variables will not be used."
    )
```

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -147,36 +151,523 @@ def _build_ds_Sv(channel, range_sample, ping_time, sample_interval):
...
+                channel=channel_index, ping_time=ping_time_index
+            )
+            # Check if depth vector is equal to echo range scaling value * echo range
+            assert np.all(
+                np.isclose(
+                    (depth_vector).data,
+                    (echo_range_scaling_value * echo_range_vector).data, 
+                    equal_nan=True
+                )
+            )
```
leewujung, line 561: Similar to [my comment above](https://github.com/OSOceanAcoustics/echopype/pull/1319/files#r1666435907), but it is also possible that i missed something.

leewujung: also applies to the same block under `test_add_depth_EK_with_beam_angles`

leewujung (dismissed):
@ctuguinay : thanks for the changes! Just a few small suggestions in the test, but I think this is good to go!

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -147,36 +151,523 @@ def _build_ds_Sv(channel, range_sample, ping_time, sample_interval):
...
+                ).data
+                # Check if depth value is equal to corresponding `echo_range` value + transducer depth value
+                assert np.isclose(
+                    depth_value,
+                    ds_Sv["echo_range"].isel(
+                        channel=channel_index, ping_time=ping_time_index, range_sample=range_sample_index
+                    ).data + transducer_depth_value,
+                    rtol=1e-10,
+                    atol=1e-10
+                )
```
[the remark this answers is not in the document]
ctuguinay: Yeah I think I can even do the entire array (channel, ping time, range sample)

ctuguinay:
@leewujung Thanks for the review! This should be ready for another one.

leewujung (approved):
@ctuguinay : the changes look great and we can merge this now! Thanks for putting in `_var_time2_to_ping_time`. I'll add an issue about your idea of adding `echodata.align_to_ping_time()` .

PR #1323, opened 2024-05-29
Added alpha-version code for new AZFP6 format

dash-uvic:
Added in initial code for the new AZFP6 sensor format for *.azfp files from ASL Environmental services.  The data is in a new format and has additional optional GPS information.  The XML file is now embedded into the data file.  No tests added.


leewujung:
Hey @dash-uvic : Thanks for the contribution! Could you fix the pre-commit errors? Also let us know what files you would like to add to the CI. As always, small files are better!

echopype/convert/set_groups_azfp6.py
```diff
@@ -0,0 +1,893 @@
...
+from typing import List
+
+import numpy as np
+import xarray as xr
+
+from ..utils.coding import set_time_encodings
+from .set_groups_base import SetGroupsBase
+
+
+class SetGroupsAZFP6(SetGroupsBase):
```
leewujung, line 14: @dash-uvic : It seems that a lot of the code here are copy-pasted here from `SetGroupsAZFP`? If that's the case, why don't you inherit from that and can only change the parts that are different? (Note I have not gone in detail to compare what was changed).

leewujung:
Hey @dash-uvic : Thanks for fixing the pre-commits.

I've just approved for the CI to run on this PR. If you could add file conversion tests to run on some example test files, that will allow us to review the changes more easily. 

The way we organized test data now is for you to share the files with us with the specific path you would like those to be in under `echopype/test_data/` (so these should correspond with what your tests use). We will add them to a google drive that the CI pulls automatically on each test run. This will change hopefully later this summer/fall to use GitHub Release Assets, but for now we will continue with this setup.

echopype/convert/set_groups_azfp6.py
dash-uvic: Since the hardware is still in flux (so Sv values, battery calculations, etc. may change) I decided to keep it separate for now, and then create a base class both ULS5/6 (both in groups and converter classes) would inherit from if we end up going with the current setup.  You're right, the AZFP6 group has not changed a lot, and I don't think it'll change much in the future so I rewrite that to inherit from the original AZFP group

leewujung: To make sure I understand, the approach you are taking is to have 2 completely separate class for now and then will merge the common components to have a base class for both after the hardware is settled?

dash-uvic: Yes, that's my intention.  Or if not a base class, inherit from the current AZFP class but as the class structure will remain the same, even if the internal function changes, so it makes sense to abstract and inherit.  If there is a preference for how you'd like the code structure, I'm flexible.

leewujung: I think both would be fine. It seems that inheriting from a common based class into separate ULS5/6 classes would make it pretty extensible (not that anyone wants to think about more changes at the moment... lol), so we can go with that. :)

leewujung:
@dash-uvic : I just "approved" the workflow to run - I think usually for a PR it just needs to be approved once and then it'll always run for new changes. If you see that stalled somehow feel free to ping me directly.

dash-uvic:
@leewujung What's the best method for adding/sending you the test files (average ~50M but there is a larger ~400M file)?

leewujung:
@dash-uvic : Is there a reason why the files need to be that large? If not, could you generated some smaller files? Large files will make CI runs a lot slower.

dash-uvic:
@leewujung Yeah I can probably truncate the larger azfp file down to fewer timesteps and remove unrequired fields from the expected output Mat file.  I think doing that I can get them down to ~10M if that's sufficient.

leewujung:
@dash-uvic : Yes ~10MB would be nice. Thanks!

dash-uvic:
@leewujung Ok sounds goods, I'll put them on a cloud and email you the link next week.

leewujung:
@dash-uvic : Thanks for sending the test files. Looks like the tests all pass now! Could you please fix that `pre-commit.ci` failure?

Do you think we should merge this afterwards and call it alpha, or do you have additional changes you would like to add?

dash-uvic:
@leewujung The pre-commit error has been fixed.  Strangely, it did not pop up when I ran pre-commit locally.  You can merge this as alpha.  I will likely have changes in the future but that won't be for a while and can be a new PR.

leewujung:
Great, thanks @dash-uvic! I wonder if that may be because of different versions of the pre-commit hooks.

I'll merge this now and add azfp6 alpha in the documentation. We will release v0.9.0 soon and this will be included in there. :)

Issue #1321, opened 2024-05-28
Use Simple Indexing for Noise and Attenuation Function Aggregations/Comparisons


ctuguinay:
This issue addresses the performance issue encountered in the #1314 noise and attenuation masking functions, which comes from the aggregations in the masking functions being those where the selection of depth is assumed to be non-uniform across channel and ping time. This forces us to basically do a 'where depth meets this criteria' check to create our aggregations instead of doing something fast and index-based like `.coarsen`.

From @leewujung in #1314 discussion:

> All of the criteria are built on the concept of comparing with neighbors, and the binning from accurate bin edges (from the coordinate labels) vs inaccurate bin edges (from just the indices) would likely result in pretty small differences. If the noise needs to be masked, it would get masked either way; if the noise is borderline, the masking is likely be at least equally sensitive to other independent parameter changes (such as changing the averaging bin size along depth or across pings).

I envision implementing a fast indexed version of these functions by allowing the user to pass in `use_index_binning=True/False` into each of the existing functions and right before an aggregation occurs in the masking function, we choose between a `.coarsen` like function and the slow aggregations that we currently have from #1314 (based on `use_index_binning`).

PR #1326, opened 2024-05-31
Add Index Aggregation Option for Cleaning Mask Functions

ctuguinay:
Addresses #1321


ctuguinay:
TODO: Still need to figure out the `use_index_binning=True` tests for impulse masking

ctuguinay:
I wasn't sure if the transient noise bins indicated in the Ryan 2015 paper were radius or diameter values, but I couldn't run some examples if my assumption was radius so I modified my assumption to diameter, and that is reflected in the default values of the transient noise mask function.

ctuguinay:
@leewujung This PR is at a good spot for review

echopype/clean/api.py
```diff
@@ -40,14 +44,20 @@ def mask_transient_noise(
         Calibrated Sv data with depth data variable.
     func: str, default `nanmean`
         Pooling function used in the pooled Sv aggregation.
-    depth_bin : str, default `20`m
-        Pooling radius along ``depth`` in meters.
-    num_side_pings : int, default `50`
+    depth_bin : str, default `10m`
```
leewujung, line 47: This is one example, but could you change all the things attached with a unit to be in double quotes? I think backtick is not the right format to be rendered in the docstring?
```suggestion
    depth_bin : str, default "10m"
```

echopype/clean/api.py
```diff
@@ -40,14 +44,20 @@ def mask_transient_noise(
...
+        Pooling radius vertically along `depth`.
+    num_side_pings : int, default `25`
         Number of side pings to look at for the pooling.
     exclude_above : Union[int, float], default `250`m
         Exclude all depth above (closer to the surface than) this value.
     transient_noise_threshold : Union[int, float], default `10.0`dB
         Transient noise threshold value (dB) for the pooling comparison.
```
leewujung, line 54: Let's change this to do "10.0dB" (also allow variations such as "10.0 dB") so that we force people to think about the units.

echopype/clean/api.py
```diff
@@ -40,14 +44,20 @@ def mask_transient_noise(
...
+    use_index_binning : bool, default `False`
+        Speeds up aggregations by assuming depth is uniform and binning based
+        on `range_sample` indices instead of `depth` values.
+    chunk_dict : dict, default `{}`
+        Dictionary containing chunk sizes for use in the Dask-Image
+        pooling function. Only used when `use_index_binning=True`.
```
leewujung, line 60: How about adding an Example section in the docstring to show how a single call would look like?

echopype/clean/api.py
```diff
@@ -65,7 +75,7 @@ def mask_transient_noise(
     transient noise masking and translated into xarray code:
     https://github.com/open-ocean-sounding/echopy/blob/master/echopy/processing/mask_transient.py # noqa
     """
-    if "depth" not in ds_Sv.data_vars:
+    if "depth" not in ds_Sv.data_vars and not use_index_binning:
```
leewujung, line 78: I think we need to allow both `echo_range` and `depth`, since direct output from `compute_Sv` has only `echo_range` unless people explicitly want to add a depth variable to it (and we know that is a can of worm). The units are the same between `echo_range` and `depth` so that should be relatively straightforward.

echopype/clean/api.py
```diff
@@ -40,14 +44,20 @@ def mask_transient_noise(
         Calibrated Sv data with depth data variable.
     func: str, default `nanmean`
         Pooling function used in the pooled Sv aggregation.
-    depth_bin : str, default `20`m
-        Pooling radius along ``depth`` in meters.
-    num_side_pings : int, default `50`
+    depth_bin : str, default `10m`
+        Pooling radius vertically along `depth`.
```
leewujung, line 48:
```suggestion
        Pooling bin size vertically along `depth`.
```

echopype/clean/api.py
leewujung: I couldn't comment on the exclude_above part, since it was not part of the changed code, but could you also do the same for `exclude_above`? like `250`m to "250m" ?

echopype/clean/utils.py
```diff
@@ -83,6 +84,95 @@ def calc_transient_noise_pooled_Sv(
...
+    """
+    Compute pooled Sv array for transient noise masking using index binning.
+
+    This function makes the assumption that within each channel, the difference
+    between depth values is uniform across all pings. Thus, computing the number of
+    range sample indices needed to cover the depth bin is a channel-specific task.
+    """
+    # Create `ds_Sv` copy
+    ds_Sv_copy = (
+        ds_Sv.copy()
```
leewujung, line 104: this is copy of the copy from the calling function (`mask_transient_noise`) -- is this needed?

leewujung: How about:
- remove the copy in the calling function, so only doing the copy within this function
- doing `range_sample=slice(min_range_sample, None)` along with the copy operation here

echopype/clean/utils.py
```diff
@@ -83,6 +84,95 @@ def calc_transient_noise_pooled_Sv(
...
+            )
+        )
+        chan_pooled_Sv = (
+            ds_Sv_copy["Sv"]
+            .copy()
+            .isel(
+                channel=channel_index,
+                range_sample=slice(min_range_sample, len(ds_Sv["range_sample"])),
+            )
+        )
```
leewujung, line 137: See my comment above. Also for the beginning or ending parts you can just use `None` in `slice`.

echopype/clean/utils.py
```diff
@@ -83,6 +84,95 @@ def calc_transient_noise_pooled_Sv(
...
+
+        # Grab channel-specific number of range sample indices for vertical binning
+        chan_num_range_sample_indices = all_chan_num_range_sample_indices[channel_index]
+
+        # Create pooling size list
+        pooling_size = [(2 * num_side_pings) + 1, (2 * chan_num_range_sample_indices) + 1]
```
leewujung, line 143: Here you have a discrepancy of 1* vs 2* of the value in `chan_num_range_sample_indices`. Can you make it consistent such that the number of pixels used is actually the closest integer to what the user wants from the calling function?

echopype/clean/utils.py
```diff
@@ -83,6 +84,95 @@ def calc_transient_noise_pooled_Sv(
...
+
+        # Rechunk Sv since `generic_filter` expects a Dask Array
+        chan_Sv = chan_Sv.chunk(chunk_dict)
+
+        # Compute `chan_pooled_Sv` values using dask-image's generic filter
+        chan_pooled_Sv.values = _lin2log(
+            dask_image.ndfilters.generic_filter(
+                chan_Sv.pipe(_log2lin).data,
+                function=np.nanmean if func == "nanmean" else np.nanmedian,
```
leewujung, line 152: It does not make sense that you allow `func` argument but here default everything to `nanmedian` if it is not `nanmean`

leewujung (commented):
For `mask_transient_noise`:
Why do we need to copy operations in the functions?
```python
# Copy `ds_Sv`
ds_Sv = ds_Sv.copy()
```
This copy is then copied again in `index_binning_pool_Sv` so I also commented there. I may be missing something, but I think we just need 1 copy that gets sliced with to exclude the shallower samples (`exclude_above`).

echopype/clean/utils.py
```diff
@@ -83,6 +84,95 @@ def calc_transient_noise_pooled_Sv(
...
+                size=pooling_size,
+            ).compute()
+        )
+
+        # Remove invalid pooled Sv values that are too close to the edge of the echogram
+        chan_pooled_Sv = chan_pooled_Sv.isel(
+            ping_time=slice(num_side_pings, -num_side_pings),
+            range_sample=slice(chan_num_range_sample_indices, -chan_num_range_sample_indices),
+        )
```
leewujung, line 161: i think you mentioned that this is because the wrap-around operation the dask filtering operation does. Is that documented somewhere? Cutting these out would result into some samples not being run through I think.

Seems like an alternative is to pad (using nearest neighbor) the outer rim of the image based on the filter size, and then trim those back to the original size?

echopype/clean/api.py
ctuguinay: On my end, I see the following when using "10m":
image

When using `10m` (with a backtick on either side), I see the following:
image

And this is what the read the docs shows (with backticks):
image

echopype/clean/api.py
ctuguinay: Yeah good idea, I'll add dB. Also, see my comment on backticks vs double quotes above.

echopype/clean/api.py
ctuguinay: Ok I think it is weird that nanmean is shown with backticks but 10m is not shown with backticks in the readthedocs.

echopype/clean/utils.py
ctuguinay: Nah, I'll remove it

echopype/clean/utils.py
ctuguinay: Added this suggestion

echopype/clean/utils.py
ctuguinay: Yeah, it would result in not filtering out the edge values. 

Under `generic_filter`, there's a parameter called `mode` that does this exact operation: https://image.dask.org/en/latest/dask_image.ndfilters.html.
I think `mode=nearest` or `mode=reflect` may be what we want here. My issue with `nearest` is that if the value at the end is something that is an obvious outlier, then it could really mess up the mean/median aggregation calculation and wouldn't be representative of the region. With `reflect`, we use many more surrounding values, and even if it is repeating some of the values that are already within the aggregation, it will still be more representative of the region as a whole (than that of `nearest`).

ctuguinay:
> For `mask_transient_noise`: Why do we need to copy operations in the functions?
> 
> ```python
> # Copy `ds_Sv`
> ds_Sv = ds_Sv.copy()
> ```
> 
> This copy is then copied again in `index_binning_pool_Sv` so I also commented there. I may be missing something, but I think we just need 1 copy that gets sliced with to exclude the shallower samples (`exclude_above`).

I agree it's not necessary. Even that second copy isn't required since when I remove all of them, the tests seem to pass.

echopype/clean/utils.py
ctuguinay: Sorry, I'm not quite understanding this.

So here's an example of what I intended the code to do:
If there is a 0.2m separation between each range sample index and the user wants a depth bin of 1m, then I calculate this `chan_num_range_sample_indices` value to be 5, since it takes 5 values to capture it. I pick a single point, and I go up 5 indices and down 5 indices of it. That single point is this `+1` term and the up and down is this `2 * chan_num_range_sample_indices` term.

echopype/clean/utils.py
ctuguinay: I think it's because I only allow `nanmean` and `nanmedian` to be used in this pooling function. I can instead modify this to be something like:
`np.nanmean(data) if func == "nanmean" else (np.nanmedian(data) if func == "nanmedian" else RaiseValueError(f"Unsupported function: {func}"))`?

echopype/clean/api.py
ctuguinay: I agree I can add that

echopype/clean/api.py
ctuguinay: I'll add that

echopype/clean/api.py
ctuguinay: Ok yeah I think I agree now that this should be double quotes instead of backticks. I'll make this change.

ctuguinay:
@leewujung Thanks for the review! This should be ready to look at again. I applied most of your suggestions, but I have a few questions/comments.

ctuguinay:
Oh one more slight **TODO**: In the new test, test the `use_swap` and `chunk_dict` `ds["Sv"]` values against Echoview values.

ctuguinay:
Oh wait I just had two more thoughts for some slight changes.

**TODO**:
* I should allow for both `db` and `dB`
* I should expand on the transient noise example. I think I can add some sort of example array to it

I'll make these changes once I get home today

echopype/clean/utils.py
ctuguinay: oh wait I think I should instead just modify the parent function `transient_noise_mask` to pass in `np.nanmean` or `np.nanmedian` instead of the string version. `transient_noise_mask` will do the assigning of the `np.nanmean` and `np.nanmedian`.

echopype/clean/utils.py
ctuguinay: Using `mode=reflect` after discussion today

ctuguinay:
@leewujung This should be ready to review again. I added this reflect logic we talked about previously

echopype/clean/api.py
```diff
@@ -185,14 +278,16 @@ def mask_attenuated_signal(
...
-    upper_limit_sl : Union[int, float], default `400`m
+    upper_limit_sl : str, default `400m`
         Upper limit of deep scattering layer line (m).
-    lower_limit_sl : Union[int, float], default `500`m
+    lower_limit_sl : str, default `500m`
         Lower limit of deep scattering layer line (m).
     num_side_pings : int, default `15`
         Number of preceding & subsequent pings defining the block.
-    attenuation_signal_threshold : Union[int, float], default `8.0`dB
+    attenuation_signal_threshold : str, default `8.0dB`
```
leewujung, line 287: Could you change ` ` here to " " like elsewhere?

echopype/clean/utils.py
```diff
@@ -6,8 +10,29 @@
...
-    ds_Sv: xr.Dataset, func: str, depth_bin: float, num_side_pings: int, exclude_above: float
+def extract_dB(dB_str: str) -> float:
+    """Extract float value from decibal string in the form of 'NUMdB'."""
+    # Search for numeric part using regular expressions and convert to float
+    if not isinstance(dB_str, str):
+        raise TypeError(
+            "Decibal input must be a string formatted as `NUMdB` or `NUMdb."
+            f"Cannot be of type `{type(dB_str)}`."
+        )
+    pattern = r"^[-+]?\d+\.?\d*(?:dB|db)$"  # Ensures exact match with format
```
leewujung, line 21: do you want to add an optional whitespace before dB/db so that inputs could be `NUMdB` or `NUM dB`?

echopype/clean/api.py
```diff
@@ -64,34 +77,87 @@ def mask_transient_noise(
...
+    # Assign string to numpy function
+    if func == "nanmean":
+        func = np.nanmean
+    elif func == "nanmedian":
+        # Warn when `func=nanmedian` since the sorting overhead makes it incredibly slow compared to
+        # `nanmean`.
+        logger.warning(
+            "Consider using `func=nanmean`. `func=nanmedian` is an incredibly slow operation due "
+            "to the overhead sorting."
+        )
```
leewujung, line 141: I agree with the statement, but the original method used `nanmedian` right? I feel that we should not recommend the change, but we can warn people that it is SLOW. We can also point them to the issue for the Fielding filter and the echopy code for that in the warning?

echopype/clean/api.py
```diff
@@ -38,16 +43,24 @@ def mask_transient_noise(
     ----------
     ds_Sv : xarray.Dataset
         Calibrated Sv data with depth data variable.
-    func: str, default `nanmean`
+    func: str, default "nanmean"
         Pooling function used in the pooled Sv aggregation.
```
leewujung, line 47: I feel we should default to "nanmedian" since that is what the original method said.

leewujung (approved):
@ctuguinay : Thanks for the changes. I had pretty minor comments, and think we can merge this afterwards. This is such an entrenched issue, and probably something the AA SI folks can enhance further later.

echopype/clean/api.py
ctuguinay: Made this change 👍

echopype/clean/utils.py
ctuguinay: I don't like having this space here. Maybe we just keep it `NUMdB` without space? This way it's also simpler and the user only has 2 accepted ways to write it, `NUMdB` and `NUMdb`, instead of 4 ways to write it

echopype/clean/api.py
ctuguinay: Made this change

echopype/clean/api.py
ctuguinay: Added this. Good idea! 👍

ctuguinay:
@leewujung Thanks for the review! This should be ready for a short review again

echopype/clean/utils.py
leewujung: i think this is fine. I just remembered that we also only allow "10m".

echopype/tests/clean/test_noise.py
```diff
@@ -124,8 +124,9 @@ def test_transient_mask_noise_func_error_and_warnings(caplog):
 
     # Check for expected warning
     expected_warning = (
-        "Consider using `func=nanmean`. `func=nanmedian` is an incredibly slow operation due to "
-        "the overhead sorting."
+        "`func=nanmedian` is an incredibly slow operation due to the overhead sorting. "
+        "In the future, Echopype will include the Fielding Transient Noise Filter "
+        "described here: https://github.com/OSOceanAcoustics/echopype/issues/1352"
```
leewujung, line 129:
```suggestion
        "`func=nanmedian` is an incredibly slow operation due to the overhead sorting. "
        "We plan to add the Fielding Transient Noise Filter in the future"
        "described here: https://github.com/OSOceanAcoustics/echopype/issues/1352"
```

echopype/clean/api.py
```diff
@@ -134,10 +134,11 @@ def mask_transient_noise(
...
     elif func == "nanmedian":
         # Warn when `func=nanmedian` since the sorting overhead makes it incredibly slow compared to
-        # `nanmean`.
+        # other non-sorting aggregations like `nanmean`.
         logger.warning(
-            "Consider using `func=nanmean`. `func=nanmedian` is an incredibly slow operation due "
-            "to the overhead sorting."
+            "`func=nanmedian` is an incredibly slow operation due to the overhead sorting. "
+            "In the future, Echopype will include the Fielding Transient Noise Filter "
+            "described here: https://github.com/OSOceanAcoustics/echopype/issues/1352"
```
leewujung, line 141:
```suggestion
            "`func=nanmedian` is an incredibly slow operation due to the overhead sorting. "
            "We plan to add the Fielding Transient Noise Filter in the future"
            "described here: https://github.com/OSOceanAcoustics/echopype/issues/1352"
```

leewujung (approved):
Looks good! I think we are ready to merge now!

ctuguinay:
@leewujung oh could you approve this again. i think committing your suggestions need further approval

leewujung (approved):

PR #1328, opened 2024-06-04
Fix problems in applying `compute_Sv` on combined `EchoData` object

leewujung:
This PR fixes various problems discovered in applying `compute_Sv` on combined `EchoData` object:
- redundancy in computing sound speed when calculating echo range
- forward-fill to take care of the case when `ping_time` extends beyond the end of `time1` for the environmental params
- add `bottleneck` as dependency as it is needed for [xr.DataArray.ffill](https://docs.xarray.dev/en/latest/generated/xarray.DataArray.ffill.html)

TODO:
- [ ] add test for cases needing forward fill


ctuguinay:
@leewujung This should be ready for review. I added a test and dropped `time1` after the interpolation and ffill.

ctuguinay (approved):

Issue #1329, opened 2024-06-07
Add `use_swap` for scaling `compute_Sv` on EK60/EK80


ctuguinay:
From #1212, Don's investigation into `compute_Sv` failure on a large combined Echodata object led to him determining that the large elementwise array operations taking place to compute Echo Range were the problem. He tested out the computation of Echo Range on a mock dataset of similar size to the original problem's arrays and the operation failed to compute:

```
ch, pt, rs = 5, 46187, 3957
sample_interval = xr.DataArray(dask.array.ones((ch, pt)), name="sample_interval", dims=["ch", "pt"], coords={"ch": np.arange(ch), "pt": np.arange(pt)})
sound_speed = xr.DataArray(dask.array.ones((ch, pt)) * 1053, name="sound_speed", dims=["ch", "pt"], coords={"ch": np.arange(ch), "pt": np.arange(pt)})
range_sample = xr.DataArray(dask.array.from_array(np.arange(rs)), name="range_sample", dims=["rs"], coords={"rs": np.arange(rs)})
res = range_sample * sample_interval * sound_speed / 2
```

I was able to replicate the failure: Trying to run this with 8 Dask workers with about 4 GiB of RAM led to memory leakage and then to a kernel crash.

When looking at the Dask Graph itself, one can see that the computation defaults to just 1 worker leading to this overload:
![image](https://github.com/OSOceanAcoustics/echopype/assets/87830138/637d1060-f8d2-47aa-b3ea-c596c115d9ed)

When chunking `pt` dimension, we have equal distribution of processing across workers; however, we see memory build-up in each worker:
![image](https://github.com/OSOceanAcoustics/echopype/assets/87830138/56b9b5e7-0795-4dbe-8849-e3c9037d76e9)

The memory build-up is due to Dask waiting for each task to finish computing before setting all values of the output array. So in our case, the final task is the division by 2, so after each chunk is done going through the entire task pipeline with the final task being that of division by 2, the divided value (and perhaps some small overhead) persists in memory until ALL chunks have gone through the same pipeline.

One way to solve this is by pointing the computation at a Zarr Store.

Creating a Dask array much larger than what Don's example had, we can do a similar elementwise operation:

![image](https://github.com/OSOceanAcoustics/echopype/assets/87830138/cb9b16be-fdad-4256-9d68-a3c717413d9c)

[...]

Issue #1212, opened 2023-11-07
Failure to perform `compute_Sv` with dask inputs


lsetiawan:
## Overview

Currently, `compute_Sv` does not work with a merged Sv file (`raw_converted_combined/2013/x0097_2_wt_20130824_232034_f0031.zarr`) when opened as dask arrays. Worker memory get's overwhelmed and even spilling to disk by dask won't cut it. This is the non descriptive dask traceback:

```console
2023-11-06 17:00:16,536 - distributed.worker.memory - WARNING - Unmanaged memory use is high. This may indicate a memory leak or the memory may not be released to the OS; see https://distributed.dask.org/en/latest/worker-memory.html#memory-not-released-back-to-the-os for more information. -- Unmanaged memory: 11.37 GiB -- Worker memory limit: 16.00 GiB
2023-11-06 17:00:16,878 - distributed.worker.memory - WARNING - Worker is at 80% memory usage. Pausing worker.  Process memory: 12.85 GiB -- Worker memory limit: 16.00 GiB
2023-11-06 17:00:16,922 - distributed.worker.memory - WARNING - Worker is at 79% memory usage. Resuming worker. Process memory: 12.70 GiB -- Worker memory limit: 16.00 GiB
...
RuntimeError: Cannot remove replica of ('transpose-4cd487e9eb8037d494b448b470058650', 0, 0, 0) while ('rechunk-split-rechunk-merge-78880722c4e511b3105b467cdecc2000', 0, 22, 2) in state 'executing'.
RuntimeError: Cannot remove replica of ('transpose-4cd487e9eb8037d494b448b470058650', 0, 0, 0) while ('rechunk-split-rechunk-merge-78880722c4e511b3105b467cdecc2000', 0, 22, 2) in state 'executing'.
```

The above traceback is not very useful, so I tried to perform `compute_Sv` without using dask and that seemed to work and I was able to profile the memory usage. It seems like there are a few spots that huge spikes in memory occurs.

<details>
<summary>Memory Line by Line Profile</summary>
<br>

```
Filename: /Users/lsetiawan/Repos/SSEC/echopype/echopype/calibrate/calibrate_ek.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
   137    222.6 MiB    222.6 MiB           1       def __init__(self, echodata: EchoData, env_params, cal_params, ecs_file, **kwargs):
[...]

PR #1331, opened 2024-06-12
Add Scalability Options (`use_swap` and chunks) for `ep.calibrate.compute_Sv`

ctuguinay:
Addresses #1329


ctuguinay:
Example Compute Sv Gist with OOI Data much larger than what was tested on in #1212: https://gist.github.com/ctuguinay/6195d318037f80eb60d9e9cb5dd7aa77

ctuguinay:
Also some remove background noise tests were failing, but that was out of the scope of this PR so I left it as an issue in #1332

ctuguinay:
TODO: The `use_swap` implementation in `open_raw` does something different with the Zarr store, but I didn't understand it so I stuck with what worked in my notebook. I need to look to see what happens when I save an Xarray Dataset object automatically to the store from `temp_zarr_store` instead of saving to the root array of the store from `temp_zarr_store`.

ctuguinay:
@leewujung This should be ready to review now

ctuguinay:
Oh one more slight **TODO**: In the new test, test the `use_swap` and `chunk_dict` `ds["Sv"]` values against Echoview values.

ctuguinay:
Investigate `zarr.sync.ThreadSynchronizer()`. Maybe also add some kwarg for storage options.

echopype/tests/calibrate/test_calibrate.py
```diff
@@ -45,7 +45,63 @@ def test_compute_Sv_returns_water_level(ek60_path):
...
+
+    if use_swap:
+        # Check that there are only 2 layers of the task graph
+        assert 2 == len(Sv.data.dask.layers)
+    elif not use_swap and chunk_dict:
+        # Check that there are many more than 2 layers of the task graph
+        assert 77 == len(Sv.data.dask.layers)
+    elif not use_swap and not chunk_dict:
+        # Check that Sv data is not a dask array and is instead a numpy array
+        assert isinstance(Sv.data, np.ndarray)
```
leewujung, line 83: A small suggestion and a question:
- suggestion: how about merging the 2nd and 3rd cases so that you have
   ```python
   if use_swap:
      ...
   else:
      if chunk_dict:
         ...
      else:
         ...
   ```
- question: could you explain what dask is doing when you have `use_swap=False` but with `chunk_dict`? I guess I am not fully understanding why `use_swap=True` would do the trick to reduce the number of graph layers, whereas if you only set a `chunk_dict` but not using swap, it will produce many more layers.

leewujung (commented):
So to follow up on my question in the test above:
For the code below
```python
if use_swap:
    # Setup temp store
    zarr_store = create_temp_zarr_store()

    # Save to zarr stores and point computation towards the `zarr_root`
    cal_ds.to_zarr(zarr_store, compute=True)

    # Open zarr stores
    cal_ds = xr.open_dataset(
        zarr_store, engine="zarr", chunks={}, synchronizer=zarr.sync.ThreadSynchronizer()
    )
```
Wouldn't this be equivalent to below?
```python
cal_ds = _compute_cal(...)
cal_ds.to_zarr(final_zarr_store)
```
Since it looks like when you reach Line 116 (right before you add water level) nothing is actually computed, and that's why the only place you need to do `.to_zarr` and then lazy-load it back is within the condition `if use_swap:`.

My question about why even without chunking, using `use_swap=True` would automatically make the number of layers = 2.

echopype/tests/calibrate/test_calibrate.py
ctuguinay: > could you explain what dask is doing when you have `use_swap=False` but with chunk_dict

My thought for this case was for the user to have the option to chunk `cal_object` without actually doing any computing, since `use_swap=True` would force the computation. In the case where the user may want to lazy load an entire pipeline from open raw to compute NASC, but now that I think about it, this could be done more simply and more explicitly by having something like:

```
ed = ed.chunk({'ping_time': 1000})
ds_Sv = ep.calibrate.compute_Sv(ed, use_swap=False)
...
ds_NASC = ...
ds_NASC.to_zarr(...compute=True)
```

> `use_swap=True` alone would do the trick to reduce the number of graph layers

Yeah you're right. In the `ep.open_raw(..., use_swap=False)` case, the `cal_object` would be fully computed and the Zarr Store wouldn't have any computation pointed towards it and it would instead just store something fully in-memory into the Zarr Store. I think you're right and I shouldn't allow for this case since it does nothing helpful. I should instead only allow `use_swap=True` if and only if the `cal_object` is chunked. And for that, I think maybe I should restrict chunking to outside of this function, like the example I show above? What are your thoughts on that?

The other option would be to restrict using `chunk_dict` to the case where `use_swap=True` and ignore `chunk_dict` in the case where `use_swap=False`.

ctuguinay:
> Wouldn't this be equivalent to below?

Oh are you thinking I move the `to_zarr` to the API wrapper functions (`compute_Sv` and `compute_TS`)? I'm not quite sure I understand the question.

And also does this code below also imply the creation of this temporary Zarr Store?
```
cal_ds = _compute_cal(...)
cal_ds.to_zarr(final_zarr_store)
```

ctuguinay:
> My question about why even without chunking, using use_swap=True would automatically make the number of layers = 2.

The order of events would go like this (in the case where `cal_object` is not chunked but `use_swap=True`:

1. `cal_object` is computed and is in memory, and is not a dask array and thus has no dask graph layers
2. `cal_object` is sent to Zarr Store
3. `cal_object` is lazy loaded back into memory with no actual computation (other than the initial 2 Zarr reading layers) since it had no real computation layers to begin with

ctuguinay:
TODOs (from conversation with @leewujung today):

- [ ] Remove `use_swap`
- [ ] Add Warning if the input `backscatter_r` variable in the Echodata object is above a certain size. Say `8 GiB`. This warning will include a message for how to lazy load and point computation to Zarr Store.
- [ ] Move chunk echodata function to class function for Echodata object

leewujung:
@ctuguinay :
- I am wondering if we should put the data size warning limit lower, say 2GB, just so that people are aware of the potential problem.
- also, since we'll have to look into the same thing for complex data, how about including both `backscatter_r` and `backscatter_i` in the size limit? so the sum of both >2GB. For EK60 the `backscatter_i` does not exist so some checks are needed.

ctuguinay:
@leewujung I agree with the above 👍

echopype/calibrate/api.py
```diff
@@ -22,6 +19,51 @@
 logger = _init_logger(__name__)
 
 
+def _check_echodata_backscatter_size(echodata: EchoData):
```
leewujung, line 22: We didn't cover this in our discussion, but I think we can move this function to be under `CalibrateBase`. The reason is that, for EK80 data where some channels transmit in CW mode and some channels transmit BB (LFM) mode, the data size can be dramatically different between the 2 beam groups, because BB mode data _must be_ stored in complex format (whereas CW mode data can be stored in either power/angle or complex format). So if you sum up the size across the 2 beam groups it may be misleading, since the CW mode data could actually be pretty small and won't cause any problems.

You can find a class attribute `ed_beam_group` in `CalibrateEK`, which will give you the right beam group to access for the calibration. So you could just grab the combined size of `backscatter_*` of that beam group. 

AZFP does not have this complication, so the combined size of `backscatter_r` would suffice.

echopype/calibrate/api.py
```diff
@@ -22,6 +19,51 @@
...
+    if total_gb > 2.0:
+        logger.warning(
+            "The Echodata Backscatter Variables are large and can cause memory issues. "
+            "Consider modifying compute_Sv workflow: "
+            "Prior to `compute_Sv` run `echodata.chunk(CHUNK_DICTIONARY) and after `compute_Sv` "
+            "run `ds_Sv.to_zarr(ZARR_STORE, compute=True)`. This will ensure that the bulk of the "
+            "computation does not overwhelm your system's memory. The former ensures that "
+            "`compute_Sv` is lazily evaluated and the latter points computation towards a Zarr "
+            "Store in Disk instead of on RAM."
+        )
```
leewujung, line 64:
```suggestion
    if total_gb > 2.0:
        logger.warning(
            "The Echodata Backscatter Variables are large and can cause memory issues. "
            "Consider modifying compute_Sv workflow: "
            "Prior to `compute_Sv` run `echodata.chunk(CHUNK_DICTIONARY) "
            "and after `compute_Sv` run `ds_Sv.to_zarr(ZARR_STORE, compute=True)`. "
            "This will ensure that the computation is lazily evaluated, "
            "with the results stored directly in a Zarr store on disk, rather then in memory."
        )
```

echopype/calibrate/api.py
```diff
@@ -22,6 +19,51 @@
 logger = _init_logger(__name__)
 
 
+def _check_echodata_backscatter_size(echodata: EchoData):
+    """
+    Extracts total gigabytes in `backscatter_r` and `backscatter_i`. If above 2 GiB,
+    then raises warning for how to proceed with the `compute_Sv` workflow without
+    destroying overwhelming the system's memory.
```
leewujung, line 26:
```suggestion
    Extracts total size of `backscatter_r` and `backscatter_i` in a beam group.
    If the size is above 2 GiB, raises a warning showing a recommended workflow
    that will not overwhelm the system memory.
```

leewujung (changes requested):
@ctuguinay : Thanks for the changes! I have a specific suggestion for `_check_echodata_backscatter_size` and otherwise these look good !

echopype/calibrate/api.py
ctuguinay: @leewujung Thanks, this is very helpful! So here's the cases that I think we have for EK80 (from some digging through the data test suite):

BB and Complex -> Grab `ed_beam_group` and calculate sum nbytes in `backscatter_i` and `backscatter_r` and these variables should have a `beam` dimension
CW and Complex -> Grab `ed_beam_group` and calculate sum nbytes in `backscatter_i` and `backscatter_r`and these variables should have a `beam` dimension
CW and Power -> Grab `ed_beam_group` and calculate sum nbytes in `backscatter_r` and this variable should not have a `beam` dimension

ctuguinay: And for EK60 and AZFP, I should not expect to use anything other than `Sonar/Beam_group1` right?

leewujung: EK60 and AZFP will only have `Sonar/Beam_group1`. I think our docs is clear about that. If not please propose changes.

ctuguinay:
@leewujung Thanks for the review! This should be ready for review again. Main thing I did was do this backscatter nbyte calculation based on sonar model, encode mode, waveform mode, and echodata beam group.

ctuguinay:
@leewujung Made your suggestion to extract encode and waveform from the calibration object

leewujung (approved):
@ctuguinay : Thanks for the changes!

Issue #1335, opened 2024-06-15
Fix Xarray Duck Array Ops `invalid value encountered in cast` Runtime Warning in `ep.open_raw`


ctuguinay:
When opening and saving `.RAW` files from [OOI](https://rawdata.oceanobservatories.org/files/CE04OSPS/PC01B/ZPLSCB102_10.33.10.143/2017/08), I encountered some `duck_array_ops` warnings:

```
# Save Echodata objects locally
def open_and_save(raw_file, sonar_model, use_swap, save_path, open_kwargs={}):
    try:
        ed = ep.open_raw(raw_file=raw_file, sonar_model=sonar_model, use_swap=use_swap, **open_kwargs)
        ed.to_zarr(save_path, overwrite=True, compute=True)
    except Exception as e:
        print("Error with Exception: ", e)

# Parse EK60 `.RAW` file and save to Zarr Store
open_and_save_futures = []
for raw_file_url in desired_raw_file_urls:
    open_and_save_future = client.submit(
        open_and_save,
        raw_file=raw_file_url,
        sonar_model='ek60',
        use_swap=True,
        save_path=single_echodata_zarr_dpath
    )
    open_and_save_futures.append(open_and_save_future)
open_and_save_futures = client.gather(open_and_save_futures)
```

```
/home/exouser/miniforge3/envs/echopype/lib/python3.9/site-packages/xarray/core/duck_array_ops.py:216: RuntimeWarning: invalid value encountered in cast
  return data.astype(dtype, **kwargs)
/home/exouser/miniforge3/envs/echopype/lib/python3.9/site-packages/xarray/core/duck_array_ops.py:216: RuntimeWarning: invalid value encountered in cast
  return data.astype(dtype, **kwargs)
/home/exouser/miniforge3/envs/echopype/lib/python3.9/site-packages/xarray/core/duck_array_ops.py:216: RuntimeWarning: invalid value encountered in cast
  return data.astype(dtype, **kwargs)
```
using xarray version `'2024.5.0'`.

I'm not quite sure what this is about, but this is probably something small and it should be fixed before any new `echopype-examples` notebook PRs get merged. I think it can be very disheartening seeing all these warnings pop up when using our package for the first time.

ctuguinay:
So the problem here is a bit convoluted. I'll detail it step-by-step here:

Values in the Parser's `ping_data_dict` can be of different lengths depending on the channel. Let's take transmit mode for example:

[...]

PR #1337, opened 2024-06-16
Fix Duck Array `RuntimeWarning` for EK60 `ep.open_raw`

ctuguinay:
Addresses #1335. I have a step-by-step explanation of the problem and my proposed solution on the linked issue.


ctuguinay:
@leewujung This is ready for review. Small fix for a warning I kept encountering when trying to do the OOI notebook PR

ctuguinay:
@leewujung I made the change that you proposed in #1335. The failing tests are unrelated to this PR and are the processing level tests mentioned in #1342.

I am curious though, why in

```
EXPECTED_VAR_DTYPE = {
    "channel": np.str_,
    "cal_channel_id": np.str_,
    "beam": np.str_,
    "channel_mode": np.float16,
    "beam_stabilisation": np.byte,
    "non_quantitative_processing": np.int16,
}  # channel name  # beam name
```

are we only checking for these subset of variable types? Are they special in some way?

echopype/tests/convert/test_convert_ek60.py
```diff
@@ -1,9 +1,12 @@
+import warnings
+
 import numpy as np
 import pandas as pd
 from scipy.io import loadmat
-from echopype import open_raw
 import pytest
 
+from echopype import open_raw
+import echopype as ep
```
leewujung, line 9: Seems like the new test uses `open_raw` only, so you don't need this?

leewujung (dismissed):
LGTM! Only a comment on an import statement.

leewujung:
> are we only checking for these subset of variable types? Are they special in some way?

In fact I don't remember exactly why. But I think it was due to similar warnings, since before these types were either not set or set to something that would cause a warning. 

I think we should at some point implement a pydantic approach for all the types, to avoid the messiness and the arbitrary patching approach you pointed out here.

Also note that our NaN-padding approach probably made many of the resultant types different from those written in the convention. Something we should also discuss, probably with the user community and convention folks.

echopype/tests/convert/test_convert_ek60.py
leewujung:
```suggestion
```

leewujung (approved):

Issue #1330, opened 2024-06-11
Modify `apply_mask` check to work for MVBS workflow


ctuguinay:
The function `apply_mask` does not work for MVBS -> frequency differencing -> masking workflow since `apply_mask` expects the `range_sample` dimension and MVBS's vertical dimension has to be either `echo_range` or `depth`.

It should also check for alignment of dimensions for the input dataset and the mask.

PR #1345, opened 2024-06-21
Allow `MVBS` into `ep.mask.apply_mask`, and add Alignment Check for Target Variable and Mask

ctuguinay:
Addresses #1330


echopype/mask/api.py
```diff
@@ -389,6 +419,9 @@ def apply_mask(
         _variable_prov_attrs(output_ds[var_name], mask)
     )
 
+    # Transpose to original dimension order
```
leewujung, line 422: I feel we should not use "transpose" in the comment here since it is a misnomer. How about "Use the original dimension order" ?

echopype/tests/mask/test_mask.py
```diff
@@ -1240,4 +1241,146 @@ def test_apply_mask_channel_variation(source_has_ch, mask, truth_da):
...
+                source_Sv=MVBS,
+                freqABEq="120000.0Hz - 38000.0Hz > 0dB",
+                storage_options={},
+            ).expand_dims(dim={"channel": MVBS["channel"][0:2].data}),
+            ep.mask.frequency_differencing(
+                source_Sv=MVBS,
+                freqABEq="38000.0Hz - 18000.0Hz > 0dB",
+                storage_options={},
+            ).expand_dims(dim={"channel": MVBS["channel"][2:6].data}),
+        ]
```
leewujung, line 1299: How about taking out the dependency on `ep.mask.frequency_differencing` here and just create a random bit map for the test? This way if `frequency_differencing` fails it would not affect this test. Same with the else condition.

echopype/tests/mask/test_mask.py
```diff
@@ -1240,4 +1241,146 @@ def test_apply_mask_channel_variation(source_has_ch, mask, truth_da):
...
+        range_bin="1m",
+        ping_time_bin="10s"
+    )
+
+    # Create frequency differencing (120kHz - 38kHz) mask
+    diff = ep.mask.frequency_differencing(
+        source_Sv=MVBS,
+        freqABEq="120000.0Hz - 38000.0Hz > 0dB",
+        storage_options={},
+    )
```
leewujung, line 1367: Same as the above -- seems better to take out the dependency on `frequency_differencing`.

echopype/mask/api.py
```diff
@@ -33,9 +33,33 @@ def _validate_source_ds(source_ds, storage_options_ds):
...
+        mask_dims = set(mask.dims)
+
+    # Grab dimensions of the target variable in source ds
+    target_variable_dims = set(source_ds[var_name].dims)
+
+    # Raise ValueError if the mask has channel dim but the target variable doesn't
+    if "channel" in mask_dims and "channel" not in target_variable_dims:
+        raise ValueError(
+            "'channel' is in Mask dimensions but not in the Source " "Dataset Variable dimensions."
+        )
```
leewujung, line 52: What is the case when the target variable does not have a channel dimension? For all echopype-generated dataarrays I think we always have a channel dimension, since it start from the EchoData object?

leewujung: If we are keeping there's seems some quotation marks mismatches? Also thinking we can use a simpler string.
```suggestion
    # Raise ValueError if the mask has channel dim but the target variable doesn't
    if "channel" in mask_dims and "channel" not in target_variable_dims:
        raise ValueError(
            "'channel' is a dimension in mask but not a dimension in source."
        )
```

echopype/mask/api.py
```diff
@@ -33,9 +33,33 @@ def _validate_source_ds(source_ds, storage_options_ds):
...
+
+    # If non-channel dimensions don't align raise ValueError
+    mask_dims.discard("channel")
+    target_variable_dims.discard("channel")
+    if mask_dims != target_variable_dims:
+        raise ValueError(
+            f"Non-Channel Mask dimensions ({mask_dims}) do not match "
+            "Non-Channel Source Dataset Variable dimensions "
+            f"({target_variable_dims})."
+        )
```
leewujung, line 62:
```suggestion
        raise ValueError(
            f"The dimensions of mask: ({mask_dims}) do not match "
            f"the dimensions of source ({target_variable_dims}) "
            "when not considering 'channel'."
        )
```

leewujung (changes requested):
@ctuguinay : Thanks for the PR!

I have a few inline comment, and outside of that there's only one thing I recommend changing.

The previous `_validate_source_ds` I think is in an over-factored situation since it simply wraps the `validate_source` and opens the dataset, and IMHO does not warrant a new function -- it was confusing when I read this, because I remember we had `validate_source` already from `utils.io`.

So I suggest we do the following:
- put these few lines directly in `apply_mask`
  ```python
  # Validate the source_ds type or path (if it is provided)
  source_ds, file_type = validate_source(source_ds, storage_options_ds)

  if isinstance(source_ds, str):
      # open up Dataset using source_ds path
      source_ds = xr.open_dataset(source_ds, engine=file_type, chunks={}, **storage_options_ds)
  ``` 
- rename the function to be `_check_mask_dim_alignment` for the new code you added

echopype/tests/mask/test_mask.py
ctuguinay: Oh that's clever I'll add this suggestion

echopype/mask/api.py
ctuguinay: Yeah I agree, I'll add that

echopype/mask/api.py
ctuguinay: > What is the case when the target variable does not have a channel dimension? For all echopype-generated dataarrays I think we always have a channel dimension, since it start from the EchoData object?

It was just in the case where somebody chose to do something like `ds_MVBS.isel(channel=0).drop("channel")` and then they chose to put in a mask created with the original number of channels that the `ds_MVBS` DataArray had before dropping channel. Maybe I am being too paranoid about that and this is too fringe (and non-sensical) a case 😆

ctuguinay:
@leewujung Thanks for the review!

> So I suggest we do the following:
> - put these few lines directly in `apply_mask`
>   ```python
>   # Validate the source_ds type or path (if it is provided)
>   source_ds, file_type = validate_source(source_ds, storage_options_ds)
> 
>   if isinstance(source_ds, str):
>       # open up Dataset using source_ds path
>       source_ds = xr.open_dataset(source_ds, engine=file_type, chunks={}, **storage_options_ds)
>   ``` 
> - rename the function to be `_check_mask_dim_alignment` for the new code you added

Yeah I agree, it was a bit much in a single function. I'll make these changes (once I get good WiFi again)

leewujung (approved):
LGTM! Feel free to merge!

PR #1353, opened 2024-07-09
Padding Short Pings

ctuguinay:
When testing SWFSC RAW files, there was a file where range was adaptive and based on bottom. When `open_raw` is called, these pings need to be padded because they are short and non-uniform with the other short pings. In this padding code, there was a snippet that relied on `complex_` to work, but `complex_` is `complex128`, while our data is `complex64`. I changed this to instead use `complex64`.

### DIFF
--- a/echopype/convert/parse_base.py
+++ b/echopype/convert/parse_base.py
@@ -639,13 +639,15 @@ class ParseEK(ParseBase):
             # Create output array from mask
             out_array = np.full(mask.shape, np.nan)
 
+            # Concatenate short pings
+            concat_short_pings = np.concatenate(data_list).reshape(-1)  # reshape in case data > 1D
+
             # Take care of problem of np.nan being implicitly "real"
-            arr_dtype = data_list[0].dtype
-            if np.issubdtype(arr_dtype, np.complex_):
-                out_array = out_array.astype(arr_dtype)
+            if concat_short_pings.dtype == np.complex64:
+                out_array = out_array.astype(np.complex64)
 
             # Fill in values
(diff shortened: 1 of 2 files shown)


echopype/convert/parse_base.py
```diff
@@ -639,13 +639,15 @@ def pad_shorter_ping(data_list) -> np.ndarray:
...
 
+            # Concatenate short pings
+            concat_short_pings = np.concatenate(data_list).reshape(-1)  # reshape in case data > 1D
+
             # Take care of problem of np.nan being implicitly "real"
-            arr_dtype = data_list[0].dtype
-            if np.issubdtype(arr_dtype, np.complex_):
-                out_array = out_array.astype(arr_dtype)
+            if concat_short_pings.dtype == np.complex64:
+                out_array = out_array.astype(np.complex64)
```
leewujung, line 647: I think my question was whether the original `astype(arr_dtype)` is needed to make sure dtypes other than `np.complex64` would be set correctly, but I am unsure.

ctuguinay: but it will only be called if the dtype is equal to complex64. i think the original intention of the astype(arr_dtype) was that (presumably) the types that are subdtype of complex_ could be complex64/128/_ etc. but since i am just checking that the dtype is equal to complex64, i dont think we will run into any other case like that

leewujung: Oh I think you're right! Sorry I completely misread the if statement.

leewujung (approved):
Thanks @ctuguinay for the diagnosis and quick fix!

PR #1355, opened 2024-07-12
docs: Merge data processing level changes from `stable`

leewujung:
This PR merges the data processing level changes from `stable`. Afterwards we can delete `stable` and just have the RTD default to using `stable` to track the latest release.

This is part of the migration to use the [trunk-based development workflow](https://www.atlassian.com/continuous-delivery/continuous-integration/trunk-based-development).


leewujung:
I will self merge this since this is something already out there in `stable`.

PR #1357, opened 2024-07-13
hotfix: Temporarily skip datagrams generated by EC150

leewujung:
This PR includes changes to `ParseEK` to skip reading and saving datagrams generated by EC150 ADCP.

We can actually parse the echosounder-equivalent data fine, but we still need 2 other pieces to carry it to the next stage, so this is a "hotfix" to just skip these datagrams for now. The pieces we need are:
- determine where to save the complex or power data: as a separate `Beam_group3` or with the previous `Beam_group1/2`?
- make sure calibration code can run through fine: not sure at the moment if all the necessary broadband calibration parameters are in place, but looks like all the necessary narrowband calibration parameters are there.

This PR still needs:
- [x] tests with small test file
- [x] an issue that document needed changes for follow-up work (#1358)


ctuguinay (approved):
Thanks looks good, just commented out a print statement!

PR #1359, opened 2024-07-15
Use latest MRU parser version; `ep.consolidate.add_location` can be based off Datagram Type

ctuguinay:
The latest EK MRU parser code from Rick Towler can be found at https://github.com/noaa-afsc-mace/pyEcholab/blob/2314b374ffe99f7c3e3c920ef3f9a240e7715789/echolab2/instruments/util/simrad_parsers.py#L569C1-L708C61. Includes latitude and longitude field values, and we now parse this.

Setting of platform groups will now set `(lat/lon)_(idx/mru/nme)`. Additionally, `ep.consolidate.add_location` will also be based off of the datagram type passed in (i.e IDX/MRU/NME).


ctuguinay:

echopype/convert/parse_base.py
```diff
@@ -564,13 +565,20 @@ def _read_datagrams(self, fid):
...
                 self.nmea["nmea_string"].append(new_datagram["nmea_string"])
 
-            # MRU datagrams contain motion data for each ping for EK80
-            elif new_datagram["type"].startswith("MRU"):
-                self.mru["heading"].append(new_datagram["heading"])
-                self.mru["pitch"].append(new_datagram["pitch"])
-                self.mru["roll"].append(new_datagram["roll"])
-                self.mru["heave"].append(new_datagram["heave"])
-                self.mru["timestamp"].append(new_datagram["timestamp"])
+            # Parse Older MRU datagrams contain motion data for each ping for EK80
```
leewujung, line 568:
```suggestion
            # MRU0 datagrams contain motion data for each ping for EK80
```

echopype/convert/parse_base.py
```diff
@@ -564,13 +565,20 @@ def _read_datagrams(self, fid):
...
+            elif new_datagram["type"].startswith("MRU0"):
+                self.mru0["heading"].append(new_datagram["heading"])
+                self.mru0["pitch"].append(new_datagram["pitch"])
+                self.mru0["roll"].append(new_datagram["roll"])
+                self.mru0["heave"].append(new_datagram["heave"])
+                self.mru0["timestamp"].append(new_datagram["timestamp"])
+
+            # Parse Newer MRU datagrams contain latitude data for each ping for EK80
```
leewujung, line 576:
```suggestion
            # MRU1 datagrams contain latitude/longitude data for each ping for EK80
```

echopype/convert/set_groups_ek80.py
```diff
@@ -303,38 +303,65 @@ def set_platform(self) -> xr.Dataset:
...
-                    np.array(self.parser_obj.mru.get("roll", [np.nan])),
+                    np.array(self.parser_obj.mru0.get("roll", [np.nan])),
                     self._varattrs["platform_var_default"]["roll"],
                 ),
                 "vertical_offset": (
                     ["time2"],
-                    np.array(self.parser_obj.mru.get("heave", [np.nan])),
+                    np.array(self.parser_obj.mru0.get("heave", [np.nan])),
                     self._varattrs["platform_var_default"]["vertical_offset"],
                 ),
```
leewujung, line 366: We may have to change these later in some way because MRU1 also contains the same set of pitch/roll/heave data. Does the lasker data still have MRU0, or does it only have MRU1? Wondering if this should be a substitution, or both should exist (`rol`, `roll_mru1`, etc).

echopype/consolidate/api.py
```diff
@@ -227,77 +234,52 @@ def add_location(
     echodata : EchoData or str or pathlib.Path
         An ``EchoData`` object or path to a file containing the ``EchoData``
         object holding the raw data
-    nmea_sentence
-        NMEA sentence to select a subset of location data (optional)
+    datagram_type : Optional[str], default None
+        The type of datagram used to select latitude and longitude.
+        Can only be used for EK sonar models and value can be either 'NMEA', 'MRU1' or 'IDX'.
+    nmea_sentence : Optional[str], default None
+        NMEA sentence to select a subset of location data
```
leewujung, line 241:
```suggestion
    datagram_type : Optional[str], default None
        Datagram type to use for latitude and longitude.
        Can be 'NMEA', 'MRU1' or 'IDX'.
        Can only be used for EK sonar models.
    nmea_sentence : Optional[str], default None
        NMEA sentence to select a subset of location data
```

echopype/consolidate/loc_utils.py
```diff
@@ -0,0 +1,173 @@
...
+            "Downstream interpolation on the position variables requires unique time values."
+        )
+
+
+def sel_interp(
+    ds: xr.Dataset, echodata: EchoData, loc_name: str, time_dim_name: str, nmea_sentence: str
+) -> xr.DataArray:
+    """
+    If NMEA sentence is not NaN, selects subset of position variable that is within NMEA sentence,
+    and interpolates position variable to match the ping_time dimension of the input dataset.
```
leewujung, line 154: Maybe rewrite this to include the NMEA=NaN case, since that would be the case if people use lat/lon_mru for example.

echopype/consolidate/loc_utils.py
```diff
@@ -0,0 +1,173 @@
...
+logger = _init_logger(__name__)
+
+
+def check_loc_vars_present_all_NaN(
+    echodata: EchoData, lat_name: str, lon_name: str, datagram_type: str
+):
+    """
+    Checks if any of the location variables missing or are all NaN. Recommends using another
+    datagram_type if the corresponding lat/lon variables are not missing and not not all
+    NaN.
```
leewujung, line 16: I think this is a good function. I could understand the code but not the docstring... There's some grammar problem and a redundant not. Could you revise it?

echopype/consolidate/loc_utils.py
```diff
@@ -0,0 +1,173 @@
...
+                error_message = (
+                    error_message
+                    + f" Consider setting datagram_type to any of {good_datagram_types}."
+                )
+
+        # Raise error
+        raise ValueError(error_message)
+
+
+def check_loc_vars_any_NaN_0(echodata: EchoData, lat_name: str, lon_name: str, datagram_type: str):
```
leewujung, line 70: I feel that you can combine this and the above function (`check_loc_vars_present_all_NaN`). They are largely identical except for the actual checks.

leewujung (commented):
Hey @ctuguinay : thanks for the PR! I made a few comments. Not major changes except for the utility function. I think for the ship pipeline we can pull this branch and just continue to update it. I wasn't able to get to the tests, but wanted to get the other parts to you first.

echopype/convert/set_groups_ek80.py
ctuguinay: Yeah, I agree. I also think both should exist where `roll` is mru0 and `roll_mru`.

echopype/consolidate/api.py
ctuguinay: Ah I see. I didn't check for `datagram_type == 'NMEA'` in the code since that would be the `else` (default) case. I agree though that making it explicit is the way to go.

echopype/consolidate/loc_utils.py
ctuguinay: Yup will do 👍

echopype/consolidate/loc_utils.py
ctuguinay: Ah I see. So the 4 cases would be:

1. If `datagram=NMEA` is used and NMEA sentence is NaN, then do nothing.
2. If `datagram=NMEA` is used and NMEA sentence is not NaN, then do the selection.
3. If `datagram!=NMEA` and NMEA sentence is NaN, then do nothing.
4. If `datagram!=NMEA` and NMEA sentence is not NaN, then raise ValueError, ie not allow

I don't think my current code actually deals with the case 4. I'll make this very explicit though.

echopype/consolidate/loc_utils.py
ctuguinay: I agree, I'll make that change 👍

echopype/consolidate/loc_utils.py
```diff
@@ -0,0 +1,162 @@
+import numpy as np
+import xarray as xr
+
+from ..echodata import EchoData
+from ..utils.log import _init_logger
+
+logger = _init_logger(__name__)
+
+
+def compute_invalid_check(lat_var: xr.DataArray, lon_var: xr.DataArray, validity_check: str):
```
leewujung, line 10:
```suggestion
def compute_invalid_check(lat_var: xr.DataArray, lon_var: xr.DataArray, validity_check: str) -> bool:
```

echopype/consolidate/loc_utils.py
```diff
@@ -0,0 +1,162 @@
...
+            raise ValueError(output_message)
+        elif validity_check in ["some_nan", "some_zero"]:
+            logger.warning(output_message)
+
+
+def check_loc_time_dim_duplicates(echodata: EchoData, time_dim_name: str) -> None:
+    """Check if there are duplicates in time_dim_name"""
+    if len(np.unique(echodata["Platform"][time_dim_name].data)) != len(
+        echodata["Platform"][time_dim_name].data
+    ):
```
leewujung, line 112: I wonder how expensive `unique` is. We can keep this in place for now. An alternative I can think of is to check if any of `diff(time_dim_name) == 0`. But of course this would only work if the duplicates happen in consecutive indices (which is most of the cases, but hard to say if all cases). 

How about we give user the option to skip this check in `add_location` and accept that it may fail?

echopype/tests/convert/test_convert_ek80.py
```diff
@@ -431,19 +431,21 @@ def test_convert_ek80_no_fil_coeff(ek80_path):
             assert f"{t}_{p}" not in vendor_spec_ds
 
 
+@pytest.mark.xfail(reason="Parsing MRU1 platform motion data is not yet implemented.")
```
leewujung, line 434:
```suggestion
@pytest.mark.xfail(reason="Setting MRU1 platform motion data to EchoData Platform group is not yet implemented.")
```

echopype/consolidate/loc_utils.py
leewujung, line 10:
```suggestion
def compute_invalid_check(lat_var: xr.DataArray, lon_var: xr.DataArray, validity_check: str) -> bool:
```

echopype/consolidate/loc_utils.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+
+def sel_interp(
+    ds: xr.Dataset,
+    echodata: EchoData,
+    datagram_type: str,
+    loc_name: str,
+    time_dim_name: str,
+    nmea_sentence: str,
```
leewujung, line 125:
```suggestion
    ds: xr.Dataset,
    echodata: EchoData,
    loc_name: str,
    time_dim_name: str,
    nmea_sentence: Union[str, None] = None,
    datagram_type: Union[str, None] = None,
```

echopype/consolidate/loc_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    Selection and interpolation for a location variable.
+
+    The selection logic is as follows, with 4 possible scenarios:
+
+    1) If datagram_type == NMEA is used and NMEA sentence is NaN, then do nothing.
+    2) If datagram_type == NMEA is used and NMEA sentence is not NaN, then do the selection.
+    3) If datagram_type != NMEA and NMEA sentence is NaN, then do nothing.
+    4) If datagram_type != NMEA and NMEA sentence is not NaN, then raise ValueError since NMEA
+    sentence selection can only be used on location variables from the NMEA datagram.
```
leewujung, line 136:
```suggestion
    The selection logic is as follows, with 4 possible scenarios:
    Note here datagram_type = None is equivalent to datagram_type = NMEA
    
    1) If datagram_type is None and nmea_sentence is None, then do nothing.
    2) If datagram_type is None and nmea_sentence is not None, then do the selection.
    3) If datagram_type is not None and nmea_sentence is None, then do nothing.
    4) If datagram_type is not None and nmea_sentence is not None, then raise ValueError since NMEA
       sentence selection can only be used on location variables from the NMEA datagram.
```

leewujung (dismissed):
Thanks @ctuguinay ! I think this one is ready to go. I just have some small changes.

echopype/consolidate/loc_utils.py
ctuguinay: True true 👍

echopype/consolidate/loc_utils.py
ctuguinay: I see what you mean. This could be very expensive since it will require everything to be loaded into memory. This would be pretty bad in the case where we have many many Echodata objects concatenated together, calibrated, and add_depth is used on that. But I'm not quite sure if this should be an option. It feels just a bit awkward and too specific.

ctuguinay: Maybe this won't be too bad though. I didn't encounter any memory issues on this specific part of adding depth when I was running this Krill NASC GPS notebook: https://github.com/OSOceanAcoustics/echopype-examples/pull/52 and I was working with Sv from 170 RAW files.

ctuguinay:
@leewujung Thanks for the review! This should be ready for review. I just had a few comments

echopype/consolidate/loc_utils.py
leewujung: Hmm interesting, thanks for trying it out. Maybe once we get to a point to memory profile all functions in the pipeline together we can see how bad that it.

leewujung (approved):
Thanks for the changes @ctuguinay . Feel free to merge!

Issue #1334, opened 2024-06-14
Add "FM" to an allowable option for `waveform_mode`


leewujung:
Add "FM" (frequency modulated) to be an alternative of  the current "BB" (broadband) in the `waveform_mode` option for `compute_Sv`.

PR #1363, opened 2024-07-18
Allow Waveform Mode FM in Calibrate API

ctuguinay:
Addresses #1334

This was the easiest way I could envision allowing FM. Otherwise, I would've reviewed all the calibrate code and added FM there. What are your thoughts @leewujung?


leewujung:
Yeah I agree with this approach. Just much easier to implement.

The only suggestion I have is to do the following right at the top of `_compute_cal`, then we don't need to change the arguments. This also makes our intension clear:
```python
# Make waveform_mode "FM" equivalent to "BB"
waveform_mode = "BB" if waveform_mode == "FM" else waveform_mode
```

ctuguinay:
@leewujung Thanks! I added your suggestion. This should be ready for another quick review.

leewujung (approved):
LGTM!

Issue #1368, opened 2024-07-25
Allow External `depth_offset` and `tilt` arrays into `add_depth`


ctuguinay:
Allow external `depth_offset` and `tilt` arrays into `add_depth` and perform `ping_time` alignment on them. The alignment is directly related to #1367. This would be beneficial in the case where depth and tilt are not included in the RAW data and are instead in a separate file. An example would be an echosounder on a glider where the acoustic data is separate from the moving platform data.

Issue #1367, opened 2024-07-25
Write a Universal Align to `ping_time` Function


ctuguinay:
There are two functions in Echopype that align non-Echosounder data (Platform and Environment data) to Echosounder data time-wise:  `_var_time2_to_ping_time` `sel_interp`, and `harmonize_env_param_time`. It would be good to have a standard function that does this time-wise alignment.

The function header can look something like this:
`def align_to_ping_time(external_da: xr.DataArray, external_time_name: str, ping_time_da: xr.DataArray)`.

Issue #1332, opened 2024-06-13
Remove Background Noise Removal Failure


ctuguinay:
Found in PR #1331, background noise removal fails at during `test_raw_to_mvbs` in the test file `echopype/tests/utils/test_processinglevels_integration.py`. Not quite sure what the problem is but I'll get to it later since it is out of the scope of #1331.

```
______________________________________ test_raw_to_mvbs[AZFP-AZFP-raw_and_xml_paths1-extras1] ______________________________________

sonar_model = 'AZFP', path_model = 'AZFP', raw_and_xml_paths = ('17082117.01A', '17041823.XML')
extras = {'latitude': 45.0, 'longitude': -60.0, 'pressure': 59, 'salinity': 27.9}
test_path = {'AD2CP': PosixPath('/home/exouser/echopype/echopype/test_data/ad2cp'), 'AZFP': PosixPath('/home/exouser/echopype/echo...me/exouser/echopype/echopype/test_data/ea640'), 'ECS': PosixPath('/home/exouser/echopype/echopype/test_data/ecs'), ...}

    @pytest.mark.test
    @pytest.mark.parametrize(
        ["sonar_model", "path_model", "raw_and_xml_paths", "extras"],
        [
            pytest.param(
                "EK60",
                "EK60",
                ("Winter2017-D20170115-T150122.raw", None),
                {},
                # marks=pytest.mark.skipif(sys.platform == "win32", reason="Test data not available on windows tests"),
            ),
            pytest.param(
                "AZFP",
                "AZFP",
                ("17082117.01A", "17041823.XML"),
                {"longitude": -60.0, "latitude": 45.0, "salinity": 27.9, "pressure": 59},
                # marks=pytest.mark.skipif(sys.platform == "win32", reason="Test data not available on windows tests"),
            ),
        ],
    )
    def test_raw_to_mvbs(
            sonar_model,
            path_model,
            raw_and_xml_paths,
            extras,
            test_path
    ):
        # Prepare the Sv dataset
        raw_path = test_path[path_model] / raw_and_xml_paths[0]
        if raw_and_xml_paths[1]:
            xml_path = test_path[path_model] / raw_and_xml_paths[1]
        else:
            xml_path = None
    
        def _presence_test(test_ds, processing_level):
            assert "processing_level" in test_ds.attrs
            assert "processing_level_url" in test_ds.attrs
            assert test_ds.attrs["processing_level"] == processing_level
    
        def _absence_test(test_ds):
[...]

PR #1369, opened 2024-07-25
Allow external arrays in add depth; Add universal ping time alignment function

ctuguinay:
Addresses #1368 and #1367.

Also, addresses #1332 since this was also related to the AZFP Glider notebook.


ctuguinay:

echopype/utils/align.py
```diff
@@ -0,0 +1,61 @@
...
+    external_da : xr.DataArray
+        External Non-Echosounder data.
+    external_time_name : str
+        Time variable name of the External Non-Echosounder data.
+    ping_time_da : xr.DataArray
+        Echosounder ping time.
+    method : str, default 'nearest'
+        Interpolation method. Not used if external time matches ping time or external
+        DataArray is a single value.
+        For more interpolation methods please visit: https://docs.xarray.dev/en/stable/generated/xarray.DataArray.interp.html # noqa
```
leewujung, line 25: Just removing some capitalizations
```suggestion
    Aligns an external DataArray to align time-wise with the echosounder ping time DataArray.

    Parameters
    ----------
    external_da : xr.DataArray
        External non-echosounder data.
    external_time_name : str
        Time variable name of the external non-echosounder data.
    ping_time_da : xr.DataArray
        Echosounder ping time.
    method : str, default 'nearest'
        Interpolation method. Not used if external time matches ping time or external
        DataArray is a single value.
        For more interpolation methods please visit: https://docs.xarray.dev/en/stable/generated/xarray.DataArray.interp.html # noqa
```

echopype/utils/align.py
```diff
@@ -0,0 +1,61 @@
...
+
+    Returns
+    -------
+    aligned_da : xr.DataArray
+        External Non-Echosounder data that is now aligned with the Echosounder ping time.
```
leewujung, line 30:
```suggestion
    aligned_da : xr.DataArray
        External non-echosounder data that is now aligned with the echosounder ping time.
```

echopype/calibrate/env_params.py
```diff
@@ -63,13 +64,9 @@ def harmonize_env_param_time(
                 f"ping_time needs to be provided for comparison or interpolating {p.name}"
             )
```
leewujung, line 65: I wonder if the checks in Line 51-65 should be part of `align_to_ping_time`. It seems that not all cases overlap, but the check on "# If there's only 1 time1 value:" seems to be in `align_to_ping_time` already?

echopype/consolidate/api.py
```diff
@@ -82,11 +84,15 @@ def add_depth(
         Source Sv dataset to which a depth variable will be added.
     echodata : Optional[EchoData, str, pathlib.Path], default `None`
         `EchoData` object from which the `Sv` dataset originated.
-    depth_offset : Optional[float], default None
+    depth_offset : Optional[Union[Number, xr.DataArray]], default None
         Offset along the vertical (depth) dimension to account for actual transducer
         position in water, since `echo_range` is counted from transducer surface.
-    tilt : Optional[float], default None.
+        If provided as an xr.DataArray, it must have a single dimension corresponding to time.
+        The data array should represent the depth offset at each time step.
```
leewujung, line 91:
```suggestion
        If provided as an xr.DataArray, it must have a single dimension corresponding to time.
        The data array should represent the depth offset at each time step, in meters.
```

echopype/consolidate/api.py
```diff
@@ -82,11 +84,15 @@ def add_depth(
...
+    tilt : Optional[Union[Number, xr.DataArray]], default None.
         Transducer tilt angle [degree]. 0 corresponds to a transducer pointing vertically.
+        If provided as an xr.DataArray, it must have a single dimension corresponding to time.
+        The data array should represent the tilt angle at each time step.
```
leewujung, line 95:
```suggestion
        If provided as an xr.DataArray, it must have a single dimension corresponding to time.
        The data array should represent the tilt angle in degrees at each time step.
```

echopype/consolidate/loc_utils.py
```diff
@@ -118,48 +118,32 @@ def check_loc_time_dim_duplicates(echodata: EchoData, time_dim_name: str) -> Non
...
-    time_dim_name: str,
     nmea_sentence: Union[str, None] = None,
     datagram_type: Union[str, None] = None,
 ) -> xr.DataArray:
     """
-    Selection and interpolation for a location variable.
+    Select location subset for a location variable based on NMEA sentence.
 
-    The selection logic is as follows, with 4 possible scenarios:
+    The selection logic is as follows, with 3 possible scenarios:
```
leewujung, line 130: still 4 in the list below? but maybe you meant the actual code that only has 3 cases? so the 2 do-nothing cases are just the `else`?

echopype/tests/calibrate/test_calibrate.py
```diff
@@ -310,53 +310,12 @@ def test_compute_Sv_combined_ed_ping_time_extend_past_time1():
...
-                            )
-                        ).data
-                    ),
-                    env_var.isel(
-                        channel=channel_index
-                    ).sel(
-                        ping_time=slice(time1.max())
-                    ).data[-1]
-                )
-
```
leewujung, line 359: By removing the above section, I gathered that basically this is now tested under `test_align` so no longer needed here?

leewujung (dismissed):
Hey @ctuguinay : Thanks for the changes! Other than some minor wording/capitalization things and small questions, I think this is good to go!

echopype/calibrate/env_params.py
ctuguinay: So that line sends back a single value, while `align_to_ping_time` will always send back an array of the same length as the ping time DataArray receives.

echopype/consolidate/loc_utils.py
ctuguinay: Oh yes yes, still 4 but the 2 do-nothing cases are part of else. I will revert back to 4 since that makes more sense since I'm describing the scenarios and not exactly the code itself. Nice catch!

echopype/tests/calibrate/test_calibrate.py
ctuguinay: Yeah, so in `echopype/tests/utils/test_align.py`, I test for multiple bfill and ffill scenarios, but I just can it extrapolation, since we can specify how we want to extrapolate and bfill and ffill are just 'nearest' extrapolations.

ctuguinay:
@leewujung Thanks for the review! This should be ready for a quick lookover again.

echopype/calibrate/env_params.py
leewujung: Ah I see, ok.

leewujung (approved):
Cool! This is ready to be merged now.

Issue #1375, opened 2024-08-16
`apply_mask` does not interpret `NaN` correctly


leewujung:
Apparently right now `echopype.mask.apply_mask` does not interpret `NaN` as 0 or `False` in the mask production, and actually treats it as 1 or `True`. 

This is because `xr.where` treat `NaN` in _cond_ as `True`... kinda unexpected behavior.

I'll follow up with a PR to fix this.

ctuguinay:
Closed by merging of #1376

PR #1376, opened 2024-08-16
Handle `NaN` in `apply_mask` input mask

leewujung:
This addresses #1375.

I added a simple test, because it doesn't look like this one would fit under any of the existing parameterized tests, but let me know if that's not the case.


ctuguinay (approved):
@leewujung Wow thanks for spotting this. Looks good to me!

Issue #1378, opened 2024-08-23
Eager computation in `compute_MVBS` setting of attributes


ctuguinay:
There exists eager computation in `compute_MVBS` when setting the `actual_range` attributes:

```
    ds_MVBS["Sv"] = ds_MVBS["Sv"].assign_attrs(
        {
            "cell_methods": (
                f"ping_time: mean (interval: {ping_time_bin_resvalue} {ping_time_bin_resunit_label} "  # noqa
                "comment: ping_time is the interval start) "
                f"{range_var}: mean (interval: {range_bin} meter "
                f"comment: {range_var} is the interval start)"
            ),
            "binning_mode": "physical units",
            "range_meter_interval": str(range_bin) + "m",
            "ping_time_interval": ping_time_bin,
            "actual_range": [
                round(float(ds_MVBS["Sv"].min().values), 2),
                round(float(ds_MVBS["Sv"].max().values), 2),
            ],
        }
    )
```
https://github.com/OSOceanAcoustics/echopype/blob/be2b86cc668b59b0b856789de29614bcc25000bd/echopype/commongrid/api.py#L164C1-L183C6

The calling of `.values` evaluates the `.min()` and `.max()`. The `actual_range` attribute should be removed so that the function can be kept entirely lazy.

leewujung:
I wonder if this is something that we can do as a side function at the end of the `compute_MVBS` code. So instead of having this and preventing the computation to be lazy, we do all ops needed to get the MVBS datasets first, and then call a function to lazily find the min/max and assign these values to the attributes before existing the `computer_MVBS` function.

PR #1380, opened 2024-08-23
Remove min max attributes for `compute_MVBS`

ctuguinay:
Addresses #1378

### DIFF
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -175,10 +175,6 @@ def compute_MVBS(
             "binning_mode": "physical units",
             "range_meter_interval": str(range_bin) + "m",
             "ping_time_interval": ping_time_bin,
-            "actual_range": [
-                round(float(ds_MVBS["Sv"].min().values), 2),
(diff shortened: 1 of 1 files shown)


ctuguinay:
@leewujung Just removed those few lines. This should be ready for a review.

leewujung (approved):
Oh wow, thanks for finding out the second one!

The only thing I can think of is whether not having this attribute would impact automatic plotting in xarray or holoviz. I seem to remember when I was doing the MVBS downsampling for the ship echoshader plotting, something wasn't happy when the new MVBS dataarray did not have attributes. But I don't remember which one tripped it.

I just did a simple test using the dataset on this hvplot Image [example page](https://hvplot.holoviz.org/reference/xarray/image.html), and it plots without problem without the `actual_range` attributes.

So I think we are good to merge! We should probably add a standalone separate function to add `actual_range` back to the MVBS dataarray. This way if it is required in some way it will be easy to produce, and can be lazily evaluated.

ctuguinay:
@leewujung Thanks for the review! I'll add an issue for the standalone function

Issue #1374, opened 2024-08-14
Errors when converting NWFSC Hake Survey Data from 2017 and 2021


ctuguinay:
Some errors I found while converting all of 2017, 2019, and 2021 files using the latest Echopype main branch:

```
Error converting noaa-wcsd-pds/data/raw/Bell_M._Shimada/SH1707/EK60/Summer2017-D20170912-T194552.raw with Exception: Short read while getting trailing raw file datagram size for check 4 != 0 @ (19315440L, 32080)
Error converting noaa-wcsd-pds/data/raw/Bell_M._Shimada/SH1707/EK60/Summer2017-D20170718-T121359.raw with Exception: Short read while getting trailing raw file datagram size for check 4 != 0 @ (1311056L, 540)
Error converting noaa-wcsd-pds/data/raw/Bell_M._Shimada/SH1707/EK60/Summer2017-D20170819-T060438.raw with Exception: Short read while getting dgram size 4 != 0 @ (1449574268L, 1458)
Error converting noaa-wcsd-pds/data/raw/Bell_M._Shimada/SH1707/EK60/Summer2017-D20170807-T171736.raw with Exception: The DType <class 'numpy.dtypes.DateTime64DType'> could not be promoted by <class 'numpy.dtypes.Float64DType'>. This means that no common DType exists for the given inputs. For example they cannot be stored in a single array unless the dtype is `object`. The full list of DTypes is: (<class 'numpy.dtypes.DateTime64DType'>, <class 'numpy.dtypes.Float64DType'>)
```

```
Error converting noaa-wcsd-pds/data/raw/Bell_M._Shimada/SH2106/EK80/Hake-D20210913-T130612.raw with Exception: cannot reindex or align along dimension 'ping_time' because the (pandas) index has duplicate values
Error converting noaa-wcsd-pds/data/raw/Bell_M._Shimada/SH2106/EK80/Hake-D20210913-T225435.raw with Exception: cannot reindex or align along dimension 'ping_time' because the (pandas) index has duplicate values
```

ctuguinay:
The `short read` error is probably the EK software abruptly ending the file write, but perhaps there's a way to retrieve the data that has been written properly instead of just losing it completely to this error.

ctuguinay:
`cannot reindex` error stems from the fact that the ping times have duplicate values in them:

![image](https://github.com/user-attachments/assets/3c0c6545-97a4-434f-b105-e5a408e4d29c)

[...]

PR #1382, opened 2024-08-27
Drop Ping Time Duplicates

ctuguinay:
Addresses the duplicate ping time part of #1374


ctuguinay:

leewujung (commented):
@ctuguinay : Thanks for looking into this. I agree this would address the problem, but one thing I think we should check is if the entries with duplicated `ping_time` actually contain the same data. If not, the duplication of `ping_time` does not make sense, but dropping it would also mean that we lose data?

Also, the same as my comment on another PR, that if both files you used in the tests have the same issues, let's just keep one of them to avoid our test suite getting too big.

leewujung:
@ctuguinay: I just got back to this and saw my previous comments, which still stands. We can merge this once those are addressed. Thanks!

ctuguinay:
@leewujung So I did a quick check of the first two duplicated ping time pairs to see if this was dropping ping times that contained duplicate data and it turns out that they were completely equal:

image

And here's a more generalized check that shows that subsequent ping time pairs are all equal:

image

> Thanks for looking into this. I agree this would address the problem, but one thing I think we should check is if the entries with duplicated ping_time actually contain the same data. If not, the duplication of ping_time does not make sense, but dropping it would also mean that we lose data?

And I was thinking about this again, and I think we would either need to drop duplicate ping times even if they contained unique data or we would need to apply some tiny time offset to them such that they did not have the same duplicate ping times, since `xr.concat` in `set_sonar()` will fail if there are any duplicates in the datasets were are concatenating.

leewujung:
> And I was thinking about this again, and I think we would either need to drop duplicate ping times even if they contained unique data or we would need to apply some tiny time offset to them such that they did not have the same duplicate ping times, since `xr.concat` in `set_sonar()` will fail if there are any duplicates in the datasets were are concatenating.
> 
> I don't have a .raw file where I can test this tiny time offset idea on since I don't I have seen a dataset where the ping time duplicates contain unique data.

I agree. In the grand scale of things, I kinda feel that dropping one ping is not a big deal, but perhaps that goes against the 0% lossless idea for the raw-converted data.

Perhaps what we can do is to drop duplicated `ping_time` regardless, but spit out a warning to user if the dropped pings contain different data compared to the retained ping?

I'll shoot an email to Simrad rep to see if he has thoughts about why the pings are duplicated in the first place, and what he thinks about your idea to add a tiny time offset. Will cc you!

leewujung:
I took a look at the file you used above, and found out that in fact ALL PINGS are duplicated (!), from the dictionary fields `low_date`/`high_date` in ParseEK80:
Screenshot 2024-12-31 at 8 38 56 AM

The data appear to be completely duplicated as well.

Maybe something was happening here with the instrument?

@ctuguinay How many files in the set you were working on encountered this problem?

ctuguinay:
@leewujung Oh huh that is truly strange. I only had 2 files from the 2021 NWFSC Pacific Hake Survey (Shimada) that displayed this problem.

I wonder if this is just a streaming issue so that duplicate datagrams were sequentially saved to the same place instead of two separate places. I also wonder what Echoview makes of this. Perhaps it just ignores the second ping?

leewujung:
Good that that was just 2 files, so not a big hit in terms of data storage.

Yeah would be good to see what Echoview handling looks like. Maybe we can ask folks with a license to help here.

ctuguinay:
@leewujung This should be ready for review again. I think the pinning worked? Not sure.

leewujung:
@ctuguinay : Could you try unpinning zarr and see if everything passes here? The failed tests in #1429 are likely in other modules and not the ones changed in this PR. (There I used the [ci tests all] PR title to trigger running _all_ tests; here only the ones in the modules that had code changes are run.)

leewujung (approved):
Looks good. Thanks!

Issue #1386, opened 2024-09-03
Problematic EK80 Environment Datagrams for Replayed Files and Large Files


ctuguinay:
From conversation with @leewujung,

Currently, the Echopype parsing code in `parse_base.py` expects the EK80 Environment datagram to be unique, i.e., we only get one XML datagram with subtype environment. However, when parsing replayed files and large files (~2GiB) with EK80 `application_version='23.6.2.0'` and `file_format_version='1.32'` (the latest versions), we encounter multiple XML datagrams with subtype environment where only the first datagram contains all the environment variables we need for calibration (temp, salinity, pressure, etc.) and the latter datagrams contain only `drop_keel_offset` and `drop_keel_offset_is_manual`. The latter datagrams override the first datagram and we end up with an EchoData object that does not contain the environment parameters we need for calibration.

To fix this, we will drop XML datagrams with subtype environment where the only data available is `drop_keel_offset` and `drop_keel_offset_is_manual`. We intend this to be a temporary fix since we believe it to be a bug in the EK80 software.

PR #1387, opened 2024-09-03
Skip Parsing Invalid EK80 Environment Datagram(s)

ctuguinay:
Temporary fix for #1386

### DIFF
--- a/echopype/convert/parse_base.py
+++ b/echopype/convert/parse_base.py
@@ -495,8 +495,9 @@ class ParseEK(ParseBase):
 
             # XML datagrams store environment or instrument parameters for EK80
             if new_datagram["type"].startswith("XML"):
-                # Check that environment datagrams aren't invalid
-                # Temporary fix for suspected EK80 software bug described in:
+                # Check that environment datagrams contain more than
+                # just drop_keel_offset and drop_keel_offset_is_manual
+                # Temporary fix for handling >1 EK80 environment datagrams described in:
                 # https://github.com/OSOceanAcoustics/echopype/issues/1386
                 if new_datagram["subtype"] == "environment" and set(
                     ["drop_keel_offset", "drop_keel_offset_is_manual"]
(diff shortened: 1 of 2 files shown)


ctuguinay:

echopype/convert/parse_base.py
```diff
@@ -495,7 +495,12 @@ def _read_datagrams(self, fid):
 
             # XML datagrams store environment or instrument parameters for EK80
             if new_datagram["type"].startswith("XML"):
-                if new_datagram["subtype"] == "environment":
+                # Check that environment datagrams aren't invalid
+                # Temporary fix for suspected EK80 software bug described in:
+                # https://github.com/OSOceanAcoustics/echopype/issues/1386
+                if new_datagram["subtype"] == "environment" and set(
+                    ["drop_keel_offset", "drop_keel_offset_is_manual"]
+                ) != set(new_datagram["environment"].keys()):
```
leewujung, line 503: So I don't think this is a "bug" per se - if the EK80 software only uses variables existing in the data to update computation, that is entirely valid. It is us that need to handle the variability.

The part that I referred to as a potential bug is that the replayed environment datagram has the same timestamp of the original time the data was collected -- but this may actually be intentional, since in a replayed scenario there is no environmental data being collected.
```suggestion
                # Check that environment datagrams contain more than
                # just drop_keel_offset and drop_keel_offset_is_manual
                # Temporary fix for handling >1 EK80 environment datagrams described in:
                # https://github.com/OSOceanAcoustics/echopype/issues/1386
                if new_datagram["subtype"] == "environment" and set(
                    ["drop_keel_offset", "drop_keel_offset_is_manual"]
                ) != set(new_datagram["environment"].keys()):
```

leewujung (commented):
I suggested some text for comments. Other than that, I think we can merge this.

ctuguinay:
@leewujung Thanks for the review! And yeah you're right, it's not a bug per se. I just added your suggestion.

leewujung (approved):

PR #1398, opened 2024-10-25
Update setup-services.py

spacetimeengineer:
The docker container name is mismatched from what is found in setup-services.py so the tests cant run without this fix.


leewujung:
Hey @spacetimeengineer - Thanks for the PR. What OS do you use? This is something I've run into on my machine (MacOS) but it does not seem to occur all the time for others. Seems that the tests have run fine on Linux. Let me ask someone in my group who uses Windows to give this a try, and then we can merge this.

leewujung:
Ok, great. @brandynlucca reported back that this work on Windows, and I just tested on MacOS too. I'll get this merged! :)

PR #1403, opened 2024-10-30
Add error warning for all NaN lat/lon in `get_distance_from_latlon`

leewujung:
When the lat/lon dataframe contains all NaN entries, the `get_distance_from_latlon` spits out errors that are hard to decipher. This PR adds a clear error message for this case.

`get_distance_from_latlon` is currently used in `compute_MVBS`.


ctuguinay (approved):
Good catch LGTM! I'll check out what's causing this error on our labels.

Issue #1406, opened 2024-11-07
[Data conversion] EA640 files fails to calibrate after upgrading from 0.8.4 to 0.9.0


torsteinnh:
### General description of problem

After upgrading from 0.8.4 to 0.9.0, by EA640 raw data files fails to compute Sv.
The bug is caused by an oversight in _check_echodata_backscatter_size, where total_nbytes remains unset for all sonar models except for EK60, EK80 and AZFP.

### Echopype version

0.9.0

### How did you install Echopype (e.g., conda or pip)

pip

### What is your operating system

macOS 14.6.1

### Minimal code example

import echopype as ep

echo_data = ep.open_raw(raw_file=path, sonar_model="EA640")
echo_data_sv = ep.calibrate.compute_Sv(echo_data, waveform_mode="CW", encode_mode="power")

### Error message printouts

```shell
Traceback (most recent call last):
  File "/Users/tnh/Documents/workspace/PGS_playground/demos/demo_pgs.py", line 14, in <module>
    echo_data_sv = ep.calibrate.compute_Sv(echo_data, waveform_mode="CW", encode_mode="power")
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tnh/.virtualenvs/echo-test/lib/python3.11/site-packages/echopype/calibrate/api.py", line 208, in compute_Sv
    return _compute_cal(cal_type="Sv", echodata=echodata, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tnh/.virtualenvs/echo-test/lib/python3.11/site-packages/echopype/calibrate/api.py", line 62, in _compute_cal
    cal_obj._check_echodata_backscatter_size()
  File "/Users/tnh/.virtualenvs/echo-test/lib/python3.11/site-packages/echopype/calibrate/calibrate_base.py", line 117, in _check_echodata_backscatter_size
    total_gb = total_nbytes / (1024**3)
               ^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'total_nbytes' where it is not associated with a value
Exception ignored in: <function EchoData.__del__ at 0x168f640e0>
Traceback (most recent call last):
[...]

PR #1407, opened 2024-11-07
Update calibrate_base.py to include EA640 in backscatter size check.

torsteinnh:
This patch solves my problem in https://github.com/OSOceanAcoustics/echopype/issues/1406.
I am unfortunately not able to fix the problem for all supported echo sounder models since I don't have test files at hand.
This fix has been tested to work with my EA640 files, and does not interfere with my EK60 or EK80 files.


leewujung:
@torsteinnh : Thanks for the PR! Sorry it took me forever to get to this! 

@ctuguinay : Pinging you here since I think you added this function? Seems like the bug is a coverage problem, that the list does not include all `sonar_model`s. What do you think if we make `self.echodata.sonar_model == "EK80"` to be the `if` condition, and lump all `else` into grabbing data only from `backscatter_r`? Since right now there aren't other echosounders that spit out complex data (hence only `Beam_group1` and `backscatter_r` exist)

torsteinnh:
Good idea, I added that fix now.

PR #1409, opened 2024-11-09
Temporarily pin flox <0.9.14 [all tests ci]

leewujung:
Seems like the function that caused the most recent test failure was `_datetime_nanmin` introduced just 4 days ago at flox=0.9.14 ([here](https://github.com/xarray-contrib/flox/blob/4f6164f74a657d95f795c374e31e91a7e7643e9f/flox/xrutils.py#L348)). Probably why the CI runs with #1403 didn't catch this when it was run a few weeks back. Let's see if this is the issue in the new CI runs.

### DIFF
--- a/docs/source/contributing.md
+++ b/docs/source/contributing.md
@@ -50,7 +50,7 @@ To create an environment for developing Echopype, we recommend the following ste
     ```shell
     # Create a conda environment using the supplied requirements files
     # Note the last one docs/requirements.txt is only required for building docs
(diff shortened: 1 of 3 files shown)


leewujung:
Ok, confirmed that this is a flox issue.

leewujung:
The bug causing errors in flox=0.9.14 has been reverted in flox=0.9.15, so I'll close this now.

Issue #1418, opened 2024-12-17
Echopype publishes python-3.12 packages eventhough it is not compatible with python-3.12


peter-urban:
Echopype provides conda-forge and pypi packages for python 3.12 (and probably even 3.13 and 3.14) even though it is not compatible (https://github.com/OSOceanAcoustics/echopype/issues/1408). This creates difficult-to-understand problems (e.g. https://github.com/OSOceanAcoustics/echopype/issues/1415). 

Given that echopype has such an extensive ci-test-suit that would easily catch such issues, the maximum version for published packages on pypi and conda should be set to the maximum python-version used in the ci-test suit. I am not familiar with build/publish procedure applied, but is it possible to specify a maximum (and also minimum) python version for the package? (e.g. the requires-python field in [project] section of the pyproject.toml (https://peps.python.org/pep-0621/)

ctuguinay:
Oh yeah let me try specifying minimum python version in `pyproject.toml`. I'll also work on supporting Python 3.12 these next few days.

leewujung:
Our current test matrix only goes up to 3.11, so the errors are not caught. We know that it is likely the datatree upgrade of xarray that triggered these conflicts, which makes this a bigger task (see #1405 and #1408). @ctuguinay thanks for taking this on!

leewujung:
Related to this - I just realized that the batch of latest changes in the installation doc page is not in the `stable` RTD version folks see, but in `latest` since we have not released for a while. In [that page](https://echopype.readthedocs.io/en/latest/installation.html) we have a warning about 3.12 not being supported. Since it's been so long, I think we can just wait for #1419 and then release.

Issue #1408, opened 2024-11-09
Support for python 3.12


leewujung:
Currently our tests only run on python 3.9, 3.10, 3.11, and from various previous error reports (https://github.com/OSOceanAcoustics/echopype/issues/1370, https://github.com/OSOceanAcoustics/echopype/issues/1371, https://github.com/OSOceanAcoustics/echopype/issues/1391), looks like we will need to unpin the `datatree` version to do this. Coincidentally (which is probably not a coincidence :P)`datatree` is now officially supported by xarray (https://github.com/pydata/xarray/discussions/9680), so looks like this is a good time for us to resolve these 2 problems together!

Related issue: #1405

ctuguinay:
This might get somewhat involved because Xarray DataTree is being more strict in its alignment between children and parent nodes, so I'll detail some of the steps that I'll take once I've discovered all alignment mismatches that pose problems in our CI steps.

leewujung:
Yeah, given the error messages I suspected that this was because of the datatree upgrade (#1405) and it would become a large task. Let me know if I can help somewhere.

PR #1419, opened 2024-12-26
Support Python 3.12 and Migrate to Xarray DataTree

### LINKED ISSUE #1405: Update datatree support
Xarray v2024.10.0 brings on official support for xarray.DataTree !
https://github.com/pydata/xarray/discussions/9680 

We need to update our use of the previous pinned datatree version to ensure long-term compatibility.

**Edit**:
- Looking into a few previous issues below, looks like we need to update datatree in order to get to python=3.12. See #1408.
- Related to #1420.

ctuguinay:
Still in draft.

Resolves #1405, #1408, #1418.

Following the DataTree Migration Guide detailed here: https://github.com/pydata/xarray/blob/main/DATATREE_MIGRATION_GUIDE.md.


ctuguinay:
I commented out `TestEchoData` in `echopype/tests/echodata/test_echodata.py` since it was failing and it had most likely something to do with the backward compatibility issue #1420.

ctuguinay:
This should be ready for review @leewujung

echopype/convert/set_groups_base.py
```diff
@@ -149,14 +149,14 @@ def set_nmea(self) -> xr.Dataset:
...
+                    ["nmea_time"],
                     raw_nmea,
                     {"long_name": "NMEA datagram"},
                 )
             },
             coords={
-                "time1": (
-                    ["time1"],
+                "nmea_time": (
+                    ["nmea_time"],
```
leewujung, line 159: How about `time_nmea` to be consistent with the noun-modifier pattern that's used elsewhere?

leewujung: Ugh just saw that there's `ping_time` that deviates from this pattern, and it would be odd to make it `time_ping`. Personally I feel it is cleaner for the other ones to be all like `time_X`...

echopype/echodata/echodata.py
```diff
@@ -242,7 +242,7 @@ def group_paths(self) -> Set[str]:
     def __get_dataset(node: DataTree) -> Optional[xr.Dataset]:
         if node.has_data or node.has_attrs:
             # validate and clean dtypes
-            return sanitize_dtypes(node.ds)
+            return sanitize_dtypes(node.to_dataset())
         return None
```
leewujung, line 246: I found it a bit curious that `__get_datset` is set as a static method. I wonder if this has anything to do with how it is used in multiple places, like `__getitem__`, `__setitem`, `__setattr__` -- so this function would get evaluated multiple times, which seems not the most efficient thing to do. Does decorating it as a static method remove the redundancy??

leewujung (approved):
@ctuguinay : Thanks for the investigation and fixes! I think this is ready to be merged once the time name is settled. My comment about `__get_dataset` is probably better addressed separately, if needed.

echopype/echodata/echodata.py
leewujung: @oftfrfbf: pinging you here since you may be running into these in #1420, and also this PR won't be changing the use of `__get_dataset`. Tanks!

echopype/convert/set_groups_base.py
ctuguinay: Oh huh I never thought about it that way. I personally see them both as a standalone/compound noun since 'ping' + 'time' refer to different things but together form this one distinct thing and 'ping time' could be operated on like "short + 'ping time'", and 'nmea time' I feel could be operated on similarly.

echopype/echodata/echodata.py
ctuguinay: I may be wrong, but I'm fairly sure the static decorator is there because the function itself does not need the class to be used. It could be a standalone function, but because it's used so often in the class, it's better to just put it in the class, so I don't think it removed this redundancy.

ctuguinay: So yeah, I think I agree with you on that. It really only should be set on the stuff that actually sets the items, not fetches them. Fetching them for view we can send a `DatasetView` object, and for something to be actually modified or set in the `EchoData` object, `Dataset` should be sent instead.

leewujung: That's my understanding of static methods too, so yeah, probably a good idea to remove that redundancy. I'll put it in an issue.

leewujung: Here we go! https://github.com/OSOceanAcoustics/echopype/issues/1427

echopype/convert/set_groups_base.py
leewujung: Ah I see what you mean, so ping time is when a ping is issues, and nmea is when an nmea sentence is issues. I agree with that. Perhaps we can just keep it this way here. We will at some point need a larger breaking change to keep up with convention changes once more feedback from the community is discussed, but that'll be at least a few months down the road.

Issue #1428, opened 2025-01-16
Echopype not compatible with zarr 3.0.0 !?


hentatisundberg:
Hi, 
It seems like the present version of echopype (0.9.1) is not compatible with the recent release of zarr (3.0.0). I installed echopype with pip in a virtual environment (including the requirements file), and when running open_raw() I get: 

_ImportError: cannot import name 'GroupNotFoundError' from 'zarr.errors' (/Users/jonas/Documents/Programming/python/echopype/venv/lib/python3.13/site-packages/zarr/errors.py)_

Run code and link to example .raw file below. Thanks in advance! 

import echopype as ep
file_path = 'SLUAquaSailor2020V2-Phase0-D20210501-T111709-1.raw'
ed = ep.open_raw(file_path, sonar_model="EK80", use_swap=True)

raw file: 
https://drive.google.com/file/d/1elkyzejWWkceh997VaJhKJD_2FITXii5/view?usp=sharing

leewujung:
Hey @hentatisundberg : I suspect this is related to the pending upgrade to python=3.12 (#1405 and #1420 are in the works...). If you install from conda at v0.9.1 it should pull an older version of zarr. But let me know if that's not the case.

hentatisundberg:
Yes you are probably right, I should have mentioned I also upgraded to python 3.12. Happy to see this moving forward!

oftfrfbf:
For those who are installing using pip installers would it make sense to pin the zarr version temporarily with:
```
zarr>=2,<3
```
as mentioned in the migration guide here:

https://zarr.readthedocs.io/en/latest/user-guide/v3_migration.html

leewujung:
Thanks @oftfrfbf - I wonder if we should bump a release with zarr pinned (and potentially netcdf4 too because of #1426).

leewujung:
So typically we try to pin as few dependency as possible, unless absolutely necessary, especially when we do < some version. I was hoping we can have the zarr 3.0 advantages "for free," but I guess not! Since the breaking changes would likely require some work/time for us to catch up, it's best to pin it so that folks don't keep on erroring out. 

I'll do another release with zarr and potentially netcdf4 pinned on in the next couple days.

We should try to unpin it as soon as possible, since zarr is one of our key dependencies. This way we don't become outdated and the package slowly becomes unusable since all other packages continue to move forward (kinda like the py>=3.12 case).

leewujung:
[...]

Issue #1426, opened 2024-12-30
pip install failed with HDF5_DIR environment variable issue


leewujung:
With python=3.9-3.11, `pip install echopype` failed with error `HDF5_DIR environment variable not set` shown below.

Some searches led to the same issues reported by others:
- https://github.com/Unidata/netcdf4-python/issues/901
- [on OS-X](https://stackoverflow.com/questions/18107298/cannot-install-netcdf4-python-package-on-os-x)
- [on Windows](https://github.com/Unidata/netcdf4-python/issues/1184): though an earlier version than our pinned >1.6.0 so not a problem for us

I'll try manually setting the env variable to see if that resolves the issue.

```shell
(echopype_20241230_py39) wujung@pseudorca ~ % pip install echopype
Collecting echopype
  Using cached echopype-0.9.1-py3-none-any.whl.metadata (8.7 kB)
Collecting bottleneck (from echopype)
  Using cached Bottleneck-1.4.2-cp39-cp39-macosx_11_0_arm64.whl.metadata (7.7 kB)
Collecting dask[array,distributed] (from echopype)
  Using cached dask-2024.8.0-py3-none-any.whl.metadata (3.8 kB)
Collecting dask-image (from echopype)
  Using cached dask_image-2024.5.3-py3-none-any.whl.metadata (2.8 kB)
Collecting jinja2 (from echopype)
  Using cached jinja2-3.1.5-py3-none-any.whl.metadata (2.6 kB)
Collecting netCDF4>1.6 (from echopype)
  Using cached netcdf4-1.7.2.tar.gz (835 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [36 lines of output]
      reading from setup.cfg...
          HDF5_DIR environment variable not set, checking some standard locations ..
      checking /Users/wujung/miniconda3/envs/echopype_20241230_py39/include...
      hdf5 headers not found in /Users/wujung/miniconda3/envs/echopype_20241230_py39/include
      checking /Users/wujung/miniconda3/envs/echopype_20241230_py39/Library/include...
      hdf5 headers not found in /Users/wujung/miniconda3/envs/echopype_20241230_py39/Library/include
      checking /Users/wujung/include...
      hdf5 headers not found in /Users/wujung/include
      checking /usr/local/include...
[...]

PR #1429, opened 2025-01-21
Pin `zarr` and `netcdf4` temporarily [all tests ci]

leewujung:
This is a temporary measure for #1426 and #1428. 
We should try to unpin them as soon as possible, especially to migrate to Zarr-Python 3.

To make it easier to debug what's failing, I also addressed a few warnings:
- use `np.all` in place of `np.alltrue` -- this is one of the numpy>=2 changes
- use `np.argmin(da.data)` instead of `da.argmin()`, which will be deprecated at some point
- use "S" instead of "s" for setting MVBS `ping_time_bin` in tests and in `testing.py`
  > Warning: 'S' is deprecated and will be removed in a future version, please use 's' instead.
- explicitly extract the single element when computing `nsamples` and `L` in `tapered_chirp`
  > Warning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)

**NOTE: Remaining warnings are recorded in #1439. We need to resolve those very soon.**


leewujung:
Remaining warnings are recorded in #1439. We need to resolve those very soon.

PR #1441, opened 2025-02-03
Use _ for typing extensions

leewujung:


leewujung:
For whatever reason the `-` version now works, even though it failed when #1440 was first merged. This can therefore be closed now.

PR #1444, opened 2025-02-03
Resolve critical warnings from #1439

ctuguinay:
Resolves #1439

**TODO:** 

These tasks seem to be lagged in the CI test:

```
echopype/tests/calibrate/test_calibrate.py::test_compute_Sv_azfp 
echopype/tests/calibrate/test_calibrate.py::test_compute_Sv_returns_water_level 
Warning: divide by zero encountered in log10
Warning: divide by zero encountered in log10
```

Issue #1301, opened 2024-04-12
Add support for new AZFP ULS6 format


leewujung:
As the name suggested!

This involves 3 parts:
- [ ] add/update parser for ULS6 format
- [ ] determine storage location for data variables in the new format (plus reviewing the old one)
  - [ ] add code to actually set the SONAR-netCDF4 groups
- [ ] update calibration code if ULS6 involves any changes to the computation and data variables

leewujung:
An alpha version including the first 2 tasks were merged via #1323.

leewujung:
This is addressed by #1581 .

PR #1450, opened 2025-02-16
Fix deprecation warning for truth value of an empty array [all tests ci]

ctuguinay:
Fixes deprecation warning which arose in the azfp6 parser when checking valid parameters for temperature and tilt which was discovered in #1444:

> Warning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.

### DIFF
--- a/echopype/convert/parse_azfp6.py
+++ b/echopype/convert/parse_azfp6.py
@@ -308,9 +308,10 @@ class ParseAZFP6(ParseBase):
         def _test_valid_params(params):
             # If parameters are all empty return True.
             if all([len(self.parameters[p]) == 0 for p in params]):
-                return True
-            # The above was added because of the deprecation warning that arose due to
+                return False
+            # The above was added because of a deprecation warning that arose due to
             # the code below:
+            # https://github.com/OSOceanAcoustics/echopype/pull/1450
             # 'DeprecationWarning: The truth value of an empty array is ambiguous.
             # Returning False, but in future this will result in an error.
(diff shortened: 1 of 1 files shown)


ctuguinay:
@leewujung This one is ready for review

leewujung:
Since the one fixed is under `parse_azfp6` which I think came from `parse_azfp`, I took a look there too and found the same  `_test_valid_params` function there. This repetition is not supposed to be like this, but it is part of the unfinished #1301 and #1323. I'll push the same change as what you had for the azfp counterpart.

Also, I think the condition you tested for should output False? Since in this case, `self.parameters[[]` does not contain valid parameters? I'll push this change too and you could review.

leewujung:
Actually, I was wrong, the parameters were saved differently in azfp and azfp6 format.
In azfp, if some parameters are missing, the values are set to 0.
In azfp6, if some parameters are missing, the values simply do not exist.
So @ctuguinay you changes made sense.

The only thing I changed was to go from True to False.

leewujung:
Oops I messed this up. Will look again tomorrow.

leewujung:
Ok, turns out that this is a bug from #1323 that the parameter checking was somehow moved to _before_ the XML part of the file is parsed in `ParseAZFP6` (it was the correct order in `ParseAZFP`). With that fixed, the scenario that produces the warning should never happen.

Let me know what you think.

PR #1452, opened 2025-02-16
Compute dask array before np array equal [all tests ci]

ctuguinay:
Resolves:

> Warning: The `numpy.array_equal` function is not implemented by Dask array. You may want to use the da.map_blocks function or something similar to silence this warning. Your code may stop working in a future release.

which was discovered in #1444.


ctuguinay:
@leewujung This one should be ready for review

leewujung (approved):
Ah ok, makes sense. Thanks! Merging now.

PR #1453, opened 2025-02-16
Dataset.dims to Dataset.sizes [all tests ci]

ctuguinay:
This resolves a deprecation warning found in #1444:

> Warning: The return type of `Dataset.dims` will be changed to return a set of dimension names in future, in order to be more consistent with `DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.


ctuguinay:
@leewujung This one also should be ready for review

leewujung (approved):
@ctuguinay : Looks great, thanks! Merging now.

PR #1454, opened 2025-02-16
Chunks as dictionaries in `_get_auto_chunk` [all tests ci]

ctuguinay:
Resolves deprecation warning when chunks are passed in as tuples discovered in #1444:

> Warning: Supplying chunks as dimension-order tuples is deprecated. It will raise an error in the future. Instead use a dict with dimension names as keys.


leewujung (approved):
Looks great! I'll merge this now.

PR #1455, opened 2025-02-16
Replace previous arXiv citation with recent ICES paper citation

ctuguinay:


leewujung:
Ah great that you updated this! I only remembered the cff file.

Issue #1413, opened 2024-11-20
[Data conversion] ValueError when attempting to import raw EK80 data


tobitobitobi123:
### General description of problem

Really excited to use echopype more once I have this running - thanks everyone. I have some EK80 data stored as .raw files (you can find a link to example files further down). When I attempt to open them using ep.open_raw I'm getting a ValueError. It seems it can't collect some information about the channel. I know that some of the channels/transducers ran in CW mode, some in FM mode, in case that's the issue. For another dataset that has a similar mixture of CW and FM modes there seems to be no issue.

### Echopype version

echopype v0.9.0

### How did you install Echopype (e.g., conda or pip)

conda

### What is your operating system

Windows 10

### Minimal code example

```
import echopype as ep
file = "DY150-D20220505-T060019.raw"
ed = ep.open_raw(file, sonar_model='EK80')
```

### Error message printouts

```shell
Traceback (most recent call last):

  Cell In[5], line 1
    ed = ep.open_raw(file, sonar_model='EK80')

  File ~\anaconda3\envs\ek80\Lib\site-packages\echopype\utils\prov.py:237 in inner
    dataobj = func(*args, **kwargs)

  File ~\anaconda3\envs\ek80\Lib\site-packages\echopype\convert\api.py:536 in open_raw
    tree_dict["Vendor_specific"] = setgrouper.set_vendor()

  File ~\anaconda3\envs\ek80\Lib\site-packages\echopype\convert\set_groups_ek80.py:1367 in set_vendor
    ds = ds.pipe(self._add_filter_params, coeffs_and_decimation)

  File ~\anaconda3\envs\ek80\Lib\site-packages\xarray\core\common.py:775 in pipe
    return func(self, *args, **kwargs)

  File ~\anaconda3\envs\ek80\Lib\site-packages\echopype\convert\set_groups_ek80.py:1435 in _add_filter_params
    return dataset.assign(coeffs_xr_data)

  File ~\anaconda3\envs\ek80\Lib\site-packages\xarray\core\dataset.py:6160 in assign
    data.update(results)

  File ~\anaconda3\envs\ek80\Lib\site-packages\xarray\core\dataset.py:5026 in update
    merge_result = dataset_update_method(self, other)

  File ~\anaconda3\envs\ek80\Lib\site-packages\xarray\core\merge.py:1104 in dataset_update_method
    return merge_core(

[...]

Issue #1317, opened 2024-05-07
Error storing filter coefficients for sequenced ek80 file


julianblanco:
When trying to open ek80 data (2 channel system 70khz and 200khz) i get the following value error
ValueError: conflicting sizes for dimension 'channel': length 1 on 'WBT_filter_i' and length 2 on {'channel': 'frequency_nominal', 'pulse_length_bin': 'sa_correction'}

calling code
echodata = ep.convert.api.open_raw('/three_ensemble-Phase0-D20240506-T053349-0.raw', sonar_model='EK80')

version 0.8.4 mac, installed via pip in pycharm venv python ver 3.11

link to file: https://drive.google.com/file/d/1doLT8V3CFXNxkulIKtV1-BXMwvm0ln5N/view?usp=sharing

ctuguinay:
@leewujung This is interesting and I was able to get the same message. What are your thoughts on this? One of the channels is missing the filter values, or they're assumed to be shared for both?
These are the transducer names from the datagram: `['ES70-18CD', 'ES200-7CDK-Split']`

leewujung:
Hmm, the filters are specific to each channel, so could not be shared. I think I've seen files before that have a subset of filter values missing, and those parsed fine before. It is possible that some later changes we introduced made those not allowable, and we did not have tests covering that.

julianblanco:
Thanks for the quick attention. This is from a WBT mini system saving to a usb drive rather than collected via the ek80 software. 

Let me know if i can help in any way. I wrote something to filter out some of the stuff and stitch a echogram but id be interested in helping to fix the underlying issue and learn how to make contributions back upstream.

ctuguinay:
@julianblanco That would be great if you could help us out with this!

Here's the doc for how to contribute: https://echopype.readthedocs.io/en/stable/contributing.html
We're currently no longer using the `dev` branch so point any PR to `main` instead.

Here's the doc for how data is organized in the raw Echodata object: https://echopype.readthedocs.io/en/stable/data-format-raw.html. Scroll down to the EK80 part and take a look at `Vendor_specific`. A valid `Vendor_specfic` dataset should be one such that the number of individual `WBT_filter_n` arrays matches the number of individual channels/tranducers.

In the code itself, I would first look at the call to create the `Vendor_specific` dataset and work backward from there:
```
echopype/echopype/convert/api.py:487:
...
tree_dict["Vendor_specific"] = setgrouper.set_vendor()
```

[...]

PR #1458, opened 2025-02-17
Fill in NaN for missing EK80 coefficients [all tests ci]

leewujung:
This PR change EK80 `set_vendor` to fill in `np.nan` for missing filter coeffs across all channels.

Currently we have below in our test file:
- `D20210330-T123857.raw` that does not contain any filter coeffs
- `three_ensemble-Phase0-D20240506-T053349-0.raw` that only contains filter coeffs from 1 of the 2 channels, which seems to be due to ping sequencing. This file is from #1317. 
  - #1413 seems to be due to a similar reason.

This addresses #1317 and #1413.

### DIFF
--- a/echopype/calibrate/ek80_complex.py
+++ b/echopype/calibrate/ek80_complex.py
@@ -62,11 +62,11 @@ def filter_decimate_chirp(coeff_ch: Dict, y_ch: np.array, fs: float):
 
     # WBT filter and decimation
     ytx_wbt = signal.convolve(y_ch, coeff_ch["wbt_fil"])
-    ytx_wbt_deci = ytx_wbt[0 :: coeff_ch["wbt_decifac"]]
+    ytx_wbt_deci = ytx_wbt[0 :: int(coeff_ch["wbt_decifac"])]
 
     # PC filter and decimation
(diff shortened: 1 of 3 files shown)


ctuguinay:
Yeah, I tried one of the failed tests with an older scipy version and it passed, but after upgrading to the latest scipy that test failed.

ctuguinay:
So when trying to calibrate the three ensemble raw file I'm getting the following error:
image
I'ma grab some lunch and investigate this.

ctuguinay:
I pushed something small to cast numpy float arrays to integers when indexing.

ctuguinay (dismissed):
@leewujung I think what you have looks good. Let me know what you think of these small changes I made.

leewujung:
@ctuguinay : Thanks for the changes and the added test! Great that you caught this indexing problem. I had forgotten that having `nan` would force the array to be float.

The only thing I have is github warns that there's no `@pytest.mark.test` -- I can't find it anywhere either. Was this a typo?

ctuguinay:
Oh yes my bad that was for singling out a single test. Let me remove that and squash merge this. It was hidden at line 619: 
image

ctuguinay (approved):

Issue #1439, opened 2025-02-02
Resolve critical warnings revealed in tests


leewujung:
- [x] https://github.com/OSOceanAcoustics/echopype/actions/runs/13098048680/job/36542811253?pr=1429#step:15:139
   **PR:** #1451
  ```
  Warning: Exception ignored in: <function EchoData.__del__ at 0x7f32eba4d440>
  
  Traceback (most recent call last):
    File "/home/runner/work/echopype/echopype/echopype/echodata/echodata.py", line 105, in __del__
      self.cleanup_swap_files()
    File "/home/runner/work/echopype/echopype/echopype/echodata/echodata.py", line 94, in cleanup_swap_files
      fs = zarr_stores[0].fs
           ~~~~~~~~~~~^^^
  IndexError: list index out of range
  ```

- [x] https://github.com/OSOceanAcoustics/echopype/actions/runs/13098048680/job/36542811253?pr=1429#step:15:500
   **PR:** #1452 
  ```
  Warning: The `numpy.array_equal` function is not implemented by Dask array. You may want to use the da.map_blocks function or something similar to silence this warning. Your code may stop working in a future release.
  ```
  
- [x] https://github.com/OSOceanAcoustics/echopype/actions/runs/13098048680/job/36542811253?pr=1429#step:15:616
   **PR**: #1453
  ```
  Warning: The return type of `Dataset.dims` will be changed to return a set of dimension names in future, in order to be more consistent with `DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.
  ```
- [x] https://github.com/OSOceanAcoustics/echopype/actions/runs/13098048680/job/36542811253?pr=1429#step:15:956
   **PR**: #1454
  ```
  Warning: Supplying chunks as dimension-order tuples is deprecated. It will raise an error in the future. Instead use a dict with dimension names as keys.
  ```
- [x] https://github.com/OSOceanAcoustics/echopype/actions/runs/13098048680/job/36542811253?pr=1429#step:15:380
  ```
  Warning: 'S' is deprecated and will be removed in a future version, please use 's' instead.
  ```
  

- [x] https://github.com/OSOceanAcoustics/echopype/actions/runs/13098048680/job/36542811253?pr=1429#step:15:1020
   **PR**: #1450
  ```
  Warning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
  ```

[...]

PR #1462, opened 2025-02-17
Set decode timedelta to False since it will default to this in later xarray version [all tests ci]

ctuguinay:
Resolves https://github.com/OSOceanAcoustics/echopype/actions/runs/13098048680/job/36542811253?pr=1429#step:15:944 found in #1439:

> Warning: In a future version of xarray decode_timedelta will default to False rather than None. To silence this warning, set decode_timedelta to True, False, or a 'CFTimedeltaCoder' instance.


ctuguinay:
@leewujung Really short change, ready for review

leewujung:
I was curious what would happen if we set `decode_timedelta = True`. Turned out that the "MatlabTimeStamp" variable gets messed up:
Screenshot 2025-02-17 at 9 58 20 PM

leewujung (approved):
Looks great, thanks! Merging now.

PR #1463, opened 2025-02-18
Use normalization to compute echo range scaling

ctuguinay:
Enforces normalization of beam direction z if normalization of the beam direction vector is not already present. If vector is all zeros, return NaN.

### DIFF
--- a/echopype/consolidate/ek_depth_utils.py
+++ b/echopype/consolidate/ek_depth_utils.py
@@ -109,6 +109,4 @@ def ek_use_beam_angles(beam_ds: xr.Dataset) -> xr.DataArray:
 
     # For channels with near-zero norm, we return NaN. Otherwise, we return the normalized
     # z component.
-    normed_z = xr.where(norm < tolerance, np.nan, beam_direction_z / norm)
-    echo_range_scaling = normed_z
-    return echo_range_scaling
+return xr.where(norm < tolerance, np.nan, beam_direction_z / norm)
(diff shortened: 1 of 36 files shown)


ctuguinay:

echopype/consolidate/ek_depth_utils.py
```diff
@@ -74,36 +74,41 @@ def ek_use_platform_angles(platform_ds: xr.Dataset, ping_time_da: xr.DataArray)
...
+            "Beam direction vector was not normalized; applying normalization. "
+            "By definition, it should have been normalized."
+        )
+
+    # Warn if any channel has a (nearly) zero vector
+    if (norm < tolerance).any():
+        logger.warning("Some beam direction vectors are zero. Outputting NaN for those channels.")
+
+    # For channels with near-zero norm, we return NaN. Otherwise, we return the normalized
+    # z component.
```
leewujung, line 111: This statement is reversed, right?

echopype/consolidate/ek_depth_utils.py
```diff
@@ -74,36 +74,41 @@ def ek_use_platform_angles(platform_ds: xr.Dataset, ping_time_da: xr.DataArray)
...
+    normed_z = xr.where(norm < tolerance, np.nan, beam_direction_z / norm)
+    echo_range_scaling = normed_z
     return echo_range_scaling
```
leewujung, line 114: I guess this could be substituted with the below?
```suggestion
return xr.where(norm < tolerance, np.nan, beam_direction_z / norm)
```

leewujung (dismissed):
@ctuguinay : Two minor comments, but otherwise this looks great. Feel free to merge once those are addressed!

echopype/consolidate/ek_depth_utils.py
ctuguinay: I don't think so since `xr.where(norm < tolerance, np.nan, beam_direction_z / norm)` will return NaN if norm is small
image

leewujung: oh i know what I read wrong! I read "near-zero" as "non-zero" 😵
This is resolved then!

PR #1470, opened 2025-02-24
Enhance `compute_MVBS` feasability 

ctuguinay:
Allow `reindex` to be visible flox parameter in `compute_MVBS` and default to `reindex=False`, allowing computations to be more memory feasible.

Added `range_var_max` optional parameter which allows a user to specify range variable maximum, which in turn allows `compute_MVBS` to be lazily computed.

Tested the new `compute_MVBS` code on a large 40 GiB Sv dataset on a Dask Local Cluster with 2 workers each with 2 threads and 2 GiB RAM (4 GiB RAM total), shown on the following Gist: https://gist.github.com/ctuguinay/4cafda0c604fe6873b8f18c79cccdd1c

### DIFF
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -3,7 +3,7 @@ Functions for enhancing the spatial and temporal coherence of data.
 """
 
 import logging
-from typing import Literal
+from typing import Literal, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


ctuguinay:
@leewujung This should be ready for review now

echopype/commongrid/api.py
```diff
@@ -86,10 +103,15 @@ def compute_MVBS(
     if not isinstance(ping_time_bin, str):
         raise TypeError("ping_time_bin must be a string")
 
-    # create bin information for echo_range
-    # this computes the echo range max since there might NaNs in the data
-    echo_range_max = ds_Sv[range_var].max()
-    range_interval = np.arange(0, echo_range_max + range_bin, range_bin)
+    # Create bin information for the range variable
+    if range_var_max is None:
+        # This computes the range variable max since there might NaNs in the data
```
leewujung, line 108:
```suggestion
        # This computes the range variable max since there might be NaNs in the data
```

leewujung (approved):
@ctuguinay : Sorry it took me forever to look through this. Your finding with `reindex` is interesting. I think this is ready to be merged. I only have a tiny change on a comment.

I read the flox docs though not sure if I fully understood everything. Was what you found that with `reindex=True` we typically run out of memory for larger datasets on a regular-sized machine, and when the machine has larger memory, we can set `reindex=True` to increase the speed?

Issue #1377, opened 2024-08-22
`apply_mask` should only take in boolean masks


ctuguinay:
After talking to @leewujung, we should modify `apply_mask` to enforce that the masks passed in are boolean masks.

Users can currently pass in masks that are `1`/`0` and `1`/`NaN`, which stand for `True`/`False`; however, more obscure configurations can be passed in, such as `arbitrary float`/`NaN`, with the current way that `apply_mask` handles and applies masks onto Sv/MVBS data, and this makes it a bit confusing for what the user should pass in.

leewujung:
@spacetimeengineer - this is another one that's also more singular that I think could be addressed quickly.

spacetimeengineer:
@leewujung

When considering echodata masks, I understand the feature and the intent behind it. However, I wanted to ask about a potential edge case where, in the future, someone might use these masks for more complex operations.

Specifically, I’m thinking about scenarios where multiple mask objects exist within a given session. In such cases, a user may need to perform boolean operations, such as unions or intersections, on those masks. I’ve encountered this myself when working with both lines and circular masks—boolean operations became essential for handling them properly.

I agree that having only True and False values makes for a cleaner input/output system. However, I wonder if allowing additional temporary states could be useful for certain 2D operations on xarray or masks that otherwise wouldn’t be possible. The final result should always resolve to True or False, but intermediate states might facilitate calculations.

What are your thoughts on this? I will try to finish this by the end of the week.

leewujung:
@spacetimeengineer : We already allow users to put in a list of masks they would like to combine (element-wise AND) in `apply_mask`. For other possible combinations, I think it'll be better to leave it to the users to go through their own processing to come up with clean masks before using this function.

spacetimeengineer:
Acknowledged. Thankyou.

leewujung:
@shivam096 : I added you to this issue too along with other `apply_mask` issues. @spacetimeengineer has previously put together #1476 but seems like tests are failing. If you could take a look the tests there that'll be super helpful. Thanks!

shivam096:
Hi @leewujung,

I have updated the PR with my comments. Also, I would like to check if the users are expected to provide a boolean input mask?  Or can the input mask be of any type?

leewujung:
Resolved by #1502.

PR #1476, opened 2025-03-10
Attempt at fixing #1377

spacetimeengineer:
Attempt at fixing #1377.

### DIFF
--- a/echopype/mask/api.py
+++ b/echopype/mask/api.py
@@ -136,6 +136,11 @@ def _validate_and_collect_mask_input(
                     "('channel', 'ping_time', 'depth')"
                     "('channel', 'ping_time', 'echo_range')"
                 )
+            # **Enforce boolean type on mask values**
+            if not np.issubdtype(mask[mask_ind].dtype, np.bool_):
+                raise TypeError(
+                    f"Mask at index {mask_ind} must be boolean (dtype=bool), but got dtype {mask[mask_ind].dtype}"
+                )
 
         # Check for the channel dimension consistency
         channel_dim_shapes = set()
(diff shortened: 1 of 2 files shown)


leewujung:
@spacetimeengineer : Many of the original tests in `test_mask` are failing because of the changes you added. Could you fix them? If you don't have time, let me know.

cc @jmjech

shivam096:
Hi @leewujung and @spacetimeengineer ,

After reviewing the implementation, it seems that the main issue is the newly added Boolean check, which fails with a TypeError triggered when non-boolean inputs are passed through it.

The issue code -
```
 # **Enforce boolean type on mask values**
            if not np.issubdtype(mask[mask_ind].dtype, np.bool_):
                raise TypeError(
                    f"Mask at index {mask_ind} must be boolean (dtype=bool), but got dtype {mask[mask_ind].dtype}"
)
```
This fails even before the asserts are hit, causing the test to fail. A possible solution would be to update the inputs or use the check in a different way.

leewujung:
@shivam096 : thanks for looking into this. Could you push up a PR to fix this? A separate PR on the upstream (OSOceanAcoustics) repo would be fine, if you branch out from this one, then @spacetimeengineer’s commits will be already included there. Thanks!

shivam096:
Sure @leewujung ,

I will go ahead and update the code. Just wanted to check if the input has to be a boolean array? Or the users can send in any type which related to a boolean representation, and then we do the processing at the backend to make it work ?

leewujung:
@shivam096: I think we should restrict the input to be a boolean array, and explicitly error people out if they send in something else. This way there's no ambiguity on the user side what would happen when they send in various numbers (since 0 and 1 are clearly False and True, but other numbers are less clear). Other numbers do occur in our own processing since we sometimes store masks with a dimension being the scatterer type (like 0-background, 1-large fish, 2-small fish).

shivam096:
@leewujung I have started working on this

leewujung:
Closing as this is superseded by #1502.

PR #1484, opened 2025-04-03
Replace `utcnow()` with timezone-aware object

leewujung:
This PR addresses the warning
`datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).`
 that occurs in various places in our codebase.

### DIFF
--- a/echopype/consolidate/api.py
+++ b/echopype/consolidate/api.py
@@ -224,7 +224,7 @@ def add_depth(
     used_platform_angles = use_platform_angles and not tilt
     used_beam_angles = use_beam_angles and not tilt
     history_attr = (
-        f"{datetime.datetime.utcnow()} +00:00. depth` calculated using:"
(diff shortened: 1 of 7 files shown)


leewujung:
Ugh @ctuguinay sorry this is not ready for review yet. I'll ping again once it is.

leewujung:
Turns out that `datetime.UTC` was [only added in python=3.11](https://docs.python.org/3/library/datetime.html#datetime.UTC) so the 3.10 test is failing!

Onward with a version handler.

leewujung:
@shivam096 : I haven’t had a chance to get back to this, but perhaps this would be a standalone issue that you can jump on easily? I haven’t looked at the latest failing report, but if the tests that are failing are actually not related to `utcnow`, let me know and we can see whether to address it in another PR or this one.

leewujung:
This is superseded by #1497. Closing now.

PR #1492, opened 2025-04-20
Added support for split-beam angle computation with mixed channel beam types in complex mode

beatfactor:
This PR adds support for computing split-beam angles in datasets where only a subset of channels have valid `beam_type` values (e.g., `beam_type = 1, 17, 49, 65, 81`). It applies specifically to CW complex data.

Previously, the presence of an unsupported `beam_type` would raise a `ValueError`.

**Changes:**
- updated `get_angle_complex_samples()` to process each channel individually if `beam_type` is not uniform;
- skips channels with unsupported `beam_type` values and logs a warning instead of failing.

### DIFF
--- a/echopype/consolidate/split_beam_angle.py
+++ b/echopype/consolidate/split_beam_angle.py
@@ -8,7 +8,26 @@ from typing import List, Tuple
 import numpy as np
 import xarray as xr
 
+from ..utils.log import _init_logger
 from ..calibrate.ek80_complex import compress_pulse, get_norm_fac, get_transmit_signal
+
+
+logger = _init_logger(__name__)
+
+# Beam type identifiers
+BEAM_TYPE_SPLIT_4_SECTOR = 1      # 4-sector split-beam (common Simrad type)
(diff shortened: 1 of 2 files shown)


beatfactor:
I couldn't figure out how to download the test data used in the existing tests, so the test file used in the unit test from this PR is a best guess.

leewujung:
@beatfactor: the first command in [this section](https://echopype.readthedocs.io/en/latest/contrib_setup.html#running-the-tests) will pull all the test files down, and you can stop the other services afterwards.

beatfactor:
hey @leewujung, any updates here? could you re-trigger the build please?

leewujung:
@beatfactor : Thanks for the reminder - I just re-triggered the tests and looks like everything runs fine. 

I am pretty swamped for the past weeks and hopefully will be able to look at the code changes later next week.

beatfactor:
> @beatfactor : somehow the pre-commit.ci pushes were skipped. Could you please add the suggested changes into the code?

Ok, done. Yes, it seems that the build won't push the changes if the PR is sent from a Github organization, which it is in this case. Thanks!

PR #1497, opened 2025-05-09
Fix utcnow

shivam096:
Update datetime format


leewujung (approved):
@shivam096 : Thanks for finishing up this PR! 

Looks like the tests all passed once xarray is pinned to <2025.4.0, so I'll merge this now. :) 

The xarray version problem with python 3.10 os weird, but there's an inconsistency in how timeX is encoded in the platform group. I'll write a quick issue so that we can don't forgot about it.

leewujung:
This superseded #1484 .

Issue #1498, opened 2025-05-12
Timestamps encoding inconsistency in the `Platform` group


leewujung:
While working on #1497, I found this weird phenomenon that time encoding was inconsistency in the `Platform` group like below:

![Image](https://github.com/user-attachments/assets/2db894ce-699c-46f9-b446-3566c856ac92)

`time1` has weird numbers while `time2` and `time3` are nicely formed timestamps.

The test was `test_add_location_lat_lon_0_NaN_warnings` using `echopype/test_data/ek80/D20170912-T234910.raw`.

This test failed with python=3.10 and xarray=2025.4.0, but the tests passed for python=3.11-12. Pinning xarray<2025.4.0 in solved the problem.

This might be some leftover from fixing nanosecond int64 encoding while upgrading to pandas 2.0, but I am not certain.

leewujung:
@shivam096 : could you take a look? since you're now familiar with the related tests...  Thanks!

PR #1499, opened 2025-05-14
Add time decode [all tests ci]

shivam096:
Add fix for issue - https://github.com/OSOceanAcoustics/echopype/issues/1498

### DIFF
--- a/echopype/convert/set_groups_base.py
+++ b/echopype/convert/set_groups_base.py
@@ -400,10 +400,10 @@ class SetGroupsBase(abc.ABC):
             },
         )
         timestamp_array = xr.coding.times.decode_cf_datetime(
-                timestamp_array,
-                units=DEFAULT_TIME_ENCODING["units"],
-                calendar=DEFAULT_TIME_ENCODING["calendar"],
-            )
+            timestamp_array,
+            units=DEFAULT_TIME_ENCODING["units"],
+            calendar=DEFAULT_TIME_ENCODING["calendar"],
+        )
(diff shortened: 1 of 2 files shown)


leewujung:
Hey @shivam096 : Thanks for the PR. I checked the tests and found that the first file did have proper encoding but `time4` in the third file seen below has weird time encoding again.

How about you 1) fix it in an ad-hoc way like what you did for `time1` for the first test file (so then we can merge this PR), and 2) write an issue for checking all time encoding in the not very distant future? Part of the reason is that I want to spend some time thinking about the naming of the time coordinates, and it seems something we can do in one go (i.e. find a good way to ensure proper time encoding and then rename the `time*` coordinates).

Screenshot 2025-05-15 at 4 48 28 PM

leewujung:
@shivam096 : FYI - I just added `[all tests ci]` at the end of the PR title, and Close/Reopen the PR. This would trigger the tests to run through the entire test suite (otherwise only the tests that are "touched" by the code changes in this PR would be run -- but sometimes there would be downstream). I usually do this when the code changes may have downstream consequences.

leewujung (dismissed):
@shivam096 : Looks good! Thanks for finding the `time4` entry. I'll remove the xarray pinning and see if everything runs though, and merge the PR.

Could you write an issue for needing an overall timestamp encoding/decoding handling (see my previous comment)?

leewujung:
Looks like this resolves that xarray problem! I'll merge this now.

PR #1502, opened 2025-05-28
Update `apply_mask` to only take in boolean or 0/1 masks

shivam096:
Changes:

echopype/mask/api.py:
Added strict validation to ensure masks do not contain NaN or non-boolean values. Now, a TypeError is raised if the mask contains any NaNs or values other than 0, 1, True, or False.

echopype/tests/mask/test_mask.py:
Updated the test for masks with NaN values to expect a TypeError, reflecting the new validation logic that rejects non-boolean masks.


leewujung:
This fixes #1377 and supersedes #1476.

echopype/mask/api.py
```diff
@@ -138,6 +138,18 @@ def _validate_and_collect_mask_input(
                     "('channel', 'ping_time', 'echo_range')"
                 )
 
+            # **Validate boolean-like values - reject NaN**
+            # First check if there are any NaN values
+            if np.any(np.isnan(mask[mask_ind].values)):
+                raise TypeError("Mask must be boolean")
```
leewujung, line 144:
```suggestion
                raise TypeError("Mask cannot contain NaN")
```

echopype/mask/api.py
```diff
@@ -161,6 +173,18 @@ def _validate_and_collect_mask_input(
...
+        if np.any(np.isnan(mask.values)):
+            raise TypeError("Mask must be boolean")
+
+        # Get unique values and check if they are boolean-like
+        unique_vals = np.unique(mask.values)
+
+        # Check if all values are boolean-like (0, 1, True, False)
+        if not np.all(np.isin(unique_vals, [0, 1, True, False])):
+            raise TypeError("Mask must be boolean")
+
```
leewujung, line 187: Not specific to the changes you made, but this entire `else` block seems completely redundant. Can you absorb this into the `if isinstance(mask, list)` section by simply making a single input mask a list and just use the code for list?

echopype/mask/api.py
```diff
@@ -312,7 +336,8 @@ def apply_mask(
     source_ds: xr.Dataset, str, or pathlib.Path
         Points to a Dataset that contains the variable the mask should be applied to
     mask: xr.DataArray, str, pathlib.Path, or a list of these datatypes
-        The mask(s) to be applied.
+    The mask(s) to be applied. Must contain only boolean values (True/False).
```
leewujung, line 339:
```suggestion
    The mask(s) to be applied. Must contain only boolean values (True/False or 1/0).
```

leewujung (changes requested):
@shivam096 : thanks for the PR! I only had some minor edits for error messages, but made a request to remove redundant code in `_validate_and_collect_mask_input`. It is an easy change so even though it is different from the core intention of this PR, I think it's better to just get it done here.

echopype/mask/api.py
```diff
@@ -138,6 +138,18 @@ def _validate_and_collect_mask_input(
                     "('channel', 'ping_time', 'echo_range')"
                 )
 
+            # **Validate boolean-like values - reject NaN**
```
leewujung, line 141: @shivam096 could you also remove all the ** in the comments? thanks.

Issue #1507, opened 2025-06-13
Allow processing CW/FM interleaving data


leewujung:
Currently all raw complex data from CW or FM transmissions are stored in the same `Beam_group` under `Sonar`. This is fine, but after discussing with @ctuguinay , we think in the case with interleaving CW/FM transmissions, it is more convenient on the processing side to have these data separately stored in different `Beam_group`s. Therefore, we will update the `open_raw` function to store FM and CW data separately in different `Beam_group`s.

We might need to think about backward compatibility -- on the other hand, because of the issue in #1506, I don't think anyone was able to use echopype to correctly process interleaving CW/FM data (the processing would run through, but the results are incorrect since the filter coefficients are not the right match to the pgins). So we might not have to worry about backward compatibility issues...

leewujung:
This is addressed in #1508 - remaining issues documented in #1653.

Issue #1506, opened 2025-06-13
Store ping-by-ping filter coefficients


leewujung:
In the current code only one set of filter coefficients are stored. This is problematic since in cases like FM transmission with different frequency combinations, or interleaving CW/FM transmissions, the coefficients can change on a ping-by-ping basis. We should correct this as part of being able to handle processing ping sequence patterning data.

leewujung:
This is addressed by #1508.

PR #1508, opened 2025-06-13
Enable processing CW/FM interleaving data [all tests ci]

leewujung:
This addresses #1506 and #1507 in one go, since they are tightly linked.

@ctuguinay : I think I've only gotten to store ping-by-ping filter coefficients (FIL1 datagram), but it's been a while so my memory is pretty faded. Please check. Thanks a lot for picking this up!


ctuguinay:

echopype/convert/set_groups_ek80.py
```diff
@@ -1187,7 +1210,19 @@ def set_beam(self) -> List[xr.Dataset]:
...
+            "LFM" in ds_beam["transmit_type"] or "FMD" in ds_beam["transmit_type"]
+        ):
+            # Create BB and CW masks and select beam subsets from mask coordinates
+            mask_BB = ds_beam["transmit_type"].isin(["LFM", "FMD"])
+            mask_BB = mask_BB.where(mask_BB, drop=True)
+            mask_CW = ds_beam["transmit_type"] == "CW"
+            mask_CW = mask_CW.where(mask_CW, drop=True)
+            ds_beam_BB = ds_beam.sel(channel=mask_BB["channel"], ping_time=mask_BB["ping_time"])
+            ds_beam_CW = ds_beam.sel(channel=mask_CW["channel"], ping_time=mask_CW["ping_time"])
+            return [ds_beam_BB, ds_beam_CW, ds_beam_power]
```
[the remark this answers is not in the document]
ctuguinay: Yeah I think I know what you mean. I'll try to do the splitting of the complex data into FM and CW right after we call `_assemble_ds_complex` and I can follow that up with `_add_freq_start_end_ds` on the FM data, that way I don't have to do the assignment that you talked about in the previous comment.

echopype/convert/set_groups_ek80.py
```diff
@@ -743,17 +752,25 @@ def _add_freq_start_end_ds(self, ds_tmp: xr.Dataset, ch: str) -> xr.Dataset:
...
-            freq_stop = freq_start
-        else:
-            return ds_tmp
+        # pulse_form: 0 = CW, 1 = FM, 5 = FMD
+        # cw_idx = np.array(self.parser_obj.ping_data_dict["pulse_form"][ch]) == 0
+        fm_idx = np.array(self.parser_obj.ping_data_dict["pulse_form"][ch]) == 1
+        freq_start = np.ones(self.parser_obj.ping_time[ch].size) * np.nan
+        freq_stop = np.ones(self.parser_obj.ping_time[ch].size) * np.nan
+        freq_start[fm_idx] = np.array(self.parser_obj.ping_data_dict["frequency_start"][ch])
+        freq_stop[fm_idx] = np.array(self.parser_obj.ping_data_dict["frequency_end"][ch])
```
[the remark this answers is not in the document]
ctuguinay: Yeah I can avoid this assignment by following what I said in this comment: https://github.com/OSOceanAcoustics/echopype/pull/1508#discussion_r2844253001

echopype/convert/parse_base.py
```diff
@@ -501,6 +497,10 @@ def _read_datagrams(self, fid):
                 new_datagram["timestamp"].replace(tzinfo=None), "[ns]"
             )
 
+            # Remove \00t from channel_id
+            if "channel_id" in new_datagram:
+                new_datagram["channel_id"] = new_datagram["channel_id"].replace("\00t", "")
```
leewujung: @ctuguinay: just curious from what sources of files you saw these trailing \00t

echopype/echodata/echodata.py
```diff
@@ -728,3 +730,29 @@ def chunk(self, chunk_dict: dict):
                     self[echodata_group] = group.chunk(subset_chunks)
 
         return self
+
+    def copy(self, deep: bool = True):
```
leewujung: @ctuguinay : I only see this used in `test_multiple_filter_times_calibration` -- do we still need it? Not sure if I missed something.

echopype/echodata/echodata.py
```diff
@@ -104,7 +106,7 @@ def cleanup_swap_files(self):
     def __del__(self):
         # TODO: this destructor seems to not work in Jupyter Lab if restart or
         #  even clear all outputs is used. It will work if you explicitly delete the object
-        if self.converted_raw_path is None:
+        if self.converted_raw_path is None and self.original:
```
leewujung: @ctuguinay : is `self.original` still needed if we remove `echodata.copy()` (if that can be removed)?

ctuguinay: No this is also a remnant. I can remove this.

echopype/echodata/echodata.py
ctuguinay: No this is also a remnant. I will remove this.

echopype/convert/parse_base.py
ctuguinay: It's in this new file that I recorded at APL that I placed in: `echopype/test_data/ek80_bb_complex_multiplex/CW_FM_power_complex_repeating.raw`

I'm not quite sure why this is happening, but the `channel_id` looks like this: `EKA 293104-07 ES38-18DK-Split\x00t`. This weird appending is not the case for any of the other combinations (i.e, 70 single, 120 split, 200 single).

Also I made a modification to the string replacement. It should be "\x00t" and not "\00t".

leewujung: sounds good - let's keep it since it doesn't impact other parts. I'll add a comment to say we've only seen it once.

leewujung:
@ctuguinay : I have made a number of changes to clean a few things up - the main things are:
- the logic surrounding calibrate using using a single or multiple different `filter_time` in `calibration/api.py::_compute_cal()`
  - there are many remaining TODOs that I don't think I can get to in the next few days, so I think we can go ahead to merge this (because the code works) to get a release ready, and then debug/fix in a later release 
- simplify `_retrieve_correct_beam_group_EK80` to return only 1 beam group name
  - I think `_retrieve_correct_beam_group_EK60` and `_retrieve_correct_beam_group_EK80` can be merged to leave just one function, but am also running out of time to do that -- will fix in a later release

I also have a question about this code block: I couldn't immediately figure out why this block would not work when there is only 1 `filter_time` in the Vendor-specific group, but it fails `test_ek80_BB_complex_multiplex_NaNs_and_non_NaNs` for both files:
https://github.com/leewujung/echopype/blob/62aeaf19100ed39a07ca4e2e4b7c049b0ef54531/echopype/calibrate/api.py#L127-L164
For the first file "NYOS2105-D20210525-T213648.raw" , instead of producing Sv with 4 channels, it only produces data for 2 channels. I think something is not quite right for the stack operation, but I couldn't figure out why `dropna("pairs")` drops those channels. Maybe you know?

Issue #1263, opened 2024-02-25
Upgrade handling of test data files to use GitHub Release Assets


leewujung:
Right now test data files are stored in a private google drive folder and copied to the virtual machine when running tests. This is somewhat brittle and makes this project less open to contributions. 

@ocefpaf suggested the option to use GitHub Release Assets to store test data files, which keeps the data in the same repo but does not contaminate the code, with the additional advantage that the test data can then be versioned. Since none of our test data files >2G so will not cause problem with the assets, this should work pretty well.

@lsetiawan has given this a try here [here](https://github.com/lsetiawan/echopype-test-data/releases), so seems indeed we can go this route to handle our test data files.

dextrot:
Hi @leewujung 
I want to work on this issue, can you please assign it to me?

HQ786:
Hi @leewujung
I am interested to work on this issue, can you please assign this to me?

oftfrfbf:
@leewujung  just pinging this issue to say that I am currently working on addressing the testing resource limitations. I have my development environment set up and am currently researching other projects use of the _pooch_ utility mentioned to see if it will work here.

> @ocefpaf suggested the option to use GitHub Release Assets to store test data files

With respect to the "github release assets." Is the expected workflow to create a consolidated release of test_data files embedded within the echopype codebase as git LFS objects? I am not entirely familiar with the process yet, the best I can interpret is that releases can contain LFS objects as mentioned here:

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-git-lfs-objects-in-archives-of-your-repository

So when tagging a release, such as "v0.9.1," this could conceivably include all the 2.95 GB of needed test_data assets as LFS objects.

Anyway, I will be working on a prototype to see if I can consolidate files so we don't need google-drive and docker integration to serve test files.

ocefpaf:
@oftfrfbf here is an example of what I do: https://github.com/pyoceans/pocean-core/releases

You can see two releases, [one for data](https://github.com/pyoceans/pocean-core/releases/tag/2024.04) and one for the code. The data release uses CalVer to avoid clashing with the library that uses SemVer. All the data files are added as assets to the release and fetched/cached at test time [using pooch](https://github.com/pyoceans/pocean-core/blob/main/pocean/tests/download_test_data.py). Let me know if you have any questions.

Note that, while I prefer to keep them both in the same repository, one can have a repo for data and another one for the code.

[...]

PR #1511, opened 2025-07-09
[1485] Moved the check for duplicate times to after NMEA selection

### LINKED ISSUE #1485: Select nmea sentence type first before checking duplicates
Currently in `consolidate.add_location`, `check_loc_time_dim_duplicates` is called before the subsetting based on NMEA sentence type happens. However, often the duplications are from different NMEA sentence types, so we should reverse the order of these operations.

dbashford-NOAA:
Resolves #1485 

Moved the check for duplicate times to after selecting the NMEA subset when adding location to sv dataset.
Also updated the tests for add_location()

### DIFF
--- a/echopype/tests/consolidate/test_add_location.py
+++ b/echopype/tests/consolidate/test_add_location.py
@@ -1,5 +1,5 @@
 import math
-
+import os
 import pytest
 import numpy as np
 import xarray as xr
@@ -7,6 +7,7 @@ import xarray as xr
 import echopype as ep
 from echopype.consolidate.loc_utils import sel_nmea
 
+TEST_DATA_DIR = "../../test_data/"
 
 @pytest.mark.unit
 def test_sel_nmea_value_error():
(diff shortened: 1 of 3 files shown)


echopype/tests/consolidate/test_add_location.py
```diff
@@ -361,7 +360,7 @@ def test_add_location_lat_lon_missing_all_NaN_errors(
             ]
         ),
         (
-            "echopype/test_data/ek80/RL2407_ADCP-D20240709-T150437.raw",
+            os.path.abspath(TEST_DATA_FOLDER / "ek80/RL2407_ADCP-D20240709-T150437.raw"),
```
leewujung, line 363: You could keep the paths as `Path` objects in all cases?

leewujung (approved):
Hey @dbashford-NOAA : Looks good! I made some changes to use fixture for the test data folder, just so that things are more consistent - you can take a look.

Hopefully we can get to using mechanisms like github release assets for the test data (#1263) since copying data from the google folder takes the longest time in running the tests. If you have other suggestions let me know!

PR #1517, opened 2025-07-23
[1486] Support computing MVBS with frequency_nominal and fix some tests [all tests ci]

### LINKED ISSUE #1486: Construct MVBS dataset using either `channel` or `frequency_nominal`
Right now the `compute_MVBS` output is hard-coded to align with `("channel", "ping_time", range_var)`, but it should also be able to produce `("frequency_nominal", "ping_time", range_var)`.  Probably useful to allow it to be align with other coordinates too, as long as they are also the same dimension as `channel` and `frequency_nominal`.

dbashford-NOAA:
Resolves #1486 

- Generalized instances of `"channel"` string to support `frequency_nominal` or others in dim 0.
- Added test that includes swapping dims.
- Fixed broken tests.

### DIFF
--- a/echopype/commongrid/api.py
+++ b/echopype/commongrid/api.py
@@ -137,13 +137,16 @@ def compute_MVBS(
         **flox_kwargs,
     )
 
+    # Generalize the first dimension name to support multiple like channel and frequency_nominal
+    dim_0 = list(raw_MVBS.dims.keys())[0]
+
     # create MVBS dataset
     # by transforming the binned dimensions to regular coords
     ds_MVBS = xr.Dataset(
-        data_vars={"Sv": (["channel", "ping_time", range_var], raw_MVBS["Sv"].data)},
+        data_vars={"Sv": ([dim_0, "ping_time", range_var], raw_MVBS["Sv"].data)},
         coords={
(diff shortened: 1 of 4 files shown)


dbashford-NOAA:
@leewujung it looks like the tests are failing due to test data that's not supported. Does that need to be fixed for this PR?

leewujung:
@dbashford-NOAA - hmm this is interesting. Could you try to create a new environment using conda/mamba and see which version they pulled there? Pip often has this type of problems. I think that if conda can solve this combination of dependencies, we can merge this, and just wait for the packages to catch up on their dependencies...

dbashford-NOAA:
@leewujung I created a new environment and it's pulling moto 5.1.8, s3fs 2025.7.0, boto3 1.39.8, botocore 1.39.8, and aiobotocore 2.23.2. These tests pass in that dev environment but I don't think they even use these packages. Let me know if there is something else I should test.

leewujung:
@dbashford-NOAA : I think it's likely that we can ignore the failure. I'll trigger this to run again and see how things look.

Could you update the use of `.dims` so that it is compatible with the next changes in xarray? You can check the detailed message under the Files changed tab.
Screenshot 2025-07-28 at 1 52 18 PM

Feel free to ping me to review once it's ready.

PR #1520, opened 2025-07-30
[1488] Support frequency nominal dim in consolidate functions

### LINKED ISSUE #1488: Support dimension alignment with `frequency_nominal` for all `ep.consolidate` functions
`add_splitbeam_angle` should allow alignment with `frequency_nominal` since in the workflow people may call `swap_dims_channel_frequency` before doing this, then everything will be aligned with `frequency_nominal` instead `channel`.

Need to make sure this is supported for all `ep.consolidate` functions.

dbashford-NOAA:
Resolves #1488 

- Use a generalized `dim_0` so frequency nominal can be supported in place of channel
- Added testing for the swapped dims
- Update tests to use fixture for the test data folder


echopype/consolidate/api.py
```diff
@@ -455,12 +456,14 @@ def add_splitbeam_angle(
...
+    dim_0 = list(source_Sv.sizes.keys())[0]
 
-    # Select ds_beam channels from source_Sv
-    ds_beam = echodata[ed_beam_group].sel(channel=source_Sv["channel"].values)
+    if dim_0 in ["channel", "frequency_nominal"]:
+        ds_beam = echodata[ed_beam_group].sel({dim_0: source_Sv[dim_0].values})
+    else:
+        raise ValueError(
+            "The input source_Sv Dataset must have a channel or frequency_nominal dimension!"
+        )
```
leewujung, line 466: In this implementation `dim_0` can be anything that is in the first position (index 0) in the `ds` dataset keys. In theory`ds` can come from different sources, so there is no guarantee that the first key would be `channel` or `frequency_nominal`. Since you check `if dim_0 in ["channel", "frequency_nominal"]` downstream anyways, how about checking if the dimension contains one of these two options before proceeding?

Also, the current implementation
```
            if echodata["Sonar/Beam_group1"][dim_0].equals(ds[dim_0]):
                beam_group_name = "Beam_group1"
            else:
                beam_group_name = "Beam_group2"
```
is a weird setup in terms of Beam_groupX selection. This is a larger issue so no immediate actions required under this PR, but I will note this in a separate issue.

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -286,6 +286,147 @@ def test_add_splitbeam_angle(sonar_model, test_path_key, raw_file_name, test_pat
...
+        "ek60_CW_power",
+        "ek60_CW_power_Sv_path",
+        "ek80_CW_complex",
+        "ek80_BB_complex_no_pc",
+        "ek80_CW_power",
+    ],
+)
+def test_add_splitbeam_angle_w_dim_swap(sonar_model, test_path_key, raw_file_name, test_path,
+                             paths_to_echoview_mat, waveform_mode, encode_mode,
+                             pulse_compression, to_disk):
```
leewujung, line 357: This new test is largely identical with the existing `test_add_splitbeam_angle`, and the content of that test was to make sure the split beam angle computation is correct. I think it's a bit meddling the water to repeat the same thing here, since dimension swap is not part of the split beam calculate.

Could you slim this down to just testing the dimension swap?

leewujung (changes requested):
@dbashford-NOAA : Thanks for the PR. Sorry it took a while for me to get to this.

I made a few suggestions that I think will improve the robustness and the tests.

Could you also add a test for `add_location` to make sure things flow through even with dimension swap? Thanks.

Please request my review once you're done with the above.

echopype/tests/consolidate/test_consolidate_integration.py
dbashford-NOAA: There is already a test for the `swap_dims_channel_frequency` and I didn't make any changes to that function but do you think it needs more coverage? This test was meant to cover the changes I made to 'test_add_splitbeam_angle' and make sure the split beam angle computation was correct when the dims are swapped. Maybe it would be better to just add a dim swap test case to the existing `test_add_splitbeam_angle` test?

leewujung: adding one test to go through the `swap_dims_channel_frequency` step and then `add_splitbeam_angle` would be great. I think making it a separate test from `test_add_splitbeam_angle` which tests the function outputs would be a good idea.

echopype/tests/consolidate/test_add_location.py
```diff
@@ -184,6 +184,75 @@ def _tests(ds_test, location_type, nmea_sentence=None):
...
+        "pressure": extras["pressure"],
+    }
+
+    ds = ep.calibrate.compute_Sv(echodata=ed, env_params=env_params)
+
+    ds = ep.consolidate.swap_dims_channel_frequency(ds)
+    for group in ed["Sonar"]['beam_group'].values:
+        ed[f"Sonar/{group}"] = ep.consolidate.swap_dims_channel_frequency(ed[f"Sonar/{group}"])
+
+    ds_all = ep.consolidate.add_location(ds=ds, echodata=ed)
```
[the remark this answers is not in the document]
leewujung: (Somehow I missed your response...)

Yes I think all `ep.consolidate` functions should work that way. Basically the functions should work with either `channel` or `frequency_nominal` as a coordinate.

echopype/consolidate/api.py
```diff
@@ -205,11 +205,36 @@ def add_depth(
...
-            else:
-                beam_group_name = "Beam_group2"
+            # Check beam groups to find which one contains the matching dimension
+            dim_0 = list(ds.sizes.keys())[0]
+            beam_group_name = None
+            for idx in range(echodata["Sonar"].sizes["beam_group"]):
+                if dim_0 in list(echodata[f"Sonar/Beam_group{idx + 1}"].sizes):
+                    if echodata[f"Sonar/Beam_group{idx + 1}"][dim_0].equals(ds[dim_0]):
+                        beam_group_name = f"Beam_group{idx + 1}"
+                        break
```
leewujung: The block would fail if the number of channels/transducers in the 2 beamgroups are the same.

I think a cleaner approach is to simply check if `dim_0` is `"channel"` or `frequency_nominal`, since this is _within_ echopype and the first dimension should be one of these 2, or the function should error out. Later in line 225-226 you explicitly do an if-else for these 2 allowed dimension names anyway, so no need to broadening this up to other possible dimension names.

echopype/consolidate/api.py
```diff
@@ -456,12 +480,22 @@ def add_splitbeam_angle(
...
+            dim_0 = dim
+            break
 
-    # Select ds_beam channels from source_Sv
-    ds_beam = echodata[ed_beam_group].sel(channel=source_Sv["channel"].values)
+    if dim_0:
+        if dim_0 not in list(echodata[ed_beam_group].sizes) and "channel" in list(
+            echodata[ed_beam_group].sizes
+        ):
+            echodata[ed_beam_group] = swap_dims_channel_frequency(echodata[ed_beam_group])
```
leewujung: We don't want to swap "frequency_nominal" with "channel" -- we want to _preserve_ whatever the dimension the input ds has and just add the depth variable to the dataset.

echopype/tests/consolidate/test_consolidate_integration.py
```diff
@@ -286,6 +286,63 @@ def test_add_splitbeam_angle(sonar_model, test_path_key, raw_file_name, test_pat
...
+    ds_Sv = ep.calibrate.compute_Sv(ed, waveform_mode=waveform_mode, encode_mode=encode_mode)
+
+    ds_Sv = ep.consolidate.swap_dims_channel_frequency(ds_Sv)
+
+    ds_Sv = ep.consolidate.add_splitbeam_angle(source_Sv=ds_Sv, echodata=ed,
+                                               waveform_mode=waveform_mode,
+                                               encode_mode=encode_mode,
+                                               to_disk=False)
+
+    # Check that channel dim has been swapped to frequency_nominal
```
leewujung: make sure to change this too since we want to _preserve_ whatever the input ds has

leewujung (changes requested):
@dbashford-NOAA : Sorry it again took a while for me to get to this!

I think there may have been some misunderstandings on what #1488 aims to achieve, but hopefully my comments below and inline would clear it up.

The goal of #1488 is to enable users to use _all_ functions in `ep.consolidate` on Sv dataset regardless of if they have called `swap_dims_channel_frequency` or not.

With this in mind, I have a couple suggestions for this PR:
- for all `ep.consolidate` functions, check if either "channel" or "frequency_nominal" exists in the input `ds`, and error out if not. 
  - I'd suggest making a small utility function for this and returns `dim_0`, so that the same routine can be used across all functions. This would help with future maintenance since everything is in one place.
- make sure all `ep.consolidate` functions have a test that checks if the function would run through on Sv dataset _both before and after_ calling `swap_dims_channel_frequency` -- this embodies the goal of #1488

Also, please pull in all upstream changes to make sure your tests will run with the right assets. We've changed the CI setup pretty significantly in the past month or so. Those changes shouldn't impact the components you're working on, but just in case since I saw some directory changes in this PR.

dbashford-NOAA:
@leewujung 
Do you know if there is any test data I could use for `add_depth` and `add_splitbeam_angle` that has the `frequency_nominal` dim instead of `channel`?
I think part of the issue is that I was trying to adapt the existing test data to test with a `frequency_nominal` dim.

leewujung:
@dbashford-NOAA : ah I see. Echopype generated echodata object and its derivative like the Sv dataset would always have channel as a dimension initially. It’s only when users use swap_dims_channel_frequency will they get that dimension changed to frequency_nominal. We made that swap function because most people refer to different channels by their central frequency - like “the 120 kHz channel.” 

What we can’t control is when users would do the swap, and therefore we want to get all consolidate functions to be able to work with both the before and after swap data. 

So, you can get that by using the swap function.

dbashford-NOAA:
@leewujung 
I think I've addressed everything you mentioned. I left the dim swap in the test code just as a way to adapt the data to test when dim_0 is frequency_nominal. Let me know if there's anything else that needs to change.

ctuguinay:
Superseded by #1734

PR #1521, opened 2025-08-02
Implementations of basic seafloor detection in mask subpackage [all tests ci]

LOCEANlloydizard:
I have tested two possible ways to integrate the seafloor detection into the current workflow. Both functions are currently implemented in the mask subpackage (the second one is temporarily located in the seafloor_detection subpackage within mask).

An example notebook demonstrating and testing the feature will be submitted soon to the echopype-examples repository for review.


leewujung (commented):
@LOCEANlloydizard : It's great to see the implementation! I have gone through the structure of your code (both implementations) and the notebook. Below are my suggestions.

What I think would work well now (and easy to change afterwards if we decide to move the various detection algorithms for seafloor and swarm and single targets somewhere else), is to have a function similar to your `compute_bottom()` in `mask/api.py`, which will act as an intermediary to detect the bottom line (depth for each ping).

I would suggest a function signature as below:
```python
def detect_bottom(
  ds: xr.Dataset,  # Sv dataset
  method: str,  # string to indicate the detection method
  params: Dict,  # dictionary holding parameters for the selected detection method
) -> xr.DataArray:
```
This function will call functions sitting under `mask/seafloor_detection` to perform the operation.

The output of this function would just be a simple `xr.DataArray` object, _without_ channel as a dimension, and has 1 depth value along the `ping_time` dimension.

This DataArray can then be loaded into the Echoregions `Lines` object (see [here](https://echoregions.readthedocs.io/en/latest/Lines_functionality.html#creating-echogram-mask-with-bottom)) to make a mask. The reasons why I think reusing the Echoregions function would be a good idea are:
- That function has smoothing and interpolation capability built in, and can therefore handle missing pings. For example, if the 18 kHz channel is used to detect the bottom, but it does not ping as frequently as the other channels, then there's a mismatch in the `ping_time` dimension given the way Echopype stores Sv data (that missing pings are stored as NaN).
- It allows separation of functionalities, so that we can go fancy on the exact algorithms used for creating masks (this is probably more important for regions rather than lines), without affecting what Echopype contains.

Additionally I have the following suggestions for the code:
- I think using `*` in function input arguments can lead to hard-to-read code and increase probability of user error. Using `**` is much better. In the above function signature I suggested using a dictionary to pass in parameters for different methods, because it allows one to put together a quick check to see if all required params exist before proceeding on the computation
- I don't see a clear need for setting up class scaffolding for the algorithms you have in `seafloor_detection` since there's no inheritance and the number of parameters needed for each algorithm is not too large. For example, this section below is pretty awkward:
  Screenshot 2025-08-02 at 8 58 19 PM
  and would be a lot simpler without the `self.` part. I'd suggest changing these to just be flat functions that calls private functions (which are your class methods)

Lastly, I haven't tried to run the notebook so I am not sure why the apply_mask function is slow. It shouldn't be, since we've used that function extensively in our own workflow. I wonder if it has something to do with whether data are on disk or in memory, and the chunk size if it's a dask array. We can resolve this later though. I suggest getting the functions all in place first before trying to optimize the performance, since that latter is often non-trivial.

Right now the `Llines` object [takes EVL or CSV as inputs](https://github.com/OSOceanAcoustics/echoregions/blob/c2b5b70bd8bd9f37cfae2385ed8b646febc6abbf/echoregions/lines/lines.py#L32-L38). I think we can add reading a dataframe (with `ping_time` as index and a column `depth`) or the DataArray output from above to initiate the `Lines` object.

cc @ctuguinay to keep him in the loop since he's the main Echoregions maintainer.

LOCEANlloydizard:
Thanks for the review @leewujung!

I’ve refactored the code following your feedback: removed the class scaffolding, added a detect_bottom() dispatcher, and ensured the output is a 1D DataArray. Let me know what you think (I’ll commit shortly)!

If this structure looks good to you, I’ll move on to add the Blackwell method next, and shoal detection should integrate easily using the same approach.

@ctuguinay I’ve also prepared a notebook that uses this basic bottom detector, builds a mask with echoregions from this detection, and applies it using apply_mask() from the echopype.mask subpackage. I’ll commit the notebook to the echopype-examples repo shortly!

Cheers!

leewujung:
@LOCEANlloydizard : The changes look great - thanks!

Do you want to add the Blackwell algorithm to this PR too? Seems good to just have them together?

echopype/mask/seafloor_detection/utils.py
```diff
@@ -0,0 +1,105 @@
...
+        variable = np.array([variable])
+
+    if isinstance(variable, int):
+        variable = np.float64(variable)
+
+    mask = np.ma.masked_less_equal(variable, 0).mask
+    variable[mask] = np.nan
+    log = 10 * np.log10(variable)
+    log[mask] = -999
+    return log
```
leewujung, line 105: Can use `echopype/utils/compute.py::_log2lin / _lin2log` for these?

Also, is -999 really the "convention"? I know that's what echoview uses, but wouldn't NaN be better?

echopype/mask/api.py
```diff
@@ -678,23 +677,85 @@ def detect_seafloor(
...
+    -------------------------
+    basic
+        var_name : str
+            Name of Sv variable (dB), e.g. ``"Sv"``.
+        channel : str
+            Channel identifier to select (e.g., ``"GPT  38 kHz ..."``).
+        threshold : float, default -50.0
+            Sv threshold in dB (lower bound). If the implementation also
+            supports an upper bound, a scalar is interpreted as a lower bound
+            only.
```
leewujung, line 710: Seeing what's under the blackwell method, I grabbed what you have in bottom_basic and modified it slightly below. Please check if it is correct.
```suggestion
        threshold : float, default -50.0
            Sv threshold(s) in dB. If a single float is given, it is treated as the
            lower bound and the upper bound is set to be 10 dB above the 
            lower bound. If a 2-tuple `(tmin, tmax)` is provided, both the lower
            and upper bounds are used directly.
```

leewujung (approved):
@LOCEANlloydizard I think we are good to go! I made some minor suggestion for the docstrings. Feel free to change what I modified. I'll merge once you give me a go ahead!

leewujung (approved):

Issue #1423, opened 2024-12-29
Change default PR-triggered test behavior


leewujung:
Right now the PR-triggered tests are determined by the existence of the string `[all tests ci]` in the PR title. If this doesn't exist, only tests associated the module that was modified are run. This has caused quite a bit of problems since some code changes would break downstream cascading processing but would not be caught in this way. 

I am thinking of changing the default behavior to:
- running all tests whenever there is a code change
- if there are only docs changes, only the docs build are run

Or that we can make it such that:
- a label like "run CI" will trigger all test runs
- a label like "build docs" will trigger docs build
- otherwise nothing will be done

@oftfrfbf @ctuguinay : Any thoughts or suggestions? 

Also, there was a label "Needs Complete Testing" that did not have any effect and was probably a relic from long time ago. I have removed it.

oftfrfbf:
Sorry for the slow response, I have been thinking about this and some other things.

I have not triggered any doc builds yet so I am unfamiliar with the process.

I think that if you are struggling to keep your github-action-hours under some limit per month then it is worth while to keep the fail-fast and targeted testing structure in place. If you are not constrained by the test computing resources, then keeping everything as simple as possible is my recommendation and in that case just run the full test suite each time. Less lines of code in the testing infrastructure means more reliable results. Developers should in most cases be approaching pull requests with already tested code, by allowing them to only test small segments of the code we will instead be allowing them to introduce some very difficult to debug problems.

Running the full test suite each time also allows us to enforce a contract where if you break something you need to fix it.

...

These are some other unsolicited test recommendations as well:

Some of the test resources are pretty large, especially the ek60 and ek80 resources. It might be worth finding replacements for test files that have the same required and unique test attributes but are just smaller file sizes. I can can scan the S3 _noaa-wcsd-pds_ bucket to find files if needed, but there should be a goal to keep the test suite small so that developers can more eaily run the full suite as needed. If the full test suite could run in <5 minutes then it would feels more accessible and useful.

Here is my latest scan of asset sizes by directory:

https://github.com/oftfrfbf/echopype-test-data/releases/tag/2024.12.23.10.10

...

[...]

PR #1524, opened 2025-08-08
Update pr.yaml to run full test suite for all PR [all tests ci]

LOCEANlloydizard:
This PR updates the pr.yaml GitHub Actions workflow to always run the full test suite for all pull requests, regardless of the PR title, by removing the conditional logic that previously limited full test runs to PRs titled with [all tests ci].

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -70,13 +70,8 @@ jobs:
       - name: Print Changed files
         run: echo "${{ steps.files.outputs.added_modified_renamed }}"
       - name: Running all Tests
-        if: contains(github.event.pull_request.title, '[all tests ci]')
         run: |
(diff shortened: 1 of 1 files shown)


LOCEANlloydizard:
Made this proposition following our discussion @leewujung, and I guess that if we refer to @oftfrfbf’s opinion in #1423 ("If you are not constrained by the test computing resources, then keeping everything as simple as possible is my recommendation and in that case just run the full test suite each time."), then it should be ok like this for now?

leewujung (approved):
@LOCEANlloydizard : Thanks for the PR! I'll merge this now. I think this means that the PR test behavior will change. Let's see how that goes. The main time sink I think has been to download data from google drive, so once we move to release assets it should be much more straightforward.

PR #1530, opened 2025-08-13
Add regrid mask functionality

### LINKED ISSUE #1516: Add functionality for mask regridding
Similar to what was done in the Hake Labels repo, add Dask code in the mask subdirectory to regrid masks. This will replace the current ad hoc mask regrid operation that uses compute_MVBS and a bit of data wrangling.

ctuguinay:
This resolves #1516.

The `ep.mask.regrid_mask` function takes in a `xr.DataArray` and outputs a `xr.DataArray`. Regridding is done using flox `xarray_reduce` and has two aggregation options: `logical-OR` and strict `logical-AND`. These aggregations are not native to flox, so implement them, I turned 0/False into NaN and used flox built-in `nanmean` and `mean` aggregations.


ctuguinay:
TODO: Try looking at using 1/0 with mean and if output strictly 1 and 0 for logical-AND then become 0 and if logical-OR then become 1.

ctuguinay:
**TODO:** Refactor the tests to work with 2D range variable implementation.

ctuguinay:
Jupyter-book recently got an update and our currently implementation seems to be no longer compatible with it so I pinned it in requirements.

ctuguinay:
TODO figure out why logical-OR produces 1's near NaNs. 

Edit: Fixed this by modifying `xarray_reduce` `fill_value` for empty bins. The `fill_value` was originally NaN, but since the logical-OR logic used `xr.where(mask_regridded_da != 0.0, 1, 0)`, this would turn NaNs into 1s instead of 0. So instead of NaN, I used `0.0` for the `fill_value`.

ctuguinay:
@leewujung This PR is ready for review

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+    ping_time_bin : str, default '20s'
+        bin size along 'ping_time'
+    third_dim : str, default None
+        Name for non-binned region. If None, use 'xarray_reduce' to operate on 2D array. Else,
+        use 'xarray_reduce' to operate on 3D array with unbinned 3rd dimension.
+    func: {'logical-AND', 'logical-OR'}, default 'logical-AND'
+        When func == 'logical-AND', the output value is set to 1/True only if all corresponding
+        values in the regrid bin are 1/True.
+        When func == 'logical-OR', the output value is set to 1/True if any value in the regrid
+        bin is 1/True.
```
leewujung:
```suggestion
    func: {'logical-AND', 'logical-OR'}, default 'logical-AND'
        When func == 'logical-AND', the output value is set to 1 or ``True`` only if 
        all values in the regrid bin are 1 or ``True``.
        When func == 'logical-OR', the output value is set to 1 or ``True`` if 
        any value in the regrid bin is 1 or ``True``.
```

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+
+    if third_dim is not None and len(mask_da.dims) != 3:
+        raise ValueError("Mask must have only 3 dimensions.")
+
+    # Check if mask is binary (True/False or 1/0)
+    if not np.isin(mask_da.data, [1, 0]).all():
+        raise ValueError("Mask must be binary True/False or 1/0.")
+
+    # Check aggregation function
+    if func != "logical-AND" and func != "logical-OR":
```
leewujung:
```suggestion
    if func not in ["logical-AND", "logical-OR"]:
```

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+        If False, reindex after the blockwise stage. If True, reindex at the blockwise stage.
+        Generally, `reindex=False` results in less memory at the cost of computation speed.
+        Can only be used when method='map-reduce'.
+        See flox `documentation <https://flox.readthedocs.io/en/latest/implementation.html>`_
+        for more details.
+    closed: {'left', 'right'}, default 'left'
+        Which side of bin interval is closed.
+    range_var_max: str, default None
+        Maximum value of range variable. If this value is known, the user should input it as
+        it will skip the maximum computation and this keeps this function lazy.
```
leewujung:
```suggestion
    range_var_max: str, default None
        Maximum value of the range variable. If this value is known, it is recommended to provide it 
        as an input as it makes the underlying computation more efficient 
        (by skipping a `max` computation and keeping the data lazily-loaded).
```

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+            coords={
+                "ping_time": np.array([v.left for v in mask_regridded_da.ping_time_bins.values]),
+                range_da.name: np.array(
+                    [v.left for v in mask_regridded_da[f"{range_da.name}_bins"].values]
+                ),
+            },
+        )
+
+    # For logical-AND, if output is not strictly 1, then the value becomes 0.
+    # For logical-OR, if output is not strictly 0, then the value becomes 1.
```
leewujung: I removed "strictly" since it can make it confusing
```suggestion
    # For logical-AND, if output is not 1, then the value becomes 0.
    # For logical-OR, if output is not 0, then the value becomes 1.
```

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+        "n": {"nptd64": "ns", "unitstr": "nanosecond"},
+        "ns": {"nptd64": "ms", "unitstr": "millisecond"},
+    }
+    ping_time_bin_td = pd.Timedelta(ping_time_bin)
+    # res = resolution (most granular time unit)
+    ping_time_bin_resunit = ping_time_bin_td.resolution_string.lower()
+    ping_time_bin_resvalue = int(
+        ping_time_bin_td / np.timedelta64(1, timedelta_units[ping_time_bin_resunit]["nptd64"])
+    )
+    ping_time_bin_resunit_label = timedelta_units[ping_time_bin_resunit]["unitstr"]
```
leewujung: This section overlaps with what's in commongrid.compute_MVBS.
How about making this an util function so that if we need to update code it's all in one place?

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+        raise ValueError(f"Passing in reindex={reindex} is only allowed when method='map_reduce'.")
+
+    if not isinstance(ping_time_bin, str):
+        raise TypeError("ping_time_bin must be a string")
+
+    if third_dim is None and len(mask_da.dims) != 2:
+        raise ValueError("Mask must have only 2 dimensions unless 'third_dim' is specified.")
+
+    if third_dim is not None and third_dim not in mask_da.dims:
+        raise ValueError(f"Mask must contain '{third_dim}' as dimension.")
```
leewujung:
```suggestion
    if third_dim is not None and third_dim not in mask_da.dims:
        raise ValueError(f"Mask must contain the specified '{third_dim}' as a dimension.")
```

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+
+    if third_dim is not None and len(mask_da.dims) != 3:
+        raise ValueError("Mask must have only 3 dimensions.")
```
leewujung:
```suggestion
    if third_dim is not None and len(mask_da.dims) != 3:
        raise ValueError("Mask must have 3 dimensions when 'third_dim' is specified.")
```

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+    closed: Literal["left", "right"] = "left",
+    range_var_max: str = None,
+    **flox_kwargs,
+) -> xr.DataArray:
+    """
+    Regrid mask through downsampling via a logical AND operation.
+
+    Parameters
+    ----------
+    mask : xr.DataArray
```
leewujung:
```suggestion
    mask_da : xr.DataArray
```

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+    method: str, default 'map-reduce'
+        The flox strategy for reduction of dask arrays only.
+        See flox `documentation <https://flox.readthedocs.io/en/latest/implementation.html>`_
+        for more details.
+    reindex: bool, default False
+        If False, reindex after the blockwise stage. If True, reindex at the blockwise stage.
+        Generally, `reindex=False` results in less memory at the cost of computation speed.
+        Can only be used when method='map-reduce'.
+        See flox `documentation <https://flox.readthedocs.io/en/latest/implementation.html>`_
+        for more details.
```
leewujung, line 720:
```suggestion
    reindex: bool, default False
        If False, reindex after the blockwise stage. If True, reindex at the blockwise stage.
        Generally, `reindex=False` results in less memory usage at the cost of computation speed.
        Can only be used when method='map-reduce'.
        See flox `documentation <https://flox.readthedocs.io/en/latest/implementation.html>`_
        for more details.
```

echopype/mask/api.py
```diff
@@ -668,6 +671,225 @@ def _create_mask(lhs: np.ndarray, diff: float) -> np.ndarray:
...
+            data=mask_regridded_da.data,
+            dims=[third_dim, "ping_time", range_da.name],
+            coords={
+                third_dim: mask_da[third_dim].values,
+                "ping_time": np.array([v.left for v in mask_regridded_da.ping_time_bins.values]),
+                range_da.name: np.array(
+                    [v.left for v in mask_regridded_da[f"{range_da.name}_bins"].values]
+                ),
+            },
+        )
```
leewujung: Here and also in the else case below, a new DataArray is constructed by taking the data in `mask_regridded_da` with new coordinates. I was wondering if it'll be more efficient to just rename the coordinates and update the values. But, looking at the I am not sure in practice if the underlying implementation - but it looks like a new object is created when rename is called anyway, so maybe it doesn't matter. Any thoughts here?

leewujung (changes requested):
Hey @ctuguinay : Thanks for the PR! I think it's about ready to go - my inline comments are mostly just wording changes. 

One thing I think that may be useful is to separate out the tests from `test_mask.py` to something more descriptive, now that we have so many tests in that module. We should split out the tests for seafloor and swarm detection, but maybe we do that when we know where the functions themselves would go (@LOCEANlloydizard and I were thinking they should probably go somewhere else instead of sitting in `echopype.mask`.

echopype/mask/api.py
ctuguinay: Yeah I agree that it's cleaner to just rename and update values. I'll make that change.

echopype/mask/api.py
ctuguinay: Good point, just made this change.

leewujung (dismissed):

leewujung:
Thanks for the changes @ctuguinay! I'll merge this now.

leewujung (approved):

Issue #1512, opened 2025-07-12
Unpin numpy<2


leewujung:
Currently we have numpy pinned to <2 in the requirements file. It's a bit outdated, so we should upgrade!

leewujung:
Reopening this since there are issues related to netcdf4 that's not resolved.

PR #1531, opened 2025-08-13
[1428/1512] Migrate to numpy v2 and zarr v3 with updated test_data path format [all tests ci]

### LINKED ISSUE #1428: Echopype not compatible with zarr 3.0.0 !?

dbashford-NOAA:
Resolves #1428 and #1512 

- Migration changes for numpy v2 and zarr v3
- Updated tests to use standard test_data path


leewujung:
@dbashford-NOAA : I am finally able to look at this.

Could you fix the pre-commit error?

Separately, a month later, it seems that the error that occurred has changes (which may be that something got upgraded properly). Python 3.10 fails with `ImportError: cannot import name 'BloscCodec' from 'zarr.codecs' (/opt/hostedtoolcache/Python/3.10.18/x64/lib/python3.10/site-packages/zarr/codecs.py)` -- could you look into that?

Am still waiting for Python 3.11-12 tests to run through, but I suspect they would be fine.

If it's only Python 3.10 that has the issue, it would be interesting to see why it doesn't show up in 3.11-12. Once we know, we can decide if to discontinue supporting 3.10 or to put in a fix there.

echopype/convert/set_groups_ek60.py
```diff
@@ -146,7 +146,7 @@ def set_env(self) -> xr.Dataset:
             ds_env.append(ds_tmp)
 
         # Merge data from all channels
-        ds = xr.merge(ds_env)
+        ds = xr.merge(ds_env, compat="no_conflicts", join="outer")
```
leewujung, line 149: it seems that `compat="no_conflicts"` and `join="outer"` are the default values at least in the recent xarray versions. What did you run into that motivated you to add them here explicitly?

leewujung (changes requested):
@dbashford-NOAA : I merged the latest `main` changes necessary for xarray compatibility and added a few inline comments in the code. 

Could you point me to documentation on how zarr v3 handles consolidated metadata? I saw the following changes and am wondering how things are handled now.:
- your comments on that flag is removed in `to_zarr`
- `check_consolidated` is no longer used (though still exists)
- changes in `echopype/tests/echodata/utils.py`

dbashford-NOAA:
> @dbashford-NOAA : I merged the latest `main` changes necessary for xarray compatibility and added a few inline comments in the code.
> 
> Could you point me to documentation on how zarr v3 handles consolidated metadata? I saw the following changes and am wondering how things are handled now.:
> 
> * your comments on that flag is removed in `to_zarr`
> * `check_consolidated` is no longer used (though still exists)
> * changes in `echopype/tests/echodata/utils.py`
> 
> [Update: I also just pushed a change to force zarr>=3 in the requirements to not have to worry about backward compatibility issues.]

I can't find documentation on how the file structure is changed for local stores I just noticed the change in my testing and found info on how the consolidated metadata changed in v3. These links have some info on consolidated metadata.
- [`xarray.Dataset.to_zarr`](https://docs.xarray.dev/en/latest/generated/xarray.Dataset.to_zarr.html) has the consolidated arg
- [Consolidated Metadata](https://zarr.readthedocs.io/en/stable/user-guide/consolidated_metadata.html#user-guide-consolidated-metadata) in the zarr user guide

PR #1543, opened 2025-09-25
Add weill to codespell ignore list

LOCEANlloydizard:
Extend .pre-commit-config.yaml to ignore the word "weill", which is used as a method in the shoal detection module. If not, codespell incorrectly flags and auto-corrects it to "will".


leewujung (approved):
Thanks for catching this!

PR #1544, opened 2025-09-25
Add Fielding and Matecho-like transient noise removal methods

### LINKED ISSUE #1352: Add Fielding transient noise mask
The transient noise (TN) filter from [Ryan et al. 2015](https://doi.org/10.1093/icesjms/fsv121) is slow due to the need to sort in calculating median for each window surrounding a given pixel.

Alejandro Ariza (@alejandro-ariza) and Sophie Fielding in British Antarctic Survey (BAS) has figured a new approach that works pretty well. It is implemented [in echopy here](https://github.com/open-ocean-sounding/echopy/blob/96bb25f83490529a5373aeb3b423f03c9605f7a6/echopy/processing/mask_transient.py#L87) and documented in the docstrings. It would be good to add this functionality.

LOCEANlloydizard:
This commit introduces two new transient noise removal methods:
* Fielding-based filter
* Matecho-style deep transient spike filter
* Add tests for both

For now, both methods are implemented in a new subpackage transient_noise. The existing Ryan method remains in api.py. In a follow-up commit, we can either (a) move the Ryan method into the same subpackage, or (b) re-export these new methods from api.py for consistency.

The tests are also stored in their own test_transient_noise.py for now. They can easily be move to the test_noise.py, or vice-versa.

Closes #1352

### DIFF
--- a/echopype/tests/clean/test_transient_noise.py
+++ b/echopype/tests/clean/test_transient_noise.py
@@ -30,7 +30,7 @@ def ds_small_chunked(ds_small, request):
 @pytest.mark.unit
 def test_dispatcher_rejects_unsupported_method(ds_small):
     with pytest.raises(ValueError, match="Unsupported transient noise removal method"):
-        ep.clean.mask_transient_noise_dispatch(
+        ep.clean.detect_transient(
             ds_small, method="not_a_method", params={}
         )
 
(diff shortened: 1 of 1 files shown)


leewujung (commented):
@LOCEANlloydizard : Thanks for the PR, it looks great! I only have some minor function naming suggestions below. Once those are done we can go ahead to merge this. I'll add an issue in echopype example repo to keep track of the need to make a notebook to demo this (might be better to make a separate notebook so that the demo could use a more suitable dataset?). I think you may have already done this and just haven't pushed yet? 

[Update] I added https://github.com/OSOceanAcoustics/echopype-examples/issues/69 and it appears that we don't have any existing notebook actually demoing transient noise removal. I think it was because we didn't have a suitable dataset, and the Ryan method was slow. These new methods would be much better for a demo notebook!

I think it's cool that you're already using the `apply_ufunc` for this, so the transition to using coordinate aware computation should be very straightforward via the same xarray function.

For the function names, what do you think about the following? Just for consistency.
- `mask_transient_noise_fielding` -> `transient_noise_fielding`
- `mask_transient_noise_matecho` -> `transient_noise_matecho`
- `_fielding_core_numpy`: no change
- `mask_transient_noise_matecho_core` -> `_matecho_core_numpy`

Issue #1549, opened 2025-09-30
Adapt to latest xarray update v2025.9.1


LOCEANlloydizard:
Xarray’s latest release introduced a couple of changes that affect echopype:

- Dataset.update() now returns None (in-place only). So code that did ds = ds.update({...}) breaks.

- Default netCDF engine preference changed to h5netcdf/scipy instead of netcdf4. Grouped netCDF writes in our code rely on netcdf4.

I'll put in a small PR right now, but more details here if needed [https://docs.xarray.dev/en/stable/whats-new.html](https://docs.xarray.dev/en/stable/whats-new.html)

PR #1550, opened 2025-09-30
Modifications to match xarray update to v2025.9.1

LOCEANlloydizard:
This PR updates two lines to be compatible with the latest xarray release v2025.9.1.

- Dataset.update() now returns None (in-place only).
- The default netCDF engine preference changed to h5netcdf/scipy, which breaks grouped writes unless we explicitly use netcdf4.

Maybe there are other changes to make too?
Mentions: #1549

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -10,7 +10,8 @@ jobs:
   test:
     name: ${{ matrix.python-version }}--${{ matrix.runs-on }}
     runs-on: ${{ matrix.runs-on }}
-    if: ${{ !contains(github.event.pull_request.title, '[skip ci]') }}
+    # safe on workflow_dispatch (no PR object)
+    if: ${{ !contains(github.event.pull_request.title || '', '[skip ci]') }}
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
@@ -18,8 +19,8 @@ jobs:
         python-version: ["3.11", "3.12"]
         runs-on: [ubuntu-latest]
         experimental: [false]
+
     services:
-      # TODO: figure out how to update tag when there's a new one
       minio:
         image: cormorack/minioci:latest
         ports:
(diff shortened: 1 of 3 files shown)


leewujung (commented):
Hey @LOCEANlloydizard : Thanks for looking into the CI failures and pointing to the sources in xarray v2025.9.1 updates!

For the changes re `.update`, the now deleted comments seemed to hint that before if `platform_var` already exists, it needs to be dropped first before a new one with the same name can be added. I don't remember that being the case (ie I thought it's fine to just overwrite an existing variable). Could you confirm simple overwrite is fine (ie the code changes you just made would always go through)?

For the changes in `echopype/utils/io.py` where you force the engine to `netcdf4`, what do you think to just change this mapping below to not force the netcdf4 engine? then it'll use the current ones that usually have better performance.
https://github.com/OSOceanAcoustics/echopype/blob/0dffd519e930a370f6b9bd111d31d737e08cfee1/echopype/echodata/echodata.py#L27-L30

The xarray updates are also very inspiring. I haven't read those for a while. I suspect that we can pretty much remove the entire EchoData object component handling in Echopype and just use native `xr.DataTree` for that. We created it back then when `DataTree` didn't exist. This would be a large change, so obviously out of scope here, but just wanted to mention it. I'll write an issue as a reminder.  I suspect most of it would be pretty straightforward, but we'll need to be careful on the merging component (the current `ep.combine_echodata`.

LOCEANlloydizard:
Hi @leewujung, thanks for pointing out these points!

I switched back to using .update(), so basically we just removed the reassignment. I also restored the comment noting that with .update() there’s no need to drop the variable first, since overwrite is fine.

For utils/io.py, I went with engine=engine rather than pulling from XARRAY_ENGINE_MAP. The reason is that io.py already validates against its own SUPPORTED_ENGINES dict ({"netcdf4": {"ext": ".nc"}, "zarr": {"ext": ".zarr"}}), so importing the other map could desynchronise things? For example, if echodata set ".nc" -> "h5netcdf", io.py would reject it as unsupported?

Maybe we should unify both, but that might be better tracked in #1551?
Cheers!

leewujung (approved):
Sound good for keeping `.update`.

For `engine=engine`: actually that's much nicer. I think we should get rid of `XARRAY_ENGINE_MAP` to avoid confusion, and agree with you that #1551 is probably a better place to take care of this. We can do with netcdf4 for now.

I'l merge this now!

PR #1553, opened 2025-10-02
[1542] Support python 3.13

### LINKED ISSUE #1542: Add support for Python 3.13
After #1541 that discontinues support for Python 3.10, let's bring on support for Python 3.13 so that we can always 3 supported versions.

This is a priority after #1541 and #1531 are merged.

dbashford-NOAA:
Resolves #1542

### DIFF
--- a/.ci_helpers/py3.13.yaml
+++ b/.ci_helpers/py3.13.yaml
@@ -1,0 +1,8 @@
+name: echopype
+channels:
+  - conda-forge
+dependencies:
+  - python=3.13
+  - pip
+  - pip:
+    - -r ../requirements.txt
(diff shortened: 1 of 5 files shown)


dbashford-NOAA:
@leewujung It looks like `netCDF4-1.6.5` isn't compatible with python 3.13. Do you know why netCDF4 has a max version set in requirements? Could we use a later version?

leewujung:
You can check the requirements for what we pinned it too. How about you investigate why we pinned it at that version, and see what could be done to resolve it? 

This is related to #1550.

dbashford-NOAA:
@leewujung It looks like it was pinned due to a wheel problem that has been resolved. I removed the max version pin and it installed fine with all tests passing!

PR #1554, opened 2025-10-06
Update docstrings for transient noise and shoal detection functions

LOCEANlloydizard:
Update docstrings for transient noise and shoal detection functions, and add credit to Echopy for the original implementations.

### DIFF
--- a/echopype/clean/transient_noise/transient_fielding.py
+++ b/echopype/clean/transient_noise/transient_fielding.py
@@ -119,11 +119,29 @@ def transient_noise_fielding(
     start: int = 0,
 ) -> xr.DataArray:
     """
-    Build a Fielding-style (modified from Echopy) transient-noise mask from an xarray Dataset.
+    Transient noise detector modified from the "fielding"
+    function in `mask_transient.py`, originally written by
+    Alejandro ARIZA for the Echopy library (C) 2020.
 
-    This wrapper extracts a 1-D vertical coordinate, then applies a NumPy core
-    over the last two dims using `apply_ufunc` (vectorized across leading dims,
-    e.g., channel). The result is returned as a boolean mask aligned to `Sv`.
+    Overview
+    -------------------
+    This algorithm identifies deep transient noise in echosounder
+    data by comparing the echo level of each ping with its local
+    temporal neighbourhood in a deep water window.
+    It operates in linear Sv space and uses a two-stage decision:
+
+    1. Deep window test – In a specified depth interval (e.g., 900–1000 m),
+    compute the ping median and the median over neighbouring pings.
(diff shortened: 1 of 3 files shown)


echopype/clean/transient_noise/transient_matecho.py
```diff
@@ -123,13 +123,29 @@ def transient_noise_matecho(
...
-    (computed in linear units) exceeds a local percentile + delta_db within a
-    deep window.
+    Overview
+    --------
+    Flags entire pings as transient when, within a deep window, the ping’s
+    mean Sv (computed in linear units, then converted back to dB) exceeds a
+    local reference percentile by `delta_db`.
+
+    1) Deep window: Use a vertical slice from `start_depth` to
+       `start_depth + window_meter`, limited by a local bottom (if provided;
```
leewujung, line 135: with the `start_depth` to `start_depth + window_meter` specification, this does not have to be "deep" per se, right? How about changing it to just "Depth window"?

echopype/clean/transient_noise/transient_matecho.py
```diff
@@ -123,13 +123,29 @@ def transient_noise_matecho(
...
+       otherwise r[-1]). Skip if usable height < `min_window`.
+    2) Local reference: For ping j, form a temporal neighborhood
+       [j - window_ping/2, j + window_ping/2] and compute the chosen `percentile`
+       (in dB) over that neighborhood within the deep window.
```
leewujung, line 139: since log is linear, it doesn't matter whether it is in dB or linear, right?

echopype/clean/transient_noise/transient_fielding.py
```diff
@@ -119,11 +119,29 @@ def transient_noise_fielding(
     start: int = 0,
 ) -> xr.DataArray:
     """
-    Build a Fielding-style (modified from Echopy) transient-noise mask from an xarray Dataset.
-
-    This wrapper extracts a 1-D vertical coordinate, then applies a NumPy core
-    over the last two dims using `apply_ufunc` (vectorized across leading dims,
-    e.g., channel). The result is returned as a boolean mask aligned to `Sv`.
```
leewujung, line 126: Perhaps add these back like in the Matecho case?

leewujung (changes requested):
@LOCEANlloydizard : Thanks for the PR! I made some wording change suggestions via inline comments. The main things I'd suggest are: 1) making things more consistent across algorithms, and 2) be clear what the function arguments are when describing the algorithms.

If you could do another pass on this I can merge afterwards.

On a related note, I am thinking we could do a sweep for RTD rendering check at some point, since sometimes docstrings don't necessary render the way we thought it would, but it's kinda messy to have that dragging the PRs. Perhaps we do that at the same time we do the github release asset changes?

PR #1556, opened 2025-10-07
fix(ci): free disk space to resolve "No space left on device"

LOCEANlloydizard:
Added cleanup steps (free-disk-space and docker prune) in docker GitHub Actions workflow to prevent System.IO.IOException: "No space left on device" errors during multi-arch builds.


leewujung (approved):
Oh awesome! Let's give it a try!

PR #1581, opened 2025-12-02
AZFP ULS6 beta version, ULS5 fixes

dash-uvic:
ULS5 Bug fixes:
* fixed recursive bug for corrupted files
* fixed RepeatedField not applying correctly
ULS6 updates:
* Added AZFP Nano (single channel) compatibility
* Added compatibility for beta version of ULS6 standard
* Refractored code to inherit ULS5 as a base class
* Added additional tests for ULS6 to include Paros sensors

PR #1582, opened 2025-12-06
Update azfp6 test data

leewujung:
Update azfp6 sha256 and version tag for pooch data, since the new files are in a pre-release v0.11.1a1.


leewujung:
@LOCEANlloydizard : I made this change to get the azfp6 test files updated for PR #1581. I didn't want to change the asset file we have in v0.11.0 so that we can keep the test data versioned too, so I made a pre-releasee v0.11.1a1. This seems to be the recommended way to stage file for new releases, since external PRs cannot access assets in draft release.

Issue #1532, opened 2025-08-29
Add initial single target detection


leewujung:
Add single target detection algorithms.

For the first stage let's add at least those existing in ESP3. We can add the other more advanced/specific ones at a later stage.

PR #1588, opened 2025-12-19
Single target detection CW

LOCEANlloydizard:
This PR introduces the public API and output schema for single target detection! Ref #1532 

It focuses on:
- defining the function signature and dispatcher integration
- validating required inputs and parameters
- returning a stable xr.Dataset with a target-based schema
- issuing warnings when optional metadata are missing

This provides a foundation for upcoming algorithm development and allows early review of the API design, data model, and test coverage!


echopype/tests/mask/test_mask.py
```diff
@@ -2006,3 +2010,180 @@ def test_echoview_mincan_no_linking():
...
+        "Angle_major_axis",
+        "StandDev_Angles_Minor_Axis",
+        "StandDev_Angles_Major_Axis",
+        "Heave",
+        "Roll",
+        "Pitch",
+        "Heading",
+        "Dist",
+        "TSfreq_matrix",
+    ]
```
leewujung: changing these to using snake_case?

echopype/tests/mask/test_mask.py
```diff
@@ -2006,3 +2010,180 @@ def test_echoview_mincan_no_linking():
     # Label the mask and confirm they are separate components (not connected)
     _, nlab = ndi.label(mask.values, structure=np.ones((3, 3), dtype=bool))
     assert nlab == 2
+
+# test for single target detections
+
+REQ = {
+    "channel": "chan1",
+    "pulse_length": 1e-3,             # 1 ms dummy
+    "timeSampleInterval_usec": 40.0,  # dummy
```
leewujung: Not sure where this would be used, but maybe also change this to using snake_case?

echopype/tests/mask/test_mask.py
leewujung: I'll put my comments for the variables themselves in `detect_matecho.py`

echopype/mask/single_target_detection/detect_matecho.py
```diff
@@ -0,0 +1,1005 @@
...
+            Pitch=("target", _arr("Pitch", float), {"units": "rad"}),
+            Heading=("target", _arr("Heading", float), {"units": "rad"}),
+            Dist=("target", _arr("Dist", float), {"units": "m"}),
+            TSfreq_matrix=(("target", "tsfreq_col"), tsfreq, {}),
+        ),
+        coords=dict(
+            target=target,
+            tsfreq_col=np.arange(tsfreq.shape[1], dtype=int),
+            Ping_number=("target", _arr("Ping_number", int)),
+            Time=("target", _arr("Time", np.dtype("datetime64[ns]"))),
```
leewujung: I don't see any variables above aligned with `Ping_number` or `Time`, so think these are redundant and can be removed?

Also, ping number and time are themselves the same thing, unless the timestamps are not derived from ping time (though I think for ping-by-ping methods like this it _is_ derived from ping time?)

leewujung (commented):
Hey @LOCEANlloydizard : Thanks for putting together this PR for discussion! I have a few overarching suggestion:
- use xarray data container (dataarray or dataset) instead of dict when possible -- for example at the output of `_compute_core_matrices`, instead of a Dict, a xr.Dataset would work well. BUT see my comment below about that function [1].
- add more inline code comments to enhance code readability
- using xarray operation would help with code readability too, compared to when doing vanilla numpy operations, since it's xarray is label-aware and it's easier to follow the operations.

-------------------------
I have some thoughts between `_cond_8_build_outputs` and `detect_matecho` that would affect the implementation:

If I read the code right, what you have in the double loop in `_cond_8_build_outputs` is to go through all pings (`ip`) that contain detected single targets, and then all ranges (`pidx`) of the detected targets within that ping. The `out` assembly and refinement does 3 things:
1. "flatten" the ping-range structure to just a 1-D array indexed by `target`
2. refine range in meters, though it's still index-based and not on a continuous scale
3. compute some derived variables, such as `positionBeamCoord`

So, at the core the single target detection output has 1 dimension `target` (which is also a coordinate), and 2 variables: the index number along `ping_time`, and the index number along `range_sample` (or `range_meter`, depending on what the detection is operating on [2]). All other information can be extracted or computed from the original dataarray the detection runs on using these 2 index numbers for each target. `isel` or `sel` allow passing multiple indices at once, so seems more efficient than looping (i.e. vectorizing the code).

If the above is true, how about breaking things up into 2 steps? The first function focuses on getting this "core" single target detection output that's flattened (outputA), and the second function focuses on extracting and computing the required values in the full output (outputB)? Then users can decide if they want to store outputA and stop there, or proceed to compute outputB. 

The reasons why I think it may be useful to separate outputA and outputB are:
- it gives a good progression in what the code is trying to do in a workflow
- derived values in outputB likely change more dramatically if calibration or something else change in the source dataset, but outputA is less variable
- I am not exactly sure about this one: this may provide us a way for scaling the single target detection operation in the future, along with the idea that storing an intermediate data product to disk and then read it back for lazy computation is more scalable than trying to do everything in memory at once.

I am also fine with keeping outputA just as an intermediate produce _within_ `detect_matecho` and not exposing it to the user, though thinking it's good to have a structure that's easy to follow.

----------------------------
[1] `_compute_core_matrices` computes quantities that largely overlap with current echopype `compute_Sv` and `compute_TS` functions, and I think it's better to use those directly (and fix any errors there if any!)

[2] I didn't read in detail to see if the current code includes the range decimation you mentioned earlier, so this comment may have no bearing: as part of `compute_Sv/TS` there's a function that computes sample range and store it into the `range_meter` variable in the Sv dataset. Could we compute the refined range from there? This is so that we don't have redundancy in code that does the same thing scattered in different parts of the codebase, which would make maintenance challenging.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+            raise ValueError(f"ds_sp missing required variable: {v}")
+
+    sp_mat = ds_sp["Sp"]
+
+    if sp_mat.dims != ("ping_time", "range_sample"):
+        raise ValueError("Expected ds_sp['Sp'] dims exactly ('ping_time', 'range_sample').")
+
+    for v in ["echo_range", "angle_alongship", "angle_athwartship"]:
+        if ds_sp[v].dims != ("ping_time", "range_sample"):
+            raise ValueError(f"Expected ds_sp['{v}'] dims exactly ('ping_time', 'range_sample').")
```
leewujung, line 65: Do you intend for these to be _of the exact order_ for the dimensions? as in: is `('ping_time', 'range_sample')` allowed?

If people didn't have the original Sp data generated from echopype, they might have the reversed order. Even for our code i think this has happened before -- I seem to remember there's an open issue to ensure our calibration outputs would always go the same order of dimensions.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+    alpha = ds_sp["sound_absorption"]
+
+    if alpha.ndim == 1:
+        ds_sp = ds_sp.assign(sound_absorption=alpha.broadcast_like(sp_mat))
+    elif alpha.ndim == 0:
+        ds_sp = ds_sp.assign(sound_absorption=xr.zeros_like(sp_mat) + alpha)
+    elif alpha.ndim == 2:
+        pass
+    else:
+        raise ValueError("sound_absorption must be scalar, 1D(ping_time), or 2D like Sp.")
```
leewujung, line 76:
```suggestion
        raise ValueError(
            "sound_absorption must be one of the following: "
            "a scalar, a 1D vector (with dimension ping_time), "
            "or a 2D matrix of the same dimension as Sp."
        )
```

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+            raise ValueError(f"Expected ds_sp['{v}'] dims exactly ('ping_time', 'range_sample').")
+
+    alpha = ds_sp["sound_absorption"]
+
+    if alpha.ndim == 1:
+        ds_sp = ds_sp.assign(sound_absorption=alpha.broadcast_like(sp_mat))
+    elif alpha.ndim == 0:
+        ds_sp = ds_sp.assign(sound_absorption=xr.zeros_like(sp_mat) + alpha)
+    elif alpha.ndim == 2:
+        pass
```
leewujung, line 74: only checking for ndim may be insufficient:
- for ndim=1, what if the vector is not along `ping_time`?
- for ndim=2, what if it is not `ping_time` and `range_sample`?

Would be good to change the code to actually check these.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+            break
+
+    return m, last
+
+
+def _beam_comp_db(ds_sp: xr.Dataset, params: dict) -> xr.DataArray:
+    model = params["beam_comp_model"]
+
+    if model == "none":
+        return xr.zeros_like(ds_sp["Sp"])
```
leewujung, line 156: Why wouldn't this be `model == None`? Seems more natural.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+
+        bw_al = ds_sp["beamwidth_alongship"].broadcast_like(th_al)
+        bw_at = ds_sp["beamwidth_athwartship"].broadcast_like(th_at)
+
+        # Angles from add_splitbeam_angle() are already offset-corrected.
+        x = 2.0 * th_al / bw_al
+        y = 2.0 * th_at / bw_at
+
+        beam_comp_db = 6.0206 * (x**2 + y**2 - 0.18 * x**2 * y**2)
+        return beam_comp_db.broadcast_like(ds_sp["Sp"])
```
leewujung, line 175: Suggest changing this to just `model=="simrad"` or `model=="kongsberg"`.

Also along with my comments on the spectrum PR (#1664), I think we should have a single place where this type of simrad/kongsberg-specific things live, instead of having them duplicated and sprinkled all over the place in our codebase. 

This section is identical to `_get_beam_correction()` used by `_get_beam_compensated_gain()` in `CalibrateEK80`, which is also identical to `_get_B_theta_phi_m()` also under `CalibrateEK80`.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+            "beamwidth_alongship",
+            "beamwidth_athwartship",
+            "angle_offset_alongship",
+            "angle_offset_athwartship",
+        ]:
+            if v not in ds_sp:
+                raise ValueError(f"ds_sp missing required variable for beam compensation: {v}")
+            ds_sp[v] = ds_sp[v] * deg2rad
+
+    beam_comp_db = _beam_comp_db(ds_sp, params)
```
leewujung, line 538: I think it'll be cleaner to have all the beampattern compensation-related code under a consolidated function that lives elsewhere (see https://github.com/echostack-org/echopype/pull/1588/changes#r4074931399). 

Also better to not modify the data variables in place. I think the new consolidated function should handle the angle unit conversion under the hood in a way that does not alter the input dataset.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+    "max_sd_major_deg",
+}
+
+OPTIONAL_PARAMS = {
+    "dec_tir_samples",
+    "bottom_offset_m",
+    "exclude_above_m",
+    "exclude_below_m",
+    "allow_nans_inside_envelope",
+}
```
leewujung, line 22: I think it'll be better to use fully names for these params, except for the easy-to-decipher ones. It is very hard to figure out what `pldl_db` and `dec_tir_samples` are.

The `detect_from_Sp` function needs a thorough docstring section.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+
+    peak = (plike_mat > prev) & (plike_mat >= nxt) & ~(plike_mat == prev)
+    peak = peak & xr.apply_ufunc(np.isfinite, prev)
+    peak = peak & xr.apply_ufunc(np.isfinite, nxt)
+    peak = peak & xr.apply_ufunc(np.isfinite, plike_mat)
+
+    return peak
+
+
+def _nech_p_samples(ds_sp: xr.Dataset) -> np.ndarray:
```
leewujung, line 102: What is this function doing? docstring?

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+    isup_list = []
+    pulse_len_samples_list = []
+    norm_pulse_len_list = []
+    plike_peak_list = []
+    target_range_list = []
+    beam_comp_db_list = []
+    angle_minor_sd_deg_list = []
+    angle_major_sd_deg_list = []
+
+    for it in range(plike_np.shape[0]):
```
leewujung, line 250: Some quick documentation of what this loop is doing would be very useful. The code as is is very impenetrable, especially with all the direct indexing on specific dimensions.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+        off = float(params["bottom_offset_m"])
+        bottom2d = ds_sp["bottom"].broadcast_like(ds_sp["Sp"])
+        cand_mask = cand_mask & (ds_sp["echo_range"] <= (bottom2d - off))
+
+    plike_np = plike_mat.values
+    cand_np = cand_mask.values
+    al_np = ds_sp["angle_alongship"].values
+    ath_np = ds_sp["angle_athwartship"].values
+    range_np = ds_sp["echo_range"].values
+    beam_comp_np = beam_comp_db.values
```
leewujung, line 227: `.value` loads data into memory. `.data` would be be better practice for lazy-loading.

echopype/mask/single_target_detection/detect_from_Sp.py
```diff
@@ -0,0 +1,619 @@
...
+            },
+            attrs=dict(method="from_Sp"),
+        )
+
+    it = feats["ping_index"].values.astype(np.int64)
+    p = feats["range_sample"].values.astype(np.int64)
+
+    ping_time = ds_sp["ping_time"].values[it]
+    single_target_alongship_angle = ds_sp["angle_alongship"].values[it, p] * 180.0 / np.pi
+    single_target_athwartship_angle = ds_sp["angle_athwartship"].values[it, p] * 180.0 / np.pi
```
leewujung, line 415: There are a lot of back and forth conversion between degrees and radians. I think it would be more efficient and less confusing to keep everything in degrees, and just make sure the criteria used and the compensation function does take in degrees as inputs.

echopype/mask/single_target_detection/detect_from_Sp.py
leewujung: See my later comments about keeping inputs in degrees, and keep radians under the hood of functions, to avoid lots of back and forth conversion that current appear in the code.

leewujung (changes requested):
Hey @LOCEANlloydizard : Thanks for the PR! Apologies for my again very late review after your last changes.

I made a quick direct push for the empty output dataset constructor -- please check I didn't break anything.

I see two things needed for this PR, and we can merge -- one is the same as my comments on #1664 re duplicated code that literally does the same thing. 
- beampattern compensation code currently resides in 3 places:
   - `_beam_comp_db`
   - `_get_beam_correction` used by `_get_beam_compensated_gain`
   - `_get_B_theta_phi_m`
  
  The latter two are even within the same class `CalibrateEK80`! This is not good. We should extract these code into a single place that handles kongsberg/simrad-specific stuff and call that function from there in all uses. 

- If I understand correctly, the filtering of targets in `_reject_overlaps_per_ping` can actually occur in `_phase1_simple`, so that things are more efficient
  - right now you assemble a nice dataset and then filter it by converting everything into pure numpy in memory, which is not always the best depending on the data volume. 
  - I think it will be more efficient if you have the target filtering also in the same function. This way to speed things up, we likely would only need to figure out how to send a "block" of pings to a worker, so that detection of multiple blocks can be run at the same time. (This is likely faster than each worker takes 1 ping, because of comms overhead.)

PR #1589, opened 2025-12-22
Update azfp asset sha256

leewujung:
Update azfp.zip in staged release asset.

### DIFF
--- a/.ci_helpers/docker/setup-services.py
+++ b/.ci_helpers/docker/setup-services.py
@@ -29,7 +29,7 @@ COMPOSE_FILE = BASE / "docker-compose.yaml"
 
 def get_pooch_data_path() -> Path:
     """Return path to the Pooch test data cache."""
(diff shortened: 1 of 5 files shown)


LOCEANlloydizard (approved):

LOCEANlloydizard (changes requested):
Hi @leewujung, I think we can try this! In echopype/tests/conftest.py we can change the key 

`key: ep-data-${{ env.ECHOPYPE_DATA_VERSION }}`

to

`key: ep-data-${{ env.ECHOPYPE_DATA_VERSION }}-azfp43976d3e`


The "-azfp43976d3e" part is just an identifier to force a new cache (but it could be anything).

From what i understand,  since azfp.zip was updated under the same v0.11.1a1 tag, the current cache key can restore the data.. even when we try disabling the cache, that doesn’t actually delete existing files on disk or force re-extraction, so cached AZFP data can still be reused in the job.

It seems that on the long term, we should bump the data release tag whenever assets change (but not sur 100%..) but this should fix the CI issue right now!

Cheers!

PR #1591, opened 2025-12-25
Update test_commongrid_api.py

### LINKED ISSUE #1590: Numpy update raise error
Just noting this here to address shortly: NumPy 2.4.0 was released on December 20, and we now see a test failure:

```
echopype/tests/commongrid/test_commongrid_api.py:337: TypeError
TypeError: only 0-dimensional arrays can be converted to Python scalars
```

the reason is that ["Conversion of an array with ndim > 0 to a scalar was deprecated in NumPy 1.25. Now, attempting to do so raises TypeError. Ensure you extract a single element from your array before performing this operation."](https://numpy.org/doc/stable/release/2.4.0-notes.html#raise-typeerror-on-attempt-to-convert-array-with-ndim-0-to-scalar)

LOCEANlloydizard:
Closes #1590

### DIFF
--- a/.ci_helpers/docker/setup-services.py
+++ b/.ci_helpers/docker/setup-services.py
@@ -29,7 +29,7 @@ COMPOSE_FILE = BASE / "docker-compose.yaml"
 
 def get_pooch_data_path() -> Path:
     """Return path to the Pooch test data cache."""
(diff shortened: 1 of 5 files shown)


LOCEANlloydizard:
Thanks @leewujung for the review and for sorting out the azfp data issue! 🎉 cheers!

PR #1597, opened 2026-01-20
feat: Add regrid functionality based on match geometry

### LINKED ISSUE #726: Add function to regrid Sv data

eliascapriles-NOAA:
Implements regrid_all_channel function that would allow for users to match the sampling rates of a channels in a ds_Sv dataset to a specific channel


leewujung:
Thanks @eliascapriles-NOAA ! I'm going to ask @LOCEANlloydizard to help review this since you had most of the discussions with him.

@LOCEANlloydizard - Could you please take a look? feel free to ping me for discussions. Thanks!

echopype/tests/commongrid/test_commongrid_api.py
LOCEANlloydizard: This error "TypeError: only integer scalar arrays can be converted to a scalar index" is due to pandas update to version 3.0. Could you merge the latest version of the main echopype repo (pandas is pinned there!)

LOCEANlloydizard (dismissed):
Hey @eliascapriles-NOAA, thx for the PR!
Following our discussion, a small recap (you probably noted more!):

- the errors from CI are now gone with the merge of pandas < 3, just need to address the assertionError
- remove the +20 padding at the end of the function
- double-check that pings are actually aligned before regridding (there’s already a helper in echopype that does this, see getting_started notebook)
- update the notebook to call the function from your PR, so I can run it on my side too
- we can go from there for other points !

One thing I wanted to flag more generally: right now the regridding is done by looping over channel in Python, and inside that running an apply_ufunc that vectorises over ping_time × range_sample.

This is more a design question than a bug: are we happy with looping over channels in Python and vectorising over pings with apply_ufunc, or should we try to make the whole regridding run as one channel-aware vectorised operation? I know you looked into it.., and i could have a look as well! and maybe @leewujung would have recommendations for this?

Cheers!

leewujung:
Hey @eliascapriles-NOAA @LOCEANlloydizard : Not sure if I am really making a good suggestion, but since `apply_ufunc` is already configured to be run in parallel, it seems to me that it would be useful to have all channels to be processed at once (instead of in a loop). My thought is then this way the parallelization component is delegated to the xarray-dask combination, rather than us making specific choices ourselves (unless we know for sure that it is better). Let me know what you think!

leewujung:
Sounds good, thanks @eliascapriles-NOAA! Let us know if that ended up working out. It may be useful to try to benchmark the performance. Sometimes we _think_ it's faster one way but that may or may not be true, depending on how the data is chunked, if the data is lazy-loaded, the total data size, and the worker memory...

echopype/commongrid/utils.py
```diff
@@ -649,3 +649,171 @@ def assign_actual_range(ds_MVBS: xr.Dataset) -> xr.Dataset:
...
+        The xarray Dataset containing the variable of interest.
+    channel : str
+        The label used for the channel to access within the Dataset.
+    deepest_ping : int
+        The index of the ping that contains the deepest valid sample in the specified variable.
+
+    Returns
+    -------
+    deepest_ping: int
+        index of the "deepest" sample
```
LOCEANlloydizard: something like

"Index of the ping containing the deepest valid range sample."

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
     return ds_NASC
 
 
-def regrid():
-    return 1
+def regrid(
```
LOCEANlloydizard: just thinking.. maybe regrid is a bit too broad term?
what about something more specific, like "resample_to_geometry" or something like that?

echopype/commongrid/utils.py
```diff
@@ -649,3 +649,171 @@ def assign_actual_range(ds_MVBS: xr.Dataset) -> xr.Dataset:
...
+    ds: xr.Dataset, target_channel: str = None, target_grid: xr.DataArray = None
+) -> int:
+    """
+    Finds the integer indices of the last non-NaN value in a DataArray.
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        The xarray Dataset containing the variable of interest.
+    channel : str
```
LOCEANlloydizard: target_channel

echopype/commongrid/utils.py
```diff
@@ -649,3 +649,171 @@ def assign_actual_range(ds_MVBS: xr.Dataset) -> xr.Dataset:
...
+        The label used for the channel to access within the Dataset.
+    deepest_ping : int
+        The index of the ping that contains the deepest valid sample in the specified variable.
```
LOCEANlloydizard: delete/move in Returns

echopype/commongrid/utils.py
```diff
@@ -649,3 +649,171 @@ def assign_actual_range(ds_MVBS: xr.Dataset) -> xr.Dataset:
         round(float(ds_MVBS["Sv"].max().values), 2),
     ]
     return ds_MVBS.assign_attrs({"actual_range": actual_range})
+
+
+def get_valid_max_depth_ping(
+    ds: xr.Dataset, target_channel: str = None, target_grid: xr.DataArray = None
```
LOCEANlloydizard, line 655: need to define in parameters

echopype/commongrid/utils.py
```diff
@@ -649,3 +649,171 @@ def assign_actual_range(ds_MVBS: xr.Dataset) -> xr.Dataset:
         round(float(ds_MVBS["Sv"].max().values), 2),
     ]
     return ds_MVBS.assign_attrs({"actual_range": actual_range})
+
+
+def get_valid_max_depth_ping(
```
LOCEANlloydizard: how about 

"get_deepest_valid_ping_index"? to mention the index?

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
...
+    ds_Sv, target_variable: str = None, target_channel: str = None, target_grid: xr.DataArray = None
```
LOCEANlloydizard: we should not have the target_variable as None?
and also make the target_channel / targt_grid optional?

something like ?

def regrid(
    ds_Sv: xr.Dataset,
    target_variable: str, 
    target_channel: str | None = None,
    target_grid: xr.DataArray | None = None,
) -> xr.Dataset:

like this we can remove then also the 

"if target_variable is None:
    raise ValueError(...)"


Or we allow None, and switch to Sv directly if thats the case?

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
...
+) -> xr.Dataset:
+    """
+    Regrids all channels in the EchoData object to match the geometry
+    of the specified target channel index.
+
+    Parameters
+    ----------
+    ds_Sv : xr.Dataset
+        Input Dataset containing Sv data
+    target_channel : string, optional
```
LOCEANlloydizard: proposition:

target_channel : str, optional
Channel used as reference grid. Must be provided if target_grid is None.

target_grid : xr.DataArray, optional
Custom grid. Must be provided if target_channel is None.

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
...
+    Returns
+    -------
+    xr.Dataset
+        A new Dataset where all channels share the same `ping_time`,
+        `range_sample`, and `echo_range` as the target.
+    """
+    if target_variable is None:
+        raise ValueError("Provide a target variable.")
+
+    if target_channel is not None and target_grid is not None:
```
LOCEANlloydizard: if we pass for instance

```
aligned_ds_channel = regrid(
    ds_Sv.chunk({"channel": 1, "ping_time": 1000, "range_sample": -1}),
    target_variable="Sv"
)
```

then it pass this steps, and we get a downstream error in the copy(). We culd check here that both are not None as well !

e.g.

```
if (target_channel is None) == (target_grid is None):
    raise ValueError(
        "Provide exactly one of target_channel or target_grid."
    )
```

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
...
+        raise ValueError("Provide only one of target_channel or target_grid, not both.")
+
+    if exist_reversed_time(ds_Sv, "ping_time"):
+        coerce_increasing_time(ds_Sv)
+
+    channels = ds_Sv.channel.values
+    ds_var_names = ds_Sv.keys()
+
+    if target_variable not in ds_var_names:
+        raise IndexError(f"{target_variable} is not part of the variable names in : {ds_var_names}")
```
LOCEANlloydizard: the IndexError is a bit hardr to read I find, maybe ValueError is shorter? thoughts?

e.g.

```
raise ValueError(
    f"target_variable must be one of {list(ds_var_names)}, got {target_variable!r}."
)
```

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
...
+    expected_dims = {"channel", "ping_time", "range_sample"}
+    actual_dims = set(ds_Sv[target_variable].dims)
+    if actual_dims != expected_dims:
+        raise ValueError(
+            f"Target variable '{target_variable}' must have exactly the dimensions "
+            f"('channel', 'ping_time', 'range_sample'). Found: {ds_Sv[target_variable].dims}"
+        )
+
+    if target_channel and target_channel not in channels:
+        raise IndexError(f"{target_channel} is not part of the channel names in : {channels}")
```
LOCEANlloydizard: same, maybe use ValueError?

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
...
+            .isel(ping_time=deepest_ping_index)
+            .values
+        )
+
+    # Target grid is given
+    else:
+        target_range_da = target_grid.copy()
+
+        deepest_ping_index = get_valid_max_depth_ping(ds_Sv, target_grid=target_grid)
+        valid_range_sample = np.argmax(target_grid.isel(ping_time=deepest_ping_index).values)
```
LOCEANlloydizard: I feel here we could be more concise if 1) we don't trim ? (to discuss) and no need for copy()?

something like:
```
if target_channel is not None:
    target_range_da = ds_Sv["echo_range"].sel(channel=target_channel)
else:
    target_range_da = target_grid
```

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
...
+
+    # List to hold the aligned DataArrays
+    aligned_arrays = []
+
+    for i, channel in enumerate(channels):
```
LOCEANlloydizard: if we don't use `i` it could be 

`for channel in channels:`

echopype/commongrid/api.py
```diff
@@ -441,5 +446,128 @@ def compute_NASC(
...
+
+        ds_source = da_var.sel(channel=channel)
+
+        if target_variable == "Sv":
+            source_linear = _log2lin(ds_source)
+        else:
+            source_linear = ds_source.copy()
```
LOCEANlloydizard: maybe .copy() not needed

echopype/tests/commongrid/test_commongrid_api.py
```diff
@@ -90,7 +93,280 @@ def test__groupby_x_along_channels(request, range_var, lat_lon):
...
+    def calculate_total_energy(ds, channel):
+        """
+        Calculates the total integrated energy (Linear Sv * Thickness) 
+        per ping for a specific channel.
+        
+        Parameters
+        ----------
+        ds : xr.Dataset
+            Dataset containing 'Sv' and 'echo_range'.
+        channel_idx : int
```
LOCEANlloydizard: should be str here ? maybe specify type in arguments?

echopype/tests/commongrid/test_commongrid_api.py
```diff
@@ -90,7 +93,280 @@ def test__groupby_x_along_channels(request, range_var, lat_lon):
...
+    """Testing the regrid_all_channels_robust wrapper function"""
+    if er_type == "regular":
+        ds_Sv = request.getfixturevalue("ds_Sv_echo_range_regular")
+    else:
+        ds_Sv = request.getfixturevalue("ds_Sv_echo_range_irregular")
+    channel = ds_Sv["channel"].values[0]
+
+    ds_regridded = ep.commongrid.regrid(ds_Sv.chunk({"channel": 1, "ping_time": 1000, "range_sample": -1}), target_variable="Sv", target_channel=channel)
+
+    def calculate_total_energy(ds, channel):
```
LOCEANlloydizard: defined twice (in two test), we could have as a sngle pytest fixture?

echopype/tests/commongrid/test_commongrid_api.py
```diff
@@ -90,7 +93,280 @@ def test__groupby_x_along_channels(request, range_var, lat_lon):
...
+    echodata = ep.open_raw(ek80_raw_path, sonar_model='EK80')
+    ds_Sv = ep.calibrate.compute_Sv(
+        echodata, waveform_mode='CW', encode_mode='complex'
+    )
+    
+    channel = ds_Sv["channel"].values[0]
+
+    ds_regridded = ep.commongrid.regrid(ds_Sv, target_variable="Sv", target_channel=channel)
+
+    channel = ds_Sv["channel"].values[0]
```
LOCEANlloydizard: this need to be deleted (duplicate)

LOCEANlloydizard:
Hey @eliascapriles-NOAA, thanks for the modifications! I’ve left a few comments above, and also added some broader thoughts below to check first =>

1. Naming: maybe `resample` would be clearer than `regrid`, which feels a bit generic? thoughts?

2. Do we actually want the trimming step after the resampling? forgot if we discussed this
- If not, we could remove this entirely
- If yes, we could likely still replace get_valid_max_depth_ping with a xarray based approach?

3. Output dataset: 

we now returns a new xarray object which is nice, but we might want to align it more with other regridding like compute_MVBS?

-propagate key variables:
-- water_level (to allow add_depth() downstream)
-- frequency_nominal
-- position variables (lat/lon, etc)

-add provenance:
-- processing_function = "commongrid.regrid" (or resample)?

-add variable-level attrs on Sv:
-- resampling_mode ("target_channel" / "target_grid")
-- target_channel
-- short description?

4. .copy() usage: here (and other places, see comments)

```
ds_target = ds_Sv.sel(...).copy()
target_range_da = target_grid.copy()
```

does not seem necessary here since nothing is mutated afterwards? (but could be missing the obvious!)

5. Some small fixes in the tests

- in some places there's a mix of positional indexing and label-based selection? It could be better to consistently use label selection?
- Maybe we can compare the target channel element by element before and after regridding onto its own geometry? should be the same
- If we add water_level, frequency_nominal, lat/lon, attrs, or processing_function, we’ll probably want explicit tests for metadata/provenance too?
- Maybe it would be worth adding a test for cases where target_grid has NaNs/partially valid geometry? just to check the expected behaviour?

It's the final adjusments! Thank you very much for this!
Cheers!

echopype/commongrid/utils.py
eliascapriles-NOAA: How about:

deepest_ping: int
        Index of the ping containing with most amount of data before trailing NaN values.

echopype/commongrid/api.py
eliascapriles-NOAA: I like resample_to_geometry !

echopype/commongrid/api.py
eliascapriles-NOAA: I think it be better to have the user always specify the target variable. Good catch !

echopype/commongrid/api.py
eliascapriles-NOAA: Don't know why I even used enumerate here. Thank you for the suggestion

echopype/commongrid/api.py
eliascapriles-NOAA: The reason for trimming was to originally save memory and have a visual way to notice that the regridding initially worked by printing the dimension size. However, the helper function masks each nan ping to only take into account the pings. I am going to test the output without the trimming to make sure everything still works, and get back to you.

I agree that copy does not contribute anything. I will delete them.

echopype/commongrid/api.py
eliascapriles-NOAA: Done

echopype/commongrid/api.py
eliascapriles-NOAA: After testing the regrid without any trimming everything still works and regrid simply. @leewujung and @LOCEANlloydizard which of these approaches do you think better reflects echopype's current functions.

echopype/commongrid/api.py
eliascapriles-NOAA: done

echopype/commongrid/api.py
LOCEANlloydizard: I guess we're going with the no trimming for now?

LOCEANlloydizard (dismissed):
Hello @eliascapriles-NOAA, @leewujung, i've made a few final small adjustments! There will probably still be some edge cases to handle over time but I think we're ready to merge 🙂
A few of the final additions:
- support for Sp and TS through the same log -> linear -> log resampling pathway used for Sv
- warning when resampling angle variables
- a few extra tests covering Sp/TS handling, angle warnings, and target-geometry validation

I think it would be nice to eventually add a regression test against Echoview if we can process the same file and match all channels to the 200 kHz geometry, similar to what is shown in the notebook: https://github.com/echostack-org/echopype-examples/pull/101
@eliascapriles-NOAA I can help with this, but I'd need an output from Echoview for this file to compare against.. let me know if you would have time to generate it! Cheers!

LOCEANlloydizard:
Ok thank you @eliascapriles-NOAA for the file! I’ve added the Echoview regression test using the GitHub asset added in #1679. The comparison between Echoview and echopype shows very close output grids, with Sv differences below 0.003 dB outside the near-transducer region
(See test: `echopype/tests/commongrid/test_commongrid_api.py::test_resample_matches_echoview_match_geometry`).

We also now raise warnings if users try to resample TS, Sp, or angle variables, since these cases should be interpreted with caution. The [notebook](https://github.com/echostack-org/echopype-examples/pull/101) has also been updated.
I think we’re good to merge! Cheers

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+    """
+    Regrids all channels in the EchoData object to match the geometry
+    of the specified target channel index.
+
+    Parameters
+    ----------
+    ds_Sv : xr.Dataset
+        Input Dataset containing Sv data
+    target_channel : str, optional
+    Channel used as reference grid. Must be provided if target_grid is None.
```
leewujung:
```suggestion
        Channel used as reference grid. Must be provided if target_grid is None.
```

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+
+    target_grid : xr.DataArray, optional
+    Custom grid. Must be provided if target_channel is None.
+    Data array must have dimension ('ping_time', 'range_sample').
```
leewujung:
```suggestion
        Custom grid. Must be provided if target_channel is None.
        Data array must have dimension ('ping_time', 'range_sample').
```

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+    LOG_VARIABLES = {"Sv", "Sp", "TS"}
+
+    if target_variable != "Sv":
+        warnings.warn(
+            f"Resampling '{target_variable}' with overlap-weighted averaging. "
+            "This function is primarily intended and validated for Sv. "
+            "Variables such as Sp and TS are resampled in the linear domain, "
+            "but interpretation should be done with caution. "
+            "Angle variables are resampled geometrically and may not be "
+            "physically equivalent to recomputing angles from complex data.",
```
leewujung, line 496:
```suggestion
            f"Resampling '{target_variable}' with overlap-weighted averaging. "
            "This function is primarily intended and validated for Sv. "
            "Variables such as Sp and TS are resampled in the linear domain, "
            "but interpretation should be done with caution. "
            "Angle variables are resampled geometrically and is not "
            "physically equivalent to recomputing angles from complex data.",
```

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+            UserWarning,
+            stacklevel=2,
+        )
+
+    if (target_channel is not None) == (target_grid is not None):
+        raise ValueError("Provide only one of target_channel or target_grid, not both.")
```
leewujung: Shouldn't the `==` be `and` ?
Otherwise, if `target_channel` is None and `target_grid` is also None, these 2 checks would be equal and then the value error would not make sense.
This is not caught by the current tests also, so would be good to add a test for this.

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+
+    if (target_channel is None) == (target_grid is None):
+        raise ValueError("Provide exactly one of target_channel or target_grid.")
```
leewujung, line 502: Same comment as https://github.com/echostack-org/echopype/pull/1597/changes#r3564628818

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+
+    if exist_reversed_time(ds_Sv, "ping_time"):
+        coerce_increasing_time(ds_Sv)
```
leewujung, line 511: Suggest adding a user warning here to let them know that there are reversed ping times and they have been corrected.

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
     return ds_NASC
 
 
-def regrid():
-    return 1
+def resample_to_geometry(
+    ds_Sv, target_variable: str, target_channel: str | None = None, target_grid: xr.DataArray = None
```
leewujung: `target_variable` is missing in the Parameters list.
Also, should we set its default to "Sv"?

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+    aligned_arrays = []
+
+    for channel in channels:
+
+        ds_source = da_var.sel(channel=channel)
+
+        if target_variable in LOG_VARIABLES:
+            source_linear = _log2lin(ds_source)
+        else:
+            source_linear = ds_source
```
leewujung: This assumes that unless the user wants to regrid a variable in LOG_VARIABLES, the regridding will  just be done on the provided values of the variable. Since we allow users to use this function on arbitrary `target_variable`, I feel we should expose a boolean input argument to give user control of whether the variable they want to regrid should be done in the linear or log domain.

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+            source_linear,
+            input_core_dims=[["range_sample"], ["range_sample"], ["range_sample"]],
+            output_core_dims=[["range_sample"]],
+            vectorize=True,
+            dask="parallelized",
+            output_dtypes=[np.float64],
+        )
+
+        # Convert back to log domain
+        result_linear = result_linear.where(result_linear > 0)
```
leewujung: I am curious -- what are the conditions when `result_linear` would be =0 or <0? Given that this is a weight interpolation, the `result_linear` would always be >=0, right? So is this just to remove =0 results? and if so, why?

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+    ds_combined = xr.concat(aligned_arrays, dim="channel")
+    echo_range_aligned = target_range_da.broadcast_like(ds_combined)
+
+    new_ds = xr.Dataset(
+        data_vars={
+            target_variable: ds_combined,
+            "echo_range": echo_range_aligned,
+            "frequency_nominal": ds_Sv["frequency_nominal"],
+        }
+    )
```
leewujung, line 622: We should also have `channel` as a dimension? Other parts of echopype anchors on using `channel` as the main coordinate (since there could be 2 different channels of the same frequency, say in a echosounder configuration containing 2 transducers of the same frequencies pointing in 2 different directions), so it'd be good to do that here as well.

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+    new_ds[target_variable].attrs = ds_Sv[target_variable].attrs
+    range_var_attrs = np.round(target_range_da.diff("range_sample").mean().values, decimals=4)
+    if target_channel:
+        new_ds[target_variable].attrs["resampling_mode"] = "target_channel"
+        new_ds[target_variable].attrs["target_channel"] = target_channel
+        new_ds[target_variable].attrs["grid_spacing"] = str(range_var_attrs) + "m"
+    else:
+        new_ds[target_variable].attrs["resampling_mode"] = "target_grid"
+        new_ds[target_variable].attrs["target_grid"] = target_grid
+        new_ds[target_variable].attrs["grid_spacing"] = str(range_var_attrs) + "m"
```
leewujung: This works right now because echopype currently does not produce range vectors that are of unequal spacing (because right now we only allow constant temperature and salinity when computing sound speed to generate Sv). This will change soon (hopefully) as we incorporate the capability to compute Sv based on a sound speed profile (ie varying along range or depth).

I am not sure what is the goal for including the `grid_spacing` attribute, but renaming it to `mean_grid_spacing` would be more accurate.

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+
+    prov_dict = echopype_prov_attrs(process_type="processing")
+    prov_dict["processing_function"] = "commongrid.resample_to_geometry"
+    new_ds = new_ds.assign_attrs(prov_dict)
+    new_ds = insert_input_processing_level(new_ds, ds_Sv)
+
+    if "echo_range" in ds_Sv:
+        new_ds["echo_range"].attrs = ds_Sv["echo_range"].attrs
```
leewujung: I don't understand the logic here -- the regridding computation is entirely based on `echo_range`, so is there any cases in which `echo_range` does not exist in `ds_Sv`?

echopype/commongrid/utils.py
```diff
@@ -696,3 +696,170 @@ def ping_time_bin_parsing_and_conversion(ping_time_bin: str):
     )
     ping_time_bin_resunit_label = timedelta_units[ping_time_bin_resunit]["unitstr"]
     return ping_time_bin_resvalue, ping_time_bin_resunit_label
+
+
+def get_valid_max_depth_ping_index(
```
leewujung: Where is this used?

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
 
 
-def regrid():
-    return 1
+def resample_to_geometry(
+    ds_Sv, target_variable: str, target_channel: str | None = None, target_grid: xr.DataArray = None
+) -> xr.Dataset:
+    """
+    Regrids all channels in the EchoData object to match the geometry
+    of the specified target channel index.
```
leewujung: Added some clarification:
```suggestion
    Regrids a variable across all channels in the EchoData object to match the geometry
    along range of the specified target channel. 
    Ping time is assumed identical for all input channels.
```

echopype/commongrid/api.py
```diff
@@ -416,5 +421,151 @@ def compute_NASC(
...
+    Channel used as reference grid. Must be provided if target_grid is None.
+
+    target_grid : xr.DataArray, optional
+    Custom grid. Must be provided if target_channel is None.
+    Data array must have dimension ('ping_time', 'range_sample').
+    Returns
+    -------
+    xr.Dataset
+        A new Dataset where all channels share the same `ping_time`,
+        `range_sample`, and `echo_range` as the target.
```
leewujung, line 476:
```suggestion
    xr.Dataset
        A new Dataset where all channels share the same `ping_time`,
        `range_sample`, and `echo_range` as the target. 
        `ping_time` is assumed identical across all input channels 
        and preserved throughout the resampling process.
```

leewujung (changes requested):
@eliascapriles-NOAA @LOCEANlloydizard : Thanks for this PR, awesome to have this in echopype! I think it's in great shape with just a few things needing attention. Please take a look on the inline comments. Some are very minor textual changes and some would require some (small) code changes.

echopype/commongrid/api.py
LOCEANlloydizard: thx! i've added target_variable and set it to default!

echopype/commongrid/api.py
LOCEANlloydizard: I've actually removed this line as it was a duplicate of the one below
`if (target_channel is None) == (target_grid is None):`

So I think it's OK like this? we're raising if both are None (True == True) or if both are provided (False == False)? And we have a small test (test_resample_requires_exactly_one_target) but I've added the match for the error message, thanks!

echopype/commongrid/utils.py
LOCEANlloydizard: this was a leftover from an earlier implementation before we simplified the logic! I've removed it, thx!

echopype/commongrid/api.py
LOCEANlloydizard: see above

echopype/commongrid/api.py
LOCEANlloydizard: i've added

```
        warnings.warn(
            "Reversed ping_time values detected. The dataset has been reordered "
            "to increasing ping_time before resampling.",
            UserWarning,
        )
```

but we don't have a test for this.. should we?

echopype/commongrid/api.py
LOCEANlloydizard: yes agreed! I've replaced LOG_VARIABLES with an is_log boolean argument, giving users control over the resampling domain!

echopype/commongrid/api.py
LOCEANlloydizard: thx! i now moved `result_linear = result_linear.where(result_linear > 0)` only inside the is_log condition to prevents zero linear values from becoming -inf when converted back to the log domain.. what do you think?

echopype/commongrid/api.py
LOCEANlloydizard: yep thanks! I’m investigating how the regridding handles irregular grids across pings, and I agree that in this case the mean spacing is not informative.. so I’ve removed the attribute!

echopype/commongrid/api.py
LOCEANlloydizard: thx! i think channel is preserved as a dimension through xr.concat? (`ds_combined = xr.concat(aligned_arrays, dim="channel")`) but I’ve now also made it explicit as a coordinate in the output dataset!

echopype/commongrid/api.py
LOCEANlloydizard: true! removed the condition

LOCEANlloydizard:
Hey @leewujung @eliascapriles-NOAA, i've applied the changes and fixed the incoherence/errors! I've also updated the notebook (https://github.com/echostack-org/echopype-examples/pull/101). We should be good, let me know if something seems off! cheers!

echopype/commongrid/api.py
leewujung: I guess my question is why we would have negative values in the first place. Since these are integration, shouldn't the output be all positive?

echopype/commongrid/api.py
```diff
@@ -416,5 +421,168 @@ def compute_NASC(
...
+        new_ds["water_level"] = ds_Sv["water_level"]
+
+    # Attach attributes
+    new_ds[target_variable].attrs = ds_Sv[target_variable].attrs
+
+    if target_channel:
+        new_ds[target_variable].attrs["resampling_mode"] = "target_channel"
+        new_ds[target_variable].attrs["target_channel"] = target_channel
+    else:
+        new_ds[target_variable].attrs["resampling_mode"] = "target_grid"
```
leewujung, line 633: Thinking about the "target_grid" option - I wonder if it's useful to preserve `target_grid` as a variable? in a sense it is not necessary, but it might be useful to have it around?

echopype/commongrid/api.py
```diff
@@ -416,5 +421,168 @@ def compute_NASC(
...
+            stacklevel=2,
+        )
+
+    if (target_channel is None) == (target_grid is None):
+        raise ValueError("Provide exactly one of target_channel or target_grid.")
+
+    if exist_reversed_time(ds_Sv, "ping_time"):
+        warnings.warn(
+            "Reversed ping_time values detected. The dataset has been reordered "
+            "to increasing ping_time before resampling.",
```
leewujung:
```suggestion
            "Reversed ping_time values detected. The ping_time variable "
            "has been modified to increase monotonically before resampling. "
            "See echopype.qc.coerce_increasing_time() for detail.",
```

echopype/commongrid/api.py
leewujung: Sounds good to have a test for this!

leewujung (changes requested):
Hey @eliascapriles-NOAA  @LOCEANlloydizard: I've just reviewed the code changes again, and went through [the notebook](https://echopype-examples--101.org.readthedocs.build/en/101/resample_to_geometry.html) -- good job! I made a few inline comments that should be straightforward to address. There are some small wording changes that I would do direct push to the notebook.

A couple bigger items:

1. This one is from the notebook [section that builds a custom grid](https://echopype-examples--101.org.readthedocs.build/en/101/resample_to_geometry.html#build-a-custom-200-khz-spacing-grid): I feel that the restriction of having the length of the custom grid to be shorter than that of the `range_sample` dimension is artificial. Since the underlying regridding function is coded in numpy, I think we should allow `target_grid` to be of any length. If it needs to be an xarray object, it should be possible to use `.broadcast_like` or similar functions to expand the `range_sample` dimension, in the case when `target_grid` is longer?

2. This is about the function itself: I think we should allow `target_grid` to potentially only contain the `echo_range` dim but not `ping_time`. Take the custom grid built [here](https://echopype-examples--101.org.readthedocs.build/en/101/resample_to_geometry.html#create-the-target-grid) for example, allowing this would remove the need for the user to have to create the entire `ping_time` dimension by tiling the same vector, making the function a lot more convenient to use, since probably >90% of the use case people would want a uniform grid across channel and ping time.

echopype/commongrid/api.py
LOCEANlloydizard: mhmm, good question! the thing is target_grid is preserved in the output as echo_range, so do you mean keeping it as a separate variable as well? or we could instead document more explicitly how to retrieve the grid from echo_range? like `target_grid = regridded_ds["echo_range"]`

echopype/commongrid/api.py
LOCEANlloydizard: ok thx! I added the test _test_resample_warns_and_corrects_reversed_ping_time_ to verify that the warning is raised!

echopype/commongrid/api.py
leewujung: Agree that if we document target_grid = regridded_ds["echo_range"] it'll be better!

LOCEANlloydizard:
Thx @leewujung for the review! Both points (which are indeed great enhancements!) are addressed:

- target_grid can be shorter or longer than the input range_sample dimension (and I added a test covering a longer grid)
- target_grid can be either 1D (range_sample) or 2D (ping_time, range_sample): a 1D grid is automatically applied across all pings and channels

(i also replied to the inline comments about the reversed ping_time test and preserving target_grid through the output echo_range)

I hope that covers what you had in mind; and the notebook has also been updated to reflect the changes! Cheers!

echopype/commongrid/api.py
LOCEANlloydizard: ok just added it!

echopype/commongrid/api.py
```diff
@@ -505,17 +523,32 @@ def resample_to_geometry(
     if target_channel and target_channel not in channels:
         raise ValueError(f"{target_channel} is not part of the channel names in : {channels}")
 
-    if target_grid is not None and target_grid.dims != ("ping_time", "range_sample"):
-        raise ValueError("target_grid dimensions do not match expected dimensions.")
+    valid_target_grid_dims = {
+        ("range_sample",),
+        ("ping_time", "range_sample"),
+    }
```
leewujung, line 530: Just a tiny point that we've seen before that somehow some datasets became `("range_sample", "ping_time")` (reversed order of dims). Should we add this additional one?

echopype/commongrid/api.py
```diff
@@ -429,8 +441,8 @@ def resample_to_geometry(
     target_grid: xr.DataArray | None = None,
 ):
     """
-    Regrids a variable across all channels in the EchoData object to match the geometry
-    along range of the specified target channel.
+    Regrids a variable across all channels in the EchoData object to
+    a common range geometry specified by either a target channel or a custom target grid.
     Ping time is assumed identical for all input channels.
```
leewujung: Somehow couldn't add suggested edits -- but this would be on a variable in a Sv or Sp dataset and not in the EchoData object, right?

leewujung (dismissed):
@LOCEANlloydizard : The changes look great! I only made 2 tiny comments. Feel free to merge once those are addressed!

leewujung:
Oh also, I am backtracking my comments about potentially using xcDat for this operation. It could be useful, but if the current algorithm gives us the same values as the Echoview outputs and `xr.apply_ufunc` is already used with `dask="parallelized"`, I think it's ok to just keep this as is.

echopype/commongrid/api.py
LOCEANlloydizard: oh yes right! will fix now, thx!

echopype/commongrid/api.py
LOCEANlloydizard: yes sure, i'll add it now!

LOCEANlloydizard:
ok thx @leewujung @eliascapriles-NOAA, merging this now! cheers!

eliascapriles-NOAA:
This has been a team effort y'all. Thank you for all the help and advice !

PR #1621, opened 2026-03-10
Return tau_effective from compute_Sv and update tau_effective handling

LOCEANlloydizard:
Adding CW validation tests for tau_effective and return tau_effective from compute_Sv

- add tau_effective in the power-encoding path
- add EK80 CW validation tests for power-encoded data against Matecho and pyEcholab
- add EK80 CW validation tests for complex-encoded data against Matecho and Echoview
- compare both tau_effective and Sv outputs
- return tau_effective in compute_Sv for both power and complex encode paths


echopype/tests/calibrate/test_calibrate_ek80_CW.py
```diff
@@ -0,0 +1,371 @@
...
+    freq_khz = (ds_ep["frequency_nominal"] / 1000).astype(int)
+    freq_khz_vals = [int(x) for x in freq_khz.values.tolist()]
+
+    n_ping = sv_ep.sizes["ping_time"]
+    n_range = sv_ep.sizes["range_sample"]
+
+    mats = []
+
+    for f in freq_khz_vals:
+        mats.append(_pad_or_crop_2d(ref["Sv"][f], n_ping, n_range))
```
leewujung: Do we actually need `_pad_or_crop_2d`? I checked the dimensions and the size of `ref` and `sv_ep` datasets are the same along the `ping_time` and `range_sample` dimension.

echopype/tests/calibrate/test_calibrate_ek80_CW.py
```diff
@@ -0,0 +1,371 @@
...
+def ek80_ext_path(test_path):
+    return test_path["EK80_EXT"]
+
+
+# --- Shared tolerances / comparison settings
+
+_SKIP_RS = 100
+_SV_ATOL_DB = 1e-2
+_TAU_RTOL = 1e-3
+_TAU_RTOL_PYECHOLAB = 1e-6
```
leewujung: `_SV_ATOL_DB` is set to 0.01 -- this is due to the results from one particular combination of data type (power or CW) and package (matecho or pyecholab), right? I think it may be better to set a different tolerance for different comparison, so that it does not give the false impression that the differences are all so large

Similar questions for `_TAU_RTOL` and `_TAU_RTOL_PYECHOLAB`.

For `_SKIP_RS`,  i am not sure if both matecho and pyecholab skips the same number of samples?

echopype/tests/calibrate/test_calibrate_ek80_CW.py
```diff
@@ -0,0 +1,371 @@
...
+    calibration matches Matecho reference values."""
+    
+    ds = cw_power_ds
+
+    tau_ep = _tau_ep_to_freq_vector(ds)
+
+    ref = _load_power_ref(ek80_ext_path, "matecho")
+    tau_ref = _ref_tau_to_da(ref)
+
+    tau_ep_aligned, tau_ref_aligned = xr.align(tau_ep, tau_ref, join="exact")
```
leewujung: Do we need `xr.align` for the comparison to work? If my check is right, `tau_ep` and `tau_ref` are already aligned? 

Also, by setting `join="exact"`, an error would already be raised if `tau_ep` and `tau_ref` are not already aligned, right?

Maybe this is just a precautionary measure, but I thought I'd ask.

echopype/tests/calibrate/test_calibrate_ek80_CW.py
```diff
@@ -0,0 +1,371 @@
...
+        name="Sv",
+    )
+
+
+# --- POWER SAMPLES CW
+
+_HAKE_RAW_REL = "ncei-wcsd/SH2306/Hake-D20230811-T165727.raw"
+
+
+def test_ek80_CW_power_tau_effective_matches_matecho(cw_power_ds, ek80_ext_path):
```
leewujung: Between `test_ek80_CW_power_tau_effective_matches_matecho` and `test_ek80_CW_power_tau_effective_matches_pyecholab`, since most of the code are the same, seems it'd be easier to manage using @pytest.mark.parametrize?

The same the `test_ek80_CW_power_Sv_matches_*` and `test_ek80_CW_complex_Sv_matches_*` tests.

echopype/tests/calibrate/test_calibrate_ek80_CW.py
```diff
@@ -0,0 +1,371 @@
...
+            "ping_time": sv_ep["ping_time"].values,
+            "range_sample": sv_ep["range_sample"].values,
+        },
+        name="Sv",
+    )
+
+
+# --- POWER SAMPLES CW
+
+_HAKE_RAW_REL = "ncei-wcsd/SH2306/Hake-D20230811-T165727.raw"
```
leewujung: Since `_HAKE_RAW_REL` and `_COMPLEX_ZARR_REL` are used only once in the fixtures, it seems better to just have them in the fixture, so that there's no need to hunt for these constants.

echopype/tests/calibrate/test_calibrate_ek80_CW.py
leewujung: If using @pytest.mark.parametrize, you could still set different tolerance by setting it as one of the parameters passed into the function.

echopype/calibrate/calibrate_ek.py
```diff
@@ -78,17 +79,54 @@ def _cal_power_samples(self, cal_type: str) -> xr.Dataset:
...
+                    channel=self.chan_sel,
+                    ping_time=beam["ping_time"],
+                )
+            except Exception as e:
+                logger.warning(
+                    "Could not compute tau_effective from transmit signal in power path; "
+                    "falling back to transmit_duration_nominal. Error: %s",
+                    repr(e),
+                )
+                tau_effective = beam["transmit_duration_nominal"].isel(ping_time=0)
```
leewujung, line 134: Curious - do you run into files that do not contain the parameters needed to compute `tau_effective`? Would be a good idea to note when that is the case.

Also, what do you mean by "power path"?
in the warning message: 
> Could not compute tau_effective from transmit signal in power path; ...

leewujung (changes requested):
@LOCEANlloydizard : Thanks for the investigation and the PR! Sorry it took me a long while to get to this...  I have a few comments that are not very substantial, but I think they would help keep the code tighter.

Also, by just looking at this, I don't think there will be merge conflict with #1508 🤞, so once these are addressed we can merge it.

echopype/tests/calibrate/test_calibrate_ek80_CW.py
```diff
@@ -0,0 +1,371 @@
...
+        ep_vals = ep_vals[..., skip_range_samples:]
+        ref_vals = ref_vals[..., skip_range_samples:]
+
+    m = _finite_mask(ep_vals) & _finite_mask(ref_vals)
+    assert m.any(), "No overlapping finite Sv values to compare."
+
+    assert np.allclose(ep_vals[m], ref_vals[m], rtol=rtol, atol=atol_db)
+
+
+def _load_power_ref(ek80_ext_path, which: str) -> dict:
```
leewujung: I'd suggest renaming `which` to `pkg_name` for readability. `which` to me feels too much like a system-reserved word...

echopype/tests/calibrate/test_calibrate_ek80_CW.py
```diff
@@ -0,0 +1,371 @@
...
+        atol_db=_SV_ATOL_DB,
+        skip_range_samples=_SKIP_RS,
+    )
+
+
+# --- COMPLEX
+_COMPLEX_ZARR_REL = "RV_Svea/ci_subset_D20230626-T235835.zarr"
+
+
+def _load_complex_ref(ek80_ext_path, which: str) -> dict:
```
leewujung: I'd suggest renaming `which` to `pkg_name` for readability. `which` to me feels too much like a system-reserved word...

echopype/calibrate/calibrate_ek.py
leewujung: Also, I am wondering about `.isel(ping_time=0)` -- this assumes that `transmit_duration_nominal` is a constant across all pings within the data file, right? but this is not necessarily true.

echopype/tests/calibrate/test_calibrate_ek80_CW.py
LOCEANlloydizard: done!

echopype/tests/calibrate/test_calibrate_ek80_CW.py
LOCEANlloydizard: agreed & done!

echopype/tests/calibrate/test_calibrate_ek80_CW.py
LOCEANlloydizard: yeap, thx! done!

echopype/tests/calibrate/test_calibrate_ek80_CW.py
LOCEANlloydizard: yes i agree! 
I replaced global tolerances with package and encoding specific tolerances for both Sv and tau_effective!
I've adapted the values based on observed max differences.

For the _SKIP_RS is to exclude near-surface bins because echopype vs references packages handle TVG / surface differently, so echopype has NaNs in shallow bins. I've set it to 50 (the test still pass), if I go below some test start to fail!

echopype/tests/calibrate/test_calibrate_ek80_CW.py
LOCEANlloydizard: yep, it was mainly a defensive check.. I’ve simplified the test and removed the xr.align call!

echopype/tests/calibrate/test_calibrate_ek80_CW.py
leewujung: Got it, thanks!

echopype/tests/calibrate/test_calibrate_ek80_CW.py
LOCEANlloydizard: yeap great point! done!

echopype/tests/calibrate/test_calibrate_ek80_CW.py
LOCEANlloydizard: for now i've i kept it because PyEcholab outputs have different range_sample lengths across frequencies! so this allows us to stack them into a common shape, but happy to refactor the test if we prefer a per frequency comparison instead? (or trim echopype/matecho)

echopype/calibrate/calibrate_ek.py
LOCEANlloydizard: * no I haven’t seen a file where tau_effective computation fails yet but it was mainly defensive to catch it if it happens.. can remove!
* for “power path” I meant the power encoding mode => I’ve updated with "encoding mode"
* For the last point, I haven’t changed it yet but we can discuss it! I just replicated the behavior from the complex path for consistency

`tau_effective[ch_GPT] = self.beam["transmit_duration_nominal"][ch_GPT].isel(ping_time=0)`

leewujung: I suspect doing `tau_effective[ch_GPT] = self.beam["transmit_duration_nominal"][ch_GPT]` without selecting the first ping_time will go through since previously without computing the specific `tau_effective`, the computation of `CSv` already uses the full `self.beam["transmit_duration_nominal"]`. I'll create an issue to track this so that we fix it soon.

leewujung: Tracked in https://github.com/OSOceanAcoustics/echopype/issues/1655

leewujung (approved):

PR #1624, opened 2026-03-12
fix: prevent crashes on zero-length dimensions in echodata conversion

beatfactor:
Partial/truncated raw files can produce channels with zero pings, causing:
- Division by zero in set_zarr_encodings() chunk calculation
- ValueError in _nan_timestamp_handler() on empty ping_time arrays
- Encoding failure on all-NaN time variables

Changes:
- utils/coding.py: Skip all-NaN time vars in set_time_encodings(); handle zero-length and scalar dims in set_zarr_encodings()
- convert/set_groups_base.py: Return [np.nan] cleanly when ping_time arrays are empty in _nan_timestamp_handler()


LOCEANlloydizard:
Hi @beatfactor, sorry for the delay... I can review this PR and https://github.com/echostack-org/echopype/pull/1635 this week, if that works for you? If so, could you also take a look at the pre-commit failures and resolve the merge conflicts? I'd be happy to help, but I can't push to those PRs... cheers!

beatfactor:
Sure, I'll take a look over the next few days.

LOCEANlloydizard:
Hi @beatfactor, i'm adding this for the next release, due at the end of the month! cheers!

LOCEANlloydizard (changes requested):
Hey @beatfactor looks good to me! Could we/should we just add a few regression tests covering the zero-length/all-NaN cases fixed here? I would do it myself, but I can’t push directly to the PR.. (maybe we could enable the “Allow edits from maintainers” option if not?) cheers!

LOCEANlloydizard (approved):
thx @beatfactor for this and for adding the tests too! i'll merge this now, thx!

PR #1627, opened 2026-03-12
fix: deduplicate time values in add_location instead of raising ValueError

### LINKED ISSUE #1478: add_location() fails if times aren't unique
I have quite a few EK80 files that appear to have duplicate timestamps. Thus, when I try to run the following...

```
# Open and convert the raw file
ed = ep.open_raw(os.path.join(raw_file_dir,raw_file), sonar_model='EK80')
ed.to_zarr(save_path=converted_file, overwrite=True)
ed.to_netcdf(save_path=converted_file.replace('.zarr','.nc'), overwrite=True)

# ed = ep.open_converted(converted_file)
# ed = ep.open_converted(converted_file.replace('.zarr','.nc'))

# Calibrate the converted file
ds_Sv = ep.calibrate.compute_Sv(ed, waveform_mode="CW", encode_mode="complex") 
ds_Sv.to_zarr(calibrated_file, mode="w")
ds_Sv.to_netcdf(calibrated_file.replace('.zarr','.nc'), mode="w")

ds_Sv_new = ep.consolidate.add_location(ds_Sv, ed)
```

I get the following error.  

> ValueError: The ``echodata["Platform"]["time1"]`` array contains duplicate values. Downstream interpolation on the position variables requires unique time values.

I've been able to confirm that there are repeated time steps using...

```
time_diff = ed['Platform'].time1.diff(dim='time1')
time_diff.plot()
is_unique = ed['Platform'].time1.to_index().is_unique
print(f"Platform.time1 is unique: {is_unique}")
```

This seems to be an issue with almost every file I've spot checked.  Does anyone have suggestions on how to deal with this?  Not sure if it's a bug or merely a data cleaning issue.

beatfactor:
Fixes #1478. EK80 data commonly has duplicate timestamps in Platform time dimensions (e.g., from multiple NMEA sentences — GGA, GLL, RMC — at the same timestamp). Previously add_location() raised a ValueError, requiring manual deduplication before use.

Now duplicate time values are automatically removed (keeping the first occurrence) with a UserWarning, and interpolation proceeds normally.


LOCEANlloydizard (commented):
Hi @beatfactor, nice! thx for the PR! Just small thoughts:

- Should we rename the function to something like check_and_drop_loc_time_dim_duplicates() to better reflect its new behavior?

- For the test, should we also assert that latitude/longitude match the ping_time dimension (e.g. same length)? This would confirm the deduplication path still preserves the expected alignment (although I think it's already tested elsewhere? so not sure)

Cheers!

LOCEANlloydizard (dismissed):
Hi @beatfactor, thx for the changes! 
It looks like the PR accidentally includes a local virtual environment (`venv/`). Could you remove it from the PR? 
There are also pre-commit and RTD errors. I’m not sure whether the RTD one is related to the venv or not? lets see if it still raise after! Cheers

LOCEANlloydizard:
Hi @beatfactor, we discussed this with @leewujung and thought it could be useful to make the warning more explicit when deduplicating timestamps, so it’s clear whether duplicate come from mixed NMEA sentence types or from a single one?

I tried to push this directly to the PR branch but don’t have permission, so sharing here!
So, **prior to** 

```
    # Select NMEA subset (if applicable) and interpolate location variables and place
    # into `interp_ds`.
    for loc_name, interp_loc_name in [(lat_name, "latitude"), (lon_name, "longitude")]:
```

I added

```
# Build contextual warning message for duplicate timestamps.
# In the default NMEA case, multiple sentence types may be mixed,
# which can produce duplicate timestamps due to differing resolution.
extra_msg = ""

if nmea_sentence is None and datagram_type is None:
    sentence_types = np.unique(echodata["Platform"]["sentence_type"].values)
    sentence_types = [str(s) for s in sentence_types]

    if len(sentence_types) > 1:
        extra_msg = (
            f" Multiple NMEA sentence types detected ({', '.join(sentence_types)}), "
            "which may have different resolution and produce duplicate timestamps. "
            "Consider specifying `nmea_sentence` to select a single GPS message type."
        )
    elif len(sentence_types) == 1:
        extra_msg = (
            f" Duplicate timestamps found within NMEA sentence type {sentence_types[0]}."
        )
```

and that message is passed to the warning:

```
warnings.warn(
    f'Dropped {n_total - n_unique} duplicate value(s) in "{time_dim_name}".' + extra_msg,
    UserWarning,
    stacklevel=2,
)
```

That only requires to change the function signature

```
def check_and_drop_loc_time_dim_duplicates(
    da: xr.DataArray,
    time_dim_name: str,
    extra_msg: str = "",
) -> xr.DataArray:
```

and calling

`loc_var = check_and_drop_loc_time_dim_duplicates(loc_var, time_dim_name, extra_msg)`

That it, just suggestions, happy to hear your thoughts on this, or if you’d prefer another approach!
And sorry, I would have preferred to push these modifications directly.. Cheers!

echopype/consolidate/api.py
leewujung:
```suggestion
   Returns
   -------
   The input dataset with location data added

   Notes
   ------
   If duplicated time values are found in the latitude/longitude data, only the first entry is kept 
   for the interpolation operation, and a warning is issued.
   """
```

echopype/consolidate/api.py
```diff
@@ -306,6 +306,26 @@ def add_location(
...
+
+        if len(sentence_types) > 1:
+            extra_msg = (
+                f" Multiple NMEA sentence types detected ({', '.join(sentence_types)}), "
+                "which may have different resolution and produce duplicate timestamps. "
+                "Consider specifying `nmea_sentence` to select a single GPS message type."
+            )
+        elif len(sentence_types) == 1:
+            extra_msg = (
+                f" Duplicate timestamps found within NMEA sentence type {sentence_types[0]}."
```
leewujung:
```suggestion
                f"Duplicate timestamps found within NMEA sentence type {sentence_types[0]}. "
                "Only the first entry with the same timestamp will be used for interpolation."
```

echopype/consolidate/api.py
```diff
@@ -306,6 +306,26 @@ def add_location(
...
+
+    if nmea_sentence is None and datagram_type is None and "sentence_type" in echodata["Platform"]:
+        sentence_types = np.unique(echodata["Platform"]["sentence_type"].values)
+        sentence_types = [str(s) for s in sentence_types]
+
+        if len(sentence_types) > 1:
+            extra_msg = (
+                f" Multiple NMEA sentence types detected ({', '.join(sentence_types)}), "
+                "which may have different resolution and produce duplicate timestamps. "
+                "Consider specifying `nmea_sentence` to select a single GPS message type."
```
leewujung:
```suggestion
                f" Multiple NMEA sentence types detected ({', '.join(sentence_types)}), "
                "which may have different resolution and produce duplicate timestamps. "
                "Consider specifying `nmea_sentence` to select a single GPS message type. "
                "Only the first entry with the same timestamp will be used for interpolation."
```

echopype/consolidate/loc_utils.py
```diff
@@ -107,13 +108,28 @@ def check_loc_vars_validity(
...
-            "Downstream interpolation on the position variables requires unique time values."
+def check_and_drop_loc_time_dim_duplicates(
+    da: xr.DataArray,
+    time_dim_name: str,
+    extra_msg: str = "",
+) -> xr.DataArray:
+    """Check for and drop duplicates in time_dim_name.
+
+    If duplicate time values are found, they are removed (keeping the first
+    occurrence) and a warning is logged. The deduplicated DataArray is returned.
```
leewujung: Removing the comments since they should be in the main function and not here.
Also made suggestions in the main function for the warning messages.
```suggestion
```

leewujung (dismissed):
@beatfactor @LOCEANlloydizard : Thanks for the PR and the review! 

@beatfactor: I made some minor inline suggestions on warning messages and where the docstring appears. Please apply the changes if you agree.

Afterwards I think we can merge this!

leewujung:
@beatfactor - checking in again to see if you can make the above changes. Thanks!

ctuguinay:
Hi @beatfactor, thanks for these changes. Can you pull from the echostack upstream? I think there's something up there that potentially may be missing that's messing up the docs. Also, could you fix the pre-commit problem. It should be something small like whitespace.

ctuguinay (approved):
Looks good, thanks!

Issue #1421, opened 2024-12-29
Migrate to use `pyproject.toml` fully


leewujung:
Right now the project specification is a confusing combination between `pyproject.toml` and `setup.cfg`, plus `requirements.txt`. It'll be great to migrate to using `pyproject.toml` fully. 

Some resources from a quick look:
- dynamic versioning https://github.com/pypa/setuptools-scm
- https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#python-requirement
- https://github.com/pydata/xarray/blob/main/pyproject.toml
- https://github.com/scipy/scipy/blob/main/pyproject.toml
- https://docs.xarray.dev/en/stable/contributing.html#creating-a-python-environment

oftfrfbf:
Not sure if this is within the scope of this issues' intent but just want to give a shout out to the "uv" package and project manager as a possible solution.

[https://docs.astral.sh/uv/](https://docs.astral.sh/uv/)

Just as a comparison to the more traditional "conda" and "pip" managers I am enjoying using 'uv' so far for personal projects — support was just added with Pycharm's latest updates. The biggest feature is fast package management (it's written in Rust) so it has shaved a lot of time off my github action run time. It also starts to consolidate most all of the random python fluff back into the pyproject.toml file, so I can define dependencies, dev-dependencies, and test-dependencies as needed in one central file. It can handle the virutal environment. It can handle building/packaging, testing, and test-matrices. There is now a ".lock" file associated with package installation (similar to what NodeJS has), so package specifications are more structured and you aren't accidentally using packages without properly adding them to the project. There are a ton of other features I'm still exploring but I really like it.

All that said, it would be a bit of a lift to make the changes mentioned above.

If the intent is to just move requirements.txt and requirements-dev.txt to the pyproject.toml file I can help with that — just let me know.

leewujung:
@oftfrfbf : uv does look very interesting! But you are right that we may not be able to do this right now... Maybe we can take it with stages -  The first step getting everything into pyproject.toml (this issue), and at a later time support uv?

One thing I haven't looked up on is how to make it such that people can create a regular or dev environment using conda based on pyproject.toml, since we need to include that in the docs. A quick search landed on this: https://github.com/conda/conda/issues/10633 and it's funny that the last comment pointed to `uv`. I saw another thread ended with that too. Any insights here?

Would you be willing to write an issue for using `uv` so that we can re-consider at some point?

oftfrfbf:
[...]

PR #1656, opened 2026-04-18
Fix dev setup issues with pooch data fetch and cache handling

### LINKED ISSUE #1652: Update docs and code for setting up dev environment
The goal of this issue is to update docs and code for setting up a dev environment. 

Let's streamline the sequence presented in the [Development setup](https://echopype.readthedocs.io/en/latest/contrib_setup.html) so it is easy to follow.

Below are items I ran into when trying to go through the current sequence:
- `pooch` is not found in the dev environment
  - may be good to consolidate all dependencies into `pyproject.toml` to enable just `pip install -e ".[dev]"` or something similar
- `get_pooch_data_path` errors out since cache does not exist in local path

LOCEANlloydizard:
ok had a look & these small fixes should address the issues encountered in https://github.com/OSOceanAcoustics/echopype/issues/1652! I was able to install echopype from scratch and run the test suite locally, so should be good!

- Add pooch to dev requirements
- Ensure pooch cache directory is created if missing
- Improve data fetch (retry + timeout)
- Document USE_POOCH setup for local environments

@leewujung did not migrate dependencies to pyproject.toml as it would require broader refactoring, but this is trackded as an issue anyway in #1421 
Cheers!

Fix #1652

### DIFF
--- a/echopype/tests/conftest.py
+++ b/echopype/tests/conftest.py
@@ -69,8 +69,10 @@ if os.getenv("USE_POOCH") == "True" and os.getenv("PYTEST_XDIST_WORKER") is None
         base_url=base,
         version=ver,
         registry=registry,
-        retry_if_failed=1,
+        retry_if_failed=5,
     )
+
+    downloader = pooch.HTTPDownloader(timeout=120)
 
     def _unpack(fname, action, pooch_instance):
         z = Path(fname)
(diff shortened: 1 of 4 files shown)


echopype/tests/conftest.py
```diff
@@ -69,9 +69,11 @@
         base_url=base,
         version=ver,
         registry=registry,
-        retry_if_failed=1,
+        retry_if_failed=5,
     )
 
+    downloader = pooch.HTTPDownloader(timeout=300)
```
leewujung, line 75: Hey @LOCEANlloydizard : I am curious - is this a safety measure because you saw whole bunch of failed attempts?

LOCEANlloydizard: yes, after I deleted the local cache directory, it started downloading again and I got several timeout errors! But I mostly saw this locally, not on CI (if I remember correctly). Maybe it's because the asset files are getting too large..?

In the long term, maybe we could reduce the number of retries and the timeout? or maybe that altogether?

PR #1657, opened 2026-04-18
Change compressor (zarr v2) to compressors (zarr v3)

leewujung:
This is a bug surfaced in running the krill_freq_diff.pynb notebook on echopype-examples: https://github.com/OSOceanAcoustics/echopype-examples/issues/92

It turns out Zarr v3 supports condec chaining so instead of compressor, it uses compressors (plural)! The example is given in the Zarr documentation: https://docs.xarray.dev/en/stable/user-guide/io.html#zarr-compressors-and-filters

I also found this test is not doing anything. It might be something that was functioning but then something got changed so the test is no longer needed -- @LOCEANlloydizard if you have a moment could you confirm I am not mistaking, that the `lazy_encodings` is indeed empty? If so, maybe we drop it?

@ctuguinay : I think this is why your temp fix to do `compress=False` worked in https://github.com/OSOceanAcoustics/echopype-examples/pull/89, but it's not the root of the problem.

### DIFF
--- a/echopype/tests/echodata/test_echodata_combine.py
+++ b/echopype/tests/echodata/test_echodata_combine.py
@@ -402,7 +402,8 @@ def test_combined_encodings(ek60_test_data):
 
     combined = echopype.combine_echodata(eds)
 
-    encodings_to_drop = {'chunks', 'preferred_chunks', 'compressor', 'filters'}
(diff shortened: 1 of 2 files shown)


leewujung:
Looks like this solves the reported problem! I'll go ahead to merge this now. Thanks @ctuguinay for checking this. Onward to fix the next warning!

PR #1658, opened 2026-04-19
Change string encoding to object

leewujung:
This PR changes string encoding to object following the recommendation for Zarr v3 [here](https://docs.xarray.dev/en/stable/user-guide/io.html#zarr):
> To store variable length strings, convert them to object arrays first with dtype=object.

and the string data type implementation detail under `string` here:
https://github.com/zarr-developers/zarr-extensions/tree/main/data-types

This is based on warnings raised when running the OOI_eclipse.ipynb notebook in echopype-examples after fixing the compressors issue (#1657):
```
/srv/conda/envs/echopype_20260418_compressors/lib/python3.12/site-packages/zarr/core/dtype/npy/string.py:249: 
UnstableSpecificationWarning: The data type (FixedLengthUTF32(length=39, endianness='little')) does not have a Zarr V3 
specification. That means that the representation of arrays saved with this data type may change without warning in a future 
version of Zarr Python. Arrays stored with this data type may be unreadable by other Zarr libraries. Use this data type at your 
own risk! Check https://github.com/zarr-developers/zarr-extensions/tree/main/data-types for the status of data type 
specifications for Zarr V3.
```

Also addressed in this PR are test failures caused by xarray DataArray not being updated properly with NaN in tests.

### DIFF
--- a/echopype/tests/consolidate/test_add_depth.py
+++ b/echopype/tests/consolidate/test_add_depth.py
@@ -248,7 +248,7 @@ def test_ek_depth_utils_group_variable_NaNs_logger_warnings(caplog, ek80_path):
     ds_Sv = ep.calibrate.compute_Sv(ed, waveform_mode="CW", encode_mode="power")
 
     # Set first index of group variables to NaN
-    ed["Platform"]["water_level"].values = np.nan # Is a scalar
+    ed["Platform"]["water_level"][...] = np.nan # Is a scalar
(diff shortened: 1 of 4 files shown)


leewujung:
@ctuguinay : I am so puzzled by the change of how xarray data arrays are updated. Somehow the assignment you had before just wouldn't work, so I was trying all kinds of weird things...

Issue #1651, opened 2026-04-09
Change `waveform_mode` default vocabulary to "FM" for broadband EK80 data


leewujung:
Right now for broadband data the default vocabulary for `waveform_mode` is "BB". It would be good to change it to "FM" to be consistent with Kongsberg terminology, but continue to allow "BB" as an alias.

LOCEANlloydizard:
[This PR](https://github.com/echostack-org/echopype/pull/1664/changes#diff-2c7fac606e4cfa27932e78d103a89b43b13774afe288176757fa53b7854fb01eR40) adds a deprecation warning for "BB" while maintaining it as a backward-compatible alias. The remaining internal use of "BB" still need to be cleaned up in a future follow-up!

Issue #1034, opened 2023-04-21
Add broadband Sv and TS code


leewujung:
Currently in echopype we have implemented calculation for band-averaged Sv, but not the broadband spectrum. Let's add functions for broadband Sv and TS spectrum. 

This involves the following tasks:
- add broadband spectrum calculation
- determine appropriate API call signature since before we just have `compute_Sv` without indication of whether it is band-averaged or broadband
- revise current API call signature and angle compensation for TS calculation: the current implementation does not compensated for position in the beam, so not true target strength but point scattering strength Sp

leewujung:
Dug up the old discussion when we first made the Sp to TS name change: #582. For some reason at that time we did not discuss beampattern modulation...

Issue #1033, opened 2023-04-21
Test broadband processing against CRIMAC outputs


leewujung:
Currently in echopype we test against pyEcholab outputs for broadband processing. These tests are fine (except the issue discovered in #998), but CRIMAC code is much more simplistic for reproducing the test files and are the reference from the [Andersen et al. 2021](https://arxiv.org/abs/2104.07248) paper. It will be good to switch to test against CRIMAC outputs during the test overhaul.

PR #1664, opened 2026-05-17
Add functions for broadband TS spectrum

LOCEANlloydizard:
- Implement a CRIMAC-equivalent broadband TS(f) processing pipeline, including pulse compression, matched-filter autocorrelation, normalized spectra, frequency-dependent calibration, and beam compensation (#1033)

- Add public compute_TS_spectrum() and compute_Sp() APIs, and extend the calibration framework to support frequency-dependent TS(f) products (#1034)

- Update compute_TS() for EK80 complex data to return a beam-compensated band-averaged TS product, while keeping compute_TS_spectrum() as the frequency-dependent TS(f) implementation (#1034, discussion from #582)

- Return TS(f) as an xarray Dataset indexed by (channel, target_id, frequency) (#1034)

- Deprecate the public "BB" waveform mode alias in favor of "FM" while preserving backward compatibility (#1651)

Refs: #582, #1033, #1034, #1651


leewujung:
Hey @LOCEANlloydizard, I thought of a few other params after our discussion the other day so thought I'd lay them out and see what you think. Feel free to ping for more discussions!
```python
def compute_Sv_spectrum(
    self,
    NFFT: int,
    range_step: float = 0.5,
    window: str = "hann",  # TODO: should this be default to None?
    frequency_resolution: Optional[float] = None,
) -> xr.Dataset:
    """Compute broadband EK80 Sv spectrum, or Sv(f).

    Parameters
    ----------
    NFFT : int
        Number of points in the FFT window for computing spectrum.
    range_step : float, default 0.5
        Range spacing in meters for the output Sv(f) window centres.
    window : str, default "hann"
        Type of window function to apply to each FFT segment.
        Options include "hann", "hamming", etc.
    frequency_resolution : optional[float], default None
        Frequency spacing in Hz for the output spectral grid.
        When this is provided, the output Sv(f) will be interpolated to this frequency grid.
        If None, the output Sv(f) will be on the original FFT frequency grid.

    Returns
    -------
    xr.Dataset
        A Dataset containing frequency-dependent volume backscattering strength.
    """


def compute_TS_spectrum(
    self,
    point_locations: xr.DataSet,
    NFFT: int,
    split_front: float = 0.25,
    window: str = "hann",
    frequency_resolution: Optional[float] = None,
) -> xr.Dataset:
    """Compute broadband EK80 Sv spectrum, or Sv(f).

    Parameters
    ----------
    point_locations : xr.DataSet
        Locations of the points for which to compute the spectrum.
        This is required to be an xarray.DataSet with dimensions ("target_id") and
        data variables "ping_time" and "depth" for each ``target_id``.
    NFFT : int
        Number of points in the FFT window for computing spectrum.
    split_front : float, default 0.25
        Each echo spectrum is computed from a segment of the complex echo signal.
        This parameter specifies how to identify the segment around the specified point location.
        For example, if split_front=0.25, then 25% the NFFT length will be before the point location,
        and the remaining 75% will be after the point location.
    window : str, default "hann"
        Type of window function to apply to each FFT segment.
        Options include "hann", "hamming", etc.
    frequency_resolution : optional[float], default None
        Frequency spacing in Hz for the output spectral grid.
        When this is provided, the output Sv(f) will be interpolated to this frequency grid.
        If None, the output Sv(f) will be on the original FFT frequency grid.

    Returns
    -------
    xr.Dataset
        A Dataset containing frequency-dependent point backscattering strength.
    """

```

LOCEANlloydizard:
also look if we can close https://github.com/echostack-org/echopype/issues/629

echopype/calibrate/api.py
```diff
@@ -30,9 +32,20 @@ def _compute_cal(
...
+    # Ref: https://github.com/echostack-org/echopype/issues/1651
+    if waveform_mode == "BB":
+        warnings.warn(
+            "'BB' is deprecated and will be removed in a future release. "
+            "Please use 'FM' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    waveform_mode = "BB" if waveform_mode in ("FM", "BB") else waveform_mode
```
leewujung, line 50: This basically makes BB the default instead of FM. I think it'll be better to just wire everything to be based on FM right now, so that when we think it's time to remove support for the BB syntax, we just need to remove this line and the warning.

echopype/calibrate/api.py
```diff
@@ -80,13 +93,27 @@ def _compute_cal_ds(echodata, slice_dict):
...
+        try:
+            method_name = compute_methods[cal_type]
+        except KeyError:
+            raise ValueError(f"Unsupported calibration type: {cal_type}") from None
+
+        compute_method = getattr(cal_obj, method_name, None)
+
+        if compute_method is None:
+            raise ValueError(
+                f"{cal_type} calibration is not supported for " f"{echodata.sonar_model} data."
```
leewujung, line 115:
```suggestion
                f"{cal_type} computation is not supported for " f"{echodata.sonar_model} data."
```

echopype/calibrate/api.py
```diff
@@ -80,13 +93,27 @@ def _compute_cal_ds(echodata, slice_dict):
...
-        # Perform calibration
-        if cal_type == "Sv":
-            cal_ds = cal_obj.compute_Sv()
-        else:
-            cal_ds = cal_obj.compute_TS()
+        compute_methods = {
+            "Sp": "compute_Sp",
+            "TS": "compute_TS",
+            "Sv": "compute_Sv",
+            # add Sp_spectrum??
```
leewujung, line 102: Yes I think there should be `compute_Sp_spectrum` and `compute_Sv_spectrum`. 

**updates**: from our chat sounds like they will get added after the single target detection PR gets updated (since it's tied to this one in some specific aspects)?

echopype/calibrate/api.py
```diff
@@ -161,7 +188,10 @@ def _compute_cal_ds(echodata, slice_dict):
 
                         # Calibrate and drop filter_time
                         cal_ds_iteration = _compute_cal_ds(echodata, slice_dict)
-                        cal_ds_list.append(cal_ds_iteration.drop_vars("filter_time"))
+                        if "filter_time" in cal_ds_iteration:
+                            cal_ds_iteration = cal_ds_iteration.drop_vars("filter_time")
```
leewujung, line 195: What are the cases that `cal_ds_iteration` would not have `filter_time`? I have forgotten since I review this section last time, but is it from dataset where there is only 1 `filter_time`?

leewujung: Maybe this is related to #1653?

echopype/calibrate/api.py
```diff
@@ -201,12 +231,27 @@ def _add_attrs(cal_type, ds):
...
         ds[cal_type].attrs = {
             "long_name": {
-                "Sv": "Volume backscattering strength (Sv re 1 m-1)",
+                "Sp": "Point scattering strength (Sp re 1 m^2)",
                 "TS": "Target strength (TS re 1 m^2)",
+                "Sv": "Volume backscattering strength (Sv re 1 m-1)",
+                "TS_spectrum": "Frequency-dependent target strength spectrum (TS(f) re 1 m^2)",
             }[cal_type],
             "units": "dB",
         }
```
leewujung, line 260: I think it'll be nice for us to have a consolidated place to specify all the attributes for variables. Right now we define these in an ad-hoc way and missed many of them. It seems would be a good idea to just have a YAML or similar to list the corresponding long_name, units, and other attributes for specific variables, and in functions that generates organized datasets, always pass the dataset through a function that calls the YAML to add attributes -- so like a decorator? I'll add an issue for this.

leewujung: tracked in #1747.

echopype/calibrate/api.py
```diff
@@ -345,6 +390,29 @@ def compute_Sv(echodata: EchoData, **kwargs) -> xr.Dataset:
...
+    raise NotImplementedError("compute_Sv_spectrum is not yet implemented.")
+
+
+def compute_Sp(echodata: EchoData, **kwargs) -> xr.Dataset:
+    """
+    Compute point scattering strength (Sp) from raw data.
+
+    For CW data, Sp is computed from received power samples on the range grid.
+    For EK80 broadband/FM complex data, Sp is computed after pulse compression
+    and represents a band-averaged point-scattering-strength echogram.
```
leewujung, line 414: I am not exactly sure what you mean between "on the range grid" vs "point-scattering-strength echogram". Aren't both still on the range grid, either it's the original one or the one after pulse compression?

echopype/calibrate/api.py
```diff
@@ -82,13 +95,27 @@ def _compute_cal_ds(echodata, slice_dict):
...
+        try:
+            method_name = compute_methods[cal_type]
+        except KeyError:
+            raise ValueError(f"Unsupported calibration type: {cal_type}") from None
+
+        compute_method = getattr(cal_obj, method_name, None)
+
+        if compute_method is None:
+            raise ValueError(
+                f"{cal_type} calibration is not supported for " f"{echodata.sonar_model} data."
```
leewujung, line 115: I am not following the logic here - wouldn't it be better to just do a check if `cal_type` exists in `compute_methods` instead of have it error out? And also `from None` suppresses the trace back (I just learned this!) - is this intentional?

And then there's a second `getattr` check that seems to have a very similar intention.

Maybe these can be consolidated to something simpler, that checks if `cal_type` is not None nor valid, and if otherwise spit out clear error messages?

echopype/calibrate/api.py
leewujung: one more take on this:
since `_add_attrs` is only used once and these vars are "under our control," wouldn't it be better to expect that ALL of these vars would exist so should error out when they don't exist (and therefore there shouldn't be if-else statement)? Or, if some of these only exist for some cal_type, I think it would be better to have these assignments to go as expected behaviors instead of if-else.

echopype/calibrate/ek80_complex.py
```diff
@@ -9,6 +9,272 @@
...
+        phi = np.rad2deg(np.arcsin(np.arctan2(np.imag(y_phi), np.real(y_phi)) / gamma_athwartship))
+    else:
+        raise NotImplementedError(
+            f"Split-beam angle calculation is not implemented for {pc.sizes['beam']} sectors."
+        )
+
+    theta.name = "angle_alongship"
+    phi.name = "angle_athwartship"
+
+    return theta, phi
```
leewujung, line 63: Maybe these could be combined and refactored with what was implemented in `add_splitbeam_angle` so that we have a single place to maintain for these calculations? Also because the code there handles 3-sector transducers as well from a recent PR.

echopype/calibrate/ek80_complex.py
```diff
@@ -9,6 +9,272 @@
...
+    signal: xr.DataArray,
+) -> xr.DataArray:
+    """Average complex signal over transducer sectors.
+
+    Equivalent to CRIMAC ``calcAverageSignal``.
+    """
+    out = signal.mean(dim="beam")
+    out.name = "average_signal"
+
+    return out
```
leewujung, line 264: do we need to keep this as a function? Seems too small and easier to keep track of if the .mean operation is used directly instead.

echopype/calibrate/ek80_complex.py
```diff
@@ -9,6 +9,272 @@
...
+    Equivalent to CRIMAC ``calcPulseCompressedSignals``.
+    """
+    pc = compress_pulse(
+        backscatter=beam["backscatter_r"] + 1j * beam["backscatter_i"],
+        chirp=matched_filter,
+    )
+    pc = pc / get_norm_fac(chirp=matched_filter)
+    pc.name = "pulse_compressed_signal"
+
+    return pc
```
leewujung, line 251: The same operation is used in `split_beam_angle.get_angle_complex_samples()` - maybe use this function there too. This would be part of the merging/refactoring to have a single place to host split-beam angle calculation code as commented [here](https://github.com/echostack-org/echopype/pull/1664/changes#r4066427662).

echopype/calibrate/ek80_complex.py
leewujung: Just to clarify: I saw your notes under `_get_beam_compensated_gain` and the differences. What I meant is that the underlying functions can be merged, so that there is only 1 copy of the operation for split beam angle calculation from complex signals. The additional compensation related to angle offset can be a flag and hence optional for the `add_splitbeam_angle` function, for example.

echopype/calibrate/calibrate_ek.py
```diff
@@ -658,53 +719,570 @@ def _cal_complex_samples(self, cal_type: str) -> xr.Dataset:
...
+        )
+
+        beam_correction_db = self._get_beam_correction(
+            theta=theta,
+            phi=phi,
+            angle_offset_alongship=angle_offset_alongship,
+            angle_offset_athwartship=angle_offset_athwartship,
+            beamwidth_alongship=beamwidth_alongship,
+            beamwidth_athwartship=beamwidth_athwartship,
+        )
```
leewujung, line 842: This part of the code that does the scaling of parameters based on frequency overlaps with what's in `cal_params.get_cal_params_EK()`. Similar to the split-beam angle calculation, maybe it makes sense for us to come up with a module that does the scaling, and just use it in both here and cal parameter preparation?

echopype/calibrate/ek80_complex.py
```diff
@@ -9,6 +9,272 @@
...
+    left_samples = idx_peak_target
+    right_samples = len(pc_target) - idx_peak_target
+
+    idx_start = max(0, idx_peak_auto - left_samples)
+    idx_stop = min(len(mf_auto), idx_peak_auto + right_samples)
+
+    return mf_auto[idx_start:idx_stop]
+
+
+def _extract_target_from_range_gate(
```
leewujung, line 113: I feel this function should probably go to a separate module and not in `ek80_complex.py` since it's not related to the complex data operations. Not exactly sure how to call it.

leewujung: Looking at the ops in this function, I am not sure I agree with this approach to change everything to pure numpy array and then use pure indexing to find the values corresponding with a specific range gate. I think the same operations can be completed in xarray and the intention would be clearer because the dimensions/coordinates are explicit.

echopype/calibrate/calibrate_ek.py
```diff
@@ -658,53 +719,570 @@ def _cal_complex_samples(self, cal_type: str) -> xr.Dataset:
...
+                        theta_raw=theta_raw,
+                        phi_raw=phi_raw,
+                        target_range=target_range,
+                        target_range_min=target_range_min,
+                        target_range_max=target_range_max,
+                        split_front=split_front,
+                        n_fft=n_fft,
+                    )
+                except ValueError:
+                    continue
```
leewujung, line 1010: This is very weird - silent failure without any warning???

echopype/calibrate/calibrate_ek.py
```diff
@@ -658,53 +719,570 @@ def _cal_complex_samples(self, cal_type: str) -> xr.Dataset:
...
+                    mf_auto=mf_auto,
+                    pc_target=pc_target,
+                )
+
+                if mf_auto_red.size < n_fft:
+                    mf_pad = np.zeros(n_fft, dtype=complex)
+                    mf_pad[: mf_auto_red.size] = mf_auto_red
+                    mf_auto_red = mf_pad
+                elif mf_auto_red.size > n_fft:
+                    mf_auto_red = mf_auto_red[:n_fft]
```
leewujung, line 1034: This padding or truncation operation is very important. I think we should have it in the docstring and also the example notebook, to demonstrate what would happen with the different length specified when computing the spectrum.

echopype/calibrate/ek80_complex.py
```diff
@@ -9,6 +9,272 @@
...
+    """Convert normalised TS(f) spectrum to received power spectrum.
+
+    Equivalent to CRIMAC ``calcPowerFreqTS``.
+    """
+    return _compute_complex_power(
+        normalized_spectrum=normalized_spectrum,
+        n_beams=n_beams,
+        z_et=z_et,
+        z_er=z_er,
+    )
```
leewujung, line 211: I think this is over factoring, since `_compute_complex_power` is only used once here. `_compute_complex_power` also overlaps with what's in `_compute_power_from_complex_signal`. I think these should all be combined and refactored to use the same function.

echopype/calibrate/calibrate_ek.py
```diff
@@ -658,53 +719,570 @@ def _cal_complex_samples(self, cal_type: str) -> xr.Dataset:
...
+
+                ts = _compute_ts_spectrum_calibrated(
+                    power_spectrum=power_spectrum,
+                    target_range=target_range,
+                    frequency=frequency,
+                    sound_speed=sound_speed,
+                    absorption_f=absorption_f,
+                    transmit_power=transmit_power,
+                    gain_f=gain_f,
+                )
```
leewujung, line 1066: I'd suggest keeping the "gut" exposed here for readability. The docstring for the function can just become inline comments.

echopype/calibrate/api.py
```diff
@@ -450,3 +518,48 @@ def compute_TS(echodata: EchoData, **kwargs):
...
+        then 25% of the NFFT window is placed before ``target_range`` and the remaining
+        75% after it.
+
+    window : str, tuple, float or None, default None
+        Window passed directly to ``scipy.signal.get_window``. If ``None``,
+        a rectangular/boxcar window is used.
+
+    frequency_resolution : float, optional
+        Desired spacing of the output frequency grid in Hz. Used to define the
+        frequency grid on which TS(f) is evaluated.
```
leewujung, line 558: I think allowing `n_f_points` and `frequency_resolution` is confusing when they are together with `NFFT` and `split_front` -- these two are messy enough. I would suggest that we show people how to do interpolation _outside_ by operating on the output of this function, instead of doing it silently inside the function. Only the selected signal, NFFT, and the related padding really matters, the interpolation is cosmetic. I worry that by giving all these options people could get lost in what is really important for processing the data.

I also suggest `split_front` to be remained as `fraction_before_point_location` since "split" could be confused with split-beam.

echopype/calibrate/calibrate_ek.py
```diff
@@ -658,53 +719,570 @@ def _cal_complex_samples(self, cal_type: str) -> xr.Dataset:
...
+                    "frequency": frequency_ref,
+                },
+            )
+
+            out_by_channel.append(ds_ch)
+
+        if not out_by_channel:
+            raise ValueError("No valid TS_spectrum targets produced.")
+
+        return xr.concat(out_by_channel, dim="channel")
```
leewujung, line 1118: What happens to the `frequency` coordinate (specific to the channel) when you do `xr.concat`? Does it concat the multiple `frequency` vectors sequentially?

echopype/calibrate/calibrate_ek.py
```diff
@@ -658,53 +719,570 @@ def _cal_complex_samples(self, cal_type: str) -> xr.Dataset:
...
+
+                ts_list.append(ts)
+                target_range_list.append(target_range)
+                theta_list.append(theta_t)
+                phi_list.append(phi_t)
+                ping_time_list.append(point_ping_time)
+                target_id_list.append(target_id)
+
+            if not ts_list:
+                continue
```
leewujung, line 1081: Better to not have silent skip like this since the user may end up having an empty dataset and it can be hard to keep track of why.

echopype/calibrate/calibrate_ek.py
```diff
@@ -658,53 +719,570 @@ def _cal_complex_samples(self, cal_type: str) -> xr.Dataset:
...
+                NFFT=NFFT,
+                n_f_points=n_f_points,
+                split_front=split_front,
+                window=window,
+                frequency_resolution=frequency_resolution,
+            )
+
+        raise ValueError(f"Unsupported calibration type: {cal_type}")
+
+    def _compute_cal(self, cal_type, **kwargs) -> xr.Dataset:
```
leewujung, line 1165: I think this was in place before this PR already, but looking back at this I think we can actually do without the `_compute_cal` structure and just dispatch the following 3 functions directly in compute_Sv/Sp/TS/TS_spectrum:
- `_cal_power_samples`
- `_cal_complex_samples`
- `_cal_complex_samples_f`

Is there an advantage in keeping this `_compute_cal` function?

echopype/calibrate/calibrate_ek.py
```diff
@@ -658,53 +719,570 @@ def _cal_complex_samples(self, cal_type: str) -> xr.Dataset:
...
+            )
+
+            out_by_channel.append(ds_ch)
+
+        if not out_by_channel:
+            raise ValueError("No valid TS_spectrum targets produced.")
+
+        return xr.concat(out_by_channel, dim="channel")
+
+    def _cal_complex_samples_f(
```
leewujung, line 1120: Seems like you can just absorb what's in `_cal_complex_samples_TS_spectrum` here? I don't see much advantage of separating these 2. But I will make some comments about what I think could be factor out in `_cal_complex_samples_TS_spectrum`.

echopype/calibrate/ek80_complex.py
```diff
@@ -9,6 +9,272 @@
...
+            f"Split-beam angle calculation is not implemented for {pc.sizes['beam']} sectors."
+        )
+
+    theta.name = "angle_alongship"
+    phi.name = "angle_athwartship"
+
+    return theta, phi
+
+
+def _compute_power_from_complex_signal(
```
leewujung, line 66: This overlaps with `_compute_complex_power`

leewujung (changes requested):
Hey @LOCEANlloydizard : Thanks for the PR! Sorry it took a very long time for me to review this!

Below are a few higher-level things that may make the inline comments easier to read:
- There are some redundancy within the new code and with the old code that I think would benefit from combining and refactoring, such as:
   - frequency-based scaling of calibration parameters
   - split-beam angle estimation between beampattern compensation and `add_splitbeam_angle`
   - computing power from complex signals with the right scaling related to impedance
- I'd suggest removing `n_f_points` and `frequency_resolution` as input arguments when computing spectrum, so that the function output would just be what people would get based on `NFFT` or `target_range_min/max`; any downstream interpolation the users will do outside of the function, so that there's no potential hidden/unintended errors.
- refactoring and commenting code in `_cal_complex_samples_TS_spectrum`
  - it was hard to follow the logic because the code jumps between doing different things -- I think those can be better bunched up to be functions called by this method, to make things easier to follow
  - there is very minimal inline comments in this function -- it would be better to add the intention for the operations and any important caveats as comments
- would it make sense to just always first compute the split-beam angles before estimating the TS? That is the logical step and we can enforce it in the code, so that the `_cal_complex_samples_TS_spectrum` function for example does not need to jump between computing the spectrum and the angles. It'll then just be computing the spectrum and grab the corresponding angles to get the right beampattern compensation.

Maybe we can talk through the function when we meet to make a plan for the above changes? I think it would be great to make this PR a one-time thing to get clean code, since EK80 is stable now and the code likely won't change much later, so it'll be a worthwhile investment.

PR #1676, opened 2026-06-07
check_file_permissions: probe inside target store, not its parent (#1098)

### LINKED ISSUE #1098: Incorrect Path for Temporary File Generation in `check_file_permissions` Method
The check_file_permissions method in the codebase is used to verify file permissions by generating a temporary file in the target path. However, it has been observed that this method is incorrectly creating the temporary file in the parent directory (one level up) of the target path instead of the intended location.

image

File Directory Root :  /Users/soham/Desktop/EchoWorkSpace/echoflow/notebooks/combined_files
Test file generated at :  /Users/soham/Desktop/EchoWorkSpace/echoflow/notebooks/.f0605b17-da68-45e8-95b4-f40281cdca22

gaoflow:
Closes #1098.

## Problem

`check_file_permissions` writes its probe file to the **parent** of the target directory instead of the target itself when given an `FSMap` (the zarr/cloud path). As reported in #1098, for a target of `.../combined_files` the probe was created at `.../` :

```
File Directory Root :  .../notebooks/combined_files
Test file generated at :  .../notebooks/.f0605b17-...
```

This means the permission check validates the wrong directory, and (when parent and target permissions differ) can pass or fail incorrectly.

## Cause

The `FSMap` branch used `base_dir = os.path.dirname(FILE_DIR.root)`, which strips the last path component, so the probe lands one level up. The `Path`/`str` branch right below it correctly writes *inside* the target (and `mkdir(parents=True)` first).

`os.path.dirname` was presumably used because the target store may not exist yet (e.g. a not-yet-written zarr store), and writing inside a missing directory raises — but that's exactly the case the `Path`/`str` branch already handles by creating the directory first.

## Fix

Write the probe inside `FILE_DIR.root` and `makedirs(..., exist_ok=True)` first, mirroring the `Path`/`str` branch. This keeps the check on the actual target and still works when the store doesn't exist yet (it gets created, as the local-path branch already does).

## Tests

Added two regression tests (red before, green after):
- `test_check_file_permissions_fsmap_writes_inside_target` — the probe is opened inside the target store, not its parent.
- `test_check_file_permissions_fsmap_creates_missing_store` — a not-yet-created store is created and passes the check.

Full `test_utils_io.py` passes (47 passed, 2 xfailed).

### DIFF
--- a/echopype/tests/utils/test_utils_io.py
+++ b/echopype/tests/utils/test_utils_io.py
@@ -325,3 +325,41 @@ def test_init_ep_dir(monkeypatch):
     assert echopype.utils.io.ECHOPYPE_DIR.exists() is True
 
     temp_user_dir.cleanup()
+
+
+def test_check_file_permissions_fsmap_writes_inside_target(tmp_path):
+    """check_file_permissions must probe inside the target store, not its parent (#1098)."""
+    from echopype.utils.io import check_file_permissions
+
+    fs = fsspec.filesystem('file')
+    target = tmp_path / 'combined_files'
+    target.mkdir()
+
+    # record where the probe file is opened (it is deleted again, so it can't
+    # be observed after the fact)
+    opened = []
+    real_open = fs.open
+
+    def _record_open(path, *args, **kwargs):
+        opened.append(path)
+        return real_open(path, *args, **kwargs)
+
+    fs.open = _record_open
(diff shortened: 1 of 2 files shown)


LOCEANlloydizard:
Hi @gaoflow, thanks! the fix and tests both look good to me! cheers!

LOCEANlloydizard (approved):

Issue #1682, opened 2026-06-16
frequency_nominal / channel mismatch in compute_MVBS


LOCEANlloydizard:
While testing `compute_MVBS` on an ONC AZFP dataset (device `ASLAZFP55196`), a mismatch was observed (thanks @slonimer!) between channel labels, frequencies, and the resulting echograms. The 38, 120, and 200 kHz channels appear to be shifted in the output MVBS dataset.

Dataset information:

* Source: ONC
* Device: `ASLAZFP55196`
* Data product: `AAPTS`
* Time span: `2026-06-14T00:00:02Z` to `2026-06-14T23:59:57Z`

Input Sv dataset:

```python
channel = ['55196-38-1', '55196-120-2', '55196-200-3']
frequency_nominal = [38000., 120000., 200000.]
```

Output MVBS dataset:

```python
channel = ['55196-38-1', '55196-120-2', '55196-200-3']
frequency_nominal = [120000., 200000., 38000.]
```

The input `ds_Sv` appears correct, suggesting the mismatch may occur during MVBS computation or when `frequency_nominal` is propagated to the output dataset. In this case, `frequency_nominal` is stored as a `(ping_time, channel)` variable in the input dataset, which may be related to the issue.

TODO:
- [x] fix
- [ ] add test in gthub assets
- [ ] add test with this file

### Also check notebooks

`compute_MVBS` is used in several example notebooks. Once this issue is fixed, verify that these notebooks still behave correctly and that channel / `frequency_nominal` metadata remain consistent:

- [ ] Getting started with Echopype
- [ ] Watching a solar eclipse using a moored, upward-looking OOI echosounder
- [ ] Label-aware subselection of echo data based on ship tracks
- [ ] Perform frequency-differencing and masking to compute NASC attributed to krill from ship echosounder data
- [ ] Incorporate an external mask to compute NASC attributed to Pacific hake from ship echosounder data
- [ ] Seafloor detection

PR #1698, opened 2026-06-30
Fix channel-frequency mapping in compute_MVBS

LOCEANlloydizard:
Fixes a channel/frequency mismatch in `compute_MVBS` for datasets where `frequency_nominal` is stored as `(ping_time, channel)` (e.g. AZFP). The output now preserves the channel order returned by `compute_raw_MVBS` and aligns `frequency_nominal` accordingly.  Also added regression tests but I'll keep #1682 open, as I'd like to add a few more tests! (like with AZFP file)

Ref #1682


leewujung (dismissed):
@LOCEANlloydizard : Thanks for this PR! Look great and I think it's ready to be merged. 

The only thing I wonder about is why the `ping_time` dimension shows up in `frequency_nominal`. Is there something that should be fixed upstream?

leewujung (approved):
@LOCEANlloydizard - perfect, thanks for the changes! I'll merge this now.

PR #1702, opened 2026-07-14
Pad parameters with different lengths across channel in vendor group

leewujung:
This PR adds a short code section to pad parameters with different lengths across channel in the vendor group.

This was only encountered so far in an EA640 setup in which the lower freq channels contain a smaller number of pulse length choices than the higher freq channels.


ctuguinay (approved):
This looks good!

PR #1709, opened 2026-07-19
Fix docstring sections that weren't getting into the rendered docs

gavinmacaulay:
Part of the docstrings for `clean/api.py` and `mask/api.py` use unsupported section names (`Common expectations` and `Method-specific arguments`) that are not rendered in the final documentation.

This PR moves the text in those sections to be under sections that are rendered in the documentation.


LOCEANlloydizard (approved):
that's great, thank you @gavinmacaulay; merging now!

PR #1710, opened 2026-07-20
Make the bibliography appear in the generated docs

gavinmacaulay:
The bibliography appears in v0.11.1, but not in latest. I'm not sure what changed in the config to cause that (perhaps the difference between running jupyter-book and sphinx-build?). Anyway, these small changes should make things work again.

[skip-CI]


LOCEANlloydizard (approved):
indeed it fixes it! thx for finding and fixing this @gavinmacaulay ! merging now!

PR #1711, opened 2026-07-20
docs: fix broken contributing page reference

praneethratna:
Fix broken contributing page reference.

The old reference points to a non-existent page.


LOCEANlloydizard (approved):
that's great, thx you for fixing this @praneethratna!!

Issue #494, opened 2021-11-17
Exception of 'transceivers' when processing lots of files


oftfrfbf:
I am encountering a lot of 'transceivers' exceptions as I spot check EK60 data. From 286 different cruises I have found ~60 cruises with this specific exception. I am not sure how similar each is to one another but below is an example of one error.

Code:
```python
import echopype as ep
print(ep.__version__) # 0.5.5.dev49+g8180da1

raw_file_s3path = f"s3://noaa-wcsd-pds/data/raw/Falkor/FK180415/EK60/D20180416-T073530.raw"
try:
	ed = ep.open_raw(raw_file_s3path, sonar_model='EK60', storage_options={'anon': True})
	ds_Sv = ep.calibrate.compute_Sv(ed)
except Exception as e:
	print(f"Problem with: {raw_file_s3path}, {e}")
```
Output:
```python
23:48:59  parsing file D20180416-T073530.raw, time of first ping: 2018-Apr-16 07:35:30
Problem with: s3://noaa-wcsd-pds/data/raw/Falkor/FK180415/EK60/D20180416-T073530.raw, 'transceivers'
```

https://github.com/OSOceanAcoustics/echopype/blob/1548f46cf7f163188b2858e9a6838290fd25b7ec/echopype/convert/set_groups_ek60.py#L20

In the "set_groups_ek60.py" the parser_obj.config_datagram has no "transceivers" key to reference. The config_datagram:
```
{
	'type': 'XML0',
	'low_date': 2127786896,
	'high_date': 30659925,
	'timestamp': numpy.datetime64('2018-04-16T07:35:30.059'),
	'bytes_read': 11104,
	'subtype': 'configuration',
	'configuration': {
		'GPT 0090720674d4-1 ES18': {
			'transceiver_name': 'GPT 0090720674d4',
			'ethernet_address': '0090720674d4',
			'ip_address': '157.237.15.3',
			'transceiver_version': 'GPT-Q18(2)-S 1.0 0090720674d4',
			'transceiver_software_version': '070413',
			'transceiver_number': 1,
			'market_segment': 'Scientific',
			'transceiver_type': 'GPT',
			'serial_number': '0',
			'impedance': 1000,
			'channel_id': 'GPT 0090720674d4-1 ES18',
			'channel_id_short': 'Falkor ES18',
			'max_tx_power_transceiver': 2000,
			'pulse_duration': [0.000512, 0.001024, 0.002048, 0.004096, 0.008192],
[...]

PR #1724, opened 2026-07-21
Add first datagram type check for sonar_model EK60 vs EK80

leewujung:
This addresses #494 by checking if the first datagram is of the right type as expected from the user-provided `sonar_model`:
- EK60 and related sonar model should have "CON0"
- EL80 and related sonar model should have "XML0"

Also added tests for when the specification is reversed.


LOCEANlloydizard (approved):
all good thx @leewujung !!

PR #1726, opened 2026-07-21
Add usage to `coerce_increasing_time` docstring

### LINKED ISSUE #1572: `exist_reversed_time` before `coerce_increasing_time`
Dear Echopype Team,

I had an issue with `coerce_increasing_time` because I thought it will check for `exist_reversed_time` internally. Because in the docs is written "if... necessary,..."

> If coercion is necessary, the input ds will be directly modified.

I always got the error "index 0 is out of bounds for axis 0 with size 0", and I wondered where this came from since I figured out, that this `neg_idx = np.argwhere(time_old_diff < np.timedelta64(0, "ns")).flatten()` index seems to have size 0 and therefore the line `c_diff = np.cumsum(time_old_diff[neg_idx[0] :], axis=0)` throws this error. 

My question is now, if this is the desired behavior, so that the framework users should always use the function like this:

```py
if eqc.exist_reversed_time(ds, "ping_time"):
    eqc.coerce_increasing_time(ds, "ping_time")
```

But then we should either mention that in the docs of `coerce_increasing_time` or we just add the check there in the frameworks code.

I can make the change, after we decided for a solution.

Thanks, Jan

leewujung:
This addresses #1572 by adding usage guide to the docstring as suggested by @jmeischner.

### DIFF
--- a/echopype/qc/api.py
+++ b/echopype/qc/api.py
@@ -52,6 +52,14 @@ def coerce_increasing_time(
         length of the local window before the reversed timestamp within which
         the median pinging interval is used to infer the next ping time
 
+    Usage
+    -----
+    >>> import xarray as xr
+    >>> from echopype.qc.api import exist_reversed_time, coerce_increasing_time
+    >>> ds = xr.open_dataset("my_dataset.nc")
+    >>> if exist_reversed_time(ds, "ping_time"):  # check for reversed time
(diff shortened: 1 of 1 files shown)


LOCEANlloydizard (approved):
looking great! thx @leewujung merging this now!

PR #1731, opened 2026-07-22
Resolve some warnings during pytest

gavinmacaulay:
Resolves or configures pytest to ignore warnings generated when running pytest.

Most of the ignored warnings are from dependency libraries that need to update themselves (i.e., echopype can't resolve them). One exclusion is probably addressable by a change in echopype, but it generates so many warnings that it obscures the less frequent ones.

The warnings that remain are now much more visible in the output that pytest produces.


LOCEANlloydizard (approved):
Thx @gavinmacaulay for tackling those warnings, this is really great!
For the filters, I think it still makes sense to have warnings show up in CI, but filtered so that the output remains readable? I went ahead and changed the filters to "once" so that we still see each warning without flooding the logs, what do you think?
One thing I think could be a nice follow-up is to add a manually triggered workflow that runs the test suite with all warnings enabled and uploads the full warning log as an artifact: that way whenever we want to work on checking/cleaning up warnings, we can easily inspect the complete list without affecting the readability of the regular CI?

I'll merge this once the tests pass, but happy to hear your thoughts and update/discuss things if needed!

gavinmacaulay:
@LOCEANlloydizard - thanks - I wasn't aware of 'once' - that's a good option!

On having a manually-triggered full warnings workflow - I wonder if simply documenting how to override the 'once' option on the pytest command line would suffice and developers can then run the tests locally to see the full warnings. The feedback loop would be much quicker that way rather than going via a workflow.

LOCEANlloydizard:
right! I noted it so we can indeed add it to the documentation, cheers!

Issue #1721, opened 2026-07-21
Alternative EchoData Beam_group structure for data from different channel/waveform


leewujung:
Right now we put complex or power-angle data from different channels (or waveforms) into the same Beam_group and pad NaNs for channels that are not transmitting or with less number of samples along the `range_sample` dimension. The advantage of this is that it makes easier to slice across different data dimensions, but now that we have a lot more complex sample datasets and the various ways to sequence different waveforms, this NaN-padding scheme becomes very inefficient. 

An alternative way is to store data from different waveforms (ping library entries) into different Beam_groups, so that there is no _cross_-channel `ping_time` padding that is required (though _within_-channel padding with shorter range would still happen). In addition to storage efficiency, this will also make finding the right matching coefficients during `compute_Sv` way more efficient: right now it needs to find the match on a ping-by-ping level, but if we store each waveform in a separate Beam_group, we just need to index the Beam_groups.

This will require a larger change to the codebase, but I think it's worth doing now that we can have a diversity of transmission schemes.

PR #1734, opened 2026-07-22
Support frequency nominal in consolidate functions

### LINKED ISSUE #1488: Support dimension alignment with `frequency_nominal` for all `ep.consolidate` functions

ctuguinay:
Additionally enforces that channel or frequency nominal exists in all datasets passed into the consolidate functions.

This is a slight modification of #1520 by @dbashford-NOAA without code conflicts with the more recent changes.

When merged, it closes #1488.


ctuguinay:
@LOCEANlloydizard @leewujung I think this is ready to review. I think I addressed all the comments from the previous PR, but let me know if I didn't catch something.

echopype/consolidate/api.py
```diff
@@ -205,8 +212,9 @@ def add_depth(
...
             echo_range_scaling = ek_use_platform_angles(echodata["Platform"], ds["ping_time"])
         elif use_beam_angles:
-            # Identify beam group name by checking channel values of `ds`
-            if echodata["Sonar/Beam_group1"]["channel"].equals(ds["channel"]):
+            dim_0 = get_dim_0(ds)
+            # Identify beam group name by checking `dim_0` values of `ds`
+            if echodata["Sonar/Beam_group1"][dim_0].equals(ds[dim_0]):
                 beam_group_name = "Beam_group1"
             else:
                 beam_group_name = "Beam_group2"
```
leewujung: I think this logic would no longer work with the ping sequencing capability on WBAT, since the same channel can be configured to transmit both FM and CW signals and record in either complex or power-angle format. On WBT this would be fine since each ping library entry will produce a _1, _2, etc postfix as the `channel_id`.

Is there another way we can find the right channel beam directions from the echodata object?

How about using the `channel_all` and `beam_group_index` variables in the `Sonar` group that we added to help index the beam group to calibrate?

echopype/consolidate/api.py
```diff
@@ -490,12 +504,8 @@ def add_splitbeam_angle(
...
     ed_beam_group = retrieve_correct_beam_group(echodata, waveform_mode, encode_mode)
 
-    # check that source_Sv at least has a channel dimension
-    if "channel" not in source_Sv.variables:
-        raise ValueError("The input source_Sv Dataset must have a channel dimension!")
-
-    # Select ds_beam channels from source_Sv
-    ds_beam = echodata[ed_beam_group].sel(channel=source_Sv["channel"].values)
+    # Select beam group based on dim_0
+    ds_beam = echodata[ed_beam_group].sel({dim_0: source_Sv[dim_0].values})
```
leewujung: Same as [my comment above](https://github.com/echostack-org/echopype/pull/1734/changes#r3640656894), this logic would not work and needs to be updated.

leewujung (changes requested):
Hey @ctuguinay : Thanks for sorting through the previous PR and consolidating the changes to this new one! I know you didn't write the previous code but please help with the following requested changes.

I see a few `.sel({dim_0: dim_0_id}` usages which try to use whatever `dim_0` is (either `frequency_nominal` or `channel`) to slice datasets or dataarrays. I think a better way is to use `get_dim_0` to find out what `dim_0` is, and if it is `frequency_nominal`, map it back to `channel` using what's in `Sonar/channel_all`, so that all the indexing is done using `channel`. This is because we align pretty everything in our code to be the channel_if, and in view of #1721, it's better to just keep all indexing using `channel` to future proof the code.

I also suggest the following to keep things a bit tidier:
- rename `loc_utils.py` to `utils_loc.py`
- move `get_dim_0` to `utils_loc.py`
- rename `ek_depth_utils.py` to `utils._ek_depth.py`

One of [my comments](https://github.com/echostack-org/echopype/pull/1734/changes#r3640656894) requires a bit of consideration to figure out a good solution. Let me know if you want to bounce ideas.

ctuguinay:
@leewujung This should be ready for another look :pray:

PR #1738, opened 2026-07-24
chore: group EK60 and EK80 model families

### LINKED ISSUE #1729: Consolidate handling of sonar_model types
Currently we check sonar_model in various places in the code in the form of:
```python
if sonar_model in ["EK60", "ES60", "ES70"]:
    ...
elif sonar_model in ["EK80", "ES80", "EA640"]:
    ...
```
This is pretty inefficient and prone to error. Now that there is a new AZFP6 sonar_model we should really clean this up.

Let's consolidate these to a dictionary in `core.py` to have the following groupings:
- EX60: EK60, ES60, ES70
- EX80: EK80, ES80, EA640

This issue substitutes #995.

praneethratna:
Closes https://github.com/echostack-org/echopype/issues/1729.

Groups `EK60` and `EK80` model into different families to avoid `if else` across the code.

### DIFF
--- a/echopype/calibrate/range.py
+++ b/echopype/calibrate/range.py
@@ -2,6 +2,7 @@ from typing import Dict
 
 import xarray as xr
 
+from ..core import SONAR_MODELS
 from ..echodata import EchoData
 from .env_params import harmonize_env_param_time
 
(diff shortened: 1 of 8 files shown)


echopype/core.py
```diff
@@ -66,6 +67,7 @@ def inner(test_ext: str):
         "parser": ParseULS6,
         "parsed2zarr": None,
         "set_groups": SetGroupsAZFP6,
+        "family": "AZFP6",
     },
```
leewujung, line 71: We could put "AZFP" and "AZFP6" both in the "AZFP" family - it is different data format versions of the same instrument.

echopype/calibrate/range.py
```diff
@@ -119,17 +120,14 @@ def compute_range_EK(
...
-        raise ValueError("The specified sonar_model is not supported!")
+    model_family = SONAR_MODELS[sonar_model]["family"]
+    if model_family not in ["EX60", "EX80"]:
+        raise ValueError(f"The specified sonar_model {sonar_model} is not supported!")
 
     if "sound_speed" not in env_params:
         raise RuntimeError(
             "sounds_speed not included in env_params, "
-            f"use echopype.calibrate.env_params.get_env_params_{ek_str}() to compute env_params "
+            "use echopype.calibrate.env_params.get_env_params_EK() to compute env_params "
```
leewujung, line 130: Thanks for catching the bug that would've resulted into errors due to deprecation `get_env_params_EK60` and `get_env_params_EK80`

leewujung (dismissed):
Hey @praneethratna : Thank you so much for the PR! This is pretty much ready to go! A tiny thing (which was my bad in my original description) -- could you change EX60 to Ex60 and EX80 to Ex80? Just wanted to be consistent with the manufacturer usage. Also it's somehow easier to read the difference since it's lower case. Thanks!

echopype/core.py
praneethratna: Updated.

leewujung (approved):
Thanks! merging now!

PR #1740, opened 2026-07-27
chore: refactor log package and add support for package level verbosity

### LINKED ISSUE #1722: Echopype logger verbose logic is flipped
Currently the Echopype verbosity logic is flipped. In many parts of the test suite, to turn on logger verbosity we have:

```
# Turn on logger verbosity
ep.utils.log.verbose(override=False)
```

But the docstring field for `log.verbose` has:

```
    override: bool
        Boolean flag to override verbosity,
        which turns off verbosity if the value is `False`.
        Default is `False`.
```

so the logic needs to be reworked.

praneethratna:
Closes https://github.com/echostack-org/echopype/issues/1722.

Refactors `log` package and adds support for package level verbosity.

### DIFF
--- a/echopype/__init__.py
+++ b/echopype/__init__.py
@@ -10,7 +10,7 @@ from .utils.io import init_ep_dir
 from .utils.log import verbose
 
 # Turn off verbosity for echopype
-verbose(override=True)
+verbose(override=False)
 
 init_ep_dir()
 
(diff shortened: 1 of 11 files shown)


LOCEANlloydizard (approved):
Hi @praneethratna thx for working on this, it looks great! I'm going to merge it now.

Just adding a broader thought here: the current logging system was originally introduced mainly to replace scattered print() statements with a more flexible verbose() interface. Following this PR (and related issues) we've started discussing whether Echopype should keep maintaining its own custom logging system at all.. For instance, Echodataflow already relies on Prefect's logging, so we're wondering whether it might make sense to simplify Echopype by relying more on warnings and exceptions for library level, while leaving workflow logging to downstream applications?

it's just an open discussion but I wanted to mention it since the logging system itself may evolve (or even eventually be deprecated?) depending on where those discussions lead..
and thanks again!!

PR #1756, opened 2026-08-04
enh: add support for parsing and storing specific channels

### LINKED ISSUE #1720: Allow selecting channels to parse and store with `open_raw`
Right now `open_raw` parses and stores all datagrams from all channels. It would be nice to have an optional argument to specify a subset of channels to parse and store based on `channel_id`. This would make some processing workflows much more efficient, especially if the data is stored by WBT* (not WBAT**) so each different ping library specification has its own `channel_id` in the CW-FM alternating configuration.

This means that users need to know the `channel_id` they want to parse ahead of time. They could get that by parsing 1 file first and check, or they have that already from their echosounder settings.

\* WBT writes data collected from the same channel but with different ping library specification as `channel_id_1`, `channel_id_2`, etc
\** WBAT writes data collected from the same channel with the same `channel_id` regardless of the ping library specification

praneethratna:
Closes https://github.com/echostack-org/echopype/issues/1720.

This PR adds support for parsing and storing specific channels for Ex60 and Ex80 model families.

### DIFF
--- a/echopype/convert/api.py
+++ b/echopype/convert/api.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import TYPE_CHECKING, Dict, Literal, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Dict, List, Literal, Optional, Tuple, Union
 
 import fsspec
 from xarray import DataTree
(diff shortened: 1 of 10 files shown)


LOCEANlloydizard (approved):
thx @praneethratna this is really great! I don’t see anything blocking, so I’ll merge this now! I also opened [a PR in echopype-examples](https://github.com/echostack-org/echopype-examples/pull/116) to demonstrate how to use the new channels argument in the Getting Started notebook. cheers!

PR #1759, opened 2026-08-07
Resolve some doc build warnings [skip ci]

gavinmacaulay:
When the docs are built there are some warnings. This PR resolves some of them:

- There is a section in the docs for the visualize functionality, but that code was removed from echopype a few versions ago
- Headings in api.rst and api.py didn't have the right number of underscore characters
- A docstring list in combine.py didn't render correctly because the list items weren't indented
- Removed an inline toc as it is already provided via the right-hand toc 

A note - these warnings can adversely affect the generated documentation but they aren't very visible during the CI runs. I only noticed them when running the build locally. There is the option to have the doc build fail on warnings, but currently there are some warnings that seem hard to prevent (they are more info than warning).

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -10,16 +10,6 @@ API components that most users will interact with.
    `0.4.1 API page <https://echopype.readthedocs.io/en/v0.4.1/api.html>`_
    if you're using a previous release. That workflow is now removed.
 
-**Content**
-
-* `EchoData class`_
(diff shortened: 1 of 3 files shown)


LOCEANlloydizard (approved):
thx @gavinmacaulay! yes I also saw those warnings when compiling the docs.. thx for tackling this, it’s great! cheers!

PR #1763, opened 2026-08-08
Fix all remaining documentation build warnings and errors [skip ci]

gavinmacaulay:
This PR fixes all warnings and errors in the sphinx-build process. Items changed include:

- Fix warning about google analytics by removing it (appears to have not been working for several years)
- Fix invalid link warnings
- Remove examples.rst file as it is not used in the docs (was created 5 years ago and has never been in the generated docs, I think)
- Fixed links in ipynb files that no longer work and corrected one spelling mistake
- Adjusted config in conf.py to get more links between doc pages working
- Resolved a warning about markdown headings jumping from H2 to H4 in a ipynb file
- Added numpydoc_show_class_members setting to conf.py, as recommended when using automodapi plugin
- Resolved duplicate index warning by combining open_raw, open_converted, and combine_echodata api docs into one section (not ideal, but it was the only way, and I see that combine_echodata will be removed soon anyway)

Many of these changes fix build warnings that were essentially telling the developer that some links in the built documentation would not work. Tentatively, I've also configured the readthedocs build to fail on warnings (in an effort to reduce these sort of non-working link problems) and added the matching -W build option to the documentation on how to build the docs.


LOCEANlloydizard (commented):
Hey @gavinmacaulay thx! all looks great, just wondering if we need the `suppress_warnings = ["etoc.toctree"]`? that aside, we can merge after, cheers!

LOCEANlloydizard (approved):
Ok thx for the explanation!  i think then we can leave it like this for now then! cheers!

PR #1774, opened 2026-08-31
Test computed values in test__groupby_x_along_channels

### LINKED ISSUE #1308: `test__groupby_x_along_channels` needs to test for value
The current `test__groupby_x_along_channels` does not test for output values of `_groupby_x_along_channels`. The output values of this function are not tested in any other test functions either. 

The test function also has an unused input argument `lat_lon`, which needs investigation to see if that is actually needed, or if that is a missed coverage issue.

KAMALESWAR06:
This pull request adds value checks to test__groupby_x_along_channels (Issue #1308).
Previously, the test only verified that the binned dimension existed. It did not check:

dimension order

output shape

correctness of aggregated values

This meant that changes to binning or averaging could silently produce incorrect results while the test suite still passed.

This PR adds:

assertions on dimension order

assertions on output shape

a brute‑force reference implementation to validate aggregated values in the linear domain

A new reference reducer was required because the existing _brute_nanmean_reduce_3d helper bins as [a, b), while xarray_reduce(..., isbin=True) uses (a, b]. This caused small mismatches at bin edges. The existing helper is left untouched because other fixtures rely on it.

The unused lat_lon parameter was replaced with coverage of the distance_nmi path, which is the actual code path requiring latitude/longitude. If the original intent was different, I’m happy to adjust.

This PR is test‑only and does not modify library code.

### DIFF
--- a/echopype/commongrid/utils.py
+++ b/echopype/commongrid/utils.py
@@ -1,6 +1,6 @@
 import re
 import warnings
-from typing import Literal, Optional, Tuple, Union
+from typing import Literal, Optional, Tuple
 
 import numpy as np
 import pandas as pd
@@ -14,8 +14,8 @@ from ..utils.compute import _lin2log, _log2lin
 
 def compute_raw_MVBS(
     ds_Sv: xr.Dataset,
-    range_interval: Union[pd.IntervalIndex, np.ndarray],
-    ping_interval: Union[pd.IntervalIndex, np.ndarray],
+    range_interval: pd.IntervalIndex,
+    ping_interval: pd.IntervalIndex,
     range_var: Literal["echo_range", "depth"] = "echo_range",
     method="map-reduce",
     reindex=False,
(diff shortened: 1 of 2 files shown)


LOCEANlloydizard (changes requested):
thx @KAMALESWAR06 this is great! (BTW I just rebased the branch to resolve the conflict caused by an unrelated recent PR)

thx for looking into this issue and actually your explanation of why the existing _brute_nanmean_reduce_3d helper could not be reused actually caught my attention and exposed what I think is an inconsistency in the current API!

compute_MVBS() and compute_NASC() explicitly control bin closure through "closed= "(left by default) and convert the bin edges to pd.IntervalIndex before passing them downstream, but compute_raw_MVBS() and compute_raw_NASC() currently also accept plain np.ndarray edges. In that case, the arrays eventually reach _groupby_x_along_channels() and xarray_reduce(..., isbin=True), so the closure is implicitly determined by flox and behaves as right-closed (a, b], as you correctly said.

I tested this with values exactly on both distance and range bin boundaries and calling compute_raw_NASC() with arrays gives the same result as IntervalIndex(closed="right"), but a different result from IntervalIndex(closed="left"). So a user calling the raw API currently gets a bin closure choice they did not explicitly make.

If you're happy to look into it, I think we could use your PR to clean this up rather than adding _reference_binned_nanmean specifically to reproduce that implicit behavior. What do you think about changing compute_raw_MVBS(), compute_raw_NASC(), and _groupby_x_along_channels() to accept pd.IntervalIndex for the bin intervals, so the closure is always explicit before reaching the reduction? and so high level functions would remain unchanged and would still allow users to select closed="left" or "right".

if you agree, I think this test could use explicit IntervalIndex bins and reuse the existing _brute_nanmean_reduce_3d helper, while keeping the value, shape, dimension-order, and distance_nmi coverage you added! Feel free to rename the PR accordingly if you agree with the approach! and of course, just let me know if you don't have the time to do this and i can help! cheers!

Issue #1663, opened 2026-05-14
Add mixed EK80 WBT/GPT CW file as integration test


LOCEANlloydizard:
We identified a good regression/integration test file: `1907RL_CW-D20190810-T215906.raw`

A mixed EK80 CW file with:
- 4 WBT channels in `Sonar/Beam_group1` → **complex**
  - ES333 / ES200 / ES70 / ES120
- 1 GPT channel in `Sonar/Beam_group2` → **power**
  - ES38B

This file exposed mixed-configuration edge cases during calibration debugging, so it would be a useful integration/regression test candidate.

TODO:
- [ ] Add file to GitHub assets (zarr format)
- [ ] Add CI integration coverage using this file
- [ ] Add file metadata/config note in the relevant test-data README (see #1565)

PR #1775, opened 2026-08-31
Add unit tests for range computation

### LINKED ISSUE #548: Upgrade tests for `echodata.compute_range()`
While working on #547 I noticed that the tests for `echodata.compute_range()` at the moment only check if the method can be executed successfully. We should test for the actual computed values.

KAMALESWAR06:
This pull request adds unit tests for `calibrate/range.py` (Issue #548).

The file `echopype/tests/calibrate/test_range.py` is new and contains eight
tests that check the core range calculations:

• compute_range_EK:
  – correct range formula
  – correct scaling with sound speed
  – NaN backscatter → NaN range
  – beam dimension dropped
  – both error paths

• range_mod_TVG_EK:
  – EX60 correction
  – EX80 correction

These tests use small synthetic xarray datasets built in memory, so they do not
depend on downloaded sonar files and run quickly in the normal unit test suite.

This PR is test-only and does not modify library code. It is marked as
“addresses” rather than “closes” Issue #548 because the issue references
`echodata.compute_range()`, which no longer exists.


LOCEANlloydizard (changes requested):
thank you @KAMALESWAR06 for picking up those old issues, it's great to see them resolved, and this one is particularly important as well! I think the current tests give good coverage of the EK range calculation!

I have a couple of small remarks before merging:

- should we add an EK80 case containing both WBT and GPT channels? range_mod_TVG_EK() has specific handling for GPT channels within EK80, while the current tests cover EK80/WBT and EK60/GPT separately
- could you also remove the old fully commented test_compute_range block from tests/echodata/test_echodata.py, since it targets the retired EchoData.compute_range() API?

Otherwise this looks good to me, cheers!

-----
i'm adding some items more as notes for when I come back to it, and potentially we can open follow-up issues for:

- it would be good to have mixed WBT/GPT data integration testing (we could benefit from #1663?)
- we also need to have tests for compute_range_AZFP()
- it would be nice to compare echo_range against reference exports from other software (e.g. Echoview, pyEcholab...)

